PR #698, opened 2013-10-22
Fix a memory leak in the Window Agent

### LINKED ISSUE #586: Windows - agent failing periodically
Windows Server 2003 Web Edition, most recent version of the agent

Event Type:    Error
Event Source:    Datadog Agent
Event Category:    None
Event ID:    1
Date:        7/18/2013
Time:        2:54:41 PM
User:        N/A
Computer:    XXXXXX01
Description:
The description for Event ID ( 1 ) in Source ( Datadog Agent ) cannot be found. The local computer may not have the necessary registry information or message DLL files to display messages from a remote computer. You may be able to use the /AUXSOURCE= flag to retrieve this description; see Help and Support for details. The following information is part of the event: 2013-07-18 14:54:41,562 | checks.collector | ERROR | Unable to fetch Windows system metrics.
  File "checks\collector.pyo", line 142, in run
  File "checks\system\win32.pyo", line 67, in check
  File "wmi.pyo", line 819, in query
  File "wmi.pyo", line 241, in handle_com_error
x_wmi: <x_wmi: Unexpected COM Error (-2147023169, 'The remote procedure call failed and did not execute.', None, None)>.

conorbranagan:
There is a _slow_ memory leak in the Agent that appears to be caused by relative imports of the WMI class and instantiating a WMI class on every check run. There is a test to check that memory isn't leaking in the Windows system checks.

I've eliminated the relative imports and fixed up all the checks so that they only create one WMI connection per host/port/username. In nearly all cases, it will be None for all three values because we're just querying the local host.

I believe this will fix #586 which, upon further investigation, was caused by the Agent taking up ~350MB and the WMI checks were failing with the following error:

``` python
2013-10-20 22:03:59,466 | ERROR | checks.iis(__init__.pyc:454) | Check 'iis' instance #0 failed
  File "checks\__init__.pyc", line 445, in run
  File "C:\Program Files\Datadog\Datadog Agent\files\..\checks.d\iis.py", line 54, in check
  File "wmi.pyc", line 1290, in connect
  File "wmi.pyc", line 241, in handle_com_error
x_wmi: <x_wmi: Unexpected COM Error (-2147024882, 'Not enough storage is available to complete this operation.', None, None)>
2013-10-2
```

### DIFF
--- a/checks.d/iis.py
+++ b/checks.d/iis.py
@@ -1,6 +1,11 @@
 '''
 Check the performance counters from IIS
 '''
+try:
+    import wmi
+except Exception:
+    wmi = None
+
 from checks import AgentCheck
 
 class IIS(AgentCheck):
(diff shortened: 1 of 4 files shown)


ajmal744:
I am using 5.1.1 of datadog windows agent and I see a slow memory leak still happening. Do you know if this issue is still not fixed ?

remh:
Thanks for the feedback @ajmal744 

The issue mentioned in this thread was fixed in 2013 but it's possible that you encoutered another bug.

Could you reach out to support _at_ datadoghq.com with as much details as you can about this leak please ?
Thanks a lot !

PR #888, opened 2014-03-28
Nginx dogstream parsing example

lc-nyovchev:
Nginx log file parsing example with gauge and counter.


remh:
Thanks for the feedback @sserbin 
We are currently evaluating new options to replace dogstream and we likely won't add new features to the current implementation unfortunately, so i'm going to close this issue. Feel free to reply or to reach out to our support team if you have more questions.

lc-nyovchev:
No problem, but hope there is something like a drop-in replacement of dogstream, because we use a lot of those (custom ones, written by us) in production.

PR #1140, opened 2014-09-29
Remove pup

remh:


suprememoocow:
Hi, has Pup been removed from Datadog? The docs on the Datadog site seem to indicate that it's still there: http://docs.datadoghq.com/dd-agent/, but this issue seems to indicate that it's gone.

I'm trying to test a configuration locally and Pup would be useful for testing.

PR #1151, opened 2014-10-07
exclude server with svname 'BACKEND'

celenechang:
otherwise, the servers are overcounted in the count_by_status metric


remh:
It's actually a bug, thanks @celenechang 
I made a few changes and rebased the branch, will merge when the tests are passing!

PR #1161, opened 2014-10-20
Add support for specifying a regex as a queue name matcher

simonmcc:
Instead of having to specify a list of potentially changing queue names, you can specify a regex in rabbitmq.yaml and have the agent match against the regex


remh:
https://github.com/DataDog/dd-agent/pull/1311 got merged. Closing this one.
Thanks again @simonmcc !

PR #1165, opened 2014-10-20
Supervisord check

isaacdd:


remh:
@isaacdd Any update on the comments made ?

remh:
@isaacdd any update on the travis test ?

isaacdd:
@remh have not had the chance to tackle it, and I don't think I will be able to tackle anytime soon.

remh:
Ok changing the milestone then.

checks.d/supervisord.py
```diff
@@ -0,0 +1,158 @@
...
+
+class SupervisordCheck(AgentCheck):
+
+    def check(self, instance):
+        server_name = instance.get('name')
+
+        if not server_name or not server_name.strip():
+            raise Exception("Supervisord server name not specified in yaml configuration.")
+
+        supervisor = self._connect(instance)
```
ckrough: Does this overwrite the 'supervisor' instantiated at import on line 9?

isaacdd: Good catch. It actually does, but it does not cause issues because supervisor (the library) is used only in the _connect method. But worth renaming it a different name to avoid future conflicts.

remh:
@isaacdd any chance to work on this to add a test ?

checks.d/supervisord.py
```diff
@@ -0,0 +1,159 @@
...
+    AgentCheck.UNKNOWN: 'unknown'
+}
+
+SERVER_TAG = 'supervisord_server'
+
+PROCESS_TAG = 'supervisord_process'
+
+FORMAT_TIME = lambda x: time.strftime('%Y-%m-%d %H:%M:%S', time.localtime(x))
+
+@attr(requires='supervisord')
```
remh: This shouldn't be part of the checks.d, but part of the test.

LeoCavaille:
CI passed ! Merging ! Thanks @isaacdd !

PR #1213, opened 2014-11-21
Add a varnish service check.

conorbranagan:
The service check uses the `varnishadm debug.health` command to
get the health of available backends. One check per backend will
be submitted, tagged by the backend name.

The BIG question here is how we will let users enable this because
the varnishadm command requires access to the secret key which has
root permissions by default (and is owned by root).

I don't know what the answer is yet but this will open up the discussion.

### DIFF
--- a/checks.d/varnish.py
+++ b/checks.d/varnish.py
@@ -1,12 +1,29 @@
 # stdlib
-import xml.parsers.expat # python 2.4 compatible
+from collections import defaultdict
 import re
 import subprocess
+import xml.parsers.expat # python 2.4 compatible
 
 # project
 from checks import AgentCheck
 
+
+class BackendStatus(object):
+    HEALTHY = 'healthy'
+    SICK = 'sick'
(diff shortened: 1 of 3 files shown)


checks.d/varnish.py
```diff
@@ -88,16 +72,47 @@ def check(self, instance):
...
+            cmd.extend(['-n', name])
+            tags += [u'varnish_name:%s' % name]
+        else:
+            tags += [u'varnish_name:default']
+        proc = subprocess.Popen(cmd, stdout=subprocess.PIPE,
+                                     stderr=subprocess.PIPE)
+        output, error = proc.communicate()
+        if error and len(error) > 0:
+            self.log.error(error)
+        self._parse_varnishstat_metrics(varnishstat_path, use_xml, tags)
```
mms-gianni: you should pass "output" instead of "varnishstat_path" here. 
See pullrequest https://github.com/DataDog/dd-agent/pull/1377

PR #1214, opened 2014-11-21
Update sqlserver.py

PedroMiguelFigueiredo:
Case corrections.
This way it also works with Latin1_General_BIN collation SQL Server.


remh:
Thanks!

PR #1253, opened 2014-12-17
Extended tagging for related class & constants

rlaveycal:
We run many instances of the same executable. We want to be able to tag our process metrics with the command line params fed to each instance to be able to better identify them. The command line isn't available on the Win32_PerfFormattedData_PerfProc_Process class but is on Win32_Process.

I've extended the WMI check to "join" these classes together (via config) to allow this tagging. I also suppress the "#instance" number appended to the process name by WMI.

Secondly we want to attach some fixed tags to metrics. An extra property has been added to support this.

See wmi_check.yaml.example comments for details

### DIFF
--- a/checks.d/wmi_check.py
+++ b/checks.d/wmi_check.py
@@ -35,9 +35,11 @@ class WMICheck(AgentCheck):
         metrics = instance.get('metrics')
         filters = instance.get('filters')
         tag_by = instance.get('tag_by')
+        link = instance.get('link_tag')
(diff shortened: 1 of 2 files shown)


remh:
Thanks a lot @rlaveycal 
Our 5.2.0 release is now frozen.

Thanks a lot!

yannmh:
Thanks @rlaveycal. I rebased your work on the 5.2.0 agent release, and opened a new PR #1388.

PR #1259, opened 2014-12-23
[http_check] Add SSL certificate service check

LeoCavaille:
Now the `_check` function from NetworkCheck can return a single service check or a list of service checks.


tests/test_http_check_ssl.py
```diff
@@ -0,0 +1,90 @@
...
+            {
+            'name' : 'Warning about Cert expiring',
+            'url' : 'https://github.com',
+            'timeout' : 1,
+            'check_certificate_expiration': True,
+            'days_warning': 9999
+            },
+            {
+            'name' : 'Bad Url',
+            'url' : 'https://thereisnosuchlink.com',
```
LeoCavaille: Removed it b/c it was too flaky and slow :disappointed:

PR #1260, opened 2014-12-23
[network] Really skip excluded_interfaces

sirlantis:
Not sure if the comment is still relevant.

At least the current feature is broken and the code guarantees it stays broken.


LeoCavaille:
@sirlantis happy new year to you too!

Issue #1240, opened 2014-12-11
Marathon >= 0.7 failing


amvapor:
The Marathon plugin checks the TaskRateLimit attribute from the Marathon 0.6 API version, which has been removed https://github.com/mesosphere/marathon/blob/master/docs/docs/upgrade/06xto070.md#task-rate-limit

you can see the dependency here:
https://github.com/DataDog/dd-agent/blob/73dd75035f1982d31818bb8a4a2a6cc856cf8bb7/checks.d/marathon.py#L30

using marathon v0.7.6, the `dd-agent info` looks like this:

```
    marathon
    --------
      - instance #0 [ERROR]: KeyError('taskRateLimit',)
      - Collected 1 metric, 0 events & 1 service check
```

PR #1266, opened 2014-12-26
[marathon] service checks and improvements

gphat:
# What's This Do?

After working on the just-merged etcd plugin for #1235 I wanted to take a look at the Marathon plugin. Looks like the key problem I saw was already fixed in #1240. While I was here I:

* Refactored some of the magic numbers and arrays up into class variables
* Removed unnecessary includes
* Switched to the service check API
* Added the "disk" attribute from the metrics

# Thanks!

I appreciate you taking the time to look at this. If I can do anything to help a merge, let me know!


LeoCavaille:
Hey @gphat, looking again at your PR, there are just a few things missing:
* [ ] you report a service check only for critical state but never an OK state
* [ ] the service check is tagged by `url` in the `get_json` functino which means that we will report different service checks for the different urls, do we want that? Maybe we could just tag by the url base and not the full URL ?
* [ ] to be more readable `APP_METRICS` could be wrapped and sorted like
```
APP_METRICS = [
    'a',
    'b',
]
```
* [ ] also can you rebase and squash your commits against the current master, it looks like you have a merge conflict of some sort

Thanks

gphat:
Somehow I've botched my merge and screwed things up. Lemme resend this. :)

gphat:
Sorry for moving this, I suck. This is replaced by #1296.

PR #1270, opened 2014-12-30
Make sure all service check messages are strings.

conorbranagan:
`e.reason` is an error type and not simply a string describing the
reason. This was causing errors in the HTTP emitter because these
error types aren't serializable.


LeoCavaille:
@conorbranagan thanks for the fix. Could you rebase and fix the conflict?

conorbranagan:
The original issue for ES is fixed with the change using requests so this change just applies to the other checks now.

LeoCavaille:
:+1:

PR #1272, opened 2014-12-31
[postgres] Made BGW_METRICS version dependent

ipolishchuk:
We are using Postgres 9.1.9 and I see a lot of errors in collector,log like this:

2014-12-22 18:33:47 UTC | WARNING | dd.collector | checks.postgres(postgres.py:277) | Not all metrics may be available: ('ERROR', '42703', 'column "checkpoint_write_time" does not exist')

The problem is that  postgres 9.2 columns of a pg_stat_bgwriter view are hardcoded in postgres.py
The columns checkpoint_write_time, checkpoint_sync_time, buffers_backend_fsync are missing in version 9.1
After a BGW_METRICS query failed, all the following queries fail, too. 

The proposed fix makes the BGW_METRICS Postgres version aware. This may not be the most elegant solution, but I tried to make as fewer changes as possible, and do everything in the spirit of the existing code.

After I applied this fix, I have BGW_METRICS working, and the queries following BGW_METRICS are working too.

### DIFF
--- a/checks.d/postgres.py
+++ b/checks.d/postgres.py
@@ -36,22 +36,46 @@ SELECT datname,
         'relation': False,
     }
 
+    COMMON_METRICS = {
+        'numbackends'       : ('postgresql.connections', GAUGE),
+        'xact_commit'       : ('postgresql.commits', RATE),
+        'xact_rollback'     : ('postgresql.rollbacks', RATE),
+        'blks_read'         : ('postgresql.disk_read', RATE),
+        'blks_hit'          : ('postgresql.buffer_hit', RATE),
+        'tup_returned'      : ('postgresql.rows_returned', RATE),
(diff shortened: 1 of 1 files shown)


LeoCavaille:
@ipolishchuk thanks a lot for your contribution. We will review it shortly and consider its inclusion in the next agent release.

LeoCavaille:
@ipolishchuk this is really great, I tested your change with postgres 8.4. 9.0, 9.1, 9.2, 9.3, 9.4 (we are in the process of improving integration testing and running tests against a larger matrix of versions but we are not there yet...). And it works great. One thing that I realized though is `buffers_backend_fsync` is only available after 9.1.
Merging this change and I will add something to handle the pre-9.1 versions.
:+1: 

cc @alq666

PR #1276, opened 2015-01-03
[postgres] Fix #1275. buffers_backend_fsync only avail post 9.1

### LINKED ISSUE #1275: [postgres] buffers_backend_fsync is not available on < 9.1
Add another version test to handle this case.

LeoCavaille:


remh:
Looks good to me!

PR #1282, opened 2015-01-06
[elasticsearch] Fix host tagging for external clusters.

conorbranagan:
When using an external Elasticsearch cluster, we should tag the
hostname by the node's given hostname in the stats output. Otherwise
we end up using only the last node's values because each subsequent
gauge metric will stomp on the one before it.

The test is updated to make sure the hostname given in the stats
(which should be the FQDN) matches what we output at the end.

Finally, this change also splits the cluster health metrics into a
seperate dict. This has two benefits: (a) it will clean up a lot of
debugging logging and extra work for `_process_health_data` and
(b) makes it easy to check against just the stats metrics for tests.

### DIFF
--- a/checks.d/elastic.py
+++ b/checks.d/elastic.py
@@ -34,7 +34,7 @@ class ESCheck(AgentCheck):
 
     DEFAULT_TIMEOUT = 5
 
-    METRICS = { # Metrics that are common to all Elasticsearch versions
+    STATS_METRICS = { # Metrics that are common to all Elasticsearch versions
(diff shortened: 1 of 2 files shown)


LeoCavaille:
Looks good to me, only one thing to add maybe (but it is actually a "bug" coming from my refactoring) is to use `_is_affirmative(instance.get('is_external', False))` to cast properly any text coming from the configuration.

LeoCavaille:
Took care of the small comment above, merging.

PR #1284, opened 2015-01-07
[mysql] Fixes #1283. A custom Mysql check that collects metrics from performance_schema

### LINKED ISSUE #1283: With performance_schema available in MySQL 5.6 a number of key MySQL metrics can be captured
Performance_schema in MySQL 5.6 provides a number of performance related data points that can be used to add key metrics to Datadog. 
One of the most important metric missing in Datadog is the 95th percentile query response time. This can be computed by querying performance_schema tables.

ovaistariq:
There is a lot of data provided by performance_schema in MySQL 5.6. This custom check leverages performance_schema to currently collect the following metric:
mysql.performance.query_run_time_95th_us: This metric shows the 95th percentile query run time for all queries currently executed.

It may require performance_schema.events_statements_summary_by_digest to be truncated from time to time if there are more than 10,000 unique queries executed by the MySQL server.
For more details please read the following links:
http://dev.mysql.com/doc/refman/5.6/en/performance-schema-statement-digests.html
http://www.markleith.co.uk/2012/07/04/mysql-performance-schema-statement-digests/
https://vividcortex.com/blog/2014/11/03/mysql-query-performance-statistics-in-the-performance-schema/


ovaistariq:
I couldn't find a way to send hash as metric value instead of a single value. Otherwise, we could also report 95th-percentile query run time per schema. Kind of like how system.io.r_s reports r_s values per device

ovaistariq:
The new metric name is "mysql.performance.query_run_time_avg". Query response time values are reported for this metric, with separate tags for separate schema name.

remh:
Thanks a lot @ovaistariq our set of new features is frozen for our 5.2.0 release.

We'll review for our 5.3. release.
Thanks again!

ovaistariq:
Thanks Remi

truthbk:
closing in favor of #2116 thanks @ovaistariq !

PR #1285, opened 2015-01-08
[docker] Filter events

### LINKED ISSUE #1234: [docker] container filters do not work with events
It is possible in the Docker check configuration to filter out some containers, however this filter only applies to metrics.
It should apply to events too.

LotharSee:
Some improvements around Docker containers filtering.

* Filters also apply to events. Fix #1234
* Performance improvements around filtering and regex
* Fix the logic of our rules: what we were explaining in the documentation (the docker.yaml.example) wasn't matching the real behavior. Fix the code to make it matches.
* Clarify the explanations in the example config.

### DIFF
--- a/checks.d/docker.py
+++ b/checks.d/docker.py
@@ -133,11 +133,11 @@ class Docker(AgentCheck):
         containers, ids_to_names = self._get_and_count_containers(instance)
 
         # Report container metrics from cgroups
-        self._report_containers_metrics(containers, instance)
+        skipped_container_ids = self._report_containers_metrics(containers, instance)
 
         # Send events from Docker API
         if instance.get('collect_events', True):
-            self._process_events(instance, ids_to_names)
+            self._process_events(instance, ids_to_names, skipped_container_ids)
 
 
     # Containers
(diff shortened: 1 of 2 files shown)


checks.d/docker.py
```diff
@@ -246,19 +268,22 @@ def _new_tags_conversion(self, tag):
...
             self.warning('Timeout during socket connection. Events will be missing.')
 
-    def _pre_aggregate_events(self, api_events):
+    def _pre_aggregate_events(self, api_events, skipped_container_ids):
         # Aggregate events, one per image. Put newer events first.
         events = defaultdict(list)
         for event in api_events:
+            # Skip events related to filtered containers
+            if event['id'] in skipped_container_ids:
+                continue
```
remh: nitpick but let's log something in debug here.

checks.d/docker.py
```diff
@@ -188,23 +188,43 @@ def _get_and_count_containers(self, instance):
...
+        if not instance.get("exclude"):
+            return False
+
+        # Compile regex
+        instance["exclude_patterns"] = []
+        instance["include_patterns"] = []
+        for rule in instance.get("exclude"):
+            instance["exclude_patterns"].append(re.compile(rule))
+        for rule in instance.get("include", []):
+            instance["include_patterns"].append(re.compile(rule))
```
remh: Slightly more readable.

```python
instance["exclude_patterns"] = [re.compile(r) for r in instance.get("exclude", [])]
instance["include_patterns"] = [re.compile(r) for r in instance.get("include", [])]
```

remh:
Looks great besides the nitpicks!

PR #1288, opened 2015-01-13
[postfix] Document requiretty requirement on Redhat/Centos for using sudo.

Xon:
Various CentOS/Redhat flavours require a tty to be allocated when using sudo. Document how to by-pass this requirement per user.


remh:
Thanks a lot @Xon !

PR #1290, opened 2015-01-13
Switch dogstatsd http client to requests

### LINKED ISSUE #1112: override proxies for localhost?
We received [a report](https://datadog.desk.com/agent/case/14226) that when the `http_proxy` env var was set, `urllib.getproxies()` was bypassing local localhost resolution and getting the wrong thing, something like `localhost.resolv.conf.search`.  The reporter had discovered a [similar issue](https://github.com/kennethreitz/requests/issues/879) that requests dealt with.

I'm wondering if it might be a good idea to disable proxies for localhost resolution across the board to avoid this:

```python
import os
if 'no_proxy' not in os.environ:
    os.environ['no_proxy'] = '127.0.0.1,localhost'
```

Or even something smarter which always attempts to add `127.0.0.1,localhost` to the env `no_proxy`, even if it's already set to something else.

remh:
And fix #1112 at the same time


LeoCavaille:
The code and the CI looks good, @remh did you have the occasion to test that in the real world?

remh:
Nope not yet. Let's plan to do this in the upcoming days.

remh:
Tested on my server. Works fine. 

It will be tested at larger scale on staging.

Issue #884, opened 2014-03-25
include proxy info on agent info page


clofresh:
except credentials of course

remh:
@degemer Can you take a pass at it when you'll have a chance ?

remh:
Closed with https://github.com/DataDog/dd-agent/pull/1292

PR #1292, opened 2015-01-13
Add proxy settings on agent info display

degemer:


checks/check_status.py
```diff
@@ -625,14 +625,29 @@ def __init__(self, queue_length=0, queue_size=0, flush_count=0, transactions_rec
         self.transactions_received = transactions_received
         self.transactions_flushed = transactions_flushed
 
+        self.proxy_data = get_config().get('proxy_settings', None)
```
remh: You don't need the new line above.

remh: And you don't need the , None) as it's the default value for the dict.get method.

checks/check_status.py
```diff
@@ -625,14 +625,29 @@ def __init__(self, queue_length=0, queue_size=0, flush_count=0, transactions_rec
...
+
+        if self.proxy_data:
+            lines += [
+                "Proxy",
+                "=====",
+                "",
+                "  Host: %s" % self.proxy_data.get('host'),
+                "  Port: %s" % self.proxy_data.get('port')
+            ]
+        else:
```
remh: Let's not display anything if there is no configured proxy, it would just add useless elements to the info page.

remh:
Thanks! It looks great besides the nitpicks!

Could you add a screenshot of what it would look like ?

Also if a username and a password are set could you: 
1. display the 4 last characters of the username and a constant number of "*" ahead of them. (if the username is less than 4 characters, then maybe display just half of the letters and add the same number of "*").
2. display a constant number of stars representing the password.

Something like:
```
username: *****abcd
password: **********
```

Thanks a lot!

remh:
Also could you also do display these info on the windows info page ?
https://github.com/DataDog/dd-agent/blob/master/win32/status.html

degemer:
All changes done, and here is the screenshot :
screen shot 2015-01-13 at 4 47 39 pm

checks/check_status.py
```diff
@@ -631,8 +632,28 @@ def body_lines(self):
...
+                "=====",
+                "",
+                "  Host: %s" % self.proxy_data.get('host'),
+                "  Port: %s" % self.proxy_data.get('port')
+            ]
+            username = self.proxy_data.get('user')
+            if username:
+                hidden = len(username) / 2 if len(username) <= 7 else len(username) - 4
+                self.hidden_username = '*' * hidden + username[hidden:]
+                self.hidden_password = '*' * len(self.proxy_data.get('password'))
```
LeoCavaille: The number of stars representing the password should preferably be constant to avoid a malicious third party to know its size (and people often share screenshots or send their info page via non-secured email).

LeoCavaille:
Fixes #884

win32/status.html
```diff
@@ -244,6 +244,19 @@ <h1 id="status-header">Agent info</h1>
...
           </dl>
 
+          {% if forwarder['proxy_data'] %}
+          <h3>Proxy</h3>
+
+          <dl>
+            <dt>Host</dt>     <dd>{{ forwarder['proxy_data']['host'] }}</dd>
+            <dt>Port</dt>     <dd>{{ forwarder['proxy_data']['port'] }}</dd>
+            {% if forwarder['hidden_username'] %}
+            <dt>Username</dt>     <dd>{{ forwarder['hidden_username'] }}</dd>
```
remh: This variables "hidden_username" and "hidden_password" won't exist. They have to be set in the ForwarderStatus constructor.

checks/check_status.py
degemer: Ok, I didn't get what @remh wrote earlier. (I guess it's the same thing for the login, a constant number of star before the 4 final letters).

degemer:
With the same example, the output is now :
screen shot 2015-01-13 at 5 44 09 pm

win32/status.html
```diff
@@ -244,6 +244,19 @@ <h1 id="status-header">Agent info</h1>
             <dt>Queue length</dt>       <dd>{{ forwarder['queue_length'] }}</dd>
           </dl>
 
+          {% if forwarder['proxy_data'] %}
+          <h3>Proxy</h3>
+
+          <dl>
+            <dt>Host</dt>     <dd>{{ forwarder['proxy_data']['host'] }}</dd>
+            <dt>Port</dt>     <dd>{{ forwarder['proxy_data']['port'] }}</dd>
+            {% if forwarder['hidden_username'] %}
```
remh: This should probably be:
```python
{% if forwarder.get('hidden_username') %}
```
otherwise i think you'll get a KeyError

remh:
Looks great! Thanks!

PR #1293, opened 2015-01-14
[ci/tests] New build. New Infrastructure. New Flavors

LeoCavaille:
Much more testing.
Such multiple versions.
Very docker-blazing-speed
So hip new PG tests
:dog: 

The first build will probably a bit slow, subsequent builds should be faster thanks to pre-compiled software caches.
And more seriously see commit messages:

The builds must not rely on sudo from now on.
That means either fetching pre-compiled binaries from
upstream or fetching the sources and compiling directly
on the build machine.
However, as the released upstream sources do not change
we can benefit from the caching system and cache the
binaries once we have them.
All these software are in $INTEGRATIONS_DIR

Introduction also of running the integration tests against
different versions of the software described in the build
matrix.

Also a bit of documentation on how to use the new test system.

+ Several more fixes of the test suite codebase


checks.d/postgres.py
```diff
@@ -306,6 +316,8 @@ def _collect_stats(self, key, db, instance_tags, relations):
                     log_func("Not all metrics may be available: %s" % str(e))
                     continue
 
+                self.log.info(results)
```
clutchski: you want this at info?

LeoCavaille:
@clutchski thank you so much for the careful review and for catching that, forgot to remove this `debug` info I added to know which metrics where failing on different postgres versions.
I rebased the postgres commit with the change.

tests/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+Each flavor is defined in `ci/flavor.rb`, and we set different steps for running the flavor build :
+
+* **before_install** needs to be run before installation
+* **install** installs the 3p software
+* **before_script** generally setups the software and launch it in background
+* **script** runs nosetests with a filter (see how to write tests)
+* **cleanup** stops the software and remove unnecessary data (not running on Travis, because the buildboxes are disposable)
+
+Your test cases must be written in `tests/test_flavor.rb` and they must use the nose `attr` decorator to be filtered by the flavors.
```
remh:
```
tests/test_flavor.py
```

tests/common.py
```diff
@@ -33,7 +33,8 @@ def load_check(name, config, agentConfig):
     # init the check class
     try:
         return check_class(name, init_config=init_config, agentConfig=agentConfig, instances=instances)
-    except:
+    except Exception as e:
+        raise Exception("Check is using old API, {0}".format(e))
```
remh: You can remove the lines below then.

PR #1294, opened 2015-01-14
Use a temporary file for df output to avoid PIPE limits.

conorbranagan:
subprocess.PIPE is limited in size because the data is buffered in
memory. When running the Agent on a host that has many containers,
the number of mountpoints can be very large. Because of this buffer limit,
the shell call can hang for these huge outputs.

To workaround this, we will write the output to a temporary file and
read it from there.

This is a "quick" fix for a specific case and there might be a better
general approach. But either way, we will still need to handle the
case of large outputs from our subprocess commands otherwise the
whole Agent collector simply hangs.

refs:
- http://thraxil.org/users/anders/posts/2008/03/13/Subprocess-Hanging-PIPE-is-your-enemy/
- https://datadog.desk.com/agent/case/17823

cc @remh @LeoCavaille

### DIFF
--- a/checks/system/unix.py
+++ b/checks/system/unix.py
@@ -10,6 +10,7 @@ import socket
 import string
 import subprocess as sp
 import sys
+import tempfile
 import time
 
 # project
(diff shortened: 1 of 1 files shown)


remh:
Thanks @conorbranagan nice debugging session!

I'm more in favor of rewriting the disk check to use psutil and stop using subprocess. Those are pretty small and it should be pretty straightforward. 

But i don't want to guarantee that it would be part of the 5.2 release and would rather aim at doing this for 5.3.

If it's a critical fix we could merge it as it for 5.2 and rewrite the disk check for 5.3 ? 

I'm open to more suggestions though.

conorbranagan:
OK if we're going to go the psutil direction, then I'd prefer to get this small change in 5.2 since it's critical for a particular customer so they can get their Agents running again going forward (without having to keep a patched version around).

What do you think?

checks/system/unix.py
```diff
@@ -906,12 +907,15 @@ def _get_subprocess_output(command, log):
     Run the given subprocess command and return it's output. Raise an Exception
     if an error occurs.
     """
-    proc = sp.Popen(command, stdout=sp.PIPE, close_fds=True, stderr=sp.PIPE)
+    stdout_f = tempfile.TemporaryFile('rw')
```
remh: Nitpick but could you embed that in a "with" statement so won't have to wait for the garbage collection for the file to be deleted ?

PR #1296, opened 2015-01-14
[marathon] service checks and improvements

gphat:
# What's This Do?

After working on the just-merged etcd plugin for #1235 I wanted to take a look at the Marathon plugin. Looks like the key problem I saw was already fixed in #1240. While I was here I:

* Refactored some of the magic numbers and arrays up into class variables
* Removed unnecessary includes
* Switched to the service check API
* Added the "disk" attribute from the metrics

# Thanks!

I appreciate you taking the time to look at this. If I can do anything to help a merge, let me know!

### DIFF
--- a/checks.d/marathon.py
+++ b/checks.d/marathon.py
@@ -46,6 +46,7 @@ class Marathon(AgentCheck):
     def get_json(self, url, timeout):
         try:
             r = requests.get(url, timeout=timeout)
+            r.raise_for_status()
         except requests.exceptions.Timeout:
             # If there's a timeout
             self.service_check(self.SERVICE_CHECK_NAME, AgentCheck.CRITICAL,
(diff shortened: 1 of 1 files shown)


gphat:
Note: This is a replacement for #1266. I used git before being caffeinated and seem to've ruined my branch. :)

checks.d/marathon.py
```diff
@@ -20,98 +28,37 @@ def check(self, instance):
...
-            return None
+            self.service_check(self.SERVICE_CHECK_NAME, AgentCheck.CRITICAL,
+                message='%s timed out after %s seconds.' % (url, timeout),
+                tags = ["url:{}".format(url)])
+            raise Exception("Timeout when hitting %s" % url)
 
         if r.status_code != 200:
-            self.status_code_event(url, r, aggregation_key)
-            self.warning("Got %s when hitting %s" % (r.status_code, url))
+            self.status_code_event(url, r)
```
LeoCavaille: This function does not exist anymore?
You may want to use `r.raise_for_status()` in the
```
try: 
     r = requests.get()
     r.raise_for_status()
except timeout:
    <service_check_criticial>
except httperror:
    <service_check_criticial>
else:
    <service_check_ok>`
```
above (don't know for marathon, but other 2XX status may be OK) and you can get rid of this extra if block.

checks.d/marathon.py
```diff
@@ -20,98 +28,37 @@ def check(self, instance):
...
+            raise Exception("Got %s when hitting %s" % (r.status_code, url))
+        else:
+            self.service_check(self.SERVICE_CHECK_NAME, AgentCheck.OK,
+                tags = ["url:{}".format(url)]
```
LeoCavaille: Also did you see my comment on the previous PR of this URL tag ?

gphat: Yes, sorry! I meant to ask where you'd end up with multiple URLs and not want the whole thing. I'm assuming anyone using this wold be hitting something on localhost, which means you'd differentiate by either port or path. Where would trimming down to the base url be useful?

checks.d/marathon.py
gphat: I didn't know about that. Thanks!

gphat:
Updated with your feedback. Thanks for dealing with my dumb re-send of this. :P

LeoCavaille:
:+1:

PR #1080, opened 2014-08-19
Add check for fluentd

takus:
Add AgentCheck for [Fluentd](http://www.fluentd.org/). This check enables us to create a dashboard like this:

fluentd___datadog

PR #1301, opened 2015-01-15
Add check for fluentd

LeoCavaille:
[leo@datadoghq.com] rebased for new CI skeleton and no sudo use
Squashed a bunch of commits too

Rebased changes and work from @takus to work with the new CI (doesn't allow sudo any more), we install fluentd with `gem` easily instead. Replaces #1080.

Improved a little bit the test which did not fail if it received no metrics (because iterating on the set of metrics), so now the test expects a certain number of metrics and service checks to be output.


LeoCavaille:
Thanks @takus again, merging your code for 5.2 !

takus:
Thank you very much!!!!

kzk:
:+1:

yyuu:
:+1:

jhotta:
wao!! ;)

PR #1303, opened 2015-01-16
Make histogram aggregates & percentiles configurable globally

LeoCavaille:


aggregator.py
```diff
@@ -605,6 +617,10 @@ def __init__(self, hostname, interval=1.0, expiry_seconds=300, formatter=None, r
             'ms': Histogram,
             's': Set,
         }
+        self.histogram_aggregates = histogram_aggregates\
```
remh: That should probably be in the Aggregator class so you don't have to duplicate it in all its subclasses.

aggregator.py
```diff
@@ -719,7 +740,7 @@ class MetricsAggregator(Aggregator):
     A metric aggregator class.
     """
 
-    def __init__(self, hostname, interval=1.0, expiry_seconds=300, formatter=None, recent_point_threshold=None):
+    def __init__(self, hostname, interval=1.0, expiry_seconds=300, formatter=None, recent_point_threshold=None, histogram_aggregates=None, histogram_percentiles=None):
```
remh: Nitpick but can you add a new line here ?

aggregator.py
```diff
@@ -605,6 +617,10 @@ def __init__(self, hostname, interval=1.0, expiry_seconds=300, formatter=None, r
             'ms': Histogram,
             's': Set,
         }
+        self.histogram_aggregates = histogram_aggregates\
+           or get_histogram_aggregates()
```
remh: Why do we need to recall this function ? Hasn't it already be done when initiating the config ?

aggregator.py
```diff
@@ -644,8 +660,13 @@ def submit_metric(self, name, value, mtype, tags=None, hostname=None,
 
             if context not in metric_by_context:
                 metric_class = self.metric_type_to_class[mtype]
-                metric_by_context[context] = metric_class(self.formatter, name, tags,
-                    hostname, device_name)
+                if metric_class == Histogram:
```
remh: This makes me sad :(
Maybe we should pass the config to the Metric objects instead ?

aggregator.py
```diff
@@ -236,16 +240,22 @@ def flush(self, ts, interval):
...
+        if 'min' in self.aggregates:
+            metric_aggrs.append(('min', min_, MetricTypes.GAUGE))
+        if 'max' in self.aggregates:
+            metric_aggrs.append(('max', max_, MetricTypes.GAUGE))
+        if 'median' in self.aggregates:
+            metric_aggrs.append(('median', med, MetricTypes.GAUGE))
+        if 'avg' in self.aggregates:
+            metric_aggrs.append(('avg', avg, MetricTypes.GAUGE))
+        if 'count' in self.aggregates:
+            metric_aggrs.append(('count', self.count/interval, MetricTypes.RATE))
```
remh:
```python
aggregators = {
    'min': (min_, MetricTypes.GAUGE),
    'max': (max_, MetricTypes.GAUGE),
    'median': (med, MetricTypes.GAUGE),
    'avg': (avg, MetricTypes.GAUGE),
    'count': (self.count/interval, MetricTypes.RATE)
}

metric_aggrs = [(agg_name, agg_func, m_type)\
 for agg_name, (agg_func, m_type) in aggregators.iteritems()\
  if agg_name in self.aggregates
```

dogstatsd.py
```diff
@@ -367,7 +367,7 @@ def init(config_path=None, use_watchdog=False, use_forwarder=False, args=None):
         sleep(4)
         sys.exit(0)
 
-    log.debug("Configurating     dogstatsd")
+    log.debug("Configuring dogstatsd")
```
remh: :expressionless:

aggregator.py
```diff
@@ -213,12 +215,14 @@ def flush(self, timestamp, interval):
...
-    def __init__(self, formatter, name, tags, hostname, device_name):
+    def __init__(self, formatter, name, tags, hostname, device_name,
+            aggregates=None, percentiles=None):
         self.formatter = formatter
         self.name = name
         self.count = 0
         self.samples = []
-        self.percentiles = [0.95]
+        self.aggregates = aggregates or []
+        self.percentiles = percentiles or [0.95]
```
remh: probably a edge case but because of this, if you don't want to collect percentiles at all, it wouldn't work.

aggregator.py
remh: or even better without using a dictionary
```python
aggregators = [
    ('min', min_, MetricTypes.GAUGE),
    ('max', max_, MetricTypes.GAUGE),
    ('median', med, MetricTypes.GAUGE),
    ('avg', avg, MetricTypes.GAUGE),
    ('count', self.count/interval, MetricTypes.RATE)
]

metric_aggrs = [(agg_name, agg_func, m_type)\
 for agg_name, agg_func, m_type in aggregators\
  if agg_name in self.aggregates]
```

aggregator.py
LeoCavaille: Maybe it's unclear I guess, this is to provide a sane default (calling get with no args) when people instantiate an aggregator without getting from config a anything.

aggregator.py
LeoCavaille: :+1:

aggregator.py
LeoCavaille: That was my first approach, before realizing that no metric specific (say Histograms for instance) code was in this class, I can put that here but I find it lacking consistency a bit. Maybe I should just refactor more things :unamused:

aggregator.py
LeoCavaille: I can get rid of this default, actually I made a mistake in this code above (aggregates should be initialized to the old default). I hope it's not the case but if people rely on this `Histogram` class we should be careful about changing the default init...

LeoCavaille:
@remh this is should be ready for final review hopefully (allowing per metric type config to be passed, with no special case for histograms)

tests/test_config.py
```diff
@@ -42,7 +43,7 @@ def testBadPidFile(self):
 
         p = PidFile('test', pid_dir)
         path = p.get_path()
-        self.assertEquals(path, "/tmp/test.pid")
+        self.assertEquals(path, os.path.join(os.environ.get('TMPDIR', '/tmp'), 'test.pid'))
```
remh: even better:
```python
self.assertEquals(path, os.path.join(tempfile.gettempdir(), 'test.pid')
```

PR #1306, opened 2015-01-21
Convert fluentd.buffer_total_queued_size into a rate instead of a gauge

degemer:
`buffer_total_queued_size` seems to be a counter that keeps incrementing with every new entry, it is more logical to collect it as a rate.

### DIFF
--- a/checks.d/fluentd.py
+++ b/checks.d/fluentd.py
@@ -12,8 +12,11 @@ import simplejson as json
 
 class Fluentd(AgentCheck):
     SERVICE_CHECK_NAME = 'fluentd.is_ok'
-    GAUGES = ['retry_count', 'buffer_queue_length']
-    RATE = 'buffer_total_queued_size'
+    METRICS = [
+        ('retry_count', AgentCheck.gauge),
+        ('buffer_queue_length', AgentCheck.gauge),
+        ('buffer_total_queued_size', AgentCheck.rate),
+    ]
 
     """Tracks basic fluentd metrics via the monitor_agent plugin
     * number of retry_count
@@ -41,14 +44,10 @@ class Fluentd(AgentCheck):
             status = json.loads(res)
 
             for p in status['plugins']:
-                for m in self.GAUGES:
-                    if p.get(m) is None:
+                for m_name, m_func in self.METRICS:
+                    if p.get(m_name) is None or p.get('plugin_id') not in plugin_ids:
                         continue
-                    if p.get('plugin_id') in plugin_ids:
-                        self.gauge('fluentd.%s' % (m), p.get(m), ["plugin_id:%s" % p.get('plugin_id')])
-
-                if p.get(self.RATE) is not None and p.get('plugin_id') in plugin_ids:
-                    self.rate('fluentd.%s' % (self.RATE), p.get(self.RATE), ["plugin_id:%s" % p.get('plugin_id')])
+                    m_func(self, 'fluentd.%s' % m_name, p.get(m_name), ["plugin_id:%s" % p.get('plugin_id')])
         except Exception, e:
             msg = "No stats could be retrieved from %s : %s" % (url, str(e))
             self.service_check(self.SERVICE_CHECK_NAME, AgentCheck.CRITICAL, tags=service_check_tags, message=msg)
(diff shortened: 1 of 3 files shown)


checks.d/fluentd.py
```diff
@@ -45,6 +46,9 @@ def check(self, instance):
                         continue
                     if p.get('plugin_id') in plugin_ids:
                         self.gauge('fluentd.%s' % (m), p.get(m), ["plugin_id:%s" % p.get('plugin_id')])
+
+                if p.get(self.RATE) is not None and p.get('plugin_id') in plugin_ids:
+                    self.rate('fluentd.%s' % (self.RATE), p.get(self.RATE), ["plugin_id:%s" % p.get('plugin_id')])
```
remh: Nitpick: You could rewrite this whole part with something like:
```python
for p in status['plugins']:
    for m_name, m_func in self.METRICS:
        if p.get(m_name) is None:
            continue
        m_func(self, 'fluentd.%s' % m_name, p.get(m_name), ["plugin_id:%s" % p.get('plugin_id')])
```

assuming that:
```python
METRICS = [
    ('retry_count', AgentCheck.gauge),
    ('buffer_queue_length', AgentCheck.gauge),
    ('buffer_total_queued_size', AgentCheck.rate),
]
```
it's slightly more concise and more readable.

degemer: Sure, I wasn't really satisfied of what I wrote. Thanks!

ci/fluentd.rb
```diff
@@ -9,7 +9,7 @@
     end
 
     task :before_script => ['ci:common:before_script'] do
-      sh %(fluentd -c $TRAVIS_BUILD_DIR/ci/resources/fluentd/td-agent.conf &)
+      sh %(fluentd -c $TRAVIS_BUILD_DIR/ci/resources/fluentd/td-agent.conf > /dev/null 2>&1 &)
```
remh: Is this command really spammy ? 

Won't we be completely blind if it starts failing without us realizing ?

tests/test_fluentd.py
```diff
@@ -29,16 +29,14 @@ def test_fluentd(self):
...
-            self.assertEquals(m[3]['type'], 'gauge')
+            if m[0] == 'fluentd.buffer_total_queued_size':
+                self.assertEquals(m[3]['type'], 'rate')
+            else:
+                self.assertEquals(m[3]['type'], 'gauge')
+            self.assertEquals(m[2], 0)
             self.assertEquals(m[3]['tags'], ['plugin_id:plg1'])
 
-        self.assertEquals(len(metrics), 3, metrics)
+        self.assertEquals(len(metrics), 2, metrics)
```
remh: you can call check.run() a second time. This way you would collect rate metrics as well.

degemer:
So, after a lot of digging, I found that my affirmation "buffer_total_queued_size seems to be a counter that keeps incrementing with every new entry" was true in my case (ie a single fluentd writing all logs to a single file without a real buffer), but not in more standard cases.

I didn't find a real explanation of the metrics exposed by fluentd (the [monitoring page](http://docs.fluentd.org/articles/monitoring) is really succinct), but here is a graph from graphite (a munin [plugin](https://github.com/moaikids/munin-fluentd/blob/master/plugins/fluentd/fluentd_monitor_agent) also exist) to prove my point: 
Using munin

I also found a explanation of [fluentd buffer](http://docs.fluentd.org/articles/buffer-plugin-overview), which is probably the best indirect explanation of the metrics.

So, in order to obtain a "regular" behaviour, I used this config:
```
<match **>
  id   plg2
  type file
  path /var/log/fluentd/access1
  buffer_type file
  buffer_path /var/log/fluentd/myapp.*.buffer
  buffer_chunk_limit 10k
  buffer_queue_limit 128
  flush_interval 10s
</match>
```
and hit the fluentd with:
```
while true; do curl -X POST -d 'json={"json":"message"}' http://localhost:8888/test.hello; sleep 0.1; done
```

Other useful links: (some in Japanese)
http://inokara.hateblo.jp/entry/2014/05/31/081249
https://www.omniref.com/ruby/gems/fluentd/0.10.42/symbols/Fluent::MonitorAgentInput::MONITOR_INFO#line=182
http://docs.fluentd.org/articles/high-availability

PR #1307, opened 2015-01-22
Change postgres deadlocks from gauge to rate

### LINKED ISSUE #1289: Invalid metric type GAUGE for postgresql.deadlocks
Looking at graphs from my postgresql servers, I can see that the graphs for postgresql.deadlocks shows the total number of deadlocks since startup. Wouldn't it make more sense to use RATE instead?

https://github.com/DataDog/dd-agent/blob/2bd69dc0ee50eb1d246ad0b958efb415d6ec5c7e/checks.d/postgres.py#L54

nate:
Fixes #1289.

### DIFF
--- a/checks.d/postgres.py
+++ b/checks.d/postgres.py
@@ -51,7 +51,7 @@ SELECT datname,
     }
 
     NEWER_92_METRICS = {
-        'deadlocks'         : ('postgresql.deadlocks', GAUGE),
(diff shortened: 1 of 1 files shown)


remh:
Thanks a lot @nate it does seem to be a bug indeed.

I'll triple-check that and we'll merge it for our 5.2.0 release.

Thanks again!

remh:
Looks great! Merging.
Thanks again

PR #1171, opened 2014-10-22
Basic Teamcity agent - only supports guest authentication, default branch

AirbornePorcine:
Basic support for the TeamCity CI server (https://www.jetbrains.com/teamcity/) in a similar style to the Jenkins agent check.

Unfortunately I'm not very well-versed in Python testing techniques so I wasn't able to write any tests here. Help with that would certainly be appreciated.

PR #1308, opened 2015-01-22
[teamcity] new check

LeoCavaille:
Rebased, improved a little more the PR to fit the current agent codebase.
Also added a test to test the logic of the check : get only newer builds after the agent start. The responses saved in mocks came from a real TeamCity server at one point though.

Supersedes #1171

### DIFF
--- a/checks.d/teamcity.py
+++ b/checks.d/teamcity.py
@@ -4,87 +4,130 @@ import time
 
 # project
 from checks import AgentCheck
+from config import _is_affirmative
 
 
-class TeamCity(AgentCheck):
-    headers = {'Accept': 'application/json'}
-    server = None
+class TeamCityCheck(AgentCheck):
+    HEADERS = {'Accept': 'application/json'}
+    DEFAULT_TIMEOUT = 10
+    NEW_BUILD_URL = "http://{server}/guestAuth/app/rest/builds/?locator=buildType:{build_conf},sinceBuild:id:{since_build},status:SUCCESS"
+    LAST_BUILD_URL = "http://{server}/guestAuth/app/rest/builds/?locator=buildType:{build_conf},count:1"
 
-    def __init__(self, name, init_config, agentConfig):
-        AgentCheck.__init__(self, name, init_config, agentConfig)
+    def __init__(self, name, init_config, agentConfig, instances=None):
+        AgentCheck.__init__(self, name, init_config, agentConfig, instances)
+
+        # Keep track of last build IDs per instance
         self.last_build_ids = {}
-        if self.init_config.get("server") is None:
-            raise Exception("You must specify a server in teamcity.yaml")
-        self.server = self.init_config["server"]
 
-    def _initialize_if_required(self, instance_name, build_configuration):
-        if self.last_build_ids.get(instance_name, None) is None:
-            self.log.info("Initializing {0}".format(instance_name))
-            request = requests.get(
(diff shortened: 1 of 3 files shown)


tests/test_teamcity.py
```diff
@@ -0,0 +1,89 @@
...
+import unittest
+
+# 3p
+from mock import MagicMock, patch
+
+# project
+from tests.common import load_check
+
+CONFIG = {
+    'init_config': {'server': 'localhost:8111'},
```
remh: What if you have multiple servers in one host ? 
It seems to me that should be part of the instance configuration, not the init_config

conf.d/teamcity.yaml.example
```diff
@@ -0,0 +1,30 @@
+init_config:
+  # Specify the server name of your teamcity instance here
+  # Guest authentication must be on if you want the check to be able to get data
+  server: teamcity.mycompany.com
```
remh: My comment about "server" being in init_config actually belongs here.

checks.d/teamcity.py
```diff
@@ -0,0 +1,126 @@
...
+        if 'server' not in self.init_config:
+            raise Exception("You must specify a server in teamcity.yaml")
+        self.server = self.init_config['server']
+
+    def _initialize_if_required(self, instance_name, build_configuration):
+        # Already initialized
+        if instance_name in self.last_build_ids:
+            return
+
+        self.log.info("Initializing {0}".format(instance_name))
```
remh: We don't want that at the info level

checks.d/teamcity.py
```diff
@@ -0,0 +1,126 @@
...
+            raise
+
+        self.log.debug(
+            "Last build id for instance {0} is {1}."
+            .format(instance_name, last_build_id)
+        )
+        self.last_build_ids[instance_name] = last_build_id
+
+    def _build_and_send_event(self, new_build, instance_name, is_deployment, host, tags):
+        self.log.info("Found new build with id {0}, saving and alerting.".format(new_build["id"]))
```
remh: It should probably be in debug

checks.d/teamcity.py
```diff
@@ -0,0 +1,126 @@
...
+            raise Exception("Each instance must have a build configuration")
+
+        host = instance.get('host_affected') or self.hostname
+        tags = instance.get('tags')
+        is_deployment = _is_affirmative(instance.get('is_deployment', False))
+
+        self._initialize_if_required(instance_name, build_configuration)
+
+        # Look for new successful builds
+        new_build_url = "http://{0}/guestAuth/app/rest/builds/?locator=buildType:{1},sinceBuild:id:{2},status:SUCCESS"\
```
remh: it's a nitpick but let's declare this string as a constant
```python
BUILD_URL = "http://{0}/guestAuth/app/rest/builds/?locator=buildType:{1},sinceBuild:id:{2},status:SUCCESS"
```
and then you can do:
```python
new_build_url = BUILD_URL.format(self.server, build_configuration, self.last_build_ids[instance_name])
```

remh:
It doesn't look like we are collecting build/deployment failures. Is that a limitation of their API ?

It'd be nice to have service checks on those if it's possible to get them.

Otherwise it looks good!

LeoCavaille:
Merging as is. For improvements, we'll see if there is a demand. If you are reading, feel free to file an issue to add more features!

PR #1311, opened 2015-01-23
[rabbitmq] allow regex queue/node matching

LeoCavaille:
Supersedes #1161


remh:
Looks great! And it's adding tests :heart_eyes:

simonmcc:
woohooo! thank you @LeoCavaille & @remh

PR #1315, opened 2015-01-26
[etcd] Use a tag to mark leader/follower

degemer:
Use a tag to mark leader and followers instead of a gauge.
Add two captured metrics.

### DIFF
--- a/checks.d/etcd.py
+++ b/checks.d/etcd.py
@@ -35,7 +35,9 @@ class Etcd(AgentCheck):
 
     SELF_GAUGES = {
         'sendPkgRate': 'etcd.self.send.pkgrate',
-        'sendBandwidthRate': 'etcd.self.send.bandwidthrate'
+        'sendBandwidthRate': 'etcd.self.send.bandwidthrate',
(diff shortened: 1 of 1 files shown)


checks.d/etcd.py
```diff
@@ -88,6 +76,20 @@ def check(self, instance):
                 else:
                     self.log.warn("Missing key {} in stats.".format(key))
 
+        storeResponse = self.get_store_metrics(url, timeout)
```
LeoCavaille: This is very nitpicky, but if you could change those names @degemer with python-compliant names (snake_case) I would be :heart_eyes_cat:

degemer: Sure. Couldn't pylint check for this ?

LeoCavaille: Already does, stuff like:
```
C:  2, 4: Invalid variable name "snakeCaseVar" (invalid-name)
```
However, we don't throw pylint errors at this level (convention) otherwise I guess we would probably have 2 zillions errors to fix.
Writing up about stricter rules & guidelines for the agent code is part of this quest...

LeoCavaille:
Thanks!

PR #1318, opened 2015-01-27
SNMP new types support + bug fix

yannmh:
* Add Integer and Integer32 type support
* Change the SNMP get command logic to solve the 'Missing OID' issue
* Some PEP8 code cleaning

### DIFF
--- a/checks.d/snmp.py
+++ b/checks.d/snmp.py
@@ -121,6 +121,13 @@ class SnmpCheck(AgentCheck):
         port = instance.get("port", 161) # Default SNMP port
         return cmdgen.UdpTransportTarget((ip_address, port), timeout=timeout, retries=retries)
 
+    def raise_on_error_indication(self, error_indication, instance):
+        if error_indication:
+            message = "{0} for instance {1}".format(error_indication,
+                                                    instance["ip_address"])
+            instance["service_check_error"] = message
+            raise Exception(message)
+
     def check_table(self, instance, oids, lookup_names):
         '''
         Perform a snmpwalk on the domain specified by the oids, on the device
@@ -155,12 +162,9 @@ class SnmpCheck(AgentCheck):
                 *(oids[first_oid:first_oid + OID_BATCH_SIZE]),
                 lookupValues=lookup_names,
                 lookupNames=lookup_names)
+
             # Raise on error_indication
-            if error_indication:
-                message = "{0} for instance {1}".format(error_indication,
-                                                        instance["ip_address"])
-                instance["service_check_error"] = message
-                raise Exception(message)
+            self.raise_on_error_indication(error_indication, instance)
 
             # Continue on error_status
             if error_status:
@@ -193,11 +197,7 @@ class SnmpCheck(AgentCheck):
                     lookupNames=lookup_names)
 
                 # Raise on error_indication
-                if error_indication:
-                    message = "{0} for instance {1}".format(error_indication,
-                                                            instance["ip_address"])
-                    instance["service_check_error"] = message
-                    raise Exception(message)
+                self.raise_on_error_indication(error_indication, instance)
 
                 # Continue on error_status
                 if error_status:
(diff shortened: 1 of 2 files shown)


checks.d/snmp.py
```diff
@@ -123,49 +139,89 @@ def check_table(self, instance, oids, lookup_names):
...
+                transport_target,
+                *(oids[first_oid:first_oid + OID_BATCH_SIZE]),
+                lookupValues=lookup_names,
+                lookupNames=lookup_names)
+
+            # Raise on error_indication
+            self.raise_on_error_indication(error_indication, instance)
+
+            # Continue on error_status
+            if error_status:
```
LotharSee: This is a bit odd.
You created a function to deal with `error_indication` but you didn't with `error_status`, while the logic looks quite the same (except that one is raising an error).

What I suggest is something like this:

```python
if error_indication:
  report_error_indication(error_indication, instance)

if error_status:
  report_error_status(error_status, instance)
```

checks.d/snmp.py
```diff
@@ -123,49 +139,89 @@ def check_table(self, instance, oids, lookup_names):
...
+                message = "{0} for instance {1}".format(error_status.prettyPrint(),
+                                                        instance["ip_address"])
+                instance["service_check_error"] = message
+                self.log.warning(message)
+                continue
```
LotharSee: You don't need a continue

checks.d/snmp.py
```diff
@@ -123,49 +139,89 @@ def check_table(self, instance, oids, lookup_names):
...
+                    continue
+                for table_row in var_binds_table:
+                    complete_results.extend(table_row)
+
+            all_binds.extend(complete_results)
+            first_oid = first_oid + OID_BATCH_SIZE
+
+        for result_oid, value in all_binds:
+            if lookup_names:
+                object = result_oid.getMibSymbol()
```
LotharSee: Avoid the variable name `object` since it already means something in Python, and it doesn't indicate what is inside.

tests/test_snmp.py
```diff
@@ -235,6 +283,35 @@ def test_invalid_metric(self):
             self.assertEquals(sc['status'], self.check.CRITICAL, sc)
             self.assertEquals(sc['tags'], ['snmp_device:localhost'], sc)
 
+    def test_fallback_snmpgetnext(self):
```
LotharSee: Good test!
Should we also test that `1.3.6.1.2.1.25.6.3.1.4` and `1.3.6.1.2.1.25.6.3.1.4.0` give the same value?

checks.d/snmp.py
yannmh: `report_error_indication` is raising an Exception and stops the check execution.

The reason why I didn't factorize `report_error_status` is because it has a `continue` (c.f your comment below) statement which refers directly to the upper level `for` loop.

checks.d/snmp.py
```diff
@@ -123,49 +139,89 @@ def check_table(self, instance, oids, lookup_names):
...
+                snmp_command = self.cmd_generator.nextCmd
+                error_indication, error_status, error_index, var_binds_table = snmp_command(
+                    auth_data,
+                    transport_target,
+                    *missing_results,
+                    lookupValues=lookup_names,
+                    lookupNames=lookup_names)
+
+                # Raise on error_indication
+                self.raise_on_error_indication(error_indication, instance)
```
LotharSee: Same here.
You can even do

```python
self.report_errors(error_indication, error_status, instance)
```

checks.d/snmp.py
yannmh: We don't want to process the results in this case, but jump to the next `for` loop iteration. So I think we might want to keep the `continue` statement.
What do you think ?

LotharSee: Oh, my bad, I forgot that we were in a while. You are right!

tests/test_snmp.py
yannmh: Interesting test, I am adding it.

LeoCavaille:
Looking good!

PR #1320, opened 2015-01-27
Discourage legacy options for Docker and network checks

### LINKED ISSUE #1317: Put "skip_event: true" in in the network config example
Now that we promote service checks over events for HTTP check, we should keep the default value of skip_event to false (not to break existing systems), but put `skip_event: true` in the example file (but also in https://app.datadoghq.com/monitors#create/network) so that new users won't use the legacy feature.

https://github.com/DataDog/dd-agent/blob/5.1.1/conf.d/http_check.yaml.example#L66-L71
https://github.com/DataDog/dd-agent/blob/5.1.1/conf.d/tcp_check.yaml.example#L29

LotharSee:
* Restore the proper docker.yaml.example "new_tag_names" option (got commened on 36781988)
* Fix #1317

### DIFF
--- a/conf.d/docker.yaml.example
+++ b/conf.d/docker.yaml.example
@@ -18,7 +18,7 @@ instances:
 
     # Use tag names that don't conflict.
     # Keep it false if you have old dashboard using the old tag names. Default: false.
-    # new_tag_names: true
+    new_tag_names: true
 
     # Tag metrics with the command running inside the container.
     # tag_by_command: false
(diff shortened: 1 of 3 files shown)


LeoCavaille:
Great, but @LotharSee at least you could have make the tests pass :broken_heart:

miketheman:
Is there any more context on this? The "default: False" and toggling the condition to true here without any doc in the check leads a user down the road to confusing warnings.

Also, I think this message https://github.com/DataDog/dd-agent/blob/c9fa9af2e1f1504daf7d48cd5a675904288b45ef/checks/network_checks.py#L167 could declare what version this is is going to be removed in - the message only started appearing in Agent 5.2.0.

PR #1321, opened 2015-01-27
Clean and add options in redis.yaml.example

### LINKED ISSUE #1316: Attribute db is missing in the Redis configuration example
The example file https://github.com/DataDog/dd-agent/blob/5.1.1/conf.d/redisdb.yaml.example doesn't talk about the "db" attribute, while it is properly supported https://github.com/DataDog/dd-agent/blob/5.1.1/checks.d/redisdb.py#L99

LotharSee:
Fix #1316 with some extra cleanup.

### DIFF
--- a/conf.d/redisdb.yaml.example
+++ b/conf.d/redisdb.yaml.example
@@ -3,13 +3,30 @@ init_config:
 instances:
   - host: localhost
     port: 6379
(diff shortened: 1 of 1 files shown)


LeoCavaille:
Thanks

PR #1322, opened 2015-01-27
Determine the Linux distribution logic

yannmh:
Try `lsb_release` command to determine the Linux distribution and 
- If failing try `/etc/issue` should contain the release information
- If still failing, rely on the `/etc/*-release` files

More info:
http://unix.stackexchange.com/questions/92199/how-can-i-reliably-get-the-operating-systems-name

Also see:
https://github.com/DataDog/dd-agent-omnibus/pull/8


LeoCavaille:
Looking good!

PR #1325, opened 2015-01-28
[core] Move to versioned endpoints for the agent

remh:
Also add headers about the version of the agent and of the forwarder

Note that this test will fail until the new endpoints are created and properly configured.

### DIFF
--- a/ddagent.py
+++ b/ddagent.py
@@ -199,9 +199,13 @@ class MetricTransaction(Transaction):
         if parsed_url.netloc not in LEGACY_DATADOG_URLS:
             return default_url
 
-        # Replace https://app.datadoghq.com in https://v5-2-0.agent.datadoghq.com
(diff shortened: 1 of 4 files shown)


remh:
cc: @miketheman @alq666 @LeoCavaille

LeoCavaille:
Looks fine to me, I wonder if the `-app` is actually useful giving that we already namespace by the `.agent` subdomain.

remh:
This is to allow new endpoints like:

5.2.0-metrics.agent.datadoghq.com
5.2.0-metadata.agent.datadoghq.com 

in the future.

LeoCavaille:
Interesting! How about we create the endpoint before merging it so we can verify your test actually works.

remh:
yup @miketheman is on it

remh:
New ELBs are in place. Merging!

PR #1326, opened 2015-01-29
Enabled regex testing for content_match

chrissnell:
Here's a quick change to Ryan's http_check.py revisions that enable regex testing for content_match.  This makes content_match more flexible in accommodating situations where a simple if ... in ... match would be insufficient.

Consider this real-life example of content being tested:

```
{"status":"UP","rabbit":{"status":"UP"},"db":{"status":"UP","database":"MySQL","hello":1}}
```

If we want to test for that first ```"status":"UP"```, we need to use a regex because that string is not unique within the content string:

```
   content_match: '^\{"status":"UP"'
```

The one gotcha with regex in content_match is special metacharacters like "?".  I considered creating a content_match_re for regex matching but decided that it would be cleaner to just document in the example YAML the need to escape these characters.

### DIFF
--- a/checks.d/http_check.py
+++ b/checks.d/http_check.py
@@ -4,6 +4,7 @@ import os.path
 import socket
 import ssl
 import time
+import re
 from urlparse import urlparse
 
 # 3rd party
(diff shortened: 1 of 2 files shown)


remh:
Thanks a bunch a @chrissnell  !

Our master branch is currently frozen for the preparation of our 5.2.0 release.

Thanks again!

yannmh:
Thanks @chrissnell . I just rebased your PR on the last 5.2.1 agent release and opened a new PR #1390.

PR #1330, opened 2015-01-29
[redis] first pass at slowlog

miketheman:
@remh I've gotten pretty far.

- I still need some help with the config get for a default - or even get the length from the slowlog itself.
- `tags` aren't being passed in, no idea why. It'd be nice to have the `redis_port` tags, etc.
- I haven't figured out the persistence yet
- likely to be some idiomatic python things I'm missing


remh:
Closing in favor of.

PR #1331, opened 2015-01-29
Add support for forced SNMP data types to help w/ buggy devices

chrissnell:
Some SNMP-enabled devices return incorrect data types when certain OIDs are queried.  The F5 1600LTM load balancer is an example of such a device.   This PR allows for the forcing of data types (to 'gauge' or 'counter') in the event of an incorrect response from the remote agent.

It adds a parameter to the metric definition on the YAML, 'forced_type'.

NOTE: The changes in this PR were tested in a back-ported version of this improved snmp check that Datadog provided me.   The check patched here is from a newer version of the agent that I don't have any easy way to test right now.  The changes are the same between the versions and it *should* work but I have no way to test.

### DIFF
--- a/checks.d/snmp.py
+++ b/checks.d/snmp.py
@@ -282,6 +282,7 @@ class SnmpCheck(AgentCheck):
         tags = instance.get("tags", [])
         tags = tags + ["snmp_device:" + instance.get('ip_address')]
         for metric in instance.get('metrics', []):
+            forced_type = metric.get('forced_type')
             if 'OID' in metric:
                 queried_oid = metric['OID']
                 for oid in results:
@@ -293,7 +294,7 @@ class SnmpCheck(AgentCheck):
                                      queried_oid)
                     continue
                 name = metric.get('name', 'unnamed_metric')
-                self.submit_metric(name, value, tags)
+                self.submit_metric(name, value, forced_type, tags)
 
     def report_table_metrics(self, instance, results):
         '''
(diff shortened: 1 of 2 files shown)


remh:
Thanks a bunch a @chrissnell  !

Our master branch is currently frozen for the preparation of our 5.2.0 release.

Thanks again!

checks.d/snmp.py
```diff
@@ -386,6 +388,16 @@ def submit_metric(self, name, snmp_value, tags=[]):
 
         metric_name = self.normalize(name, prefix="snmp")
 
+        if forced_type:
+            if forced_type.lower() == "gauge":
+                value = int(snmp_value)
+                self.gauge(metric_name, value, tags)
+                return
```
remh: could you have an 
```python
if ...
elif ...
else
```
syntax instead ? You would then raise an exception in the else statement as the state wouldn't be properly set up.

remh:
It looks good beside the nitpick.

Could you also add a test for this please ?

irabinovitch:
@chrissnell Thank you for the contribution for the SNMP integration.  Were you by any chance able to review @remh 's feedback. We'd love to pull this in soon.

remh:
@truthbk can you have a look at this PR by the way ?

truthbk:
Closing in favor of #2165, thanks for your contribution @chrissnell! Your authorship has been respected in the #2165 commit history.

Issue #1243, opened 2014-12-12
[core] Allow API key to be set from environment variable


mfischer-zd:
One should be able to avoid storing secret data in the configuration file by allowing the Datadog API key to be set in an environment variable before launching any API client (e.g. dd-agent, dogstatsd, etc.).

remh:
Adding that to the 5.3.0 milestone it's not that hard and can indeed be useful.

remh:
Pusing back to 5.4.0, see https://github.com/DataDog/dd-agent/pull/1343

yannmh:
Adding you @LeoCavaille on the issue as you're assigned to #1343

olivielpeau:
See also #2095

olivielpeau:
Hi @gugahoi, this is indeed possible on Agent v6 with the env var `DD_API_KEY`

PR #1343, opened 2015-02-03
Allow environment variable substitution in conf file

delitescere:
- Enabled with the `--enable-env` command line option
- Opt-in ensures current conf files don't get into ConfigParser
  recursion issues, e.g. `home: %(HOME)s`
- Provide nicer error on exit if the conf file has bad interpolation
- New unit tests with and without command line option present

### DIFF
--- a/config.py
+++ b/config.py
@@ -322,7 +322,7 @@ def get_config(parse_args=True, cfg_path=None, options=None):
         try:
             for option in config.options('Main'):
                 agentConfig[option] = config.get('Main', option)
-        except ConfigParser.InterpolationMissingOptionError as e:
+        except ConfigParser.InterpolationMissingOptionError:
             sys.stderr.write("The configuration file contains invalid variable substitution. If you want to use environment variables, use the --enable-env option\n")
             sys.exit(3)
 
--- a/config.py
+++ b/config.py
@@ -73,6 +73,13 @@ def get_parsed_args():
                         default=False, dest='use_forwarder')
     parser.add_option('-n', '--disable-dd', action='store_true', default=False,
                         dest="disable_dd")
+
+    # Allow interpolation of environment variables
+    # This is opt-in initially to avoid recursive interpolation in ConfigParser.
+    # It could be made opt-out (--disable-env) at the next major version if the community prefers it.
+    parser.add_option('--enable-env', action='store_true', default=False,
+                        dest="enable_env")
+
     parser.add_option('-v', '--verbose', action='store_true', default=False,
                         dest='verbose',
                       help='Print out stacktraces for errors in checks')
@@ -85,6 +92,7 @@ def get_parsed_args():
                                 'dd_url': None,
                                 'clean': False,
                                 'disable_dd':False,
+                                'enable_env': False,
                                 'use_forwarder': False}), []
     return options, args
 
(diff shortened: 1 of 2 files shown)


delitescere:
The build failure is because it is claimed python 2.7 is ok to use, but CI is building with python 2.6 which doesn't support the `assertRaises` as a context manager (see http://docs.python.org/2/library/unittest.html#unittest.TestCase.assertRaises).

delitescere:
Wow this is one flakey CI. It can't make up it's mind if it's running python 2.6 or 2.7, and randomly fails anyway.

Suffice it to say, this patch is clean.

Cheers,
Josh.

LeoCavaille:
Thanks @delitescere 
We will review this change for the 5.3 release! (5.2 is now frozen for imminent release)
Regarding the CI, it's running BOTH python 2.6 and 2.7 ! Because non-EOL'd OS still run python 2.6, and anyone should be able to install from source the agent..

We have made huge steps towards non-flakiness, but I cannot agree more it sucks. Relying on so many 3rd party mirrors and services for integration testing is hard, next up will be some more caching to avoid that.

LeoCavaille:
Cross reference #1243

vincenzovitale:
+1

LeoCavaille:
This is a really nice feature, we'll work on it soon for the future release we didn't have time to really work on it yet. And we want to add some more work on top so that we revamp the `config.py` code and the check config code in the agent to allow using the environment for more things. Setting milestone to 5.4.

LeoCavaille:
Still some more work to do, working a leo/configcleanup branch. Will submit new PR for 5.5.

mattes:
+1

irabinovitch:
@LeoCavaille  Should we close this in favor of the branch you are working on?

irabinovitch:
@LeoCavaille Should we close this in favor of the branch you mentioned in https://github.com/DataDog/dd-agent/pull/1343#issuecomment-110452032 ?

irabinovitch:
@LeoCavaille checking back in?

LeoCavaille:
@irabinovitch wow somehow I missed that thread 3 times in a row. I will coordinate with the rest of the team (cc @remh) to decide what to do next for this feature and re-update the PR this week.
I think this has been milestoned for 6.X as that code has been refactored quite a lot since that PR was opened. My former branch with updates to this initial patch cannot easily be merged either.

config.py
```diff
@@ -302,12 +310,21 @@ def get_config(parse_args=True, cfg_path=None, options=None):
         path = os.path.dirname(path)
 
         config_path = get_config_path(cfg_path, os_name=get_os())
-        config = ConfigParser.ConfigParser()
+        if options is not None and options.enable_env:
+            log.info("Interpolation of environment variables in config is enabled")
+            config = ConfigParser.ConfigParser(os.environ)
```
jonhosmer: @delitescere @LeoCavaille 
You may want to add `ConfigParser.optionxform = str` before `config = ConfigParser.ConfigParser(os.environ)` so that the keys in `os.environ` are treated as case sensitive.  

For example:
```python
>>> from ConfigParser import ConfigParser
>>> import os
>>> os.environ['XY'] = 'foo'
>>> os.environ['Xy'] = 'bar'
>>> os.environ['xY'] = 'baz'
>>> print '\n'.join(sorted([':'.join((k, v)) for k, v in os.environ.items() if k.lower() == 'xy']))
XY:foo
Xy:baz
xY:bar

>>> cfgp = ConfigParser(os.environ)
>>> print '\n'.join(sorted([':'.join((k, v)) for (k, v) in cfgp.defaults().items() if k.lower() == 'xy']))
xy:foo

>>> ConfigParser.optionxform = str
>>> cfgp_cs = ConfigParser(os.environ)
>>> print '\n'.join(sorted([':'.join((k, v)) for (k, v) in cfgp_cs.defaults().items() if k.lower() == 'xy']))
XY:foo
Xy:baz
xY:bar

>>> assert([(k, v) for (k, v) in cfgp_cs.defaults().items() if k.lower() == 'xy'] == [(k, v) for (k, v) in os.environ.items() if k.lower() == 'xy'])
>>> assert([(k, v) for (k, v) in cfgp.defaults().items() if k.lower() == 'xy'] == [(k, v) for (k, v) in os.environ.items() if k.lower() == 'xy'])
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
<ipython-input-43-1f97baeb1e97> in <module>()
----> 1 assert([(k, v) for (k, v) in cfgp.defaults().items() if k.lower() == 'xy'] == [(k, v) for (k, v) in os.environ.items() if k.lower() == 'xy'])

AssertionError:
```

PR #1346, opened 2015-02-04
Update mongo.yml.example to add database

jshirley:
If you follow the mongo.yml example, you end up with a series of messages cluttering the logs rather noisily:
```
Feb  3 22:00:51 my-awesome-host dd.collector[16860]: INFO (mongo.py:183): No MongoDB database found in URI. Defaulting to admin.
```

Updating the example file would go a long way towards making this explicit, although it could be a better change to silence (or change the level to DEBUG) of the message above.

### DIFF
--- a/conf.d/mongo.yaml.example
+++ b/conf.d/mongo.yaml.example
@@ -1,7 +1,8 @@
 init_config:
 
 instances:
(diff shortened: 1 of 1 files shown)


remh:
Looks good to me! Thanks a lot @jshirley !

PR #1347, opened 2015-02-04
[haproxy] include/exclude logic for services

LeoCavaille:
Adds two new config options by instance:
  * `services_include`
  * `services_exclude`

These options are list of regexes that are used with the
following rule:
    "If a tag matches an exclude rule, it won't be included
    unless it also matches an include rule."

Ships also new unit/integration tests to verify this logic.

cc @garnermccloud


remh:
LGTM!

PR #1349, opened 2015-02-05
Add check for ZFS 

mboylevt:
I've added some basic checks for the zfs filesystem.  These checks give insights into space used/available, compression/deduplication ratio, and basic health checks.


remh:
Thanks a lot @mboylevt ! This looks like an awesome pull request!

We are currently in the process of releasing our 5.2.0 agent, hence our master branch is frozen for now.

We will review your Pull request when we will be preparing our 5.3.0 release.
Thanks again!

remh:
Also one first nitpick, could you squash these commits into a single one please ?
Thanks again!

checks.d/zfs.py
```diff
@@ -0,0 +1,232 @@
+#!/usr/bin/python
```
remh: You don't need this line

checks.d/zfs.py
```diff
@@ -0,0 +1,232 @@
...
+from __future__ import division
+from checks import AgentCheck
+import subprocess
+import re
```
remh: Coud you organize your imports the same way it's done in other checks please ?
e.g. https://github.com/DataDog/dd-agent/blob/master/checks.d/redisdb.py#L4-L13

checks.d/zfs.py
mboylevt: Removed

checks.d/zfs.py
mboylevt: Sure can!  Done.

mboylevt:
@remh at this point, I'm not sure how to squash my original commits w/o squashing the merged in commits from your master as well.  Would you prefer that I simply re-open this pull request?

mboylevt:
Any update on getting this merged?

remh:
Hi @mboylevt sorry for the long delay.
Unfortunately we won't merge this PR.

The mains are that:
 - We are trying to avoid the use of subprocess more and more in the agent
 - This pull requests requires root access to work and there is currently no nice way for a user to run the agent as root (and we don't want to recommend doing it).

In the same time, you can definitely use it as a custom check without having this merged, so it shouldn't block you.

Thanks a lot for your time and let us know if you have more questions.

Issue #915, opened 2014-04-24
Redis: fetch additional replication metrics available in 2.8


conorbranagan:
There are some new backlog metrics that are useful. We could potentially tag by slave IP as well, although it might be tricky to resolve it to a specific host.

From a master:
```
# Replication
role:master
connected_slaves:1
slave0:ip=*.*.*.*,port=6379,state=online,offset=19697,lag=0
master_repl_offset:19697
repl_backlog_active:1
repl_backlog_size:1048576
repl_backlog_first_byte_offset:2
repl_backlog_histlen:19696
```

From a slave:
```
# Replication
role:slave
master_host:*.*.*.*
master_port:6379
master_link_status:up
master_last_io_seconds_ago:4
master_sync_in_progress:0
slave_repl_offset:19837
slave_priority:100
slave_read_only:1
connected_slaves:0
master_repl_offset:0
repl_backlog_active:0
repl_backlog_size:1048576
repl_backlog_first_byte_offset:0
repl_backlog_histlen:0
```

remh:
Fixed by #1350

PR #1350, opened 2015-02-05
[redis] new post 2.8 replication metrics

hkaj:
[redis] add new replication metrics for >= 2.8 instances

New metrics:
* `redis.replication.backlog_histlen`
  shows the current replication backlog size in bytes used from the
  allocated `repl-backlog-size` in the config.
* `redis.replication.master_repl_offset`  
  this metric is available on masters, it shows the current offset of
  the replication process on the master in bytes
* `redis.replication.slave_repl_offset`
  idem for the slaves.
* `redis.replication.delay`
  on masters only, measures the difference between the last offset
  reported by the slave and the current replication offset, which
  indicates how far your slave is behind.
  This metric is tagged by `slave_ip`, `slave_port` and `slave_id`
  (internal ID in the master).

Also added on Travis a replication config between the two instances we
have to assert the presence of this new metric.
Bonus: now the redis instances that were launched by the CI are cleaned
properly by the `cleanup` task.

see issue #915

### DIFF
--- a/checks.d/redisdb.py
+++ b/checks.d/redisdb.py
@@ -11,8 +11,10 @@ from checks import AgentCheck
 # 3rd party
 import redis
 
+
 class Redis(AgentCheck):
     db_key_pattern = re.compile(r'^db\d+')
+    slave_key_pattern = re.compile(r'^slave\d+')
     subkeys = ['keys', 'expires']
 
     SOURCE_TYPE_NAME = 'redis'
@@ -63,7 +65,9 @@ class Redis(AgentCheck):
         'master_last_io_seconds_ago':   'redis.replication.last_io_seconds_ago',
         'master_sync_in_progress':      'redis.replication.sync',
         'master_sync_left_bytes':       'redis.replication.sync_left_bytes',
-
+        'repl_backlog_histlen':         'redis.replication.backlog_histlen',
+        'master_repl_offset':           'redis.replication.master_repl_offset',
+        'slave_repl_offset':            'redis.replication.slave_repl_offset',
     }
 
     RATE_KEYS = {
@@ -109,7 +113,7 @@ class Redis(AgentCheck):
 
                 # Only send useful parameters to the redis client constructor
                 list_params = ['host', 'port', 'db', 'password', 'socket_timeout',
-                    'connection_pool', 'charset', 'errors', 'unix_socket_path']
+                               'connection_pool', 'charset', 'errors', 'unix_socket_path']
 
                 # Set a default timeout (in seconds) if no timeout is specified in the instance config
                 instance['socket_timeout'] = instance.get('socket_timeout', 5)
@@ -130,7 +134,7 @@ class Redis(AgentCheck):
         if 'unix_socket_path' in instance:
             tags_to_add = ["unix_socket_path:%s" % instance.get("unix_socket_path")]
         else:
-            tags_to_add =  ["redis_host:%s" % instance.get('host'), "redis_port:%s" % instance.get('port')]
+            tags_to_add = ["redis_host:%s" % instance.get('host'), "redis_port:%s" % instance.get('port')]
 
         if instance.get('db') is not None:
             tags_to_add.append("db:%s" % instance.get('db'))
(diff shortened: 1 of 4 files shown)


checks.d/redisdb.py
```diff
@@ -215,6 +218,16 @@ def _check_db(self, instance, custom_tags=None):
                             self.warning("{0} key not found in redis".format(key))
                         self.gauge('redis.key.length', 0, tags=key_tags)
 
+        # Save the replication delay for each slave
+        for key in info.keys():
```
remh: you shouldn't use .keys () as it would create a copy of the keys.

You should just do:
```python
for key in info
```

checks.d/redisdb.py
```diff
@@ -215,6 +218,16 @@ def _check_db(self, instance, custom_tags=None):
...
+            if self.slave_key_pattern.match(key):
+                slave_offset = info[key].get('offset')
+                master_offset = info.get('master_repl_offset')
+                if slave_offset and master_offset and master_offset - slave_offset >= 0:
+                    delay = master_offset - slave_offset
+                    slave_tags = tags + [info[key]['ip']] if 'ip' in info[key].keys() else tags
```
remh: same here, you shouldn't use .keys()

remh:
Looks great besides the nitpicks and the comments i told you offline! Thanks a lot!
It looks like your tests are failing on travis, can you have a look ?

checks.d/redisdb.py
```diff
@@ -215,6 +218,17 @@ def _check_db(self, instance, custom_tags=None):
...
                         self.gauge('redis.key.length', 0, tags=key_tags)
 
+        # Save the replication delay for each slave
+        for key in info:
+            if self.slave_key_pattern.match(key) and isinstance(info[key], dict):
+                slave_offset = info[key].get('offset')
+                master_offset = info.get('master_repl_offset')
+                if slave_offset and master_offset and master_offset - slave_offset >= 0:
+                    delay = master_offset - slave_offset
+                    slave_tags = tags + ['slave_ip:%s' % info[key]['ip']] if 'ip' in info[key] else tags
```
LeoCavaille: It might be more readable this way:
```python
slave_tags = tags
for slave_tag in ('ip', 'port'):
    if slave_tag in info[key]:
        slave_tags.append('slave_{0}:{1}'.format(slave_tag, info[key][slave_tag]))
```

checks.d/redisdb.py
```diff
@@ -215,6 +218,17 @@ def _check_db(self, instance, custom_tags=None):
...
+                    slave_tags.append('slave_id:%s' % key.lstrip('slave'))
+                    self.gauge('redis.replication.delay.%s' % key, delay, tags=slave_tags)
```
LeoCavaille: see PR comment

LeoCavaille:
@hkaj thanks for taking a stab at it.
A few remarks:
* the metric should be named `redis.replication.delay` : lose the `slave0` suffix because we can have this information by breaking down by slave ID.
* can you tag this metric by `slave_port` too (usually people will run many synced redis instances on their master and slaves servers)
* you installed replication on the redis CI instances but can you add some integration tests to see if this metric is actually here (for the proper versions of redis)! see `tests/test_redis.py`
* a better commit message following our [guidelines](https://github.com/DataDog/dd-agent/wiki/Submitting-code#commit-messages) would be something like:
```
[redis] add new replication metrics for >= 2.8 instances

New metrics:
* `redis.replication.backlog_histlen`
  shows the current replication backlog size in bytes used from the
  allocated `repl-backlog-size` in the config.
* `redis.replication.master_repl_offset`  
  this metric is available on masters, it shows the current offset of
  the replication process on the master in bytes
* `redis.replication.slave_repl_offset`
  idem for the slaves.
* `redis.replication.delay`
  on masters only, measures the difference between the last offset
  reported by the slave and the current replication offset, which
  indicates how far your slave is behind.
  This metric is tagged by `slave_ip`, `slave_port` and `slave_id`
  (internal ID in the master).

Also added on Travis a replication config between the two instances we
have to assert the presence of this new metric.
Bonus: now the redis instances that were launched by the CI are cleaned
properly by the `cleanup` task.
```
* added a few comments in the code too

Thanks :+1:

hkaj:
@LeoCavaille Sorry for the delay. I think the last commits address your comments, please let me know if anything else is missing.

checks.d/redisdb.py
```diff
@@ -215,8 +219,23 @@ def _check_db(self, instance, custom_tags=None):
...
+                    slave_tags = tags[:]
+                    for slave_tag in ('ip', 'port'):
+                        if slave_tag in info[key]:
+                            slave_tags.append('slave_{0}:{1}'.format(slave_tag, info[key][slave_tag]))
+                    slave_tags.append('slave_id:%s' % key.lstrip('slave'))
+                    self.gauge('redis.replication.delay', delay, tags=slave_tags)
+
     def check(self, instance):
-        if (not "host" in instance or not "port" in instance) and not "unix_socket_path" in instance:
+        if ("host" not in instance or "port" not in instance) and "unix_socket_path" not in instance:
```
LeoCavaille: Not sure that's exactly what we want, see https://gist.github.com/LeoCavaille/801f5bb9288a94ca8d94
Especially the 4th example?

checks.d/redisdb.py
```diff
@@ -215,8 +219,23 @@ def _check_db(self, instance, custom_tags=None):
...
             raise Exception("You must specify a host/port couple or a unix_socket_path")
         custom_tags = instance.get('tags', [])
-        self._check_db(instance,custom_tags)
+        self._check_db(instance, custom_tags)
```
LeoCavaille: :cake:

LeoCavaille:
Except the last small nitpick on this config check, it looks good. Just waiting to hear back from you on that and then I'll merge it! Thanks

checks.d/redisdb.py
hkaj: @LeoCavaille I only moved the "not" keywords to the right place, that didn't change the behavior, which I think is what we want.

As long as there's a socket_path we're good, if there is no socket path, there must be a host and a port. We probably don't want to raise an exception if a user switch from a host:port couple to a socket and forget to remove the host from the config.

See https://github.com/DataDog/dd-agent/blob/haissam/redis-repl-metrics/checks.d/redisdb.py#L103-L107 for the use case.

LeoCavaille:
:+1:

PR #1354, opened 2015-02-06
[System/Linux] Add uptime metric

remh:
It was requested by a customer

* Only works on linux so far
* Also update outdated syntax of some other system checks


remh:
Merging

PR #1355, opened 2015-02-06
[windows] Catch psutil RuntimeError (which was blocking all checks)

degemer:
Emit warning when not able to fetch disk latency metrics.

### DIFF
--- a/checks/system/win32.py
+++ b/checks/system/win32.py
@@ -223,19 +223,22 @@ class Disk(Check):
                     device_name=name)
 
     def check_disk_latency(self, agentConfig):
-        disk_io_counters = psutil.disk_io_counters(True)
-        for disk_name, disk in disk_io_counters.iteritems():
-            read_time_pct = disk.read_time * 100.0 / 1000.0 # x100 to have it as a percentage, /1000 as psutil returns the value in ms
-            write_time_pct = disk.write_time * 100.0 / 1000.0 # x100 to have it as a percentage, /1000 as psutil returns the value in ms
-            self.save_sample("system.disk.read_time_pct", read_time_pct, device_name=disk_name) 
-            self.save_sample("system.disk.write_time_pct", write_time_pct, device_name=disk_name)
+        try:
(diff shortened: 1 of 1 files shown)


checks/system/win32.py
```diff
@@ -223,19 +223,22 @@ def check_disk_usage(self, agentConfig):
...
+        except RuntimeError:
+            self.logger.warning("Unable to fetch disk latency metrics")
 
 
     def check(self, agentConfig):
         self.check_disk_usage(agentConfig)
         self.check_disk_latency(agentConfig)
         return self.get_metrics()
-        
+
```
LeoCavaille: :cake:

checks/system/win32.py
```diff
@@ -223,19 +223,22 @@ def check_disk_usage(self, agentConfig):
...
-            self.save_sample("system.disk.read_time_pct", read_time_pct, device_name=disk_name) 
-            self.save_sample("system.disk.write_time_pct", write_time_pct, device_name=disk_name)
+        try:
+            disk_io_counters = psutil.disk_io_counters(True)
+            for disk_name, disk in disk_io_counters.iteritems():
+                read_time_pct = disk.read_time * 100.0 / 1000.0 # x100 to have it as a percentage, /1000 as psutil returns the value in ms
+                write_time_pct = disk.write_time * 100.0 / 1000.0 # x100 to have it as a percentage, /1000 as psutil returns the value in ms
+                self.save_sample("system.disk.read_time_pct", read_time_pct, device_name=disk_name)
+                self.save_sample("system.disk.write_time_pct", write_time_pct, device_name=disk_name)
+        except RuntimeError:
```
LeoCavaille: Should we catch more stuff? Can your confirm `disk_io_counters` can only raise `RuntimeError`?

degemer: I didn't think about this, but I just checked the source code, and it's the only exception which is raised inside this function.

LeoCavaille:
Thanks!

PR #1357, opened 2015-02-06
[config] fix integer to boolean casting helper

LeoCavaille:
In python bool is a subclass of int, so when we parse
something from the config, being an int like 0/1 or a
"true" boolean isinstance(s, int) does the job. For
additional security we cast it again as a bool.

This was the intentional behavior from the start...


remh:
Looks good. Merging.

PR #1358, opened 2015-02-06
Update postgres.py

adriandoolittle:
Adding checks for pg_statio_all_tables.

### DIFF
--- a/checks.d/postgres.py
+++ b/checks.d/postgres.py
@@ -1,5 +1,4 @@
 """PostgreSQL check
-
 Collects database-wide metrics and optionally per-relation metrics.
 """
 # project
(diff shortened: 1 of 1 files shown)


remh:
Thanks a lot @adriandoolittle ! We'll review this and aim at merging for our 5.3.0 release.

alq666:
@adriandoolittle since this generates a lot of metrics for every single relation and that may affect all our users, we'd like to make it a relation-only query, i.e. it's only enabled for relations you specify, as opposed to being for all relations by default.

alq666:
I've started a branch with that: https://github.com/DataDog/dd-agent/tree/alq666/pg-statio

PR #1361, opened 2015-02-09
[osx] Use omnibus-built OS X installer

degemer:
- install script (curl -L http://my.install.script | sudo bash)
- modify path in launchd
- create a custom script (from the previous `agent` one) to run the agent on Mac

Related: 
- https://github.com/DataDog/dd-agent-omnibus/pull/11
- https://github.com/DataDog/omnibus-software/pull/8


LeoCavaille:
That's great, thanks, going to review it properly.
Any chance you could add to this PR moving `osx` in `packaging/`, it would be great if we can flatten this directory by source without having the extra `datadog-agent` subdir, like:
```
packaging/
    debian/
    centos/
    osx/
```

degemer:
Sure, I moved the `osx` directory. 
But in order to get rid of the `datadog-agent` directory, we will have to update a lot of other things.

degemer:
I've built a Windows package using this branch, and it works just fine (I didn't test all the JMX stuff, but since I didn't touch `win32/agent.py`, it shouldn't be an issue).
This PR fixes https://github.com/DataDog/dd-agent/issues/1365.

packaging/osx/app/Info.plist
```diff
@@ -0,0 +1,24 @@
...
+        <key>CFBundleExecutable</key>
+        <string>gui</string>
+        <key>CFBundleIconFile</key>
+        <string>Agent.icns</string>
+    	<key>CFBundleGetInfoString</key>
+    	<string>5.2, Copyright © 2009-2015, Datadog Inc.</string>
+        <key>CFBundlePackageType</key>
+        <string>APPL</string>
+        <key>CFBundleVersion</key>
+        <string>5.2.0</string>
```
yannmh: Is it Datadog agent version number ? If so, does that mean that we need to update this file with every release ?

degemer: Nope, it's updated there during the build https://github.com/DataDog/dd-agent/pull/1361/files#diff-2eeaed663bd0d25b7e608891384b7298R130
(I should probably remove this default though, it was only for tests)

yannmh:
Successfully tested on OSX and Windows :ok_hand:.

Great work, thanks @degemer. Let's :shipit:

PR #1366, opened 2015-02-11
Update agent version in source installer

yannmh:


remh:
Thanks. We'll need to create the tag when will be ready to release.

PR #1368, opened 2015-02-12
[dogstatsd] Fix event transmission to the forwarder

degemer:
No event were sent by dogstatsd to the forwarder.


LeoCavaille:
Thanks :+1: 
Note to @DataDog/mars , we should add some more non-regression testing to the forwarder, this kind of error sholud be catched!

PR #1370, opened 2015-02-12
[win32_event_log] stop ignoring timezone in time conversion

degemer:
Event log datetime standard is `yyyymmddHHMMSS.mmmmmmsUUU` (see https://msdn.microsoft.com/en-us/library/aa387237%28v=vs.85%29.aspx), `sUUU` being the offset in minutes. (and `s` `+` or `-`).

`wmi.to_time` doesn't deal with the `s` information, so  here is a quick trick to add timezone.


remh:
Looks good to me!

PR #1373, opened 2015-02-13
Add some metrics to the postgres check

DorianZaccaria:
Improve postgresql.replication_delay, add postgresql.replication_delay_bytes and postgresql.total_tables.

### DIFF
--- a/checks.d/postgres.py
+++ b/checks.d/postgres.py
@@ -172,11 +172,32 @@ WHERE nspname NOT IN ('pg_catalog', 'information_schema') AND
   relname = ANY(%s)"""
     }
 
+    COUNT_METRICS = {
+        'descriptors': [
+            ('schemaname', 'schema')
+        ],
+        'metrics': {
+            'pg_stat_user_tables': ('postgresql.total_tables', GAUGE),
+        },
+        'relation': False,
+        'query': """
+SELECT schemaname, count(*)
+FROM %s
+GROUP BY schemaname
+        """
+    }
+
+    REPLICATION_METRICS_9_1 = {
+        'CASE WHEN pg_last_xlog_receive_location() = pg_last_xlog_replay_location() THEN 0 ELSE GREATEST (0, EXTRACT (EPOCH FROM now() - pg_last_xact_replay_timestamp())) END': ('postgresql.replication_delay', GAUGE),
+    }
+
+    REPLICATION_METRICS_9_2 = {
+        'abs(pg_xlog_location_diff(pg_last_xlog_receive_location(), pg_last_xlog_replay_location())) AS replication_delay_bytes': ('postgres.replication_delay_bytes', GAUGE)
+    }
+
     REPLICATION_METRICS = {
         'descriptors': [],
-        'metrics': {
-            'GREATEST(0, EXTRACT(EPOCH FROM now() - pg_last_xact_replay_timestamp())) AS replication_delay': ('postgresql.replication_delay', GAUGE),
-        },
+        'metrics': {},
         'relation': False,
         'query': """
 SELECT %s
@@ -205,6 +226,7 @@ SELECT %s
         self.bgw_metrics = {}
         self.db_instance_metrics = []
         self.db_bgw_metrics = []
+        self.replication_metrics = {}
 
     def _get_version(self, key, db):
         if key not in self.versions:
@@ -239,7 +261,7 @@ SELECT %s
         """
         # Extended 9.2+ metrics if needed
         metrics = self.instance_metrics.get(key)
-        
+
         if metrics is None:
 
             # Hack to make sure that if we have multiple instances that connect to
@@ -254,7 +276,7 @@ SELECT %s
 
             self.db_instance_metrics.append(sub_key)
 
-            
+
             if self._is_9_2_or_above(key, db):
                 self.instance_metrics[key] = dict(self.COMMON_METRICS, **self.NEWER_92_METRICS)
             else:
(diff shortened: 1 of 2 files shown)


remh:
Thanks, looks like the pg tests are failing. Can you fix accordingly please ?

remh:
@DorianZaccaria can you rebase your branch as i can't merge it for now.
Thanks!

DorianZaccaria:
@remh Any updates for this one?

checks.d/postgres.py
```diff
@@ -285,9 +302,9 @@ def _collect_stats(self, key, db, instance_tags, relations, custom_metrics):
                 self.SIZE_METRICS
             ]
 
-        # Only available for >= 9.1 due to
-        # pg_last_xact_replay_timestamp
-        if self._is_9_1_or_above(key,db):
+        # Only available for >= 9.2 due to
+        # pg_xlog_location_diff
+        if self._is_9_2_or_above(key,db):
```
remh: Looks like you will break replication metrics for PG 9.1 users.

You should append the replication_delay_bytes metrics where PG >= 9.2 and keep replication_delay in case of 9.1

tests/test_postgresql.py
```diff
@@ -107,8 +119,6 @@ def test_checks(self):
         self.check.run()
         metrics = self.check.get_metrics()
 
-        self.assertEquals(len(metrics), exp_metrics, metrics)
```
remh: Why are you removing these tests ?

DorianZaccaria: I'm testing metrics individually.

checks.d/postgres.py
DorianZaccaria: I'll split this one.

tests/test_postgresql.py
remh: why not both

tests/test_postgresql.py
```diff
@@ -61,42 +72,54 @@ def test_checks(self):
...
+        self.assertTrue(len([m for m in metrics if m[0] == u'postgresql.table_size'])                           >= 1, pprint(metrics))
+        self.assertTrue(len([m for m in metrics if m[0] == u'postgresql.index_size'])                           >= 1, pprint(metrics))
+        self.assertTrue(len([m for m in metrics if m[0] == u'postgresql.total_size'])                           >= 1, pprint(metrics))
+        self.assertTrue(len([m for m in metrics if m[0] == u'postgresql.max_connections'])                      >= 1, pprint(metrics))
+        self.assertTrue(len([m for m in metrics if m[0] == u'postgresql.percent_usage_connections'])            >= 1, pprint(metrics))
+        self.assertTrue(len([m for m in metrics if m[0] == u'postgresql.total_tables'])                         >= 1, pprint(metrics))
+        self.assertTrue(len([m for m in metrics if m[0] == u'postgresql.db.count'])                             == 1, pprint(metrics))
+        self.assertTrue(len([m for m in metrics if m[0] == u'postgresql.total_tables'])                         == 1, pprint(metrics))
```
LeoCavaille: Looks like you are asserting twice this `total_tables` metric.

tests/test_postgresql.py
```diff
@@ -113,9 +136,13 @@ def test_checks(self):
         self.check.run()
         metrics = self.check.get_metrics()
 
+        self.assertEquals(len([m for m in metrics if 'table:persons' in str(m[3].get('tags', [])) ]), 11, metrics)
+
+        pg_version = os.getenv('PG_VERSION', '9.4.0')
```
LeoCavaille: Can you use `self.check._get_version` instead, it will test this function too.
Also enables someone to test against local version of PG without having to use the rake CI command.

checks.d/postgres.py
```diff
@@ -292,6 +314,23 @@ def _get_bgw_metrics(self, key, db):
...
+    def _get_replication_metrics(self, key, db):
+        """ Use either REPLICATION_METRICS_9_1 or REPLICATION_METRICS_9_1 + REPLICATION_METRICS_9_2
+        depending on the postgres version.
+        Uses a dictionnary to save the result for each instance
+        """
+        metrics = self.replication_metrics.get(key)
+        if metrics is None:
+            # Only available for >= 9.2 due to
+            # pg_xlog_location_diff
+            self.replication_metrics[key] = dict(self.REPLICATION_METRICS_9_1)
```
LeoCavaille: Could you put the `_is_9_1_or_above` test logic here and only keep the `_get_replication_metrics` in `_collect_stats` for more consistency?

LeoCavaille:
@DorianZaccaria looks good except the few nitpicks

checks.d/postgres.py
```diff
@@ -292,6 +314,18 @@ def _get_bgw_metrics(self, key, db):
...
+    def _get_replication_metrics(self, key, db):
+        """ Use either REPLICATION_METRICS_9_1 or REPLICATION_METRICS_9_1 + REPLICATION_METRICS_9_2
+        depending on the postgres version.
+        Uses a dictionnary to save the result for each instance
+        """
+        metrics = self.replication_metrics.get(key)
+        if self._is_9_1_or_above(key, db) and metrics is None:
+            self.replication_metrics[key] = dict(self.REPLICATION_METRICS_9_1)
+            if self._is_9_2_or_above(key, db):
+                self.replication_metrics.update(self.REPLICATION_METRICS_9_2)
```
remh: shouldn't it be:
```python
 self.replication_metrics[key].update(self.REPLICATION_METRICS_9_2)
```
? 

Also this seems a bit weird, as you're then overriding a constant (self.REPLICATION_METRICS), which in result will be applied to all instances. So your code won't work in the (weird) case where you have multiples instances of postgres running different versions.

I think _get_replication_metrics should just return the replication metrics (as the name of the method seems to indicate)

remh:
@DorianZaccaria i made one last comment but it should be good after this one!

Thanks a lot !

remh:
Looks great! Waiting for the tests to finish and will merge.

Thanks a lot @DorianZaccaria !

remh:
:fr: :cake: :+1: :boom:

DorianZaccaria:
Yay!

PR #1376, opened 2015-02-17
Monitor 'handled' in nginx

kky:
Spread between “connections opened” and “connections handled” indicates
resource saturation


alq666:
LGTM.

PR #1379, opened 2015-02-18
[etcd] service check & test

degemer:
* send a service check on success and on failure (previously only on failure)
* add a basic test copied from fluentd one


checks.d/etcd.py
```diff
@@ -112,4 +112,5 @@ def get_json(self, url, timeout):
                 tags = ["url:{}".format(url)])
             return None
 
+        self.service_check(self.SERVICE_CHECK_NAME, AgentCheck.OK, tags=["url:{}".format(url)])
```
remh: Watchout, you'll end up sending two statuses per iteration, one for get_self_metrics and one for get_store_metrics.

Can you ensure to only send one status per iteration per instance ?

LeoCavaille: You need to use `{0}` otherwise this is not python 2.6 comptible

ci/etcd.rb
```diff
@@ -0,0 +1,49 @@
...
+
+namespace :ci do
+  namespace :etcd do |flavor|
+    task :before_install => ['ci:common:before_install']
+
+    task :install => ['ci:common:install'] do
+      sh %(wget --quiet https://github.com/coreos/etcd/releases/download/v2.0.3/etcd-v2.0.3-linux-amd64.tar.gz)
+      sh %(tar -xzvf etcd-* > /dev/null)
+      sh %(rm etcd-*.tar.gz)
+      sh %(mv `ls | grep etcd` etcd)
```
LeoCavaille: Could you look at other ci files like apache, to use $VOLATILE_DIR as a dir to extract and download tarballs (otherwise it makes it a mess to run the tests locally).
Also set etcd to be installed in its own integration dir.
Can definitely help you do that if needed.

tests/test_etcd.py
```diff
@@ -0,0 +1,55 @@
+import unittest
+import logging
+from nose.plugins.attrib import attr
+logger = logging.getLogger(__file__)
+
+from tests.common import load_check
+
+@attr(requires='etcd')
+class TestEtcd(unittest.TestCase):
```
LeoCavaille: That would be amazing if you can re-use the new common class AgentCheckTest in `tests.common` (you can look at rabbitmq and riakcs). Again I can help there.

checks.d/etcd.py
degemer: @remh: indeed, fixed now.
@LeoCavaille: also fixed. (and thanks for the explanation)

degemer:
Finally send one and only one service_checks, thanks to @remh .
@LeoCavaille I tried to use `AgentCheckTest`, not sure it's correct. At least it's working locally!

ci/etcd.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    task :script => ['ci:common:script'] do
+      this_provides = [
+        'etcd'
+      ]
+      Rake::Task['ci:common:run_tests'].invoke(this_provides)
+    end
+
+    task :cleanup => ['ci:common:cleanup'] do
+      sh %(rm -rf $VOLATILE_DIR/etcd* || echo 'etcd not running')
```
LeoCavaille: You don't need to clean up `VOLATILE_DIR` it's already taken care of by the common cleanup task.
However, you might want to kill the etcd daemon you launched in the `before_script` stage, either by using a pidfile if you can have one or by using an appropriate kill command.

ci/etcd.rb
```diff
@@ -0,0 +1,51 @@
...
+      sh %(curl -s -L -o $VOLATILE_DIR/etcd.tar.gz\
+            https://github.com/coreos/etcd/releases/download/v2.0.3/etcd-v2.0.3-linux-amd64.tar.gz)
+      sh %(mkdir $VOLATILE_DIR/etcd)
+      sh %(tar xzvf $VOLATILE_DIR/etcd.tar.gz\
+                    -C $VOLATILE_DIR/etcd\
+                    --strip-components=1 >/dev/null)
+    end
+
+    task :before_script => ['ci:common:before_script'] do
+      sh %(cd $VOLATILE_DIR/etcd && ./etcd >/dev/null &)
```
LeoCavaille: You should install the binaries and integration in `embedded` and not `VOLATILE_DIR`, the idea is that `embedded` is cached and we don't reinstall the integration if it's already here. `VOLATILE_DIR` on the contrary is disposable (and useful for instance as a build area).

LeoCavaille:
LGTM :+1:

PR #1381, opened 2015-02-18
Fix source install requirements, fix #1380

### LINKED ISSUE #1380: source install fails without gcc
Installing on a barebones arch linux machine with python (python2 symlinked to python to make it work) and systat.

DD_API_KEY=<key> sh -c "$(curl -L https://raw.githubusercontent.com/DataDog/dd-agent/master/packaging/datadog-agent/source/setup_agent.sh)"

Install logs here:

http://pastebin.com/Mq1ntee8

The install should have failed on failing to compile pycrypto, but continued and timed out waiting on metrics. All the supervisor logs had simplejson import errors (sorry, deleted them already).

After wiping ~/.datadog-agent and installing gcc, the source install worked.

remh:
pysnmp and pysnmp-mibs require pycrypto which requires a compiler. So
the source install would fail on systems that don’t have a compiler.

### DIFF
--- a/source-optional-requirements.txt
+++ b/source-optional-requirements.txt
@@ -2,3 +2,5 @@ kazoo==1.3.1
 pycurl==7.19.5
 psutil==2.1.1
 pymongo==2.6.3
+pysnmp-mibs==0.1.4
+pysnmp==4.2.5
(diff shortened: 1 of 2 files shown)


LeoCavaille:
LGTM. Merging now to make it available for everyone using the source install.
thanks @clofresh for reporting this.

PR #1389, opened 2015-02-20
[varnish] fix `varnishadm` subp call with sudo

LeoCavaille:
When calling subprocess with an argv-like array, argv[0] needs to
be the program to execute and not a sequence of programs like
`sudo someprogram`.

Otherwise this results in:
`OSError: [Errno 2] No such file or directory` when trying to find
the `sudo someprogram` executable in the PATH.


remh:
looks good to me

PR #1393, opened 2015-02-25
[cassandra] Add some more StorageProxy metrics

miketheman:
This adds a few more node-wide metrics that complete the `StorageProxy` section of Cassandra's JMX metrics.
See here for more: http://wiki.apache.org/cassandra/JmxInterface#org.apache.cassandra.service.StorageProxy

### DIFF
--- a/conf.d/cassandra.yaml.example
+++ b/conf.d/cassandra.yaml.example
@@ -40,7 +40,9 @@ init_config:
           - MemtableSwitchCount
           - MinRowSize
           - PendingTasks
+          - RangeOperations
           - ReadCount
+          - ReadOperations
           - RecentHitRate
           - RecentRangeLatencyMicros
           - RecentReadLatencyMicros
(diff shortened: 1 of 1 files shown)


conf.d/cassandra.yaml.example
```diff
@@ -40,7 +40,9 @@ init_config:
           - MemtableSwitchCount
           - MinRowSize
           - PendingTasks
+          - RangeOpertaions
```
remh: Looks like a typo.
Also RangeOperation seems to be in the org.apache.cassandra.service domain, not org.apache.cassandra.db.

miketheman:
There's no evidence of a typo other than your comment, so you just have traveled back in time or I've changed history. :smiling_imp: 

These tasks show up in the db domain using `datadog-agent jmx list_everything`, including the other items we already collect, like RecentRangeLatencyMicros.

remh:
Looks good after the typo fix :ghost: .
Thanks!

Issue #1374, opened 2015-02-17
No way to force SNMPv1


slushpupie:
I have a device that only speaks SNMPv1, if you attempt SNMPv2 it simply ignores it.   pysnmp supports this by setting mpModel=0 with the community, but the agent does not have a way to set this.

remh:
Thanks a lot for the feedback @slushpupie ! 
We'll try to add this by 5.3.0 but would you like to open a pull request to implement this feature ?

PR #1397, opened 2015-02-26
Allow SNMP version to be specified

slushpupie:
When setting the community_string, you must specify mpModel=0 if
you need SNMPv1.

 * Add new option to snmp instance configuration:   snmp_version
 * Set up logic to "detect" SNMP version the same way as previous
   versions
 * Update snmp.yaml.example to show how to set snmp_version

### DIFF
--- a/checks.d/snmp.py
+++ b/checks.d/snmp.py
@@ -81,11 +81,24 @@ class SnmpCheck(AgentCheck):
         configuration.
         See http://pysnmp.sourceforge.net/docs/current/security-configuration.html
         '''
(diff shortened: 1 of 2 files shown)


slushpupie:
This is in relation to #1374

Note: I dont really have a good way to test this code fully, but it worked in my one instance.

remh:
Thanks a lot @slushpupie 
I made another version similar to yours but which is just a bit more concise.

https://github.com/DataDog/dd-agent/pull/1408

Thanks a lot for your feedback and your pull request!

PR #1400, opened 2015-02-26
[Redis] Add slow log metrics

remh:
Superseed #1330

Creates a histogram per redis command (HGET, LPUSH, etc) that is in the
slow query log.

### DIFF
--- a/checks.d/redisdb.py
+++ b/checks.d/redisdb.py
@@ -4,6 +4,7 @@ Redis checks
 # stdlib
 import re
 import time
+from collections import defaultdict
 
 # project
 from checks import AgentCheck
@@ -11,6 +12,7 @@ from checks import AgentCheck
 # 3rd party
 import redis
 
+DEFAULT_MAX_SLOW_ENTRIES = 128
 
 class Redis(AgentCheck):
     db_key_pattern = re.compile(r'^db\d+')
@@ -81,6 +83,7 @@ class Redis(AgentCheck):
     def __init__(self, name, init_config, agentConfig, instances=None):
         AgentCheck.__init__(self, name, init_config, agentConfig, instances)
         self.connections = {}
+        self.last_timestamp_seen = defaultdict(int)
 
     def get_library_versions(self):
         return {"redis": redis.__version__}
@@ -127,8 +130,7 @@ class Redis(AgentCheck):
 
         return self.connections[key]
 
-    def _check_db(self, instance, custom_tags=None):
-        conn = self._get_conn(instance)
+    def _get_tags(self, custom_tags, instance):
         tags = set(custom_tags or [])
 
         if 'unix_socket_path' in instance:
@@ -140,6 +142,14 @@ class Redis(AgentCheck):
             tags_to_add.append("db:%s" % instance.get('db'))
 
         tags = sorted(tags.union(tags_to_add))
+
+        return tags, tags_to_add
+
+    def _check_db(self, instance, custom_tags=None):
+        conn = self._get_conn(instance)
+
+        tags, tags_to_add = self._get_tags(custom_tags, instance)
+        
 
         # Ping the database for info, and track the latency.
         # Process the service check: the check passes if we can connect to Redis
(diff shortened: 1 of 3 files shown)


checks.d/redisdb.py
```diff
@@ -233,9 +243,64 @@ def _check_db(self, instance, custom_tags=None):
...
+        #   'duration': 11238,
+        #   'id': 496L,
+        #   'start_time': 1422529869}
+        for slowlog in slowlogs:
+            if slowlog['start_time'] > max_ts:
+                max_ts = slowlog['start_time']
+
+            command_tag = 'command:{0}'.format(slowlog['command'].split()[0])
+            value = slowlog['duration']
+            self.histogram('redis.slowlog.micros', value, tags=tags + [command_tag])
```
conorbranagan: do you think it makes sense to convert to something other than microseconds (like milliseconds or seconds) since we don't really use that for any other metrics? If not, we might want to spell out "microseconds" or something. Maybe just me but I didn't know what "micros" meant without looking at the docs for the redis slowlog.

conorbranagan:
Cool metrics, interested to see what sort of data we can get from this on prod. Do you know if the commands in transactions will show up or will it just show up as `exec`?

checks.d/redisdb.py
conorbranagan: on second thought, a conversion might be a bad idea because it would make the axis all wonky

checks.d/redisdb.py
```diff
@@ -233,9 +243,64 @@ def _check_db(self, instance, custom_tags=None):
...
+        conn = self._get_conn(instance)
+
+        tags, tags_to_add = self._get_tags(custom_tags, instance)
+
+        if not instance.get(max_slow_entries):
+            max_slow_entries = int(conn.config_get(max_slow_entries)[max_slow_entries])
+            if max_slow_entries > DEFAULT_MAX_SLOW_ENTRIES:
+                self.warning("Redis {0} is higher than {1}. Defaulting to {1}."\
+                    "If you need a higher value, please set {0} in your check config"\
+                    .format(max_slow_entries, DEFAULT_MAX_SLOW_ENTRIES))
```
LeoCavaille: This will result in having a string like

> Redis 256 is higher than 128. Defaulting to 128. If you need a higher value please set 256 in your check config

Which is not super understandable?

conf.d/redisdb.yaml.example
```diff
@@ -30,3 +30,11 @@ instances:
     # Default: True
     #
     # warn_on_missing_keys: True
+
+    # Max number of entries to fetch from the sloq query log
+    # By default, the check wil read this value from the redis config
+    # But if it's too high, it will default to 128
+    # If you need to get more entries from the slow query logs
+    # set the value here.
+    # Warning: It maybe impact the performances of your redis instance
```
LeoCavaille: typo: It may impact ?

conf.d/redisdb.yaml.example
```diff
@@ -30,3 +30,11 @@ instances:
     # Default: True
     #
     # warn_on_missing_keys: True
+
+    # Max number of entries to fetch from the sloq query log
```
LeoCavaille: typo: slow

conf.d/redisdb.yaml.example
```diff
@@ -30,3 +30,11 @@ instances:
...
+    # By default, the check wil read this value from the redis config
```
LeoCavaille: typo: will

tests/test_redis.py
```diff
@@ -171,6 +174,32 @@ def test_redis_repl(self):
...
+            db.lpush(test_key, random.random())
+
+        db.sort(test_key)
+
+        self.assertTrue(db.slowlog_len() > 0)
+
+        self.run_check({"init_config": {}, "instances": [instance]})
+
+        assert self.metrics, "No metrics returned"
+        self.assertMetric("redis.slowlog.micros.max")
```
LeoCavaille: Can we assert the tag here too?

LeoCavaille:
Yep @conorbranagan I think only individual commands go in the slow log, even if they're grouped in a transaction.

checks.d/redisdb.py
miketheman: Agreed - the wording here should help the user understand what it is we're trying to tell them.

checks.d/redisdb.py
miketheman: I think `micros` is good enough - or even `in_micros` which us better, IMHO, than `μs` which would likely become `us` anyway. We have used micros for a long time with cassandra metrics too.

conf.d/redisdb.yaml.example
```diff
@@ -30,3 +30,11 @@ instances:
...
+    # But if it's too high, it will default to 128
```
miketheman: "too high" should be spelled out, why is too high too high? "Higher than 128, due to potential increased latency to retrieve more than 128 slowlog entries every 15 seconds"...

conf.d/redisdb.yaml.example
miketheman: performances => performance

checks.d/redisdb.py
remh: Arf i reused the same variable name, max_slow_entries was supposed to be "slowlog-max-len" good catch!

remh:
Thanks for the comments, i took care of them 6f52aeefb5990387402f7e7f8d8deb1fad39dfd1

PR #1409, opened 2015-03-03
[core] Use Timer instead of signal for Watchdog

### LINKED ISSUE #1407: [watchdog] Properly handle killing the process
See #1406

degemer:
Fix #1407 
Replaces #1406 
Send first a SIGTERM, and then a SIGKILL 10s later if nothing happened

### DIFF
--- a/util.py
+++ b/util.py
@@ -12,6 +12,7 @@ import uuid
 import tempfile
 import re
 import simplejson as json
+import threading
 import logging
 from hashlib import md5
 
@@ -398,18 +399,17 @@ class EC2(object):
 
 
 class Watchdog(object):
-    """Simple signal-based watchdog that will scuttle the current process
-    if it has not been reset every N seconds, or if the processes exceeds
-    a specified memory threshold.
+    """Simple threading.Timer-based watchdog that will scuttle the current
+    process if it has not been reset every N seconds, or if the processes
+    exceeds a specified memory threshold.
     Can only be invoked once per process, so don't use with multiple threads.
-    If you instantiate more than one, you're also asking for trouble.
     """
     def __init__(self, duration, max_mem_mb = None):
         import resource
 
         #Set the duration
         self._duration = int(duration)
-        signal.signal(signal.SIGALRM, Watchdog.self_destruct)
+        self._timer = threading.Timer(self._duration, Watchdog.self_destruct)
 
         # cap memory usage
         if max_mem_mb is not None:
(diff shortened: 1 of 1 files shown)


util.py
```diff
@@ -409,7 +410,7 @@ def __init__(self, duration, max_mem_mb = None):
 
         #Set the duration
         self._duration = int(duration)
-        signal.signal(signal.SIGALRM, Watchdog.self_destruct)
+        self._timer = threading.Timer(self._duration, Watchdog.self_destruct)
```
LeoCavaille: I might be confused, but if the collector goes crazy during the very first run, the timer is not activated because `.reset()` is only called at the end of the loop and the timer hasn't been started yet.

degemer: Actually, `.reset` is called when the `Watchdog` is created: https://github.com/DataDog/dd-agent/blob/d10e5f621186085a22f51679caf92906a2c1fd14/agent.py#L172

LeoCavaille: Cool thanks, didn't see that. :smile_cat:

LeoCavaille:
I like the new `Timer` logic. It looks better than the hacky thing we had before. Thanks @degemer

degemer:
I updated the description of `Watchdog`.
I also found this, which kind of does the opposite (but cleans a lot of other things by doing it): https://github.com/DataDog/dd-agent/commit/d18020d82d4fbfad3285492e9660ac08d3ced024

LeoCavaille:
Time to ask @alq666 then what was this commit about.

alq666:
Philosophically I would say that going the `Timer` way runs counter to the whole purpose of a watchdog.
 
A `Timer` assumes that the python runtime is still operating properly, which is the opposite of some of the conditions when the watchdog should trigger. A `Timer` will catch a particular check stuck in a blocking call but won't work if another thread has the GIL for an undue amount of time (e.g. during memory allocation in tight systems). CPython threads are pthreads, but they still need the GIL to do anything (this [presentation](http://www.dabeaz.com/python/GIL.pdf) is pretty insightful).

What I like about the signal-based approach is that it will kill the process regardless of how the collector behaves or misbehaves. It has to be effective even when everything else is failing.

As for what that commit was about, I can't recall the exact details but I would not be surprised if it were after we had witnessed the pre-signal watchdog fail.

LeoCavaille:
nice write up @alq666 thanks for the explanations, GIL </3 Looking forward to having something closer to the metal for the next agent. @degemer that makes sense to keep the good'ol POSIX signals then.

degemer:
Thanks @alq !
Yes, it totally makes sense to keep the signal-based Watchdog.
But maybe we could add a Timer-based Watchdog which would only send SIGTERM 10s before the signal-based one ? This way the agent could still exit properly in most cases, and otherwise be killed 10s later.

The only thing done during the cleanup before exiting is stopping JMX, I don't know if it's really important anyway.
The real fix for the pidfile bug is now #1418 anyway.

remh:
Closing this PR then.

PR #1410, opened 2015-03-03
[jmx] default heap size options

### LINKED ISSUE #1372: JMXFetch should have default values for its start heap size and max heap size
It can be manually configured using the java_options parameter 
(e.g.: https://gist.github.com/remh/9a45436d43f67f0dbdc7#file-cassandra-yaml-L6 )

but doesn't have any default values.

yannmh:
Fixes #1372

### DIFF
--- a/jmxfetch.py
+++ b/jmxfetch.py
@@ -26,6 +26,8 @@ JAVA_LOGGING_LEVEL = {
 }
 
 JMX_FETCH_JAR_NAME = "jmxfetch-0.4.1-jar-with-dependencies.jar"
+_JVM_DEFAULT_MAX_MEMORY_ALLOCATION = " -Xmx200m"
+_JVM_DEFAULT_INITIAL_MEMORY_ALLOCATION = " -Xms50m"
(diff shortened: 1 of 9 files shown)


jmxfetch.py
```diff
@@ -347,14 +349,19 @@ def start(cls, confd_path, agentConfig, logging_config, path_to_java, java_run_o
...
-
             subprocess_args.insert(4, '--check')
             for check in jmx_checks:
                 subprocess_args.insert(5, check)
 
-            if java_run_opts:
-                for opt in java_run_opts.split():
-                    subprocess_args.insert(1,opt)
+            # Specify a maximum memory allocation pool for the JVM
+            if "Xmx" not in java_run_opts:
```
remh: Watchout this won't work in the (rare) case where people are using 
```java
XX:MaxHeapSize
```
and 
```java
XX:InitialHeapSize
```

to set these values

remh:
Looks good ! Can you add a test ? :)

remh:
Also can you add the java_options parameter in the yaml.example files for jmx checks please ?

yannmh:
Done. Thanks for your feedback @remh

PR #1413, opened 2015-03-06
[marathon] add basic auth option

jkoppe:
I like to protect our marathon instances behind basic authentication, so I needed this configuration option.

Launch marathon with --http_credentials "user:password" to require basic authentication, then configure your marathon.yaml with the user and password.

I tested this pull request with and without the user/password fields in marathon.yaml -- both seem to work fine.

### DIFF
--- a/checks.d/marathon.py
+++ b/checks.d/marathon.py
@@ -30,11 +30,17 @@ class Marathon(AgentCheck):
 
         # Load values from the instance config
         url = instance['url']
+        user = instance.get('user')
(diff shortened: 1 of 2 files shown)


alq666:
Looks totally reasonable.

yannmh:
Thanks a lot @jkoppe.

We'll review your PR for the 5.3.0 release !

LeoCavaille:
LGTM, thanks again for contributing.

PR #1415, opened 2015-03-07
[postgres] Collect StatIO per table & revamp test

alq666:
Based on @adriandoolittle's work, but restricted to relations that are listed in the configuration to avoid blowing up the number of metrics.


alq666:
@remh I still have to update the test for it.

checks.d/postgres.py
```diff
@@ -177,7 +177,7 @@ class PostgreSql(AgentCheck):
             ('schemaname', 'schema')
         ],
         'metrics': {
-            'pg_stat_user_tables': ('postgresql.total_tables', GAUGE),
+            'pg_stat_user_tables': ('postgresql.table.count', GAUGE),
```
miketheman: This might break existing user's screens, alerts, etc. Renaming a metric that's out there could be considered a breaking change.

alq666: Agreed. I checked and no such screens and monitors exist right now. We can recheck before we release.

miketheman: :+1:

LeoCavaille: @alq666 @miketheman should have added a note.
That's normal b/c this is a new metric post 5.2 so it hasn't been released to the crowds yet. That's why I had no mercy renaming it.
See : bee8eec12d69d0936e87378e0939be2fe030eaa7

miketheman: Notes are good! :) Thanks for the explain @LeoCavaille .

tests/test_postgres.py
```diff
@@ -0,0 +1,222 @@
...
+                'host': host,
+                'port': port,
+                'username': 'datadog',
+                'password': 'datadog',
+                'dbname': 'dogs',
+                'relations': ['breed', 'kennel']
+            }
+        ]
+
+        self.run_check(dict(instances=instances))
```
remh: Let's use run_check_twice instead ?

checks.d/postgres.py
```diff
@@ -334,22 +354,31 @@ def _collect_stats(self, key, db, instance_tags, relations, custom_metrics):
...
-            self.COUNT_METRICS
+            self.COUNT_METRICS,
         ]
 
+        # These are added only once per PG server, thus the test
+        db_instance_metrics = self._get_instance_metrics(key, db)
+        bgw_instance_metrics = self._get_bgw_metrics(key, db)
+
+        if db_instance_metrics is not None:
+            self.DB_METRICS['metrics'] = db_instance_metrics
```
remh: It makes me sad to see what should be a constant being muted :( we should clean that in a next version of this check

tests/test_postgres.py
```diff
@@ -0,0 +1,222 @@
...
+            'postgresql.bgwriter.checkpoints_requested',
+            'postgresql.bgwriter.buffers_checkpoint',
+            'postgresql.bgwriter.buffers_clean',
+            'postgresql.bgwriter.maxwritten_clean',
+            'postgresql.bgwriter.buffers_backend',
+            'postgresql.bgwriter.buffers_alloc',
+        ]
+
+        for mname in COMMON_BGW_METRICS:
+            self.assertMetric(mname, count=1)
```
remh: we can't assess the tags here ?

tests/test_postgres.py
```diff
@@ -0,0 +1,222 @@
...
+            count=1, status=AgentCheck.OK,
+            tags=['host:localhost', 'port:15432', 'db:datadog_test']
+        )
+        self.assertServiceCheck('postgres.can_connect',
+            count=1, status=AgentCheck.OK,
+            tags=['host:localhost', 'port:15432', 'db:dogs']
+        )
+
+        self.coverage_report()
+        return
```
remh: useful ?

remh:
Looks great besides the tiny nitpicks. It was definitely not an easy refactoring.

tests/test_postgres.py
LeoCavaille: aha nice, of course. That didn't even existed when I submitted the PR :rewind:

tests/test_postgres.py
LeoCavaille: There are no tags actually, e.g.
```
 ('postgresql.bgwriter.buffers_checkpoint',
  1428435600,
  0,
  {'hostname': 'Leos-MacBook-Pro.local', 'type': 'count'}),
```

LeoCavaille:
CI passed, implemented your different comments @remh thanks. Merging.

PR #1416, opened 2015-03-09
[core] Cast system settings proxy port to int

### LINKED ISSUE #1414: System Proxy Settings return a string instead of an integer for port
When we pull system proxy settings from urllib we end up with this error:

```
  File "transaction.pyc", line 190, in flush_next
  File "ddagent.pyc", line 272, in flush
  File "tornado\httpclient.pyc", line 206, in fetch
  File "tornado\curl_httpclient.pyc", line 95, in fetch_impl
  File "tornado\curl_httpclient.pyc", line 233, in _process_queue
  File "tornado\curl_httpclient.pyc", line 358, in _curl_setup_request
TypeError: strings are not supported for this option
```

this is because the proxy port is being set to a string, which tornado doesn't accept:

https://github.com/DataDog/dd-agent/blob/master/config.py#L606-L608

yannmh:
Fixes #1414

From http://tornado.readthedocs.org/en/latest/httpclient.html:
```
`proxy_host` (string) – HTTP proxy hostname. To use proxies, proxy_host and proxy_port must be set; proxy_username and proxy_pass are optional. Proxies are currently only supported with curl_httpclient.
`proxy_port` (int) – HTTP proxy port
```

### DIFF
--- a/config.py
+++ b/config.py
@@ -605,7 +605,7 @@ def get_proxy(agentConfig, use_system_settings=False):
                 pass
             px = proxy.split(':')
             proxy_settings['host'] = px[0]
-            proxy_settings['port'] = px[1]
+            proxy_settings['port'] = int(px[1])
             proxy_settings['user'] = None
             proxy_settings['password'] = None
             proxy_settings['system_settings'] = True


LeoCavaille:
:+1: Nice, waiting for the CI to complete but LGTM

PR #1417, opened 2015-03-09
[snmp] add timeout and retries options in config

### LINKED ISSUE #1309: [snmp] Add timeout parameter to conf
I'd like an option to change the timeout (current default of 1 second) in the conf versus modifying the script direct:

https://github.com/DataDog/dd-agent/blob/06a4839b5fdf00da44e826b050b266571bb38fe8/checks.d/snmp.py#L38

degemer:
Fix #1309

### DIFF
--- a/checks.d/snmp.py
+++ b/checks.d/snmp.py
@@ -42,8 +42,8 @@ class SnmpCheck(AgentCheck):
 
     cmd_generator = None
     # pysnmp default values
-    RETRIES = 5
-    TIMEOUT = 1
+    DEFAULT_RETRIES = 5
+    DEFAULT_TIMEOUT = 1
 
     def __init__(self, name, init_config, agentConfig, instances=None):
         AgentCheck.__init__(self, name, init_config, agentConfig, instances)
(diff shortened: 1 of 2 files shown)


checks.d/snmp.py
```diff
@@ -238,6 +238,8 @@ def check(self, instance):
         ip_address = instance["ip_address"]
         table_oids = []
         raw_oids = []
+        timeout = instance.get('timeout') or self.DEFAULT_TIMEOUT
+        retries = instance.get('retries') or self.DEFAULT_RETRIES
```
yannmh: you can simply do 
```python
retries = instance.get('retries', self.DEFAULT_RETRIES)
```

degemer: Totally forgot this syntax, thanks @yannmh !

checks.d/snmp.py
```diff
@@ -238,6 +238,8 @@ def check(self, instance):
         ip_address = instance["ip_address"]
         table_oids = []
         raw_oids = []
+        timeout = instance.get('timeout', self.DEFAULT_TIMEOUT)
+        retries = instance.get('retries', self.DEFAULT_RETRIES)
```
LeoCavaille: It's unclear from what I've just read of the pysnmp doc (and the code is horrible to read) if a string value is OK here, so it might be a good idea to cast these values as integers?

LeoCavaille: And not your code but same here, https://github.com/DataDog/dd-agent/pull/1417/files#diff-b7a49ef3eaa09d1278637a175c5642cbR127 ?

degemer: Sure, updated all three. Thanks!

LeoCavaille:
Sweet thanks! :+1:

PR #1418, opened 2015-03-11
[core] improve detection of already running agent at startup

degemer:
Solve all problems related to the agent receiving a SIGKILL and not cleaning its pidfile.
Check if the pid in the pidfile is valid, if it is, abort, otherwise ignore and continue its start.
With this change, even if the collector (or any other agent process) receives a SIGKILL, it will just be restarted by supervisor.

So the question is: do we want to have a _Phoenix_ agent which will always arise from its ashes (but may hide any error causing its death), or an agent which need a manual restart after a *critical* error ?

----
It was the default behaviour before 5.2 because processes started by supervisor didn't have a pidfile.

### DIFF
--- a/daemon.py
+++ b/daemon.py
@@ -19,6 +19,9 @@ import logging
 import errno
 import signal
 
+# 3p
+from psutil import pid_exists
+
 log = logging.getLogger(__name__)
 
 class AgentSupervisor(object):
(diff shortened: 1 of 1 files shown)


remh:
LGTM, failing on a TCP check and couchdb.
Thanks @degemer !

PR #1424, opened 2015-03-13
[kafka] upgraded jmx keys for version 0.8.2

patrickbcullen:
This changes upgrades the jmx keys for kafka 0.8.2. This is not backwards compatible with kafka 0.8.1.

### DIFF
--- a/conf.d/kafka.yaml.example
+++ b/conf.d/kafka.yaml.example
@@ -26,47 +26,54 @@ init_config:
         # Aggregate cluster stats
         #
         - include:
-            domain: '"kafka.server"'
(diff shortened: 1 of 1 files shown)


LeoCavaille:
Hey @patrickbcullen thanks for contributing!
We'll have a look at how we can handle that, with the fact that it's not backwards compatible... and because it's a JMX check there is no proper way right now to hack a version check in it.
Also side note: there seems to be some sort of conflict preventing a direct merge.

LeoCavaille:
Thanks again @patrickbcullen I rebased your PR in #1438 adding a warning in the sample file and resolving the conflict. So closing this one and will merge the other.

PR #1336, opened 2015-01-30
Support tags for process check

amalakar:

PR #1426, opened 2015-03-13
[process] tags per instance + test suite refactoring

LeoCavaille:
Supersedes #1336.
Refactoring the test suite to use the new common `AgentCheckTest` interface.
+ some improvements in that class, mainly coverage reports.

### DIFF
--- a/tests/test_process.py
+++ b/tests/test_process.py
@@ -1,124 +1,172 @@
 import unittest
+import os
 
 from checks import AgentCheck
-from util import get_hostname
-from tests.common import load_check
+from tests.common import AgentCheckTest, load_check
 from nose.plugins.attrib import attr
 
 @attr('process')
-class ProcessTestCase(unittest.TestCase):
+class ProcessTestCase(AgentCheckTest):
+    CHECK_NAME = 'process'
 
-    offset = 0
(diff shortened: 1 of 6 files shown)


LeoCavaille:
This one should be good to merge, the broken jobs on the CI are related to jmxfetch (it's broken as well on master).
@remh can you have a look?

tests/test_process.py
```diff
@@ -1,124 +1,172 @@
...
+            # Assert metrics
+            for mname in self.PROCESS_METRIC:
+                proc_name = stub['config']['name']
+                expected_tags = [proc_name, "process_name:{0}".format(proc_name)]
 
-        config = self.build_config(config)
-        self.check.find_pids = self.find_pids
+                # If a list of tags is already there, the check extends it
+                if 'tags' in stub['config']:
+                    expected_tags = stub['config']['tags']
```
remh: Shouldn't you extend your list here instead of replacing it ?

LeoCavaille: That's a tricky one, but the answer is no because the checks extends itself the instance tags list if it exists. So here it will contain **ALL the tags**...

remh:
Looks good! Thanks!

PR #1430, opened 2015-03-15
More agressive NTP threshold by default

conorbranagan:
A host reporting more than 2-3 minutes off NTP time will throw off our alerting engine. We should set the default to 1 minute so when people enable the check they get the best results.

We should also make the change here: https://github.com/DataDog/puppet-datadog-agent/pull/73

### DIFF
--- a/checks.d/ntp.py
+++ b/checks.d/ntp.py
@@ -7,7 +7,7 @@ from checks import AgentCheck
 # 3rd party
 import ntplib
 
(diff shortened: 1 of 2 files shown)


remh:
It should also be updated here https://github.com/DataDog/dd-agent/blob/master/checks.d/ntp.py#L10

remh:
Closing in favor of #1433

PR #1433, opened 2015-03-16
Enable ntp check by default

### LINKED ISSUE #620: Stop packaging network.yaml
https://github.com/DataDog/dd-agent/pull/615 added the ability to configure excluded interfaces in `network.yaml`. Unfortunately, we package this file right now (instead of including `network.yaml.example` like we do for other checks), and we don't consider it a configuration file, so when a user updates this file will be overwritten. We should:

1. Define default behaviour for the network check and have it run regardless of whether `network.yaml` is present. If `network.yaml` is present, the file will override the check's default behaviour.
2. Stop packaging `network.yaml` and start packaging an example configuration in `network.yaml.example`.

remh:
Also:

* Add a better way to have checks enabled by default
* Fix #620
* Ensure that we cast exception as string (otherwise pickle can fail
when persisting the collector status)

Supersedes #1430


LeoCavaille:
Looks good to me, will merge once it passed the CI.

miketheman:
Hi @remh !
(Looking at https://github.com/DataDog/chef-datadog/pull/182) Curious how a file placed on disk with a similar name like `conf.d/ntp.yaml` with overrides would affect this new behavior - can you comment on how this should work?

remh:
Hey @miketheman !

If a .yaml file is present, the agent will use this one.
If a .yaml file is not present, the agent will load the .yaml.default 

Logic is here:
https://github.com/DataDog/dd-agent/blob/c7cb52d88d7f5587f0afd6f2a4d28c871f86d6aa/config.py#L785-L807

Does that make sense ?

remh:
So basically this PR https://github.com/DataDog/chef-datadog/pull/182 shouldn't cause any trouble.

miketheman:
@remh Thanks!

Issue #941, opened 2014-05-13
[core] processes run in foreground do not have PID files


celenechang:
collector status command seems to return an incorrect message. reported from a user, also tested locally:

$ dd-agent status
Log file is unwritable: '/var/log/datadog/collector.log'
2014-05-13 16:38:36,284 | DEBUG | dd.collector | config(config.py:523) | No proxy configured
2014-05-13 16:38:36,284 | DEBUG | dd.collector | util(util.py:387) | Using temporary pid file: /tmp/dd-agent.pid
2014-05-13 16:38:36,284 | INFO | dd.collector | daemon(daemon.py:250) | Agent is not running
Agent is not running

$ sudo supervisorctl -c /etc/dd-agent/supervisor.conf status
datadog-agent:collector          RUNNING    pid 6279, uptime 3 days, 21:34:41
datadog-agent:dogstatsd          RUNNING    pid 6277, uptime 3 days, 21:34:41
datadog-agent:forwarder          RUNNING    pid 6278, uptime 3 days, 21:34:41
datadog-agent:pup                RUNNING    pid 6280, uptime 3 days, 21:34:41

LeoCavaille:
This is caused by how we run these daemons through supervisor.
We ask supervisor to launch the collector or dogstatsd in foreground mode therefore bypassing the `daemonizing` methods that handle pid files.
As a consequence there are no pid files when running the agent with supervisor (and `status` will return not running), should you use `dd-agent start` instead it would return the appropriate status.

The solution would be to handle the pidfile directly in `run()` or use the same `start` interface of the daemon adding an extra flag `foreground` for instance.

PR #1435, opened 2015-03-16
[core] Fix pidfile bug for 5.2.2

degemer:
And thus add the ability to be restarted properly after a SIGKILL (since
the pid in the pidfile is now tested)

### DIFF
--- a/daemon.py
+++ b/daemon.py
@@ -19,6 +19,9 @@ import logging
 import errno
 import signal
 
+# 3p
+from psutil import pid_exists
+
 log = logging.getLogger(__name__)
 
 class AgentSupervisor(object):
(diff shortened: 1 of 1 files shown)


daemon.py
```diff
@@ -158,10 +158,23 @@ def start(self, foreground=False):
...
-            sys.stderr.write(message % self.pidfile)
-            sys.exit(1)
+            # Check if the pid in the pidfile corresponds to a running process
+            try:
+                os.kill(pid, 0)
+            except OSError, e:
+                if e.errno != errno.EPERM:
+                    log.warn('pidfile contains the pid of a not running process.'\
+                              ' Starting normally')
+                    exit_code = None
```
remh: Shouldn't it be 0 ?

daemon.py
```diff
@@ -158,10 +158,23 @@ def start(self, foreground=False):
         pid = self.pid()
 
         if pid:
-            message = "pidfile %s already exists. Is it already running?\n"
-            log.error(message % self.pidfile)
-            sys.stderr.write(message % self.pidfile)
-            sys.exit(1)
+            # Check if the pid in the pidfile corresponds to a running process
```
remh: let's use psutil.pid_exists, it's moe concise, easier to read and works on Windows (even though this part of the code is not run on Windows) http://pythonhosted.org/psutil/#psutil.pid_exists

jkoppe:
Now that datadog-agent runs under supervisord, is it necessary to write pid files like this?

alq666:
@remh is this for CentOS5 that we still carry pid files?

remh:
This pull request is to fix a bug that was introduced with 5.2. 

But you are right a lot of this logic is not needed anymore, we have a card in our board to clean this but it won't be done before 5.4.0.

degemer:
Pidfiles for processes spawned by supervisor were introduced to fix #941 (so that old commands `dd-agent`, `dogstatsd` and `dd-forwarder` still display a correct `status`).

jkoppe:
Gotcha.  This looks like an adequate, simple improvement until the day comes when this code can be ripped out.  Thanks!

PR #1274, opened 2015-01-02
Custom MySQL Metrics

obi11235:
Allow users to create custom MySQL metrics for their applications using a query. By including it in the standard MySQL check it will reuse the connection the datadog agent is already making.

PR #1436, opened 2015-03-16
Added mysql client conn info

jstelzer:
We'd like to be able to alert on a ratio of the number of current connections over max conns.

The existing data is for connection rate. This change will allow us to alert when clients hit %X of max_connections.

This PR is mostly a proof of concept. I've implemented a simple custom check using the checks.d interface, but if these stats can land in the mainline repo, then that's one less thing to manage.


remh:
Thanks a lot @jstelzer ! 

We'll review this for our 5.4.0 release.
Thanks again!

yannmh:
Thanks @jstelzer.

I think having a `Custom metric` section in MySQL check, like suggested in #1274, would make more sense: it's a more generic and flexible solution to answer this type of requests.

What do you think @garnermccloud ?

yannmh:
Closing in favor of #1274.

PR #1440, opened 2015-03-16
[windows] add a basic "Watchdog"

degemer:
- check memory consumption of all processes
- check whether the agent is blocked on a check or not

### DIFF
--- a/win32/agent.py
+++ b/win32/agent.py
@@ -1,31 +1,40 @@
-# set up logging before importing any other components
-
-import win32serviceutil
-import win32service
+# stdlib
+import logging
+import modules
+import multiprocessing
+from optparse import Values
+import servicemanager
+import sys
+import threading
+import time
+import tornado.httpclient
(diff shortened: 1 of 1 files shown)


degemer:
- Remove memory-based Watchdog as the previous experience with a memory-based Watchdog on Linux platforms was inconclusive.
- Remove automatic restart of the agent every day. It was introduced as a workaround of a memory leak on WS2003, and we are dropping support for WS2003.

win32/agent.py
```diff
@@ -1,31 +1,41 @@
 # set up logging before importing any other components
+from config import get_version, initialize_logging; initialize_logging('collector')
```
remh: So I guess we should remove this line @degemer ?

degemer: Yes, forgot it. Thanks @remh !

remh:
Thanks @degemer !

PR #630, opened 2013-08-23
Add check for php-fpm status

arosenhagen:
Add AgentCheck for PHP-FPM Status similar to nginx status

PR #1441, opened 2015-03-17
[php-fpm] new check, get perf insight of your php-fpm install

LeoCavaille:
Changelog:

## Add first version of check by @arosenhagen
This check targets an install of php-fpm, through the webserver
we query the status and ping paths of FPM, and we transform the
output of these commands in metrics and service checks.

[leo@datadoghq.com] Rebased, fixing up commits together and solved
conflicts.

## Revamp the PHP FPM integration
* Consolidated in one check ping + status queries
* Removed a lot of unused vars, simplified functions, make it look
  prettier
* Set up integration testing on Travis by:
   * installing a fresh nginx as a fast CGI passthru
   * installing a PHP version from source which ships the php-fpm
     binary
   * making them play together accepting connections to the
     /status and /ping paths

Supersedes #630 
Please review @LotharSee @remh


checks.d/php_fpm.py
```diff
@@ -0,0 +1,120 @@
...
+        'active processes': 'php_fpm.active_processes',
+        'total processes': 'php_fpm.total_processes',
+        'max active processes': 'php_fpm.max_active_processes',
+        'max children reached': 'php_fpm.max_children_reached'
+    }
+    COUNTERS = {
+        'accepted conn': 'php_fpm.accepted_conn',
+        'slow requests': 'php_fpm.slow_requests'
+    }
+
```
LotharSee: A few suggestions about these names:

* Group some of them (`processes.active`, `processes.total`, `processes.max_active`, ...)
* Make some of them more clear (what are listen_queue and listen_queue_len? Again, group them `listen_queue.something`)
* We could call prefix that `fpm.` instead of `php_fpm`
* Use service check names close to what we already have. (`is_connect` is common but maybe not relevant in the case, `is_running` sounds better and that's what we have with Gunicorn)

checks.d/php_fpm.py
```diff
@@ -0,0 +1,120 @@
...
+            try:
+                pool = self._process_status(status_url, auth, tags)
+            except Exception as e:
+                status_exception = e
+                pass
+
+        if ping_url is not None:
+            self._process_ping(ping_url, auth, tags, pool)
+
+        if status_exception is not None:
```
LotharSee: What is the reason to catch then re-raise it?

What about running `_process_ping` first, then `_process_status` without catching it? This way you don't lose the initial context in the exception and that's simpler.

checks.d/php_fpm.py
```diff
@@ -0,0 +1,120 @@
...
+            resp.raise_for_status()
+
+            data = resp.json()
+        except Exception as e:
+            self.log.error("Failed to get metrics from {0}.\nError {1}".format(status_url, e))
+            raise
+
+        pool_name = data.get('pool', 'default')
+        metric_tags = tags[:]
+        metric_tags.append("pool:{0}".format(pool_name))
```
LotharSee: `metric_tags = tags + ["pool:{0}".format(pool_name)]` is smaller and looks simpler to me. What do you think?

checks.d/php_fpm.py
```diff
@@ -0,0 +1,120 @@
...
+            # TODO: adding the 'full' parameter gets you per-process detailed
+            # informations, which could be nice to parse and output as metrics
+            resp = requests.get(status_url, auth=auth,
+                                headers=headers(self.agentConfig),
+                                params={'json': True})
+            resp.raise_for_status()
+
+            data = resp.json()
+        except Exception as e:
+            self.log.error("Failed to get metrics from {0}.\nError {1}".format(status_url, e))
```
LotharSee: We rarely put a `\n` in a `log()`, does it render well in the logfile?

ci/resources/phpfpm/php-fpm.conf
```diff
@@ -0,0 +1,537 @@
+;;;;;;;;;;;;;;;;;;;;;
```
LotharSee: What about putting the interesting lines? All these comments aren't useful (we could just keep a link to the source) and they make it harder to read and look for the interesting stuff.

checks.d/php_fpm.py
LeoCavaille: I should have shared that in a comment here, we discussed this with @remh .
The problem of running `ping` first is that we cannot use the pool name parsed from `status` to tag the service check which is :disappointed:
And at the same time we don't want `ping` to be skipped because `status` raised, they're indeed two different endpoints set up in php fpm.

checks.d/php_fpm.py
LeoCavaille: We do lots of them, `ag 'log.*\(.*\\n' ` you would be surprised :smile_cat: It's not ideal but it does the trick. Tracebacks usually appear like that too (after a newline).

checks.d/php_fpm.py
LeoCavaille: * good idea
* will try that, after re-reading the doc
* I am not sure about this one, the integration is named **PHP-FPM**, I feel like we should try to keep some consistency in names, or change everything from `php_fpm` to `fpm` then but I find it confusing (other stuff is called fpm too...)
* I think you meant `can_connect`, in this case
    * `can_connect` is correct in a way, but it's an understatement, because we got a pong from php fpm, meaning the server is ready to process requests
    * `is_running` sound a little bit like a process check to me
    * it's closer to an "haproxy-like" status heartbeat (people actually use it for that), I am open to suggestions, what do you think?

checks.d/php_fpm.py
LeoCavaille: :+1: my :snake: is tired

ci/resources/phpfpm/php-fpm.conf
LeoCavaille: Fair enough for me.

checks.d/php_fpm.py
LotharSee: Oh, sorry I didn't see that we give `pool` as an argument! They are other alternative, but I think it's not worth it.

checks.d/php_fpm.py
LotharSee: The service check name is a good question: the closest integration that we have is Gunicorn, but I am not a big fan of `is_running`.
It was just to raise this question and see if we can find something consistent. But since this is very specific, it's better preferring something with more sense. `can_ping`?

jippi:
:+1:

arosenhagen:
thx for taking this over. Happy to see that this will go forward and eventually be a part of an upcoming release.

LeoCavaille:
@LotharSee fixed your comments. Let me know if the new names are OK and I'll merge it.

checks.d/php_fpm.py
```diff
@@ -0,0 +1,119 @@
...
+        'total processes': 'php_fpm.processes.total',
+        'max active processes': 'php_fpm.processes.active.max',
+        'max children reached': 'php_fpm.processes.max_reached'
+    }
+
+    COUNTERS = {
+        'accepted conn': 'php_fpm.requests.accepted',
+        'slow requests': 'php_fpm.requests.slow'
+    }
+
```
LotharSee: Extra comments on the names.

First, since I didn't get what the `listen_queue` is, I had to look at the doc. This one is pretty good 
http://centminmod.com/phpfpm.html#browserstatus

* Try to keep the same depth per "group". For example `php_fpm.processes.active.max` and `php_fpm.processes. max_reached` should have the same number of dots.
* Also, finishing by `.max` could be confusing (with histograms) `php_fpm.listen_queue_reqs` and `php_fpm.listen_queue_reqs.max` (they also don't have the same depth). Could use `max_` instead.
* Not sure what we want to do with "max listen queue", "max active processes" and "max children reached" since they are counters initialized at PHP-FPM startup. Exception with "max children reached": it could be a counter.

jippi:
Yes! I was working on this myself, when is this released to the official deb dd-agent version ? :)

remh:
Hi @jippi . This will go out with our 5.3.0 release, which should be in around 3 weeks!

remh:
In the same time if you want to beta test it, you can just download from github the php-fpm.py file into your /etc/dd-agent/checks.d directory and the yaml file in your /etc/dd-agent/conf.d directory

jippi:
@remh shine, going to do just that then :) 

Thanks

PR #1442, opened 2015-03-17
[Jenkins] supporting the new folder naming convention

hkaj:
Starting from v1.597, the build folder names are not named with a timestamp anymore but with an ID. We have to get the timestamp from the `build.xml` file and to use the folder name to guess the build number.

### DIFF
--- a/checks.d/jenkins.py
+++ b/checks.d/jenkins.py
@@ -54,20 +54,18 @@ class Jenkins(AgentCheck):
         dir_basename = os.path.basename(dir_name)
         try:
             dir_basename = int(dir_basename)
-        except ValueError:
-            pass
-        if isinstance(dir_basename, int):
             # Parse the timestamp from the build.xml
             date_str = self._parse_timestamp(dir_name)
             return int(date_str) / 1000.0
-        else:
+        except ValueError:
+            pass
+        try:
             # Parse the timestamp from the directory name
-            try:
-                date_str = os.path.basename(dir_name)
-                time_tuple = time.strptime(date_str, self.datetime_format)
-                return time.mktime(time_tuple)
-            except ValueError:
-                raise Exception("Error with build directory name, not a parsable date: %s" % (dir_name))
+            date_str = os.path.basename(dir_name)
+            time_tuple = time.strptime(date_str, self.datetime_format)
+            return time.mktime(time_tuple)
+        except ValueError:
+            raise Exception("Error with build directory name, not a parsable date: %s" % (dir_name))
 
     def _get_build_metadata(self, dir_name):
         if os.path.exists(os.path.join(dir_name, 'jenkins_build.tar.gz')):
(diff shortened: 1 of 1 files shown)


checks.d/jenkins.py
```diff
@@ -75,13 +105,19 @@ def _get_build_metadata(self, dir_name):
...
             dirs = glob(os.path.join(job_dir, 'builds', '*_*'))
+            # versions of Jenkins > 1.597 dropped the timestamp-named folders
+            if len(dirs) == 0:
+                dirs = glob(os.path.join(job_dir, 'builds', '[0-9]*'))
             if len(dirs) > 0:
-                dirs = sorted(dirs, reverse=True)
+                # versions of Jenkins > 1.597 need to be sorted by build number (integer)
+                try:
+                    dirs = sorted(dirs, key=lambda x: int(x.split('/')[-1]), reverse=True)
+                except:
```
remh: You should almost never use bare "except:" as it would also catch System exiting exceptions which you don't want.

You should use except Exception or even better catch the actual exceptions that can be caught here.

checks.d/jenkins.py
```diff
@@ -75,13 +105,19 @@ def _get_build_metadata(self, dir_name):
...
         job_name = os.path.basename(job_dir)
-
         try:
             dirs = glob(os.path.join(job_dir, 'builds', '*_*'))
+            # versions of Jenkins > 1.597 dropped the timestamp-named folders
+            if len(dirs) == 0:
+                dirs = glob(os.path.join(job_dir, 'builds', '[0-9]*'))
             if len(dirs) > 0:
-                dirs = sorted(dirs, reverse=True)
+                # versions of Jenkins > 1.597 need to be sorted by build number (integer)
```
remh: Could you add in the comments the old and new folder names ? It would be easier to understand your code.

checks.d/jenkins.py
```diff
@@ -27,17 +28,46 @@ def __init__(self, name, init_config, agentConfig):
...
         try:
-            # Parse the timestamp from the directory name
-            date_str = os.path.basename(dir_name)
-            time_tuple = time.strptime(date_str, self.datetime_format)
-            return time.mktime(time_tuple)
+            dir_basename = int(dir_basename)
         except ValueError:
-            raise Exception("Error with build directory name, not a parsable date: %s" % (dir_name))
+            pass
+        if isinstance(dir_basename, int):
```
remh: that part should be in your try block so you don't have to test for the type

checks.d/jenkins.py
```diff
@@ -27,17 +28,46 @@ def __init__(self, name, init_config, agentConfig):
...
+            # Parse the timestamp from the build.xml
```
remh: Do we need to parse the timestamp in this case ?

If i understand, we need this timestamp just to know what was the latest build the agent analyzed. But we could just use the build number for this as we can assume that it's an always increasing number.

hkaj: We can, but the timestamp is also part of the event data (see https://github.com/DataDog/dd-agent/blob/master/checks.d/jenkins.py#L107). I can remove it if you think it isn't useful, but I think having a timestamp when looking at an event is always handy.

checks.d/jenkins.py
hkaj: sure, done above.

checks.d/jenkins.py
```diff
@@ -27,10 +28,37 @@ def __init__(self, name, init_config, agentConfig):
...
 
+    def _parse_timestamp(self, dir_name):
+        if os.path.exists(os.path.join(dir_name, 'jenkins_build.tar.gz')):
+            raise Skip('the build has already been archived', dir_name)
+
+        # Read the build.xml metadata file that Jenkins generates
+        build_file = os.path.join(dir_name, 'build.xml')
+
+        if not os.access(build_file, os.R_OK):
+            self.log.debug("Can't read build file at %s" % (build_file))
```
remh: You don't need this line

remh:
One last thing is that now you'll parse the build.xml twice.

One to get the timestamp and one to get the other metadata, can you consolidate this ?

Looks good otherwise!

checks.d/jenkins.py
remh: Ok

checks.d/jenkins.py
```diff
@@ -75,43 +94,42 @@ def _get_build_metadata(self, dir_name):
...
             if len(dirs) > 0:
-                dirs = sorted(dirs, reverse=True)
+                # versions of Jenkins > 1.597 need to be sorted by build number (integer)
+                try:
+                    dirs = sorted(dirs, key=lambda x: int(x.split('/')[-1]), reverse=True)
+                except ValueError:
+                    dirs = sorted(dirs, reverse=True)
                 # We try to get the last valid build
-                for index in xrange(0, len(dirs) - 1):
+                for index in xrange(len(dirs)):
```
remh: Nice bug fix!
We should likely just do:
```python
for dir_name in dirs:
```
unless i'm missing something ?

PR #1443, opened 2015-03-17
[Zookeeper check] bytes_outstanding metric is misleading

### LINKED ISSUE #1383: [zookeeper] bytes_outstanding metric returns number of requests served by zk node
User checking through the ZooKeeper check noticed that the bytes_outsanding metric the DD agent is scrapping is not bytes, but rather the number of inflight requests currently being serviced by the ZK node.

https://github.com/apache/zookeeper/blob/ca480231e1bdcbe6beda9142766fc3caa2031cd8/src/java/main/org/apache/zookeeper/server/ServerStats.java#L104

https://github.com/apache/zookeeper/blob/ca480231e1bdcbe6beda9142766fc3caa2031cd8/src/java/main/org/apache/zookeeper/server/ZooKeeperServer.java#L740

remh:
This value is not a value in bytes.
Fix #1383

We should remove it in a major version of the agent but for now, let’s
add a new metric name for this value.


LeoCavaille:
Agreed, looks like the CI fails for other reasons. Merging this one. Thanks :+1:

PR #1446, opened 2015-03-18
[sqlserver] add command timeout, 5sec default

### LINKED ISSUE #1247: SQL Server hang
SQL Server hanging when trying to run RECONFIGURE - caused by SQLServer check in Agent 

See: https://datadog.desk.com/agent/case/19284

degemer:
Set a timeout for each SQL command (by default 5 sec), to fix #1247 .

### DIFF
--- a/checks.d/sqlserver.py
+++ b/checks.d/sqlserver.py
@@ -48,6 +48,8 @@ class SQLServer(AgentCheck):
 
     SOURCE_TYPE_NAME = 'sql server'
     SERVICE_CHECK_NAME = 'sqlserver.can_connect'
+    # FIXME: 6.x, set default to 5s (like every check)
+    DEFAULT_COMMAND_TIMEOUT = 30
 
     METRICS = [
         ('sqlserver.buffer.cache_hit_ratio', 'Buffer cache hit ratio', ''),  # RAW_LARGE_FRACTION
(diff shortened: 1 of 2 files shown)


checks.d/sqlserver.py
```diff
@@ -185,19 +186,21 @@ def get_cursor(self, instance):
             try:
                 conn_str = self._conn_string(instance)
                 conn = adodbapi.connect(conn_str)
+                conn.CommandTimeout = int(instance.get('command_timeout',
+                                                        self.DEFAULT_COMMAND_TIMEOUT))
```
yannmh: `conn.CommandTimeout` was set to 30seconds by default. As you are changing it to 5 seconds now, it may break some users configuration ? What do you think ?

checks.d/sqlserver.py
```diff
@@ -185,19 +186,21 @@ def get_cursor(self, instance):
             try:
                 conn_str = self._conn_string(instance)
                 conn = adodbapi.connect(conn_str)
```
yannmh: You can also write:
```python
conn = adodbapi.connect(conn_str, timeout=...)
```
to set `conn.CommandTimeout`

yannmh:
Besides the concern about potential backward incompatibilities, it looks good to me.

Also it seems like you'd need to rebase your PR to be able to merge it proper

degemer:
Thanks @yannmh, updated my PR.
I bumped the default timeout to 10s (it should be more than enough).

yannmh:
Thanks ! :+1:

PR #1329, opened 2015-01-29
redisdb: add replication service check and 'down since' metric

pbitty:
This add a service checks for redis slaves, which reports CRITICAL when the slave loses a link with the master.  It will also report the amount of seconds the link has been down.

PR #1447, opened 2015-03-18
[Redis] add replication service check and 'down since' metric

remh:
This is based on #1329 

Fixing the conflicts and renaming the service check.

This PR adds a service check on the "master_link_status" and collect the replication lag.

Thanks a lot @pbitty !


yannmh:
Looks good to me !

PR #1448, opened 2015-03-19
Kill a fluentd process at test cleanup time.

yyamano:
ci/fluentd.rb invokes a fluentd process, but doesn't kill it at cleanup time. This change fixes it.


yannmh:
Thanks a lot @yyamano. Our 5.3 agent release is currently frozen. We'll review this for 5.4.0 !

PR #1449, opened 2015-03-19
Add type tag support to fluentd check.

yyamano:
This change allow us to use either fluentd plugin_id or type as a tag.

### DIFF
--- a/checks.d/fluentd.py
+++ b/checks.d/fluentd.py
@@ -29,6 +29,7 @@ class Fluentd(AgentCheck):
         try:
             url = instance.get('monitor_agent_url')
             plugin_ids = instance.get('plugin_ids', [])
+            tag_by = instance.get('tag_by', 'plugin_id')
 
             parsed_url = urlparse.urlparse(url)
             monitor_agent_host = parsed_url.hostname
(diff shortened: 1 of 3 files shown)


yannmh:
Thanks a lot @yyamano !
5.3 agent release is currently frozen, but we are looking forward reviewing your PR for the 5.4 release.

yyamano:
I will give a little explanation of the change.

Currently, I have to:

* specify plugin ids explicitly in fluentd's configuration file
* have same plugin ids in fleuntd.yml

And I get metrics per plugin id. If I have a bunch of plugins, it's really annoying. I don't want to manage all plugin ids. Also metrics per plugin type works well to me. So I introduce 'tag_by' parameter.

```
-  monitor_agent_url: http://example.com:24220/api/plugins.json
   tag_by: type
```

You can use 'tag_by: plugin_id' too. 'tag_by' is optional. Default value is 'plugin_id'.

Technically speaking, my change breaks backward comaptibility in a very rare case. I don't think it cause any real problems.

Here is a common configuration.

```
-  monitor_agent_url: http://example.com:24220/api/plugins.json
   plugin_ids:
   - plg1
   - plg2
```

It works well with both current implementation and my version. Plugin metrics that match 'plg1' or 'plg2' with a tag like 'plugin_id:plg1' are sent to datadog.

A problem might have come up with the following configuration.

```
-  monitor_agent_url: http://example.com:24220/api/plugins.json
   plugin_ids:
```

It didn't send any metrics with current, but now all plugin metrics gonna to be sent. I can't imagine someone has been using such configuration.

checks.d/fluentd.py
```diff
@@ -40,11 +41,13 @@ def check(self, instance):
...
 
             for p in status['plugins']:
+                tag = "%s:%s" % (tag_by, p.get(tag_by))
                 for m in self.GAUGES:
                     if p.get(m) is None:
                         continue
-                    if p.get('plugin_id') in plugin_ids:
-                        self.gauge('fluentd.%s' % (m), p.get(m), ["plugin_id:%s" % p.get('plugin_id')])
+                    # Filter unspecified plugins to keep backward compatibility.
+                    if tag_by == 'type' or len(plugin_ids) == 0 or p.get('plugin_id') in plugin_ids:
```
yannmh: I am not sure to understand why we are checking the value of `tag_by` here. I would just write:
```python
 if len(plugin_ids) == 0 or p.get('plugin_id') in plugin_ids:
```
Am I missing something ?

yannmh:
Thanks a lot @yyamano. 

I rebased your work on latest Datadog agent to fix some merge conflicts, and opened a new PR #1623. We plan to release your changes in the upcoming 5.4 agent release :tada:

PR #1450, opened 2015-03-19
Fix example nginx config

techgaun:
Nginx yaml configuration example includes wrong format. Please review. Thanks!

### DIFF
--- a/conf.d/nginx.yaml.example
+++ b/conf.d/nginx.yaml.example
@@ -12,8 +12,8 @@ instances:
 
   - nginx_status_url: http://localhost/nginx_status/
   #   tags:
-  #     - instance:foo
+  #     - foo
   #
   # - nginx_status_url: http://example2.com:1234/nginx_status/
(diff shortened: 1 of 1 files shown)


remh:
Hi @techgaun 
Thanks a lot for the feedback!

The format in the current yaml is right.

I think you added an extra space after the column (":") in your tag. 

i.e.:
```yaml
tags:
    - foo:bar
```

will result in the following python data structure:
```python
{'tags': ['foo:bar']}
```

While the following yaml:
```yaml
tags:
    - foo: bar
```
will result in:
```python
{'tags': [{'foo':'bar'}]}
```
You can use the following yaml parser to test your configuration files.

We should definitely enforce a better validation of the configuration though, i'll create an issue on our end about it.

Thanks again for your feedback!

techgaun:
@remh Oh yeah, right. I definitely added a space after the colon. Thanks for the feedback.

PR #1454, opened 2015-03-20
Do not parse args in info page

remh:
Otherwise it will try to parse args that are not expected
e.g. use_simple_http_client and will log some errors.

This fixes the 
```
ddagent.py: error: no such option: --use_simple_http_client
```

error on source install.


remh:
Failing on flare. Merging

PR #1456, opened 2015-03-23
[flare] fix test_endpoints (wrong error msg)

degemer:
And some pep8 cleaning.

Also setting TIMEOUT to 60s (in case of huge files).

### DIFF
--- a/utils/flare.py
+++ b/utils/flare.py
@@ -30,6 +30,7 @@ import requests
 # Globals
 log = logging.getLogger('flare')
 
+
 def configcheck():
     all_valid = True
     for conf_path in glob.glob(os.path.join(get_confd_path(), "*.yaml")):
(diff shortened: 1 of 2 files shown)


utils/flare.py
```diff
@@ -311,8 +312,8 @@ def _pip_freeze(self):
     # Check if the file is not too big before upload
     def _check_size(self):
         if os.path.getsize(self._tar_path) > self.MAX_UPLOAD_SIZE:
-            log.info('{0} won\'t be uploaded, its size is too important.\n'\
-                      'You can send it directly to support by mail.')
+            log.info('{0} won\'t be uploaded, its size is too important.\n'
+                     'You can send it directly to support by mail.')
```
LeoCavaille: Nitpick but you could use double quotes and not have to escape stuff.

LeoCavaille:
Nice! LGTM, and thanks for pep8-ing this!

PR #1464, opened 2015-03-24
Fix small bug in vSphere integration check

wang-arthur:
Fix bug in retrieving open connections to vCenter instances
(previously, `server_instance` was only defined for new connections).


LeoCavaille:
LGTM, thanks @wang-arthur !

PR #1465, opened 2015-03-24
Fix zero division bug in mysql check

wang-arthur:
- Don't try to calculate `innodb_buffer_pool_pages_utilization` when `innodb_buffer_pool_pages_total == 0`


LeoCavaille:
Thanks, next time if you can follow our naming convention for commit messages, that's even cooler :sunglasses: 

https://github.com/DataDog/dd-agent/wiki/Conventions#githubcommit-titles

PR #1466, opened 2015-03-24
[haproxy] use new test class

degemer:
- PEP8 and FLAKE8 cleaning in `haproxy.py`
- use AgentCheckTest to test haproxy (rewrite old tests to have a 100%
  metrics coverage)
- start/stop haproxy in `ci/haproxy.rb`, instead of inside the test
  itself
- add a `run_check_twice` function inside `AgentCheckTest` to avoid code
  repetition when testing rates

### DIFF
--- a/tests/common.py
+++ b/tests/common.py
@@ -4,6 +4,7 @@ import logging
 import os
 import signal
 import sys
+import time
 import traceback
 import unittest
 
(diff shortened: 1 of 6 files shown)


tests/common.py
```diff
@@ -119,6 +120,12 @@ def __init__(self, *args, **kwargs):
 
         self.check = None
 
+    # Helper function when testing rates
+    def run_check_twice(self, config, agent_config=None, mocks=None):
+        self.run_check(config, agent_config=None, mocks=None)
```
LeoCavaille: Here you should pass the options the user intended to use and not `None`.
```
self.run_check(config, agent_config, mocks)
```

LeoCavaille:
Nice cleanup :+1: 
Except the small comment above on the new function of `AgentCheckTest` it's GTG !

tests/common.py
degemer: Yep, copy/paste error, thanks !

LeoCavaille:
Thanks!

PR #1467, opened 2015-03-25
[http_check] add coverage_report

degemer:
`http_check` was already using `AgentCheckTest`, thanks to @yannmh.


LeoCavaille:
:boom:

PR #1468, opened 2015-03-25
[core] Collector and Checks metrics Governor

yannmh:
Collector and Checks metrics Governor is ready for review !

## What is a Limiter ?

A limiter is a rule set to monitor the metric stream.

It is defined by:
* a `scope`: scope where the rule applies
* a `selection`: selection where the rule applies
* an optional `limit`:  limit of selections per scope. 

`scope` and `selection` are defined as subset of `['name', 'agent', 'check', 'tags']` and allow to define 
custom rules. 
For instance
`'scope' = ('name', 'check') selection = 'tags'` monitors number of tags per (metric name, check)
`'scope' = 'agent'` `selection = 'name'` monitors globally the number of metrics names

Limiters keep tracks of 'active' selections (see point below), i.e. metrics's met selections on a sliding window, and count the number of metrics exceeding the limit set. 

## What an 'active' selection ?

A selection is considered as 'active', if it has been met at least once during the last `X` metrics stream batch processing, where `X` is a settable parameter. 

## What is a Governor ? 

A Governor supervises multiple limiters: submitting metric stream batches for checks, and reporting when an anomaly is detected.

Governors are split in two categories 
* CheckGovernor. 
Supervise `check` scoped limiters. 

* AgentGovernor.
Supervise wide scoped limiters. 
When a limiter hit the active selections limit, a warning is post to Datadog.

## Notes
For now the Governor is ready-only and does not alterate any payload.

### TO-DO
- [ ] Choose where to get Governor settings from (hardcoded for now)
- [ ] Define Governor settings
- [ ] Enable Governor report endpoint
- [ ] Add a Governor for Dogstatsd metrics

### DIFF
--- a/agent.py
+++ b/agent.py
@@ -28,6 +28,7 @@ from checks.check_status import CollectorStatus
 from config import (
     get_confd_path,
     get_config,
+    get_governor_config,
     get_logging_config,
     get_parsed_args,
     get_system_stats,
@@ -44,6 +45,7 @@ from util import (
     Watchdog,
 )
 from utils.flare import configcheck, Flare
+from utils.governor import Governor
 
 # Constants
 PID_NAME = "dd-agent"
@@ -104,10 +106,12 @@ class Agent(Daemon):
         hostname = get_hostname(agentConfig)
         systemStats = get_system_stats()
         emitters = self._get_emitters(agentConfig)
+
+        # Instantiate collector
+        self.collector = Collector(agentConfig, emitters, systemStats, hostname)
+
         # Load the checks.d checks
         checksd = load_check_directory(agentConfig, hostname)
-
-        self.collector = Collector(agentConfig, emitters, systemStats, hostname)
 
         # Configure the watchdog.
         check_frequency = int(agentConfig['check_freq'])
@@ -288,14 +292,21 @@ def main():
             # Try the old-style check first
             print getattr(checks.collector, check_name)(log).check(agentConfig)
         except Exception:
+
+            # Init Governor
+            governor_config = get_governor_config()
+            Governor.init(governor_config, agentConfig, hostname)
+
             # If not an old-style check, try checks.d
             checks = load_check_directory(agentConfig, hostname)
+
             for check in checks['initialized_checks']:
                 if check.name == check_name:
                     check.run()
                     print check.get_metrics()
                     print check.get_events()
                     print check.get_service_checks()
+                    print check.get_governor_status()
                     if len(args) == 3 and args[2] == 'check_rate':
                         print "Running 2nd iteration to capture rate metrics"
                         time.sleep(1)
@@ -303,6 +314,7 @@ def main():
                         print check.get_metrics()
                         print check.get_events()
                         print check.get_service_checks()
+                        print check.get_governor_status()
                     check.stop()
 
     elif 'configcheck' == command or 'configtest' == command:
--- a/aggregator.py
+++ b/aggregator.py
@@ -2,7 +2,6 @@ import logging
 from time import time
 
 from checks.metric_types import MetricTypes
-from config import get_histogram_aggregates, get_histogram_percentiles
 
 log = logging.getLogger(__name__)
 
@@ -55,7 +54,6 @@ class Gauge(Metric):
         self.last_sample_time = time()
         self.timestamp = timestamp
 
-
     def flush(self, timestamp, interval):
         if self.value is not None:
             res = [self.formatter(
@@ -72,6 +70,7 @@ class Gauge(Metric):
             return res
 
         return []
+
 
 class BucketGauge(Gauge):
     """ A metric that tracks a value at particular points in time.
@@ -131,6 +130,7 @@ class Count(Metric):
             )]
         finally:
             self.value = None
+
 
 class MonotonicCount(Metric):
 
@@ -213,6 +213,7 @@ class Counter(Metric):
 
 DEFAULT_HISTOGRAM_AGGREGATES = ['max', 'median', 'avg', 'count']
 DEFAULT_HISTOGRAM_PERCENTILES = [0.95]
+
 
 class Histogram(Metric):
     """ A metric to track the distribution of a set of values. """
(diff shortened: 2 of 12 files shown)


degemer:
I think you could use the new `utils/` subclass for `governor.py`.

governor.py
```diff
@@ -0,0 +1,408 @@
...
+    # Defines what's an 'active' context
+    # i.e. number of iterations for which a context (not encountered anymore)
+    # is kept stored before being flushed.
+    _CONTEXTS_TTL = 3
+
+    # Collector payload's metric format
+    _METRIC_FORMAT = ['name', 'timestamp', 'value', 'attributes']
+
+    # Governor report endpoint
+    DATADOG_REPORT_GOVERNOR_URL = 'agent_governor/report_too_many_active_contexts'
```
LeoCavaille: It would probably me more flexible to have a `agent_governor/report` endpoint and support different "events" from the governor

governor.py
```diff
@@ -0,0 +1,408 @@
...
+
+    # Governor agent-specific static variables
+    _CHECK_LIMITERS, _AGENT_LIMITERS = None, None
+    _HOSTNAME = None
+    _DD_URL = None
+    _API_KEYS = None
+
+    @classmethod
+    def init(cls, governor_config, agent_config={}, hostname=None):
```
LeoCavaille: You shouldn't use a mutable default args.

governor.py
```diff
@@ -0,0 +1,408 @@
...
+        """
+        Set Governor agent-specific variables
+        """
+        if not _is_affirmative(agent_config.get('use_governor', True)):
+            return
+
+        cls._CHECK_LIMITERS, cls._AGENT_LIMITERS = LimiterParser.parse_limiters(governor_config)
+        cls._HOSTNAME = hostname
+        cls._DD_URL = agent_config.get('dd_url')
```
LeoCavaille: You should probably use the post-formatted URL instead here? see https://github.com/DataDog/dd-agent/commit/92f0c759df0dc17d76a7d28c5cbe1e0bf159c4f8#diff-27d4010935202c09554e28f0bb3239a7R196

governor.py
```diff
@@ -0,0 +1,408 @@
...
+        cls._API_KEYS = agent_config.get('api_key')
```
LeoCavaille: Why plural? if this is to only store a single API key.

governor.py
```diff
@@ -0,0 +1,408 @@
...
+            return self._submit_metric(*args, **kw)
+
+
+class AgentGovernor(Governor):
+    """
+    Monitor entire collector's metric payload
+    """
+    def __init__(self):
+        super(AgentGovernor, self).__init__()
+        self._limiters = copy.deepcopy(self._AGENT_LIMITERS)
```
LeoCavaille: Why copy this and not let it at the class level? It won't be modified at runtime anyway no?

governor.py
```diff
@@ -0,0 +1,408 @@
...
+    def __call__(self, *args, **kw):
+        """
+        Decorator purpose around `submit_metric` method for 'real-time' analysis
+        """
+        # Shortcut when no rules are defined
+        if not self._limiters:
+            return self._submit_metric(*args, **kw)
+
+        # Really dirty trick -> to improve
+        named_args = self._name_args(args, kw)
```
LeoCavaille: Add a `FIXME` disclaimer, it's easier to grep in the code.

governor.py
```diff
@@ -0,0 +1,408 @@
...
+        Agent hit the active contexts limit set by the Governor.
+        Send a warning email to Datadog team.
+        """
+        log.warning("Agent hit the active contexts limit set by the Governor."
+                    "Sending a warning to Datadog team")
+
+        url = "{0}/{1}".format(self._DD_URL, self.DATADOG_REPORT_GOVERNOR_URL)
+
+        try:
+            data = ""
```
LeoCavaille: ? Another `FIXME`

governor.py
```diff
@@ -0,0 +1,408 @@
...
+        for m in metrics:
+            named_args = self._name_args(m[0:3], m[3], asynchronous=True)
+            self._check(named_args)
+
+        statuses = self.get_status(flush=flush)
+
+        # Report to Datadog when Governor limit is hit
+        overflow_statuses = filter(lambda s: s['trace']['overflow_metrics'], statuses)
+        if report and overflow_statuses:
+            self._post_datadog_warning(overflow_statuses)
```
LeoCavaille: This arg doesn't exist in the function below.

governor.py
```diff
@@ -0,0 +1,408 @@
+import copy
```
remh: Could you organize your imports here ?

governor.py
```diff
@@ -0,0 +1,408 @@
...
+        """
+        log.warning("Agent hit the active contexts limit set by the Governor."
+                    "Sending a warning to Datadog team")
+
+        url = "{0}/{1}".format(self._DD_URL, self.DATADOG_REPORT_GOVERNOR_URL)
+
+        try:
+            data = ""
+            log.debug("Performing post {0} to url {1}".format(data, url))
+            r = requests.post(url, data=data)
```
remh: You should get the proxy settings and use them if needed.

governor.py
degemer: Actually I moved this function into `config.py` to use it for `flare` - https://github.com/DataDog/dd-agent/blob/e33d658f25a7b8a02358dae070b12a45293125fa/config.py#L102-L114

yannmh:
Good idea @degemer. Moving `governor.py` to `utils/`

governor.py
yannmh: It will.  Limiters have a "state", in particular
* An index to keep track of active selections
* A trace, which records the number of metrics overflowing the limit set

As there should be only one `AgentGovernor` object instantiated, it would not be a problem to manipulate these class variables directly (unlike `CheckGovernor`). 
I still think it's preferable to keep them "blank" at class level, and make a copy for the instance.

remh:
Won't be implemented.

PR #1472, opened 2015-03-25
[rabbitmq] Improve test suite and fix tagging

LeoCavaille:
* because we checked the tags only to be non-`None` we were creating
  empty tags like `rabbitmq_policy:`
* adds some helper methods in `AgentCheckTest` to assert directly a
  service check by status like `assertServiceCheckCritical` without
  having to import `AgentCheck` just for that in the test file
* add coverage report to ensure we cover 100% of the metrics output
  add TODO instead of commenting the missing metrics, maybe we can do
  that at some point but we would have to create a consumer and do some
  work to get these. For instance use the `pika` python module.
* Bonus: some pep8-ing

### DIFF
--- a/checks.d/rabbitmq.py
+++ b/checks.d/rabbitmq.py
@@ -17,7 +17,9 @@ QUEUE_TYPE = 'queues'
 NODE_TYPE = 'nodes'
 MAX_DETAILED_QUEUES = 200
 MAX_DETAILED_NODES = 100
-ALERT_THRESHOLD = 0.9 # Post an event in the stream when the number of queues or nodes to collect is above 90% of the limit:
+# Post an event in the stream when the number of queues or nodes to
+# collect is above 90% of the limit:
+ALERT_THRESHOLD = 0.9
 QUEUE_ATTRIBUTES = [
     # Path, Name
     ('active_consumers', 'active_consumers'),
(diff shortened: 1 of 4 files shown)


checks.d/rabbitmq.py
```diff
@@ -220,19 +234,20 @@ def get_stats(self, instance, base_url, object_type, max_detailed, filters, auth
...
 
         for data_line in data[:max_detailed]:
             # We truncate the list of nodes/queues if it's above the limit
             self._get_metrics(data_line, object_type)
 
-
     def _get_metrics(self, data, object_type):
         tags = []
         tag_list = TAGS_MAP[object_type]
         for t in tag_list.keys():
```
remh: I know it's not your code but we shouldn't use 
```python
.keys()
``` here

File: checks.d/rabbitmq.py [hunk 2]
```diff
@@ -220,19 +234,20 @@ def get_stats(self, instance, base_url, object_type, max_detailed, filters, auth
...
             tag = data.get(t, None)
```
remh: Same thing, not your code but we should get rid of the 
```python
, None)
```

LeoCavaille:
Thanks @remh for catching that, will merge when tests have passed (apache being a known failure)

PR #1473, opened 2015-03-25
[tests] switch lighttpd check to AgentCheckTest.

hkaj:


LeoCavaille:
You removed the test with the instance `?auto` URL, was it just not useful or redundant?

hkaj:
@LeoCavaille it's redundant. The check adds the ?auto automatically in the first config after trying without it and failing to parse the result.

hkaj:
Also the 3 build failures in Travis are due to timeouts.

LeoCavaille:
nice :+1:

PR #1478, opened 2015-03-26
[core] removes extra carriage return

yannmh:


LeoCavaille:
Let's wait the Travis tests result before merging.

LeoCavaille:
That's the epitome of what we call a *sneaky line*. cc @LotharSee 
Even sneakier when it's blank.

PR #1479, opened 2015-03-26
[apache] fix tests

yannmh:


LeoCavaille:
Looks like it's waiting for some access on Apache threads to return something, so maybe if you stimulate it before testing, this metric will always be here?
See https://github.com/apache/httpd/blob/trunk/modules/generators/mod_status.c#L342-L346

Also this is not cool as it will break the coverage report if the metric is there but isn't tested.

yannmh:
Does not seem to solve the issue :disappointed:

LeoCavaille:
Have you tried hitting the default apache URL instead of the status URL? like `http://localhost:1234`

yannmh:
@LeoCavaille Looks good to merge to me

LeoCavaille:
Thanks!

PR #1484, opened 2015-03-27
[iis] review tests

yannmh:


tests/test_iis.py
```diff
@@ -1,58 +1,154 @@
...
-
-        # Make sure that we get a service check
-        self.assertEquals(len(service_checks),1)
-        self.assertEquals(check.SERVICE_CHECK, service_checks[0]['check'])
-        self.assertEquals(['site:Default Web Site'], service_checks[0]['tags'])
-
-
-if __name__ == "__main__":
-    unittest.main()
+# stdlib
```
LeoCavaille: really reaaaallllyyyyy nitpicky but `mock` in only in stdlib in Python 3.3+ I think, otherwise it's a 3p module.

LeoCavaille:
Nice thanks for rewriting this test :+1:

tests/test_iis.py
```diff
@@ -1,58 +1,153 @@
...
+
+        # Run check
+        config = {
+            'instances': [self.WIN_SERVICES_CONFIG]
+        }
+
+        self.run_check_twice(config)
+
+        # Test metrics
+        for mname, _, _ in self.check.METRICS:
```
remh: You should probably put a copy of this list in the test.
Otherwise one may remove a metric accidentally and the test would still pass.

PR #1487, opened 2015-03-30
[ElasticSearch] Improve tests and bug fixes

### LINKED ISSUE #1402: wrong url in elastic.py can.connect monitor used for version > 0.90.10
The new URLS are /_nodes/stats?all=true

But the datadog uses sometimes the old one: /_cluster/nodes/stats?all=true

elastic.py 

*Old* version: (only interesting parts)
/opt/datadog-agent/agent/checks.d/elastic.py

```python
if version >= [0,90,10]:
            if is_external:
                self.STATS_URL = "/_nodes/stats?all=true"
            else:
                self.STATS_URL = "/_nodes/_local/stats?all=true"
else:
            self.STATS_URL = "/_cluster/nodes/stats?all=true"
            self.NODES_URL = "/_cluster/nodes?network=true"```
```

*fixed version* should be: (only interesting parts)
/opt/datadog-agent/agent/checks.d/elastic.py

```python
if version >= [0,90,10]:
            if is_external:
                self.STATS_URL = "/_nodes/stats?all=true"
            else:
                self.STATS_URL = "/_nodes/_local/stats?all=true"
else:
            if is_external:
                self.STATS_URL = "/_nodes/stats?all=true"
            else:
                self.STATS_URL = "/_nodes/_local/stats?all=true"
```

remh:
- Properly handle multiple instances
- 100% metrics and service checks coverage
- Flake8 (everything but line > 80)
- Make sure that constants are constants
- Fix #1402 
- Default version to 1.0.0 as it's more common to have an ES version >= 1.0.0

Also:
- Pretty format coverage report
- Log more info when candidates assertion is failing

### DIFF
--- a/checks.d/elastic.py
+++ b/checks.d/elastic.py
@@ -14,8 +14,8 @@ from config import _is_affirmative
 from util import headers, Platform
 
 
-class NodeNotFound(Exception): pass
-
+class NodeNotFound(Exception):
+    pass
 
 ESInstanceConfig = namedtuple(
     'ESInstanceConfig', [
@@ -26,7 +26,7 @@ ESInstanceConfig = namedtuple(
         'timeout',
         'url',
         'username',
-])
+    ])
 
 
 class ESCheck(AgentCheck):
@@ -35,7 +35,7 @@ class ESCheck(AgentCheck):
 
     DEFAULT_TIMEOUT = 5
 
-    STATS_METRICS = { # Metrics that are common to all Elasticsearch versions
+    STATS_METRICS = {  # Metrics that are common to all Elasticsearch versions
         "elasticsearch.docs.count": ("gauge", "indices.docs.count"),
         "elasticsearch.docs.deleted": ("gauge", "indices.docs.deleted"),
         "elasticsearch.store.size": ("gauge", "indices.store.size_in_bytes"),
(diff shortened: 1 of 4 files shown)


remh:
Failing on Apache test.

checks.d/elastic.py
```diff
@@ -126,7 +159,7 @@ class ESCheck(AgentCheck):
         "elasticsearch.relocating_shards": ("gauge", "relocating_shards"),
         "elasticsearch.initializing_shards": ("gauge", "initializing_shards"),
         "elasticsearch.unassigned_shards": ("gauge", "unassigned_shards"),
-        "elasticsearch.cluster_status": ("gauge", "status", lambda v: {"red":0, "yellow":1, "green":2}.get(v, -1)),
+        "elasticsearch.cluster_status": ("gauge", "status", lambda v: {"red": 0, "yellow": 1, "green": 2}.get(v, -1)),
```
LeoCavaille: nitpick: but this should probably use the class constants instead of bare integers to define the service check values.

checks.d/elastic.py
```diff
@@ -476,14 +497,13 @@ def _create_event(self, status):
...
-
-
+        return {
+            'timestamp': int(time.time()),
+            'event_type': 'elasticsearch',
+            'host': hostname,
+            'msg_text': msg,
+            'msg_title': msg_title,
+            'alert_type': alert_type,
+            'source_type_name': "elasticsearch",
```
LeoCavaille: If we set the `source_type_name` as a class constant maybe that's where we want to use it actually !?

checks.d/elastic.py
remh: What class constants ? This is a mapping that is not related to service checks. This is send as a metric, it was done before we had service checks so one could alert on the cluster status.

Those values are actually the opposite of the constants we use for service checks.

checks.d/elastic.py
remh: Yeah i didn't change that parf of the code, just fixed the PEP8 violations. Fixing it

checks.d/elastic.py
LeoCavaille: Apologies, scratch that..  Should have looked more at the context.
Maybe at one point we could deprecate this guy, it's *really* confusing

tests/common.py
```diff
@@ -263,10 +265,18 @@ def assertMetric(self, metric_name, value=None, tags=None, count=None, at_least=
...
                 candidates.append((m_name, ts, val, mdata))
 
-        self._candidates_size_assert(candidates, count=count, at_least=at_least)
+        try:
+            self._candidates_size_assert(candidates, count=count, at_least=at_least)
+        except AssertionError:
+            log.error("No candidates for {0} (value: {1}, tags: {2}, count: {3}, at_least: {4}, hostname: {5})"\
+                .format(metric_name, value, tags, count, at_least, hostname))
+            raise
+
```
LeoCavaille: I added the debug headers for that reason ("Looking for...") should we keep only one or another? I am good with both, also "No candidates" might not be a clear message for some cases like `at_least=1`.

tests/test_elastic.py
```diff
@@ -1,104 +1,263 @@
...
+        self.assertServiceCheck('elasticsearch.can_connect',
+                                status=AgentCheck.OK, tags=good_sc_tags,
+                                count=2)
+        self.assertServiceCheck('elasticsearch.can_connect',
+                                status=AgentCheck.CRITICAL, tags=bad_sc_tags,
+                                count=1)
+
+
+        status = AgentCheck.OK if os.environ.get("TRAVIS") else AgentCheck.CRITICAL
+        # Travis doesn't have any shards in the cluster and consider this as green
```
LeoCavaille: Is there a way to reproduce the same env in Travis and locally?

tests/common.py
remh: I didn't find the debug log visible enough when testing, especially when you have a lot of metrics. This log line made my debugging easier. I can change the message though.

LeoCavaille: Works for me. Let's get rid of the debug lines then.

remh:
Failing on Apache. Merging.

PR #1489, opened 2015-03-31
Remove old config migration script

remh:
Using datadog.conf to configure  these checks is not supported since Agent 5.0.0


LeoCavaille:
Nice, glad to see that gone.

PR #1490, opened 2015-03-31
[Memcache] Improve tests and bug fixes

remh:
- 100% metrics and service check coverage
- Add a default tag so that multi instances will work properly
- Remove monkey patch as it’s not needed anymore (See comments about
python-memcached version impacted here:
https://github.com/DataDog/dd-agent/commit/d3bbd1ff561b3b2682ae79d308876
a4910f9ff3b )
- Flake 8

Also:
- Add a helper in the AgentCheckTest class to determine if we’re
running on travis or not
- Add a helper to “reset” a check that may be useful sometimes

### DIFF
--- a/tests/common.py
+++ b/tests/common.py
@@ -120,9 +120,13 @@ class AgentCheckTest(unittest.TestCase):
 
         self.check = None
 
+    def is_travis(self):
+        return "TRAVIS" in os.environ
+
     # Helper function when testing rates
-    def run_check_twice(self, config, agent_config=None, mocks=None):
-        self.run_check(config, agent_config, mocks)
+    def run_check_twice(self, config, agent_config=None, mocks=None,
+        force_reload=False):
+        self.run_check(config, agent_config, mocks, force_reload)
         time.sleep(1)
         self.run_check(config, agent_config, mocks)
 
--- a/tests/test_mcache.py
+++ b/tests/test_mcache.py
@@ -4,32 +4,129 @@ import os
 import time
 from subprocess import Popen, PIPE
 
-from tests.common import load_check
+from tests.common import AgentCheckTest
 
 from checks import AgentCheck
 
+# 3rd party
+import memcache
+
+
+GAUGES = [
+    "total_items",
+    "curr_items",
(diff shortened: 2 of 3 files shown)


tests/common.py
```diff
@@ -120,6 +120,12 @@ def __init__(self, *args, **kwargs):
 
         self.check = None
 
+    def reset(self):
+        self.check = None
```
LeoCavaille: @yannmh added `force_reload` in `run_check` could that be enough? Let's try to keep this class neat.

tests/test_mcache.py
```diff
@@ -114,39 +164,14 @@ def testMemoryLeak(self):
...
-            })
-        except Exception:
-            # We expect an exception here. Just ignore it.
-            pass
-        svc_checks = self.c.get_service_checks()
-        self.assertEquals(len(svc_checks), 1)
-        self.assertEquals(svc_checks[0]['check'], self.c.SERVICE_CHECK)
-        self.assertEquals(svc_checks[0]['status'], AgentCheck.CRITICAL)
-
 if __name__ == '__main__':
```
LeoCavaille: You can drop that, we should recommend only running tests with nose.

tests/test_mcache.py
```diff
@@ -4,32 +4,131 @@
...
+        self.run_check_twice(config)
+        for tags in tag_set:
+            for m in GAUGES:
+                self.assertMetric("memcache.{0}".format(m), tags=tags, count=1)
+            for m in RATES:
+                self.assertMetric(
+                    "memcache.{0}_rate".format(m), tags=tags, count=1)
+
+        good_service_check_tags = ["host:localhost", "port:{0}".format(PORT)]
+        bad_service_check_tags = ["host:unix", "port:foo/bar"]
```
LeoCavaille: copy/pasta not needed

LeoCavaille:
Nice cleanup! :+1:

LeoCavaille:
memcache tests succeded. I just fixed up your "Fix as per review" commit with the previous one to get a cleaner history :wink: 
```
git rebase -i master
```

PR #1492, opened 2015-03-31
[dev] delete log files before caching

degemer:
Previous command was working with a *zsh* shell but not bash,
see http://unix.stackexchange.com/a/62665

We might need to manually clear the cache.


LeoCavaille:
Cache cleared.

PR #1493, opened 2015-03-31
[nagios] review tests

yannmh:


tests/common.py
```diff
@@ -245,12 +244,15 @@ def _candidates_size_assert(self, candidates, count=None, at_least=1):
             self.print_current_state()
             raise
 
-    def assertMetric(self, metric_name, value=None, tags=None, count=None, at_least=1):
+    def assertMetric(self, metric_name, value=None, tags=None, additional_attr=None, count=None, at_least=1):
```
LeoCavaille: I am not 100% convinced that we should add something as generic as that. We should just handle the `device_name` attribute through a named kwarg which seems reasonable because we do it exactly like the same way in the `AgentCheck` metrics submitting methods.
Also, it's still handled differently in the backend code.

yannmh:
Thanks for your feedback @LeoCavaille. I brought some changes to my PR.

LeoCavaille:
:+1:

PR #1497, opened 2015-03-31
[ci] Use a bucket to download all tarballs :racehorse:

LeoCavaille:
This shouldn't be used too often because we have another layer of
caching (depending on dd-agent-travis-cache).
Although, when we really have to download the source and recompile it's
often when SourceForge or GitHub have an outage. By placing every egg in
the same basket we ensure that it's *our fault* if things go wrong.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,14 +30,16 @@ env:
   matrix:
     - TRAVIS_FLAVOR=default
     - TRAVIS_FLAVOR=apache
-    - TRAVIS_FLAVOR=cassandra
+    - TRAVIS_FLAVOR=cassandra FLAVOR_VERSION=2.0.13
+    - TRAVIS_FLAVOR=cassandra FLAVOR_VERSION=2.1.3
     - TRAVIS_FLAVOR=couchdb
     - TRAVIS_FLAVOR=elasticsearch FLAVOR_VERSION=0.90.13
     - TRAVIS_FLAVOR=elasticsearch FLAVOR_VERSION=1.0.3
     - TRAVIS_FLAVOR=elasticsearch FLAVOR_VERSION=1.1.2
     - TRAVIS_FLAVOR=elasticsearch FLAVOR_VERSION=1.2.4
     - TRAVIS_FLAVOR=elasticsearch FLAVOR_VERSION=1.3.7
-    - TRAVIS_FLAVOR=elasticsearch FLAVOR_VERSION=1.4.2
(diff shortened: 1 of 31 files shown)


degemer:
:+1:  for the :racehorse: !

Maybe we could keep the default URL (ie https://download.elasticsearch.org/elasticsearch/elasticsearch/elasticsearch-#{es_version}.tar.gz) somewhere to ease the upgrade to another version of the software ? (maybe you already did it on DD side ?)

LeoCavaille:
@degemer thought about this, but didn't really know where to put it, ideas? And it's also pretty straightforward to find the link from Google with a few clicks.

degemer:
@LeoCavaille maybe just add a comment in all `flavor.rb` with the source URL ?
Especially for things like `js185` in `couchdb.rb`, which are not that easy to find.

LeoCavaille:
@degemer added upstream links to `flavor.rb`. I also fixed the postgres flavor regressions. Let's wait for this one to complete and see what's left broken before merging.

.travis.yml
```diff
@@ -83,6 +90,7 @@ script:
 after_failure:
   - echo "Logs from installation process come here / DEBUG LOGS"
   - cat /tmp/ci.log
+  - cat /tmp/tokumx.log
```
degemer: Debug log not removed, or you want to keep it ? 
It's true that it would be nice to configure all integration programs to log in `/tmp/logs` and to have a `cat /tmp/logs/*` or something like this in `after_failure`.

ci/pgbouncer.rb
```diff
@@ -13,7 +12,7 @@ def pgb_rootdir
     task :install do
       Rake::Task['ci:postgres:install'].invoke
       unless Dir.exist? File.expand_path(pgb_rootdir)
-        sh %(wget -O $VOLATILE_DIR/pgbouncer-1.5.4.tar.gz https://s3.amazonaws.com/travis-archive/pgbouncer-1.5.4.tar.gz)
+        sh %(wget -O $VOLATILE_DIR/pgbouncer-1.5.4.tar.gz https://s3.amazonaws.com/dd-agent-tarball-mirror/pgbouncer-1.5.4.tar.gz)
```
degemer: I think you forgot to c/p the old download link. (and I checked, it's not the first link on Google :stuck_out_tongue: )

.travis.yml
LeoCavaille: I'll get rid of it for now, but I :+1: the idea.

LeoCavaille:
@degemer thanks for the review! ci passed! merging.

PR #1503, opened 2015-04-02
util/get_tags: use get_all_tags instead of get_only_instances

oremj:
get_only_instances is overkill for grabbing tags.

get_only_instances requires DescribeInstances, where get_all_tags only
requires the DescribeTags permission.


LeoCavaille:
Hi @oremj thanks for your contribution!
Looks legit, we'll review this soon to include it in 5.4 (the master is now frozen for 5.3 QA).

remh:
Just tested and it looks good to me! Let's release it with 5.3.0 as it's a bug fix.
Thanks @oremj !

PR #1504, opened 2015-04-02
[source] Revamp our source install script

LeoCavaille:
CHANGELOG:
* Use `-e` and `-u` options to be more resilient to errors
* Aim at full POSIX comaptibility and maximum portability:
  - The code passes w/o issues the `checkbashisms` script from debian
    or the `shellcheck` script
* Have better defined sections between variables/constants and code
* Deprecate the `IS_OPENSHIFT` env var, it's been removed from the
  cartridge
* Better error handling, use `trap` on `EXIT` which is how we to handle
  each line failure with `-e` set
* Better logging, instead of redirecting with tee every command stdout
  to a logfile and the console, we use a nice process substitution trick
  `exec 3>&1 1>>"$LOGFILE" 2>&1`
  fd 3 keeps a ref to the console stdout, and stderr/stdout are actually
  redirected to `LOGFILE`
* Misc cleanup

### DIFF
--- a/packaging/datadog-agent/source/setup_agent.sh
+++ b/packaging/datadog-agent/source/setup_agent.sh
@@ -8,7 +8,7 @@ set -u
 # SCRIPT KNOBS
 #######################################################################
 # Update for new releases, will pull this tag in the repo
-AGENT_VERSION="5.6.3"
+DEFAULT_AGENT_VERSION="5.6.3"
 # Pin pip version, in the past there was some buggy releases and get-pip.py
 # always pulls the latest version
 PIP_VERSION="6.0.8"
@@ -17,7 +17,9 @@ SUPERVISOR_VERSION="3.0b2"
 
 #######################################################################
 # OVERRIDABLE VARIABLES:
-#
+# $AGENT_VERSION
+#   The tag or branch from which the source install performs.
+#   Defaults to $DEFAULT_AGENT_VERSION
 # $DD_API_KEY
 #   Sets your API key in the config when installing.
 #   If not specified, the script will not install a default config.
@@ -39,6 +41,8 @@ SUPERVISOR_VERSION="3.0b2"
 #   your cartridge if you're using it.
 #######################################################################
 set +u # accept temporarily unbound vars, because we set defaults
+AGENT_VERSION=${AGENT_VERSION:-$DEFAULT_AGENT_VERSION}
+
 # If DD_HOME is unset
 if [ -z "$DD_HOME" ]; then
     # Compatibilty: dd_home used in lieu of DD_HOME
--- a/packaging/datadog-agent/source/setup_agent.sh
+++ b/packaging/datadog-agent/source/setup_agent.sh
@@ -434,7 +434,7 @@ if [ "$DD_START_AGENT" = "0" ]; then
 fi
 
 # on solaris, skip the test, svcadm the Agent
-if [ "$(uname)" = "SunOs" ]; then
+if [ "$(uname)" = "SunOS" ]; then
     # Install pyexpat for our version of python, a dependency for xml parsing (varnish et al.)
     # Tested with /bin/sh
     $PYTHON_CMD -V 2>&1 | awk '{split($2, arr, "."); printf("py%d%d-expat", arr[1], arr[2]);}' | xargs pkgin -y in
--- a/packaging/datadog-agent/source/setup_agent.sh
+++ b/packaging/datadog-agent/source/setup_agent.sh
@@ -406,7 +406,12 @@ else
     if [ "$(uname)" = "SunOS" ]; then
         $SED_CMD -i "s/# log_to_syslog: yes/log_to_syslog: no/" "$DD_HOME/agent/datadog.conf"
     fi
-    echo "disable_file_logging: True" >> "$DD_HOME/agent/datadog.conf"
+    # Setting up logging
+    # Needed to avoid "unknown var $prog_log_file"
+    log_suffix="_log_file"
+    for prog in collector forwarder dogstatsd jmxfetch; do
+      echo "$prog$log_suffix: $DD_HOME/logs/$prog.log" >> "$DD_HOME/agent/datadog.conf"
+    done
     chmod 640 "$DD_HOME/agent/datadog.conf"
 fi
 print_done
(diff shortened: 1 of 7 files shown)


utils/process.py
```diff
@@ -0,0 +1,52 @@
...
+    """
+    Check if a pid exists.
+    Lighter than psutil.pid_exists
+    """
+    if psutil:
+        return psutil.pid_exists(pid)
+
+    # We don't support windows for now
+    if Platform.is_windows():
+        raise NotImplementedError("pid_exists is not implemented for Windows")
```
remh: You should have a look into this:
https://github.com/DataDog/dd-agent/blob/146339a859ee81ed9b3d6d6fd19f101667d1aa3e/jmxfetch.py#L237-L248

It's a hacky but working way to check if a pid is running on windows

packaging/datadog-agent/source/setup_agent.sh
```diff
@@ -1,465 +1,529 @@
...
-# Define some helpers #
-#######################
+DD_START_AGENT=${DD_START_AGENT:-1}
 
-dogweb_reporting_failure_url="https://app.datadoghq.com/agent_stats/report_failure"
-email_reporting_failure="support@datadoghq.com"
-agent_help_page="http://docs.datadoghq.com/guides/basic_agent_usage/"
-see_agent_on_datadog_page="https://app.datadoghq.com/infrastructure"
+if [ -n "$IS_OPENSHIFT" ]; then
+    printf "IS_OPENSHIFT is deprecated and won't do anything\n"
```
remh: Should we stop the execution here and give better instructions or it doesn't matter and will just do the right thing ?

utils/process.py
LeoCavaille: Nice! Thanks, I'll add that to the new function, and we can reuse it in `jmxfetch.py` cleanly.

packaging/datadog-agent/source/setup_agent.sh
LeoCavaille: It doesn't really matter, right now it takes the same path as the _regular_ install. This warning would be mostly for us actually just to know that we should get rid of this var in other places. I plan to get rid of it as soon as we've updated our openshift cartridge.

utils/process.py
remh: JMXFetch doesn't use it anymore since we moved jmxfetch to its supervisor entry

LeoCavaille:
Sample output from a clean install:
screen shot 2015-04-01 at 7 37 06 pm

LeoCavaille:
Updated from your review @remh thanks. A little checklist (feel free to edit) for the source install
- [ ] debian sid
- [ ] debian stable
- [x] ubuntu precise
- [ ] archlinux
- [ ] solaris
- [ ] OSX
- [ ] Raspberry Pi
- [ ] OpenShift
- [ ] ...

All you have to do is run
```
sh -c "$(curl -L https://raw.githubusercontent.com/DataDog/dd-agent/leo/sourceerrors/packaging/datadog-agent/source/setup_agent.sh)"
```

remh:
Thanks! We should run it with bash on MacOS as it's what's recommended in the instructions.

degemer:
I tested on Mac (PR #1539), it works with this: https://github.com/DataDog/dd-agent/commit/0f38c5d0c2ba3fd910a454af0d2ce349665a3441#diff-940375087621857deb4576fdb2d5fb3aR428, `supervisor` is needed also on Mac, `launchd` is only responsible for the "daemonisation" of the `supervisor` process.
However when the agent fails (`echo $DDDD` for instance), it somehow displays the error twice:
```bash
* Uncompressing tarball
Done
* Setting up a datadog.conf generic configuration file
It looks like you hit an issue when trying to install the Datadog agent.
###
Done
* Setting up a datadog.conf generic configuration file
dd-agent/packaging/datadog-agent/source/setup_agent.sh: line 398: DDDDD: unbound variable
It looks like you hit an issue when trying to install the Datadog agent.
###
###
Do you want to send a failure report to Datadog (Content of the report is in /Users/quentinmadec/.datadog-agent/ddagent-install.log)? (y/n)
```

LeoCavaille:
Thanks @degemer, aware of this problem I had a WIP locally I'll merge that with yours and update the PR soon.

olivielpeau:
I've rebased the PR and fixed a few things in the metrics submission test, so I'm going to use a new checklist of tested platforms:
- [ ] debian sid
- [ ] debian stable
- [x] ubuntu precise
- [ ] archlinux
- [ ] solaris
- [ ] ~~OSX~~ (not supported anymore)
- [ ] Raspberry Pi
- [ ] OpenShift
- [ ] freebsd

olivielpeau:
I've made a few other changes and tested the script on the most common platforms (see updated checklist), this should be ready for a final review.

benmccann:
Just FYI, this needs a rebase

packaging/datadog-agent/source/setup_agent.sh
```diff
@@ -1,467 +1,515 @@
...
+set -u
+
+#######################################################################
+# SCRIPT KNOBS
+#######################################################################
+# Update for new releases, will pull this tag in the repo
+DEFAULT_AGENT_VERSION="5.6.3"
+# Pin pip version, in the past there was some buggy releases and get-pip.py
+# always pulls the latest version
+PIP_VERSION="6.0.8"
```
LeoCavaille: might want to upgrade to 6.1.1 to get the latest bugfixes of the 6.x branch (that's also the version used in omnibus software)

LeoCavaille:
Also just wondering if the agent can be installed from source in Travis, test :allthethings: (that can be done later).

olivielpeau:
@LeoCavaille Thanks for the suggestions, I've upgraded both `pip` and `supervisor` to the versions we use in omnibus-software.

@remh I've retested the script on Ubuntu, FreeBSD and SmartOS, and haven't found any issue. This should be mergeable.

packaging/datadog-agent/source/supervisor.conf
```diff
@@ -19,7 +19,6 @@ serverurl = unix://%(here)s/../run/agent-supervisor.sock
 
 [program:collector]
 command=python agent/agent.py foreground --use-local-forwarder
-stdout_logfile=supervisord/logs/collector.log
```
remh: Looks like we are just removing that without replacing that and we are ending without any logs.

remh:
We won't be able to release that at once, as it's pointing to 5.6.3 but 5.6.3 don't have the proper files. We'll need to be careful to release that properly.

packaging/datadog-agent/source/supervisor.conf
olivielpeau: @remh Actually commit 2aca54b (in this PR) sets up `datadog.conf` in `setup_agent.sh` so that all the components of the agent log to the `logs/` directory, instead of using supervisor to redirect the stdout.

This could be an issue however if the new script is used on a machine that already has a `datadog.conf` file created by the old script. In that case the `datadog.conf` file would remain untouched by `setup_agent.sh` and we'd end up with no logs from the agent.

Do you think this case is worth changing what the script currently does?

olivielpeau:
That's correct.

We could maybe use a transitional 5.6.4 source-install-only version, with the following release process:
1. Set `DEFAULT_AGENT_VERSION` in the PR to `5.6.4`
2. Cherry-pick the commits of this PR to the `5.6.x` branch
3. Do some QA on the `5.6.x` branch
4. Tag the `5.6.4` release on the `5.6.x` branch
5. Merge the PR to master

olivielpeau:
So after discussing this I'll actually split this PR in two:
- the first one will include everything but the changes on the `setup_agent.sh` script, we'll be able to merge it right away
- the second one will only include the changes on the script, with `DEFAULT_AGENT_VERSION` to `5.7.0`, we'll merge it right after the 5.7.0 release

That way, until we merge the second PR the old script will still be on master and will fetch the files from the `5.6.3` tag.

remh:
:+1:

olivielpeau:
Superseded by #2198 and #2199 in order to release the changes, closing.

packaging/datadog-agent/source/supervisor.conf
remh: Yeah we should handle this case. 
We should sed the file regardless of it was existing or not i think.

PR #1505, opened 2015-04-02
calculate dropped connections directly in the agent

kky:
To avoid any irregularities in roll-ups / averages, etc. Also fix a
typo (“conn”=>”handled”)

### DIFF
--- a/checks.d/nginx.py
+++ b/checks.d/nginx.py
@@ -93,7 +93,7 @@ class Nginx(AgentCheck):
             handled = int(parsed.group(2))
             requests = int(parsed.group(3))
             output.extend([('nginx.net.conn_opened_per_s', conn, tags, 'rate'),
-                           ('nginx.net.conn_handled_per_s', conn, tags, 'rate'),
+                           ('nginx.net.conn_dropped_per_s', conn-handled, tags, 'rate'),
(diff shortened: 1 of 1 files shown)


checks.d/nginx.py
```diff
@@ -93,7 +93,7 @@ def parse_text(cls, raw, tags):
             handled = int(parsed.group(2))
             requests = int(parsed.group(3))
             output.extend([('nginx.net.conn_opened_per_s', conn, tags, 'rate'),
-                           ('nginx.net.conn_handled_per_s', conn, tags, 'rate'),
+                           ('nginx.net.conn_dropped_per_s', conn-handled, tags, 'rate'),
```
remh: Thanks a lot for this @kky !

Tiny nitpick could you add a whitespace on each side of the - sign to respect PEP8 code style please ?

Soit would be:
```python
('nginx.net.conn_dropped_per_s', conn - handled, tags, 'rate'),
```

Thanks again!

alq666:
Will it affect existing dashboards and alerts or is this a new metric altogether?

kky:
@remh done
@alq666 this is a new metric altogether; conn_handled_per_s never made it into the wild.

remh:
Looks good to me. Thanks!

PR #1506, opened 2015-04-02
[tests] Clean before cache

degemer:
Some files are regenerated every time the test is launched, and so were
triggering the upload of a new cache.
This commit should fix this devious behaviour.
It also fixes the issue with pylint not found by Travis-caching the
python bin directory.

& Use the cache only internally
And not on forks, this was already done by Travis to protect our
"secure env variables", this commit just removes all calls to the cache
in this case (previously, it was called, but nothing happened).

Travis doc about "Secure variables":
http://docs.travis-ci.com/user/environment-variables/#Secure-Variables

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,8 +12,7 @@ python:
 
 cache:
   directories:
-    - $HOME/virtualenv/python2.6.9/lib/python2.6/site-packages
-    - $HOME/virtualenv/python2.7.9/lib/python2.7/site-packages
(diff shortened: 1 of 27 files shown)


degemer:
This is a test and will be deleted. #test

degemer:
Still playing with it, our cache is :+1: , but Travis cache uploads every time b/c of the `pip install --upgrade pip setuptools`.

LeoCavaille:
okay @degemer maybe we can get rid of that, or pin the versions if needed. In the meantime I'll merge this one thanks.

PR #1509, opened 2015-04-03
[flare] Warn user if not root

degemer:
As otherwise `supervisorctl status` won't be collected.

Ask for email even to update a case, as it is now the 'id' to check if
it is the good case (ie a case open by this customer).


degemer:
This is a test and will be deleted. #test

remh:
Looks good to me!

PR #1511, opened 2015-04-06
Improvements to the supervisor integration

remh:
- Send status OK when we can connect to the daemon
- Rename service check
- Use proper nose API


LeoCavaille:
Looks good to me. Tests passed for the `supervisord` flavor, merging! Thanks

Issue #1495, opened 2015-03-31
[docs] add a CONTRIBUTING.md


degemer:
We should take some time to write a CONTRIBUTING.md, because it is now displayed on GH when opening a PR/issue: https://help.github.com/articles/setting-guidelines-for-repository-contributors/

LeoCavaille:
:+1:

PR #1515, opened 2015-04-06
[doc] add a CONTRIBUTING.md :memo:

degemer:
Based on dd-agent wiki. Add some more content to tests/README.md.

This is only a first version, let's see what we can do with it.

cc @technovangelist 

Fix https://github.com/DataDog/dd-agent/issues/1495

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,220 @@
+# Contributing to Datadog Agent
+
+:tada: First of all, thanks for contributing! :tada:
+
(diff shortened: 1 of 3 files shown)


technovangelist:
why do two of the files have a md extension and the 3rd has a .markdown extension. Shouldnt they all be .md?

degemer:
Both extension (`.md` & `.markdown`) are valid markdown, but that's true that for consistency we should have the same extension everywhere. I will update `README.markdown` to `README.md`.

PR #1512, opened 2015-04-06
[Supervisord]Make sure to always send the can_connect service check

remh:

PR #1516, opened 2015-04-07
[Supervisord] Fix expected error message

remh:
I changed the error message in #1512 but didn’t change the test
accordingly.

This fixes that issue.


LeoCavaille:
Failed on a stupid http check test. We need to make them reliable.

PR #1519, opened 2015-04-07
[flare] use proxy settings if present

degemer:
Picked from @yannmh work on the agent governor.

### DIFF
--- a/config.py
+++ b/config.py
@@ -645,6 +645,34 @@ def get_proxy(agentConfig, use_system_settings=False):
     return None
 
 
+def format_proxy_settings(agent_config):
+    """
+    Returns proxy settings in a format compliant with `requests` package.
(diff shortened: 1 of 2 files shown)


remh:
I'm not completely sure it's needed. 
The flare command won't be run as dd-user, so the environment settings should be properly set for the user running the flare command. 

Requests will then properly get these settings.

I think it's fine to wait before really adding this feature as it would add code that i'm not sure is required.

degemer:
Sure, but since we advise them to use `sudo`, it could break the environment settings. 
It's not really important anyway (not blocking, the customer can still send logs by email), and it's already in https://github.com/DataDog/dd-agent/pull/1468 (just need to add 3 lines in `flare`), so I'm going to close and delete it.

PR #1521, opened 2015-04-08
[phpfpm] Use monotonic counts instead of counters

LeoCavaille:
Otherwise it doesn't produce the expected results, with an increment the
graph will have a monotonically increasing curve whereas with a
monotonic count it will only report the difference since the last flush
which is what we want because the value given by FPM is never reset.


hkaj:
LGTM. The failed builds are only Travis connection issues?

LeoCavaille:
Yep looks like unrelated failures. Thanks for the review! Merging

PR #1524, opened 2015-04-09
Small mistake fixed in check for Redis.

elafarge:
A very small one but it was making it impossible to monitor redis if you set the slowlog-max-len config key in the redisdb.yaml file (as shown in redisdb.yaml.example).


yannmh:
:+1: 

Can you reword your commit using these conventions ?

PR #1525, opened 2015-04-09
[redisdb] "Variable referenced before assignment" error solved

elafarge:
(wrong variable name was used).

### DIFF
--- a/tests/test_redis.py
+++ b/tests/test_redis.py
@@ -235,23 +235,66 @@ class TestRedis(AgentCheckTest):
             'port': port
         }
 
-
         db = redis.Redis(port=port, db=14)  # Datadog's test db
+        
+        # Tweaking Redis's config to have the test run faster
+        old_sl_thresh = db.config_get('slowlog-log-slower-than')['slowlog-log-slower-than']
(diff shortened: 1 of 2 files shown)


LeoCavaille:
@elafarge thanks for the fix.
Can you add one more test to the CI that exercises this code path, we should have caught it if it was the case! You can another commit on top of this one to improve our test suite (see `tests/test_redis.py`).

elafarge:
Sure, I'll do that.

tests/test_redis.py
```diff
@@ -235,6 +235,32 @@ def test_slowlog(self):
...
+            "redis_host:localhost", "redis_port:{0}".format(port)])
+
+
+    def test_custom_slowlog(self):
+        port = NOAUTH_PORT
+        test_key = "testkey"
+        instance = {
+            'host': 'localhost',
+            'port': port,
+            'slowlog-max-len': 129
```
remh: You could set it to a low number e.g. 1

then try to generate at least 2 slow queries but make sure that you are actually getting 1 ?

remh:
Nice, could you squash your commits into a single one, then we'll merge

PR #1527, opened 2015-04-09
[riak] remove custom tags from service_checks_tags

degemer:
Since we don't support custom tags for service_checks for now.


LeoCavaille:
:+1: thanks

PR #1528, opened 2015-04-09
[core] freeze pymongo & A-Z

degemer:
Since pymongo 3.0 is out.


LeoCavaille:
Thanks for fixing the build. That's cool!
What's uncool is that we need to update the mongo check to work with 3.x at some point.
What's still cool is that we can catch those regressions by unpinning.
Right now I don't have a proper solution, I'll try to think about it for the future...

PR #1533, opened 2015-04-13
Fix some pylint issues in the CI.

LeoCavaille:
See commit messages for details but overall:
1. call pylint correctly
2. fix the lint errors we didn't catch before.


LeoCavaille:
CI is passing again!

PR #1534, opened 2015-04-13
[Multi] Fix service check tags

remh:
- etc: do not use custom tags
- Docker: Remove custom tags
- kyototycoon: add missing tags on status OK
- Marathon: Remove python2.6 incompatible code
- Mysql: Add support for socket connection
- PHP-FPM: Remove custom tags
- Redis: Use only one set of tags
- Zookeeper: Add tags to service check

--
This will likely break the CI and i'll try to add a few tests so this is not ready to merge yet.

cc: @conorbranagan , @LeoCavaille

### DIFF
--- a/checks.d/php_fpm.py
+++ b/checks.d/php_fpm.py
@@ -93,9 +93,7 @@ class PHPFPMCheck(AgentCheck):
         return pool_name
 
     def _process_ping(self, ping_url, auth, tags, pool_name):
-        sc_tags = tags[:]
-        if pool_name is not None:
-            sc_tags.append("pool:{0}".format(pool_name))
+        sc_tags = ["ping_url:{0}".format(ping_url)]
 
         try:
             # TODO: adding the 'full' parameter gets you per-process detailed
--- a/checks.d/zk.py
+++ b/checks.d/zk.py
@@ -50,6 +50,7 @@ class ZookeeperCheck(AgentCheck):
         expected_mode = (instance.get('expected_mode') or '').strip()
         tags = instance.get('tags', [])
         cx_args = (host, port, timeout)
+        sc_tags = ["host:{0}".format(host), "port:{0}".format(port)]
 
         # Send a service check based on the `ruok` response.
         try:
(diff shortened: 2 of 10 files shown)


checks.d/zk.py
```diff
@@ -50,6 +50,7 @@ def check(self, instance):
         expected_mode = (instance.get('expected_mode') or '').strip()
         tags = instance.get('tags', [])
         cx_args = (host, port, timeout)
+        sc_tags = ["host:{0}".format(host), "port:{0}".format(port)]
```
conorbranagan: zookeeper's service_checks.json file will need an update for this

checks.d/php_fpm.py
```diff
@@ -93,9 +93,7 @@ def _process_status(self, status_url, auth, tags):
         return pool_name
 
     def _process_ping(self, ping_url, auth, tags, pool_name):
-        sc_tags = tags[:]
-        if pool_name is not None:
-            sc_tags.append("pool:{0}".format(pool_name))
+        sc_tags = ["ping_url:{0}".format(ping_url)]
```
conorbranagan: php-fpm service_checks.json file will need an update for this

conorbranagan:
have you confirmed that the tags changes won't break existing service check states? I'm worried about Zookeeper specifically because I think php-fpm is new

remh:
yeah php-fpm is new, haven't checked for zookeeper, but what would be the behavior now that we are adding those tags ?

conorbranagan:
Well because we're forcing `["host"]` as the groups, it won't change anything. But once we update that in the json file, it might mean that existing checks will go into a no-data state. Let me confirm that (a) this is actually true and (b) there are monitors w/ no-data for the ZK check.

conorbranagan:
Confirmed that there are at least 2 accounts that have no-data monitors set up for the Zookeeper check. But after looking at the Colvin code I think it will just ignore anything (during evaluation) that doesn't have _all_ the requested groups. So I think we're safe to make this change.

So a big :+1: from me

LeoCavaille:
Yep, looks good. The `port:unix_socket` tag makes me a bit sad though, no way around it?

LeoCavaille:
@remh if you can comment on this one, it's the last outstanding PR for 5.3.

LeoCavaille:
OK confirmed. We'll need to make the backend evolve to get better flexiblity.

PR #1535, opened 2015-04-13
[mesos] Improve mesos integration

DorianZaccaria:
* Add a check for mesos masters, only the leader will report metrics
* Add a check for mesos slave, slaves will report metrics from the selected
tasks only if the task is running on the node
* Add a mocked tests for mesos integration


alq666:
@DorianZaccaria 

1. what's the different between `cpu_percent` and `master.cpu_percent`. Should we use tags instead? (This question applies to more metrics).
1. `mesos.state` and `mesos.stats` are very similar and thus likely to trigger mistakes. What is in general, the difference between `role` and `state`?

DorianZaccaria:
I kept the metrics names from the old check to avoid breaking things during an upgrade. But we could also tag the old check as deprecated and keep it until the 6.0 release. For the later, we can reduce the number of metrics sent by the new check.
For state metrics, I think cpu_percent and master.cpu_percent are the same and are named differently for historical reasons. I have only one node in my environment so I can't check yet but I am working on it.
Stats metrics are the same across the cluster, as well as roles and frameworks. State metrics are more specific to the node itself (except for frameworks and slaves).
I noticed that some endpoints return the same metrics, for example `/metrics/snapshot` is included in `/stats.json`.

Let me know if you have any feedback .

alq666:
@DorianZaccaria let me know once you have a multi-node setup running; we can take a look at the metrics together.

DorianZaccaria:
Sure, I'm still in touch with mesosphere support to figure out why the cluster setup is failing.

remh:
@DorianZaccaria could you rebase your PR please ? it can't be merged right now.

checks.d/mesos_master.py
```diff
@@ -0,0 +1,246 @@
...
+Collects metrics from mesos master node, only the leader is sending metrics.
+"""
+# stdlib
+from hashlib import md5
+import time
+
+# project
+from checks import AgentCheck
+
+# 3rd party
```
remh: tiny tiny nitpick but as per pep8 the import order should be:
```
#stdlib
#3p
#project
```

checks.d/mesos_master.py
```diff
@@ -0,0 +1,246 @@
...
+
+    ROLE_RESOURCES_METRICS = {
+        'cpus'                                              : ('mesos.role.cpu', GAUGE),
+        'mem'                                               : ('mesos.role.mem', GAUGE),
+        'disk'                                              : ('mesos.role.disk', GAUGE),
+    }
+
+    # These metrics are aggregated only on the elected master
+    CLUSTER_TASKS_METRICS = {
+        'master/tasks_error'                                : ('mesos.cluster.tasks_error', GAUGE),
```
remh: Should these metrics be a rate instead ? Or a monotonic count ?

checks.d/mesos_master.py
```diff
@@ -0,0 +1,246 @@
...
+    def check(self, instance):
+        if 'url' not in instance:
+            raise Exception('Mesos instance missing "url" value.')
+
+        url = instance['url']
+        instance_tags = instance.get('tags', [])
+        default_timeout = self.init_config.get('default_timeout', 5)
+        timeout = float(instance.get('timeout', default_timeout))
+
+        json = self._check_leadership(url, timeout)
```
remh: You shouldn't use json as a variable name as it's also a python module name. Also it's not a very explicit variable name.

checks.d/mesos_master.py
```diff
@@ -0,0 +1,246 @@
...
+        if json:
+            tags = ['mesos_cluster:' + json['cluster'], 'mesos_pid:' + json['pid'], 'mesos_id:' + json['id'], 'mesos_node:master'] + instance_tags
```
remh: Nitpick but can you make this line a bit cleaner ? 

```python
tags = [
    'mesos_cluster:{0}'.format(json['cluster']),
    'mesos_pid:{0}'.format(json['pid']),
    'mesos_id:{0}'.format(json['id']),
    'mesos_node:master'
]
tags += instance_tags
```

remh: Also what are the mesos_id and mesos_pid ? 
Are they ephemeral ? If so we should likely remove these tags to avoid context explosion.

checks.d/mesos_master.py
```diff
@@ -0,0 +1,246 @@
...
+
+            if self.leader:
+                self.GAUGE('mesos.cluster.total_frameworks', len(json['frameworks']), tags=tags)
+
+                for framework in json['frameworks']:
+                    framework_tags = ['framework:' + framework['id'], 'framework_name:' + framework['name']] + tags
+                    self.GAUGE('mesos.framework.total_tasks', len(framework['tasks']), tags=framework_tags)
+                    resources = framework['used_resources']
+                    [v[1](self, v[0], resources[k], tags=framework_tags) for k, v in self.FRAMEWORK_METRICS.iteritems()]
```
remh: You shouldn't use list comprehension to call a function. It makes it harder to read compared to something like:

```python
for key_name, (metric_name, metric_func) in self.FRAMEWORK_METRICS.iteritems():
    metric_func(self, metric_name, resources[key_name], tags=framework_tags)
```

checks.d/mesos_master.py
```diff
@@ -0,0 +1,246 @@
...
+
+                json = self._get_master_roles(url, timeout)
```
remh: Same remark about variable names.

checks.d/mesos_master.py
```diff
@@ -0,0 +1,246 @@
...
+                if json is not None:
+                    for role in json['roles']:
+                        role_tags = ['mesos_role:' + role['name']] + tags
+                        self.GAUGE('mesos.role.frameworks', len(role['frameworks']), tags=role_tags)
```
remh: nitpick but:
```
mesos.role.frameworks.count
```
is a better name for this metric.

checks.d/mesos_master.py
```diff
@@ -0,0 +1,246 @@
...
+                        self.GAUGE('mesos.role.weight', role['weight'], tags=role_tags)
+                        [v[1](self, v[0], role['resources'][k], tags=role_tags) for k, v in self.ROLE_RESOURCES_METRICS.iteritems()]
```
remh: Same remark about list comprehension usage.

checks.d/mesos_master.py
```diff
@@ -0,0 +1,246 @@
...
+
+            json = self._get_master_stats(url, timeout)
+            if json is not None:
+                if self.leader:
+                    metrics = {}
+                    for d in (self.CLUSTER_TASKS_METRICS, self.CLUSTER_SLAVES_METRICS,
```
remh: Instead of merging all these dictionaries into one i would just iterate over the list of dictionaries, it would be more efficient and there shouldn't be any conflicts in the key names of these dictionaries anyway.

checks.d/mesos_master.py
```diff
@@ -0,0 +1,246 @@
...
+# stdlib
+from hashlib import md5
+import time
+
+# project
+from checks import AgentCheck
+
+# 3rd party
+import requests
+from version import Version
```
remh: Do we really need this library ? 

Python is pretty good to compare lists of integers already 

```python
[3,5,2] > [0,8,4] # True
[0, 9, 9] > [1,0,0] # False
```

checks.d/mesos_master.py
```diff
@@ -0,0 +1,246 @@
...
+            msg = e.message
+            status = AgentCheck.CRITICAL
+        finally:
+            if self.SERVICE_CHECK_NEEDED:
+                self.service_check(self.SERVICE_CHECK_NAME, status, tags=tags,
+                                   message=msg)
+                self.SERVICE_CHECK_NEEDED = False
+            if status is AgentCheck.CRITICAL:
+                self.warning(msg)
+                return None
```
remh: Should we raise an Exception here and stop the execution of the check run here ?

checks.d/mesos_master.py
```diff
@@ -0,0 +1,246 @@
...
+            if json is not None:
+                if self.leader:
+                    metrics = {}
+                    for d in (self.CLUSTER_TASKS_METRICS, self.CLUSTER_SLAVES_METRICS,
+                              self.CLUSTER_RESOURCES_METRICS, self.CLUSTER_REGISTRAR_METRICS,
+                              self.CLUSTER_FRAMEWORK_METRICS, self.SYSTEM_METRICS, self.STATS_METRICS):
+                        metrics.update(d)
+                else:
+                    metrics = self.SYSTEM_METRICS
+                [v[1](self, v[0], json[k], tags=tags) for k, v in metrics.iteritems()]
```
remh: Same remark here

checks.d/mesos_master.py
```diff
@@ -0,0 +1,246 @@
...
+        try:
+            r = requests.get(url, timeout=timeout)
+            if r.status_code != 200:
+                self._status_code_event(url, r, aggregation_key)
+                status = AgentCheck.CRITICAL
+                msg = "Got %s when hitting %s" % (r.status_code, url)
+            else:
+                status = AgentCheck.OK
+                msg = "Mesos master instance detected at %s " % url
+        except requests.exceptions.Timeout as e:
```
remh: Do we need these events ? Maybe the service checks would be enough ?

checks.d/mesos_slave.py
```diff
@@ -0,0 +1,200 @@
+"""Mesos Slave check
```
remh: Most of the comments on mesos_master should be applied here too.

remh:
Great work @DorianZaccaria .
How do you think this should replace the old mesos check and how do you picture the transition from the old to the new checks ?

DorianZaccaria:
@remh I think this one can be merged before the 5.4 freeze.

remh:
@DorianZaccaria what's the status of the metric type updates ?

DorianZaccaria:
@remh Done and updated.

checks.d/mesos_slave.py
```diff
@@ -0,0 +1,186 @@
...
+        stats_metrics = self._get_stats(url, timeout)
+        if stats_metrics:
+            tags = tags if tags else instance_tags
+            metrics = [self.SLAVE_TASKS_METRICS, self.SYSTEM_METRICS, self.SLAVE_RESOURCE_METRICS,
+                      self.SLAVE_EXECUTORS_METRICS, self.STATS_METRICS]
+            for m in metrics:
+                for key_name, (metric_name, metric_func) in m.iteritems():
+                    metric_func(self, metric_name, stats_metrics[key_name], tags=tags)
+
+        self.SERVICE_CHECK_NEEDED = True
```
remh: nitpick, but usually upper case variables are for constants which is not the case here.

tests/checks/mock/test_mesos_slave.py
```diff
@@ -0,0 +1,45 @@
...
+            ]
+        }
+
+        klass = get_check_class('mesos_slave')
+        with patch.object(klass, '_get_state', _mocked_get_state):
+            with patch.object(klass, '_get_stats', _mocked_get_stats):
+                check = klass('mesos_slave', {}, {})
+                self.run_check(config)
+                time.sleep(1)
+                self.run_check(config)
```
remh: Tiny nitpick, you can use the run_check_twice helper instead
https://github.com/DataDog/dd-agent/blob/master/tests/checks/common.py#L160

remh:
Looks good!
a few tiny nitpicks to fix and then it's ready to merge.

remh:
:boom: :+1:

PR #1537, opened 2015-04-13
[pgbouncer] Raise connection error instead of pass

degemer:
Otherwise it fails with
> UnboundLocalError: local variable 'connection' referenced before
> assignment

instead of the real connection error.


LeoCavaille:
Good catch. Merging.

PR #1539, opened 2015-04-14
[source] log properly (package style) & cleanup

degemer:
Previsouly logging was done using supervisor & stdout, totally different
from the package logging. There are options in the configuration to
change the log location, so let's use them instead of doing this trick!

In order to do this, these changes were made:
* remove supervisor processes logging (except `supervisor` itself)
* add a step to configure log location in the source install script

In order to clean weird conf location & others:
* move `supervisor.conf` from `supervisord/` to `agent/` with
`datadog.conf`
* update `bin/agent` to reflect this change
* remove `bin/info` as it is handled in `bin/agent` now (was probably
still here for _legacy_ reasons)
* put launchd logs in the same directory as other logs (`logs/`)
* update the source install script to reflect this change

This PR is needed for `flare` to work on source install.

### DIFF
--- a/daemon.py
+++ b/daemon.py
@@ -19,8 +19,8 @@ import logging
 import errno
 import signal
 
-# 3p
-from psutil import pid_exists
+# project
+from utils.process import pid_exists
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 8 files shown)


degemer:
Replaced by https://github.com/DataDog/dd-agent/pull/1540.

PR #1542, opened 2015-04-14
[jmx] invalid_checks cast bug :bug:

yannmh:
Save exception as a string, not a Python object

### Before
jmx_status_python.yaml
```
invalid_checks:
  jmx: !!python/object/apply:__main__.InvalidJMXConfiguration [A numeric port must
      be specified]
timestamp: 1429048613.553269
```
Printing agent info was raising an exception:
```
ConstructorError: while constructing a Python object
cannot find 'InvalidJMXConfiguration' in the module '__main__'
  in "/tmp/jmx_status_python.yaml", line 2, column 8
```

### After
jmx_status_python.yaml
```
invalid_checks: {jmx: A numeric port must be specified}
timestamp: 1429050253.38898
```

### DIFF
--- a/jmxfetch.py
+++ b/jmxfetch.py
@@ -165,7 +165,7 @@ class JMXFetch(object):
                             tools_jar_path = check_tools_jar_path
                 except InvalidJMXConfiguration, e:
                     log.error("%s check does not have a valid JMX configuration: %s" % (check_name, e))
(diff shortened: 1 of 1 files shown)


degemer:
It fixes this bug (tested), so :+1:.

PR #1543, opened 2015-04-14
Add basic mandrill agent check

brettlangdon:
This PR adds a new agent check for mandrill. If you give it an api key it will go and fetch some basic information from the mandrill api about your account for the api's user, for tags and urls.


remh:
Thanks a lot @brettlangdon .
This should likely be a server side integration.
What do you think ?

brettlangdon:
@remh not opposed to that idea.

PR #1544, opened 2015-04-15
[system] Attempt to make source install compliant

elafarge:
So far, the behaviour has been modified only for Debian based distros,
on which our fix is supposed to work.

### DIFF
--- a/packaging/datadog-agent/source/install_agent.sh
+++ b/packaging/datadog-agent/source/install_agent.sh
@@ -5,7 +5,7 @@ logfile="ddagent-install.log"
 gist_request=/tmp/agent-gist-request.tmp
 gist_response=/tmp/agent-gist-response.tmp
 
-if [ $(which curl) ]; then
+if [ $(command -v curl) ]; then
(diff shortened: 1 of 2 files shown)


packaging/datadog-agent/source/install_agent.sh
```diff
@@ -144,7 +144,11 @@ else
 fi
 
 printf "\033[34m* Starting the Agent...\n\033[0m\n"
-$sudo_cmd /etc/init.d/datadog-agent restart
+if [ $OS = "Debian" ]; then
+    $sudo_cmd service datadog-agent restart
+else
+    $sudo_cmd /etc/init.d/datadog-agent restart
+fi
```
LeoCavaille: Same questions as the other PR actuatlly. And here why do we use `service` and not `invoke-rc.d` then?

elafarge:
I revisioned the .service file according to this piece of documentation : https://datadog.zendesk.com/agent/tickets/24878 . 

The type of service is FORKED. Indeed it matches

PR #1545, opened 2015-04-15
[NTP] Increase min collection interval and log in debug if it can't conn...

remh:
...ect


LeoCavaille:
Maybe we should have passed that message to the service check?

LeoCavaille:
Otherwise I'm good, 900 seems a bit arbitrary but why not...

remh:
900 is 15 min, i.e. 4 times an hour which is kind of the max recommended http://www.pool.ntp.org/en/vendors.html

PR #1547, opened 2015-04-15
[ci] replace sleep_for by a function

degemer:
`Wait.for` can receive either:
- a port (`Wait.for 7199`), and will return when the port is open
- an url (`Wait.for 'http://localhost:8080'`), and will return when
  receiving a 2xx response code
- a file (`Wait.for '/tmp/started'`), and will return when the file
  exists

It has a default timeout of 5s, which can be configured as its second
parameter: `Wait.for 'http://localhost:8080', 10`

### DIFF
--- a/ci/etcd.rb
+++ b/ci/etcd.rb
@@ -36,7 +36,14 @@ namespace :ci do
 
     task :before_script => ['ci:common:before_script'] do
       sh %(cd $VOLATILE_DIR && #{etcd_rootdir}/etcd >/dev/null &)
-      sleep_for 10
+      # Waiting for etcd to start
+      Wait.for 'http://localhost:4001/v2/stats/self'
+      Wait.for 'http://localhost:4001/v2/stats/store'
+      10.times do
+        sh %(curl -s http://127.0.0.1:2379/v2/keys/message\
+             -XPUT -d value="Hello world" >/dev/null)
+        sh %(curl -s http://127.0.0.1:2379/v2/keys/message > /dev/null)
+      end
     end
 
     task :script => ['ci:common:script'] do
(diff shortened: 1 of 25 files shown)


ci/etcd.rb
```diff
@@ -36,7 +36,14 @@ def etcd_rootdir
...
       sh %(cd $VOLATILE_DIR && #{etcd_rootdir}/etcd >/dev/null &)
-      sleep_for 10
+      # Waiting for etcd to start
+      Wait.for 'http://localhost:4001/v2/stats/self'
+      Wait.for 'http://localhost:4001/v2/stats/store'
+      10.times do
+        sh %(curl -s http://127.0.0.1:2379/v2/keys/message\
+             -XPUT -d value="Hello world" >/dev/null)
+        sh %(curl -s http://127.0.0.1:2379/v2/keys/message > /dev/null)
+      end
```
yannmh: Why are you addding this `10.times` bloc ?

degemer: I added it to generate some metrics which otherwise weren't always there (waiting for 10s did the trick previously, but not anymore :racehorse: )

ci/supervisord.rb
```diff
@@ -49,7 +49,7 @@ def supervisor_rootdir
 
     task :cleanup => ['ci:common:cleanup'] do
       sh %(kill `cat $VOLATILE_DIR/supervisor/supervisord.pid`)
-      sh %(rm -rf $VOLATILE_DIR/supervisor)
+      # sh %(rm -rf $VOLATILE_DIR/supervisor)
```
yannmh: Should we just remove this line ?

yannmh:
Great improvement ! Thanks @degemer

ci/supervisord.rb
degemer: Rhaaa, it's a debug comment, thanks for catching it!

degemer:
Thanks @yannmh, removed my debug comment, I will `Wait.for` Travis before merging.

LeoCavaille:
Rebased on top of the current master, to solve conflicts with the new tests refactoring.

degemer:
:white_check_mark:

PR #1549, opened 2015-04-16
[swap stats] add swap in/out check to checks.d

### LINKED ISSUE #716: Agent should collect system swap metrics
When doing system checks with the agent, we don't look at /proc/vmstat, so we don't get metrics like "swapped in" and "swapped out".  We should add this to our unix system checks.

talwai:
Uses `psutil` to collect the rate of memory swapped in and swapped out. New metrics are labelled `system.swap.swapped_in` and `system.swap.swapped_out`.

Fixes #716

### DIFF
--- a/checks.d/system_swap.py
+++ b/checks.d/system_swap.py
@@ -1,17 +1,9 @@
 from checks import AgentCheck
-try:
-    import psutil
-except ImportError:
-    psutil = None
-
-B2MB  = float(1048576)
+import psutil
 
 class SystemSwap(AgentCheck):
 
     def check(self, instance):
-        if not psutil:
-            return {}
-
         swap_mem = psutil.swap_memory()
-        self.gauge('system.swap.swapped_in', swap_mem.sin / B2MB)
-        self.gauge('system.swap.swapped_out', swap_mem.sout / B2MB)
+        self.rate('system.swap.swapped_in', swap_mem.sin)
+        self.rate('system.swap.swapped_out', swap_mem.sout)
--- a/checks.d/system_swap.py
+++ b/checks.d/system_swap.py
@@ -1,0 +1,17 @@
+from checks import AgentCheck
+try:
+    import psutil
+except ImportError:
+    psutil = None
+
+B2MB  = float(1048576)
+
+class SystemSwap(AgentCheck):
+
+    def check(self, instance):
+        if not psutil:
+            return {}
+
+        swap_mem = psutil.swap_memory()
+        self.gauge('system.swap.swapped_in', swap_mem.sin / B2MB)
+        self.gauge('system.swap.swapped_out', swap_mem.sout / B2MB)
--- a/tests/test_system_swap.py
+++ b/tests/test_system_swap.py
@@ -1,0 +1,30 @@
+import mock
+import psutil
+
+from tests.common import AgentCheckTest
+
+B2MB  = float(1048576)
+
+class _PSUtilSwapStatsMock(object):
+    def __init__(self, sin, sout):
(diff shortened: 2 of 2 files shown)


checks.d/system_swap.py
```diff
@@ -0,0 +1,17 @@
...
+B2MB  = float(1048576)
+
+class SystemSwap(AgentCheck):
+
+    def check(self, instance):
+        if not psutil:
+            return {}
+
+        swap_mem = psutil.swap_memory()
+        self.gauge('system.swap.swapped_in', swap_mem.sin / B2MB)
```
remh: You should use a rate for these.
According to the psutil documentation, sin is 
```
the number of bytes the system has swapped in from disk (cumulative)
```

By submitting the metric as a rate, you will compute the # of bytes per second swapped in. 
Otherwise the graphs will just display a monotonically increasing value.

Also you should submit this value in bytes as per our discussion earlier.

checks.d/system_swap.py
```diff
@@ -0,0 +1,17 @@
+from checks import AgentCheck
+try:
+    import psutil
+except ImportError:
```
remh: I know i told you something different yesterday but after all i think we shouldn't manage the case where psutil is not present.

The only case where it's not present is when someone installs the agent from source on a system that doesn't have a compiler. And if that's the case, the check initialization will fail and report the missing import which is an acceptable behavior.

checks.d/system_swap.py
```diff
@@ -0,0 +1,11 @@
+from checks import AgentCheck
+import psutil
+
+B2MB  = float(1048576)
```
remh: looks like it's not needed anymore :)

tests/test_system_swap.py
```diff
@@ -0,0 +1,32 @@
...
+]
+
+class SystemSwapTestCase(AgentCheckTest):
+
+    CHECK_NAME = 'system_swap'
+
+    @mock.patch('psutil.swap_memory', side_effect=MOCK_PSUTIL_SWAP_STATS)
+    def test_system_swap(self, mock_swap_stats):
+        self.run_check_twice({"instances": [{}]})
+        self.assertMetric('system.swap.swapped_in', value=SWAP_IN_INCR, count=1)
```
remh: this will fail. The value will be a rate per second, so SWAP_IN_INCR divided by the number of seconds between the two iterations.

tests/test_system_swap.py
```diff
@@ -0,0 +1,32 @@
+import mock
+import psutil
+
+from tests.common import AgentCheckTest
+
+B2MB  = float(1048576)
```
remh: This isn't needed

tests/test_system_swap.py
talwai: @remh it looks like `run_check_twice` explicitly waits a second between checks which is why the rate evaluates correctly when I run the test locally
```
    def run_check_twice(self, config, agent_config=None, mocks=None,
                        force_reload=False):
        self.run_check(config, agent_config, mocks, force_reload)
        time.sleep(1)
        self.run_check(config, agent_config, mocks)
```

remh: This is not intuitive that it's relying on a sleep of 1 sec though. Can you add a comment ?

remh:
That looks great!   :boom: 
Tiny last nitpick, could you add a comment to your PR explaining what's it's trying to solve and how please ?

LeoCavaille:
:+1:

LeoCavaille:
You should also use a "Fixes #716" expression to link that to currently opened issues.

PR #1552, opened 2015-04-17
[core] pidfiles & pickles update

degemer:
It's a first proposition, what I did for the Mac Agent.
It uses `/var/run/dd-agent`. More details in the commits descriptions.


degemer:
Not working, I am going to try something else.

PR #1554, opened 2015-04-17
[jmx] fix windows service restarts

yannmh:


jmxfetch.py
```diff
@@ -105,6 +105,13 @@ def run(self, command=None, checks_list=None, reporter=None):
             log.exception("Error while initiating JMXFetch")
             raise
 
+    def should_run(self):
+        """
+        Windows Service purpose: should JMXFetch run ?
+        """
+        jmx_checks, _, _, _, _ = self._should_run(None)
```
remh: We'll call this twice if we run jmx checks. We shouldn't have to.

PR #1555, opened 2015-04-17
[deps] clean requirements files and sync them

LeoCavaille:
Now the requirements are on this repo located in 4 different files.
Change: all the versions are now pinned.
* requirements.txt
  This is the all the dependencies of the core and some checks.
  All the dependencies here are pure python and don't need anything
  to be installed (like headers or libs). So they should install on
  100% of our target machines.
  All the core dependencies *must* match this criteria to be as portable
  as possible.
* requirements-opt.txt
  All the other dependencies of the agent (mostly checks) that are
  depending on some other libraries or compilers or ...
* test-requirements.txt
  Deps needed only for the CI or to run the test suite.
* setup.py
  This file only applies for Windows-based builds, it has been updated
  to

Note: modified the source install script, but it's not using the new files
yet because it fetches the reqs file from the last tag.
I'll ship a modification so it uses the new reqs file in an upcoming source
install refactoring.

Cleanups:
- remove pinning of sub-dependencies, we shouldn't pin something in the
  requirements file if the dependency isn't explicitly imported in our
  code. Let pip resolve the mess.
- remove unused `ez_setup.py` file


LotharSee:
Great work! That's definitively a better practice and organization. Should avoid tons of issues.

LeoCavaille:
Rebased, it's mergeable again.

degemer:
It's finally easy to see/check the dependencies, thanks @LeoCavaille !
Maybe you could also rename `test-requirements.txt` in `requirements-test.txt`. (so that it's even easier to distinguish all requirements files)

Two other concerns (not 100% related to this PR):
* about the `psycopg2` issue (or any other dependency listed in `requirements-opt.txt`), it would be nice not to fail during the `setup_env` and `ci:run` tasks (maybe implement something like what you're doing with the [source install](https://github.com/DataDog/dd-agent/pull/1504/files#diff-940375087621857deb4576fdb2d5fb3aR370)).
* since we also freeze the dependencies in https://github.com/DataDog/omnibus-software, it would also be nice to have something to keep those requirements in sync. (probably update manually omnibus-software since it requires more details and automatically keep `requirements-*` in sync) Or just to have a way to check that they are synchronized.

LeoCavaille:
@degemer 2 commits have been added to the branch to implement your feedback.
For the last point, I agree, we should automate this. But it's another _nice to have_ feature that we can have in the future, I opened an issue to track it down #1578.

requirement-test.txt
```diff
@@ -0,0 +1,5 @@
+nose==1.3.4
```
degemer: An 's' is missing in the filename.

Rakefile
```diff
@@ -57,8 +57,8 @@ task 'setup_env' do
    `venv/bin/python venv/ez_setup.py`
    `wget -O venv/get-pip.py https://raw.github.com/pypa/pip/master/contrib/get-pip.py`
    `venv/bin/python venv/get-pip.py`
-   `venv/bin/pip install -r source-requirements.txt`
-   `venv/bin/pip install -r source-optional-requirements.txt`
+   `venv/bin/pip install -r requirements.txt`
+   `PIP_COMMAND=venv/bin/pip ./utils/pip-allow-failures.sh requirements-opt.txt`
```
degemer: It doesn't work (fails with `Errno::ENOENT: No such file or directory - PIP_COMMAND=venv/bin/pip ./utils/pip-allow-failures.sh requirements-opt.txt`), I managed to make it work with `export PIP_COMMAND=venv/bin/pip && ./utils/pip-allow-failures.sh requirements-opt.txt`. (first arg must be the exec ?)

utils/pip-allow-failures.sh
```diff
@@ -0,0 +1,22 @@
...
+while read dependency; do
+    dependency_stripped="$(echo "${dependency}" | sed -e 's/^[[:space:]]*//' -e 's/[[:space:]]*$//')"
+    # Skip comments
+    if [[ $dependency_stripped == \#* ]]; then
+        continue
+    # Skip blank lines
+    elif [ -z "$dependency_stripped" ]; then
+        continue
+    else
+        if $PIP_COMMAND install $PIP_OPTIONS "$dependency_stripped"; then
```
degemer: Since you are already displaying the result afterwards, maybe you can redirect the output to `/dev/null` ?

ci/common.rb
```diff
@@ -47,8 +47,10 @@ def section(name)
            -r requirements.txt\
            --cache-dir $PIP_CACHE\
            2>&1 >> $VOLATILE_DIR/ci.log)
+      sh %(PIP_OPTIONS="--cache-dir $PIP_CACHE" ./utils/pip-allow-failures.sh requirements-opt.txt\
```
degemer: I don't know how it is possible, but an error is still displayed 
```bash
PIP_OPTIONS="--cache-dir $PIP_CACHE" ./utils/pip-allow-failures.sh requirements-opt.txt           2>&1 >> $VOLATILE_DIR/ci.log
    Command "python setup.py egg_info" failed with error code 1 in /private/var/folders/fw/j46wyy091r38jhktdht1yprr0000gn/T/pip-build-Ms_ucm/psycopg2
pip install           -r requirements-test.txt           --cache-dir $PIP_CACHE            2>&1 >> $VOLATILE_DIR/ci.log
```
Maybe `pip` which always write to stdout...

degemer:
Aaaah, finally being able to run a `rake setup_env` without having to comment/delete `psycopg2`! :tada: :tada:
I think it pretty much closes https://github.com/DataDog/dd-agent/pull/1568.

:+1: for #1578

utils/pip-allow-failures.sh
LeoCavaille: I find it useful to not swallow the errors if the user still need to fix and install this or that dependency. It's just a wrapper, if someone wants to ignore the output he should do it.

Rakefile
LeoCavaille: Indeed, we need to use `ENV` instead I guess, it would be cleaner.

LeoCavaille:
Should be GTG @degemer thanks for the careful review.

degemer:
:+1: I was able to `rake setup_env` and `rake ci:run` without any issue !

LeoCavaille:
Fast-forward merge done, closing this PR.
Merged as:
* 2da6874860bfd8d02d28f3b5d3b4d912dd3f4c63
* 1c7f420d8a02c68454b733c7895d9398b5f66d62
* 26acb0cf95ad3bc9eaec42953e591bcaa3a5a6ca

PR #1561, opened 2015-04-20
[jmxfetch] fix windows bootloop :bug:

yannmh:
* [Windows] Fix JMXFetch bootloop issue when no JMX integration config is found
* [Linux][OSX] Better management of unexpected exit codes ( != 0)
* [Linux][OSX] [Windows] Limit JMXFetch attempts of restarts to 3
* [Linux][OSX][Windows] Fill a status when JMXFetch could not be launched

### DIFF
--- a/jmxfetch.py
+++ b/jmxfetch.py
@@ -1,11 +1,11 @@
 # std
-import os
-import logging
 import glob
+import logging
+import os
 import subprocess
-import time
 import sys
 import signal
+import time
 
 # datadog
 from util import get_os, yLoader, yDumper
@@ -68,6 +68,7 @@ class JMXFetch(object):
         self.check_frequency = DEFAULT_CHECK_FREQUENCY
 
         self.jmx_process = None
+        self.jmx_checks = None
 
     def terminate(self):
         self.jmx_process.terminate()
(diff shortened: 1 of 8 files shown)


jmxfetch.py
```diff
@@ -77,35 +78,59 @@ def _handle_sigterm(self, signum, frame):
...
                     log.exception("Error while writing JMX status file")
 
-            if len(jmx_checks) > 0:
-                self._start(
-                    java_bin_path, java_options, jmx_checks, command, reporter, tools_jar_path)
-                return True
+            if len(self.jmx_checks) > 0:
+                return self._start(self.java_bin_path, self.java_options, self.jmx_checks,
+                                   command, reporter, self.tools_jar_path)
+            else:
```
remh: Is that really needed if we exit with a 0 status code ?

packaging/datadog-agent/source/supervisord.conf
```diff
@@ -45,7 +45,7 @@ command=python agent/jmxfetch.py
 stdout_logfile=supervisord/logs/jmxfetch.log
 redirect_stderr=true
 priority=999
-startsecs=0
+startsecs=3
```
remh: I think what you really want to implement is "startretries".

Also you should likely implement "exitcodes" and exit with a given exit code that would be on that list when java cannot be found.

win32/agent.py
```diff
@@ -251,14 +264,21 @@ def __init__(self, agentConfig, hostname):
                       "the Agent is currently deployed.\n" % e.args[0])
 
         self.jmx_daemon = JMXFetch(confd_path, agentConfig)
+        self.jmx_daemon.configure()
+        self.is_enabled = self.jmx_daemon.should_run()
 
     def run(self):
-        log.debug("Windows Service - Starting JMXFetch")
-        self.jmx_daemon.run()
+        from config import initialize_logging; initialize_logging('windows_jmxfetch')
```
remh: the windows_jmxfetch logger is not configured. You shouldn't use that.

win32/agent.py
```diff
@@ -251,14 +264,21 @@ def __init__(self, agentConfig, hostname):
...
+        if self.is_enabled:
+            log.debug("Windows Service - Starting JMXFetch")
+            self.jmx_daemon.run()
+        else:
+            log.info("No JMXFetch integration found, not starting it.")
 
     def stop(self):
-        log.debug("Windows Service - Stopping JMXFetch")
-        self.jmx_daemon.terminate()
+        if self.is_enabled:
```
remh: This won't be executed, this should be removed.

win32/agent.py
```diff
@@ -124,6 +128,16 @@ def _restart_proc(self, proc_name):
         # Make a new proc instances because multiprocessing
         # won't let you call .start() twice on the same instance.
         old_proc = self.procs[proc_name]
+
+        # Watch JMXFetch restarts
```
remh: Can we do it in a more generic way for all the processes ?

packaging/datadog-agent/source/supervisord.conf
yannmh: By default, `startretries` is set to 3 .
The reason why I switched `startsecs` to 3, is because the Supervisor ignores `startretries` parameter when the program is able to start (i.e runs more than `startsecs` seconds).

jmxfetch.py
yannmh: I am not sure to understand your point: we don't necessarily exit with a 0 status code.

win32/agent.py
```diff
@@ -117,25 +122,49 @@ def _check_collector_blocked(self):
...
                 self._collector_failed_heartbeats = 0
 
-    def _restart_proc(self, proc_name):
+
+class ProcessWatchDog(object):
+    """
+    Monitor the attached process.
+    Restarts when it exits until the limit set is reached.
+    """
+    def __init__(self, name, process, max_restarts="inf"):
```
remh: the default shouldn't be +inf it should be a reasonable value, 5 seems good to me.

PR #1562, opened 2015-04-20
[ProcessCheck - Windows] Improve the PID retrieval

hkaj:


checks.d/process.py
```diff
@@ -11,6 +11,8 @@ class ProcessCheck(AgentCheck):
 
     SOURCE_TYPE_NAME = 'system'
 
+    CACHE_REFRESHING_FREQ = 20
```
hkaj: This is a convenient value for testing (quickly reached), but in a real situation we'll probably want to set it higher.

checks.d/process.py
```diff
@@ -42,9 +49,11 @@ def find_pids(self, search_string, exact_match, ignore_denied_access):
...
                         self.log.warning('Process disappeared while scanning')
                     except psutil.AccessDenied, e:
-                        self.log.error('Access denied to %s process' % string)
-                        self.log.error('Error: %s' % e)
-                        if not ignore_denied_access:
+                        if not ignore_denied_access and proc.pid not in catched_access_denied:
+                            self.log.error('Access denied to %s process' % string)
+                            self.log.error('Error: %s' % e)
+                            # allows to raise only once for each failing process
+                            self.catched_access_denied.append(proc.pid)
```
LeoCavaille: You always append to this list but never delete anything.
If the processes watched are rotating in a way it could be bad. You should be able to identify a unique process by using its creation time I guess?

LeoCavaille: In that case, a dict is better probably and the lookup will also be faster.

checks.d/process.py
```diff
@@ -186,9 +195,13 @@ def check(self, instance):
...
 
-        pids = self.find_pids(search_string,
-                              exact_match,
-                              ignore_denied_access)
+        if self.iteration == 0:
+            self.pid_cache = self.find_pids(search_string,
+                                            exact_match,
+                                            ignore_denied_access)
+        pids = self.pid_cache
+        self.iteration = (self.iteration + 1) % CACHE_REFRESHING_FREQ
```
LeoCavaille: Here you probably want `self.CACHE_REFRESHING_FREQ` ?

LeoCavaille:
@hkaj the idea looks good, but we could modify slightly the logic to be optimal & reliable. Also, could you rebase this branch on the current master, looks like one unrelated commit made its way there?
Thanks.

LeoCavaille:
Actually the whole logic of this is a bit :neutral_face: , because we cache the whole list of PIDs and not the "access denied" processes only which delays updates. Will reopen a new PR with the new logic.

PR #1563, opened 2015-04-20
[systemd] Package updated to be systemd compliant

elafarge:
The install script has been modified (it uses updated-rc.d to restart
the agent, which will implicitly call systemd if available) and the
datadog-agent.service systemd descriptor has been added as part of the
package (so that omnibus puts it in the .deb packages).

This service declaration is basically pluging supervisord into systemd,
nothing more (as shown here
https://gist.github.com/tonyseek/ddc7e3a89f0d78e685d6 for instance)

The supervisord start call has been ut in a separate script.
It allows us to have access to the PATH (/bin/sh is called to run the
script) while keeping a one-level forking scheme thanks to exec (/bin/sh
gets replaced by supervisord that will then fork and launch the
different components of our agent. Being a direct child of systemd, the
forked processes are too).

### DIFF
--- a/packaging/datadog-agent/source/install_agent.sh
+++ b/packaging/datadog-agent/source/install_agent.sh
@@ -5,7 +5,7 @@ logfile="ddagent-install.log"
 gist_request=/tmp/agent-gist-request.tmp
 gist_response=/tmp/agent-gist-response.tmp
 
-if [ $(which curl) ]; then
+if [ $(command -v curl) ]; then
     dl_cmd="curl -f"
 else
     dl_cmd="wget --quiet"
@@ -144,7 +144,11 @@ else
 fi
 
 printf "\033[34m* Starting the Agent...\n\033[0m\n"
-$sudo_cmd /etc/init.d/datadog-agent restart
+if command -v invoke-rc.d >/dev/null 2>&1; then
+    $sudo_cmd invoke-rc.d datadog-agent restart
+else
+    $sudo_cmd /etc/init.d/datadog-agent restart
+fi
 
 # Wait for metrics to be submitted by the forwarder
 printf "\033[32m
--- a/packaging/debian/datadog-agent.service
+++ b/packaging/debian/datadog-agent.service
@@ -1,0 +1,11 @@
+[Unit]
+Description="Datadog Agent"
+After=network.target
+
+[Service]
+Type=forking
+ExecStart=/opt/datadog-agent/bin/start_agent.sh
+ExecStop=/opt/datadog-agent/bin/supervisorctl -c /etc/dd-agent/supervisor.conf shutdown
+
+[Install]
+WantedBy=multi-user.target
--- a/packaging/debian/start_agent.sh
+++ b/packaging/debian/start_agent.sh
@@ -1,0 +1,5 @@
+#!/bin/sh
+
+PATH=/opt/datadog-agent/embedded/bin:/opt/datadog-agent/bin:$PATH
+
+exec /opt/datadog-agent/bin/supervisord -c /etc/dd-agent/supervisor.conf --pidfile /var/run/datadog-supervisord.pid


packaging/datadog-agent/source/install_agent.sh
```diff
@@ -144,7 +144,11 @@ else
 fi
 
 printf "\033[34m* Starting the Agent...\n\033[0m\n"
-$sudo_cmd /etc/init.d/datadog-agent restart
+if [ $OS = "Debian" ]; then
+    $sudo_cmd invoke-rc.d datadog-agent restart
+else
+    $sudo_cmd /etc/init.d/datadog-agent restart
+fi
```
LeoCavaille: Should we use the same test that we use in the postinst script? If `invoke-rc.d` is present use it, otherwise fallback on the init script.

packaging/debian/datadog-agent.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description="Datadog Agent"
+After=network.target
+
+[Service]
+Type=forking
+Environment=SYSTEM_PATH=/opt/datadog-agent/embedded/bin:/opt/datadog-agent/bin:$PATH
+ExecStart=/bin/bash -c "PATH=${SYSTEM_PATH} /opt/datadog-agent/bin/supervisord -c /etc/dd-agent/supervisor.conf --pidfile /var/run/datadog-supervisord.pid"
+ExecReload=/opt/datadog-agent/bin/supervisorctl reload
```
LeoCavaille: This should probably be removed in my opinion. The supervisord config never changes and the reload option won't really do anything while the user might expect a config reload.

packaging/debian/datadog-agent.service
```diff
@@ -0,0 +1,13 @@
...
+ExecStop=/opt/datadog-agent/bin/supervisorctl shutdown
```
LeoCavaille: I think you need the config file here `-c /etc/dd-agent/supervisor.conf` otherwise supervisorctl will try to connect to the default URL which might be another related supervisord daemon.

LeoCavaille:
We talked about it IRL, but I'll just put it here not to forget, we're changing the `Environment` hack by a launch shell script that will set the `PATH` accordingly with the omnibus binaries and `exec` supervisord at the end.

packaging/datadog-agent/source/install_agent.sh
elafarge: Seems like a good idea. Ultimately it would be good to remove the system related if blocks, apart from the package manager related ones in the source install script ofc.

packaging/debian/datadog-agent.service
elafarge: I actually made it. I just have to push it has soon has I am able to run omnibus again and test the built packages.

elafarge:
Tested on CentOS 6, 7, debian squeeze and jessie, Ubuntu 12.04 so far. Things are working (once the host bug on CentOS is solved) so I guess we finally are GTG :)

remh:
What's the host bug on CentOS?

elafarge:
There's a bug during setup if the hostname is "localhost" (which basically happens when there isn't a /etc/hostname file on the machine). There was a support ticket for that a couple of days ago :https://trello.com/c/IfOGx0dc/366-bug-agent-installation-fails-on-centos-7-if-hostname-is-localhost .

Maybe we could include a fix (like generating a hostname based on the distribution name, either by creating a /etc/hostname file during install or by hacking the dd-agent/util.py) in the 5.4.2 release ?

remh:
This is not a bug, it's a feature. The agent shouldn't start if it can't find a valid hostname.

elafarge:
Ouups :/ sorry. The good part is we must be GTG then.

LeoCavaille:
Looking good, merging this. :+1:

PR #1564, opened 2015-04-21
update requests library, per CVE-2015-2296

miketheman:
Versions prior to 2.6.0 are vulnerable to a specific session hijack
attack. Since we don't necessarily set cookies on Agent-based calls to
the data intake, a session-based hijacking is unlikely, but we should
still try to keep up with security patches.

More details: http://www.openwall.com/lists/oss-security/2015/03/14/4


remh:
Thanks!

LeoCavaille:
Already fixed by 2da6874860bfd8d02d28f3b5d3b4d912dd3f4c63. Thanks

PR #1565, opened 2015-04-21
Fix docker check when a container has no name

LotharSee:
Fix a bug when a Docker container has no name.
That's uncommon (since usually Docker assign at least a random name), but that's possible.


talwai:
:+1:

PR #1568, opened 2015-04-22
readme: mention more dependencies

Sirupsen:
These are hard dependencies on the `setup_env` task, although I'm sure most of the tests can be completed without them this still makes that task fail.


LeoCavaille:
Fixed by 26acb0cf95ad3bc9eaec42953e591bcaa3a5a6ca

Issue #1348, opened 2015-02-05
Run supervisor as dd-agent instead of root


LotharSee:
I guess the supervisor is running as root for legacy reasons, and now I don't see why it should run at it.
Then it should certainly run as `dd-agent`, like the collector/forwarder/dogstatsd.

As a POC, I did it on CentOS 7. It should be quite the same for all the systems.
https://gist.github.com/LotharSee/77a69f4e5c37643b50ee

The only trouble is file permission:

* Changing permission on log file
* Putting the sock / pid / lock in a directory where `dd-agent` can read/write.

remh:
Thanks a lot for those exaplanations @LotharSee !

yannmh:
We are almost there  (:+1:  #1569) .
Let's do it in the 5.5.0 agent release.

yannmh:
Fixed with #1620 !

PR #1569, opened 2015-04-22
[core] use internal `/run` for pid and pickle files

degemer:
- use `/run` for run files (supervisor, pid and pickle), move more functions to `utils/`
- start `dogstatd` and `dd-agent` only when needed
- remove legacy option clean

See commits descriptions for more details.

It should affect:
- https://github.com/DataDog/dd-agent/pull/1504 (needs to add the `mkdir $dd_agent/run` step)
- https://github.com/DataDog/dd-agent/pull/1361 (needs to rebase afterwards)

And of course https://github.com/DataDog/dd-agent-omnibus, to add a internal `/run` directory (Windows is not concerned)


degemer:
This PR should also allow us to run `supervisor` as `dd-agent`, and thus fix https://github.com/DataDog/dd-agent/issues/1348

LeoCavaille:
Makes sense to me.
However, I'll maybe still allow `Platform` (and `PidFile` although less likely) to be imported from where they were before. Nothing says they're part of a private API, and people might use those in custom checks to switch behaviors on different OSs, you can leave a compat import, clearly state that it is deprecated and add a `FIXME` for 6.x.

degemer:
Sure, I added them back. I wanted to deprecate them, but it's a pain since all `Platform` methods are class methods, so I just added a comment instead.

remh:
@degemer can you rebase please ?

agent.py
```diff
@@ -77,6 +77,7 @@ def _handle_sigusr1(self, signum, frame):
         self._handle_sigterm(signum, frame)
         self._do_restart()
 
+    @classmethod
     def info(self, verbose=None):
```
remh: Tiny tiny nitpick but if you're switching to a class method, it should be 
```
cls
```
instead of 
```
self
```

daemon.py
```diff
@@ -227,7 +227,7 @@ def run(self):
         """
         raise NotImplementedError
 
-
+    @classmethod
     def info(self):
```
remh: same here

packaging/centos/datadog-agent.init
```diff
@@ -31,7 +31,7 @@ FORWARDERPATH="/opt/datadog-agent/agent/ddagent.py"
 SUPERVISORD_PATH="PATH=/opt/datadog-agent/embedded/bin:/opt/datadog-agent/bin:$PATH /opt/datadog-agent/bin/supervisord"
 SUPERVISORCTL_PATH="/opt/datadog-agent/bin/supervisorctl"
 SUPERVISOR_CONF="/etc/dd-agent/supervisor.conf"
-SUPERVISOR_SOCK="/var/tmp/datadog-supervisor.sock"
+SUPERVISOR_SOCK="/opt/datadog-agent/run/datadog-supervisor.sock"
```
remh: It shouldn't be an issue, but we'll need to triple check that it doesn't cause any trouble while upgrading.

utils/tailfile.py
```diff
@@ -1,29 +1,16 @@
-import base64
+import binascii
 import os
+from stat import *
```
remh: Let's just import what we need ?

remh:
Great work overall! Just a bunch of tiny nitpicks and it's good to merge.

degemer:
Thanks Rémi! Updated and rebased.

PR #1570, opened 2015-04-22
[apptags] Added apptag support to dd-agent

elafarge:
The agent now sends one "dd_check:appname" tag to dogweb for each
running check (including the JMX checks) if the user enables the
create_dd_check_tags option. This enables slice/dice by app in the
backend. The apptags are sent at regular intervals (every ten minutes by
default)

The running JMX apps list is retrieved from the {tmp}/jmx_status.yaml
file. Indeed, these tags have to be sent from the collectors side, along
with host-tags and this file happens to be the only communication
channel between JMXFetch.jar and the agent.

A test had been added to our test suite specifically for this new
feature (checking that the apptag is sent).

### DIFF
--- a/checks/collector.py
+++ b/checks/collector.py
@@ -1,18 +1,19 @@
 # Core modules
 import os
 import re
-import logging
-import subprocess
 import sys
 import time
-import datetime
 import socket
+import logging
+import datetime
+import subprocess
 
 import modules
 
+from config import get_version, get_system_stats, get_config
 from util import get_os, get_uuid, md5, Timer, get_hostname, EC2, GCE
-from config import get_version, get_system_stats
 
+import jmxfetch
 import checks.system.unix as u
 import checks.system.win32 as w32
 from checks import create_service_check, AgentCheck
@@ -28,6 +29,7 @@ log = logging.getLogger(__name__)
 
 FLUSH_LOGGING_PERIOD = 10
 FLUSH_LOGGING_INITIAL = 5
+DD_CHECK_TAG = 'dd_check:{0}'
 
 class Collector(object):
     """
@@ -58,7 +60,11 @@ class Collector(object):
             'agent_checks': {
                 'start': time.time(),
                 'interval': int(agentConfig.get('agent_checks_interval', 10 * 60))
-            }
+            },
+            'dd_check_tags': {
+                'start': time.time(),
+                'interval': int(agentConfig.get('dd_check_tags_interval', 10 * 60))
+            },
         }
         socket.setdefaulttimeout(15)
         self.run_count = 0
@@ -137,13 +143,15 @@ class Collector(object):
         self.run_count += 1
         log.debug("Starting collection run #%s" % self.run_count)
 
+        if checksd:
+            self.initialized_checks_d = checksd['initialized_checks'] # is a list of AgentCheck instances 
+            self.init_failed_checks_d = checksd['init_failed_checks'] # is of type {check_name: {error, traceback}}
+        
         payload = self._build_payload(start_event=start_event)
         metrics = payload['metrics']
         events = payload['events']
         service_checks = payload['service_checks']
-        if checksd:
-            self.initialized_checks_d = checksd['initialized_checks'] # is of type {check_name: check}
-            self.init_failed_checks_d = checksd['init_failed_checks'] # is of type {check_name: {error, traceback}}
+        
         # Run the system checks. Checks will depend on the OS
         if self.os == 'windows':
             # Win32 system checks
@@ -368,7 +376,7 @@ class Collector(object):
             payload['metrics'].extend(self._agent_metrics.check(payload, self.agentConfig,
                 collect_duration, self.emit_duration))
 
-
+        # Let's send our payload 
         emitter_statuses = self._emit(payload)
         self.emit_duration = timer.step()
 
@@ -501,6 +509,17 @@ class Collector(object):
         if external_host_tags:
             payload['external_host_tags'] = external_host_tags
 
+        # If required by the user, let's create the dd_check:xxx host tags
+        if self.agentConfig['create_dd_check_tags'] and \
+                self._should_send_additional_data('dd_check_tags'):
+            app_tags_list = [DD_CHECK_TAG.format(c.name) for c in self.initialized_checks_d]
+            app_tags_list.extend([DD_CHECK_TAG.format(cname) for cname in jmxfetch._get_jmx_appnames()])
+            
+            if 'system' not in payload['host-tags']:
+                payload['host-tags']['system'] = [] 
+
+            payload['host-tags']['system'].extend(app_tags_list)
+
         return payload
 
     def _get_metadata(self):
@@ -536,5 +555,3 @@ class Collector(object):
             return True
 
         return False
-
-
--- a/config.py
+++ b/config.py
@@ -426,8 +426,12 @@ def get_config(parse_args=True, cfg_path=None, options=None):
                 agentConfig[key] = config.get('Main', key)
             else:
                 agentConfig[key] = value
+        
+        # Create app:xxx tags based on monitored apps
+        agentConfig['create_dd_check_tags'] = config.has_option('Main', 'create_dd_check_tags')\
+                and _is_affirmative(config.get('Main', 'create_dd_check_tags'))
 
-        #Forwarding to external statsd server
+        # Forwarding to external statsd server
         if config.has_option('Main', 'statsd_forward_host'):
             agentConfig['statsd_forward_host'] = config.get('Main', 'statsd_forward_host')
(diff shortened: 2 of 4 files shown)


checks/collector.py
```diff
@@ -538,3 +553,16 @@ def _should_send_additional_data(self, data_name):
         return False
 
 
+    def _get_jmx_appnames(self):
```
remh: that should probably be in jmxfetch.py

checks/collector.py
```diff
@@ -28,6 +30,7 @@
 
 FLUSH_LOGGING_PERIOD = 10
 FLUSH_LOGGING_INITIAL = 5
+APPTAG = 'app:{0}'
```
remh: we should probably use something less generic.
```
dd_check 
``` 
maybe ?

checks/collector.py
```diff
@@ -368,7 +371,19 @@ def run(self, checksd=None, start_event=True):
             payload['metrics'].extend(self._agent_metrics.check(payload, self.agentConfig,
                 collect_duration, self.emit_duration))
 
-
+        # If required by the user, let's create the app:xxx host tags
+        if self.agentConfig['create_app_tags']:
```
remh: It's not probably not the right location to send that. We shouldn't run this code at every iteration. It should probably be in this block:
https://github.com/DataDog/dd-agent/blob/etienne/per-app-tagging/checks/collector.py#L461

checks/collector.py
```diff
@@ -368,7 +371,19 @@ def run(self, checksd=None, start_event=True):
...
+            running_apps = [APPTAG.format(c.name) for c in self.initialized_checks_d]
+            running_apps.extend([APPTAG.format(cname) for cname in self._get_jmx_appnames()])
+            
+            if 'host-tags' not in payload:
```
remh: this is not needed. the
```python
_build_payload
```
method ensures that hosts-tags will be there

checks/collector.py
```diff
@@ -368,7 +371,19 @@ def run(self, checksd=None, start_event=True):
...
+                payload['host-tags'] = dict()
+            if 'system' not in payload['host-tags']:
+                payload['host-tags']['system'] = list()
+            
+            payload['host-tags']['system'].extend(running_apps)
```
remh: so you can simply rewrite this block with
```python
payload['host-tags']['system'] = running_apps
```

Also, running_apps is not the best name for this list i think.

Something like:
```
app_tags_list
``` 
would be more explicit

LeoCavaille:
@elafarge can you rebase this one on top of the current master please?

checks/collector.py
```diff
@@ -481,9 +489,9 @@ def _build_payload(self, start_event=True):
             if GCE_tags is not None:
                 payload['host-tags'][GCE.SOURCE_TYPE_NAME] = GCE_tags
 
-            # Log the metadata on the first run
-            if self._is_first_run():
-                log.info("Hostnames: %s, tags: %s" % (repr(self.metadata_cache), payload['host-tags']))
+                # Log the metadata on the first run
+                if self._is_first_run():
```
remh: why are you changing the indentation here ?

checks/collector.py
```diff
@@ -137,13 +143,15 @@ def run(self, checksd=None, start_event=True):
         self.run_count += 1
         log.debug("Starting collection run #%s" % self.run_count)
 
+        if checksd:
+            self.initialized_checks_d = checksd['initialized_checks'] # is of type {check_name: check}
```
remh: While you're at it can you update the comments ? 
```python
checksd['initialized_checks']
```
is a list of AgentCheck instances

elafarge:
@remh  There was absolutely no reason why the indentation has been changed here. I fixed that in a separate commit (I didn't rebase because I don't know if you want me to keep the second commit or not)

checks/collector.py
```diff
@@ -501,6 +509,17 @@ def _build_payload(self, start_event=True):
...
             payload['external_host_tags'] = external_host_tags
 
+        # If required by the user, let's create the dd_check:xxx host tags
+        if self.agentConfig['create_dd_check_tags'] and \
+                self._should_send_additional_data('dd_check_tags'):
+            app_tags_list = [DD_CHECK_TAG.format(c.name) for c in self.initialized_checks_d]
+            app_tags_list.extend([DD_CHECK_TAG.format(cname) for cname in jmxfetch._get_jmx_appnames()])
+            
+            if 'system' not in payload['host-tags']:
+                payload['host-tags']['system'] = {}
```
remh: I think that should be a list, not a dictionary.

config.py
```diff
@@ -532,6 +536,20 @@ def get_config(parse_args=True, cfg_path=None, options=None):
     else:
         agentConfig['ssl_certificate'] = agentConfig['ca_certs']
 
+    # Getting interval parameters for periodic metrics
```
remh: We don't want that configurable. Can you remove that please ?

datadog.conf.example
```diff
@@ -79,6 +79,21 @@ use_mount: no
 # histogram_aggregates: max, median, avg, count
 # histogram_percentiles: 0.95
 
+# Periodic metadata intervals : select the interval at which you want different
+# metadata to be sent by the Agent (in seconds)
```
remh: See above comment.

jmxfetch.py
```diff
@@ -384,6 +384,21 @@ def _get_path_to_jmxfetch(self):
...
+    """ 
+    Retrieves the running JMX checks based on the {tmp}/jmx_status.yaml file 
+    updated by JMXFetch (and the only communication channel between JMXFetch
+    and the collector since JMXFetch).
+    """
+    check_names = []
+    jmx_status_path = os.path.join(get_jmx_status_path(), "jmx_status.yaml")
+    if os.path.exists(jmx_status_path):
+        jmx_checks = yaml.load(file(jmx_status_path)).get('checks', {})
+        for name, instances in jmx_checks.get('initialized_checks', {}).iteritems():
```
remh:
```python
for name in jmx_checks.get('initialized_checks', {}).iterkeys():
```

checks/collector.py
elafarge: Done (rebased with the 3rd and last commit)

jmxfetch.py
remh: or even more pythonic:
```python
check_names = [k for k in jmx_checks.get('initialized_checks', {}).iteritems()
```
(without an explicit loop)

tests/core/test_common.py
```diff
@@ -185,6 +186,38 @@ def test_collector(self):
...
+        checks = [load_check('redisdb', redis_config, agentConfig)]
+
+        c = Collector(agentConfig, [], {}, get_hostname(agentConfig))
+        payload = c.run({
+            'initialized_checks': checks,
+            'init_failed_checks': {}
+        })
+
+        # We check that the redis DD_CHECK_TAG is sent in the payload
+        assert DD_CHECK_TAG.format('redisdb') in payload['host-tags']['system']
```
remh: Shouldn't it be:
```
dd_check:redisdb
```
instead ?

remh: also please use 
```
self.assertTrue
```
instead

remh:
Thanks that looks great overall !
I was actually looking for such a feature while demoing yesterday!

tests/core/test_common.py
elafarge: I may be wrong but I think it's the same : DD_CHECK_TG='dd_check:{0}'. The advantage is that the tag prefixed is defined only at one place (in collector.py). 

I mostly did this because at the time the tag prefix wasn't officially defined and I knew I would forget to update the test if I had to change it. But I can replace that with dd_check:redisdb if you think it's important to check that we didn't change the apptag prefix by accident in our test suite. What should I do ?

config.py
elafarge: My bad, I thought we did because of this agentConfig.get call : https://github.com/DataDog/dd-agent/blob/master/checks/collector.py#L60 . I'll remove that at once.

elafarge:
I updated the PR with all your suggestions. Waiting for Travis to test it all, I'm testing it locally meanwhile.

tests/core/test_common.py
```diff
@@ -185,6 +186,38 @@ def test_collector(self):
             tag = "check:%s" % check.name
             assert tag in all_tags, all_tags
 
+
+    @attr(requires='sysstat')
```
remh: why does it require sysstat ?

elafarge:
@remh : Useless requirements removed. The CI succeeded : https://travis-ci.org/DataDog/dd-agent/builds/62118869 . The Travis Github status is currently "CI build in progress" since I just rebased & pushed in order to remove a couple of extra blank lines.

remh:
Nice job!

PR #1573, opened 2015-04-22
[ci] fix Travis cache reupload everytime

degemer:
It seems that pip doesn't delete some artefacts when upgrading itself
and setuptools. :facepunch: 

Build failing because of transient Travis cache issues.


LeoCavaille:
:+1:, just as a side :memo: if we can get rid of the `.9` hack at some point. I'd be delighted.

PR #1574, opened 2015-04-23
[tests] Improve tests for fluentd check.

hkaj:


tests/test_fluentd.py
```diff
@@ -1,73 +1,55 @@
-import unittest
-import logging
 from nose.plugins.attrib import attr
-logger = logging.getLogger(__file__)
 
-from tests.common import load_check
+from checks import AgentCheck
```
degemer: A debug import ? (not used anywhere)

tests/test_fluentd.py
```diff
@@ -1,73 +1,55 @@
...
                 {
                     "monitor_agent_url": "http://localhost:24220/api/plugins.json",
-                    "plugin_ids": [ "plg1" ],
+                    "plugin_ids": ["plg1"],
                 }
             ]
         }
 
-        agentConfig = {
+        self.agentConfig = {
```
degemer: It's the default configuration, you don't need to put it in the test.

tests/test_fluentd.py
```diff
@@ -1,73 +1,55 @@
...
-            if m[0] == 'fluentd.forward.retry_count':
-                self.assertEquals(m[2], 0)
-            elif m[0] == 'fluentd.forward.buffer_queue_length':
-                self.assertEquals(m[2], 0)
-            elif m[0] == 'fluentd.forward.buffer_total_queued_size':
+    def test_fluentd(self):
+        self.run_check(self.config, agent_config=self.agentConfig)
+        self.assertServiceCheckOK(self.check.SERVICE_CHECK_NAME,
+                                  tags=['fluentd_host:localhost', 'fluentd_port:24220'])
+        for m in self.metrics:
```
degemer: I don't think it's a good way to test the metrics: if there is no metric, the test will pass.

tests/test_fluentd.py
```diff
@@ -1,73 +1,55 @@
...
+            if m[0] in self.CHECK_GAUGES:
                 self.assertEquals(m[2], 0)
             self.assertEquals(m[3]['type'], 'gauge')
```
degemer: `self.assertMetric()` should be better.

tests/test_fluentd.py
```diff
@@ -1,73 +1,55 @@
...
             self.assertEquals(m[3]['tags'], ['plugin_id:plg1'])
 
-        self.assertEquals(len(metrics), 3, metrics)
+        self.assertEquals(len(self.metrics), 3)
```
degemer: You should put a `self.coverage_report()` at the end instead.

tests/test_fluentd.py
```diff
@@ -1,73 +1,55 @@
...
 
-        service_checks = check.get_service_checks()
-        service_checks_count = len(service_checks)
-        self.assertTrue(type(service_checks) == type([]))
+        service_checks_count = len(self.service_checks)
+        self.assertTrue(isinstance(self.service_checks, list))
```
degemer: You don't really need all of this, just a `self.assertServiceCheckOK()` and a `self.coverage_report()`.

tests/test_fluentd.py
```diff
@@ -1,73 +1,46 @@
...
-                self.assertEquals(m[2], 0)
-            elif m[0] == 'fluentd.forward.buffer_queue_length':
-                self.assertEquals(m[2], 0)
-            elif m[0] == 'fluentd.forward.buffer_total_queued_size':
+    def test_fluentd(self):
+        self.run_check(self.config)
+        self.assertServiceCheckOK(self.check.SERVICE_CHECK_NAME,
+                                  tags=['fluentd_host:localhost', 'fluentd_port:24220'])
+        for m in self.CHECK_GAUGES:
+            self.assertMetric('%s.%s' % (self.CHECK_NAME, m))
```
degemer: You should also check the tags.

tests/test_fluentd.py
```diff
@@ -1,73 +1,46 @@
...
+
+        for m in self.metrics:
+            if m[0] in self.CHECK_GAUGES:
                 self.assertEquals(m[2], 0)
             self.assertEquals(m[3]['type'], 'gauge')
             self.assertEquals(m[3]['tags'], ['plugin_id:plg1'])
```
degemer: This is not needed anymore, replaced by `self.assertMetric` (if you add the tags test).

tests/test_fluentd.py
```diff
@@ -1,73 +1,46 @@
...
-        check = load_check('fluentd', config, agentConfig)
-        self.assertRaises(Exception, check.run())
+        self.assertRaises(Exception, lambda: self.run_check({"instances": [{
+            "monitor_agent_url": "http://localhost:24222/api/plugins.json",
+            "plugin_ids": ["plg2"]}]}))
 
-if __name__ == '__main__':
-    unittest.main()
+        self.assertServiceCheckCritical(self.check.SERVICE_CHECK_NAME,
+                                        tags=['fluentd_host:localhost', 'fluentd_port:24222'])
```
degemer: You should `self.coverage_report()`, we never know.

degemer:
You're not lucky @hkaj, this check was updated by this PR: https://github.com/DataDog/dd-agent/pull/1623, so you have to rebase and update two more tests. :stuck_out_tongue:

tests/test_fluentd.py
degemer: Also, I know we don't really care about python 3 compatibility, but you should probably use `format` instead of `%`.

tests/checks/integration/test_fluentd.py
```diff
@@ -83,17 +51,12 @@ def test_fluentd_with_tag_by_type(self):
...
-            'version': '0.1',
-            'api_key': 'toto'
-        }
-
-        check = load_check('fluentd', config, agentConfig)
-        check.run()
-        metrics = check.get_metrics()
-        for m in metrics:
-            self.assertEquals(m[3]['tags'], ['type:forward'])
+        self.coverage_report()
```
degemer: You didn't test the service_check, the coverage report will fail.

tests/checks/integration/test_fluentd.py
```diff
@@ -106,16 +69,11 @@ def test_fluentd_with_tag_by_plugin_id(self):
...
-            self.assertEquals(len(m[3]['tags']), 1)
-            self.assertTrue(p.match(m[3]['tags'][0]))
+        self.run_check(config)
+        for m in self.CHECK_GAUGES:
+            try:
+                self.assertMetric('{0}.{1}'.format(self.CHECK_NAME, m), tags=['plugin_id:plg1'])
+            except:
+                self.assertMetric('{0}.{1}'.format(self.CHECK_NAME, m), tags=['plugin_id:plg2'])
+
+        self.coverage_report()
```
degemer: Same here.

tests/checks/integration/test_fluentd.py
```diff
@@ -106,16 +69,11 @@ def test_fluentd_with_tag_by_plugin_id(self):
...
-        p = re.compile('plugin_id:plg[12]')
-        for m in metrics:
-            self.assertEquals(len(m[3]['tags']), 1)
-            self.assertTrue(p.match(m[3]['tags'][0]))
+        self.run_check(config)
+        for m in self.CHECK_GAUGES:
+            try:
+                self.assertMetric('{0}.{1}'.format(self.CHECK_NAME, m), tags=['plugin_id:plg1'])
+            except:
+                self.assertMetric('{0}.{1}'.format(self.CHECK_NAME, m), tags=['plugin_id:plg2'])
```
degemer: You only test half of the metrics, why not assert both tags ?
```
self.assertMetric('{0}.{1}'.format(self.CHECK_NAME, m), tags=['plugin_id:plg1'])
self.assertMetric('{0}.{1}'.format(self.CHECK_NAME, m), tags=['plugin_id:plg2'])
```

PR #1577, opened 2015-04-24
Talwai/agent dev mode

talwai:
This adds a "developer mode" to the agent, that can be enabled by the `--profile` command line flag or by the `developer_mode` setting in `agentConfig`. 

Developer mode can be enabled at the `check` level or the `collector` level. When enabled at the check
level, e.g. `./agent.py check nginx --profile`, the `check.run()` function is profiled with `cProfile` and the `pstats` output is dumped to the log. 

When enabled at the `collector` level, the following behavior is enabled:
- Every run of the collector loop is profiled with `cProfile`, in the same way as above
- The `AgentMetrics` check (`checks.d/agent_metrics.py`) is run at the end of each collector loop. This check can be configured to run a set of additional `psutil.Process` methods on the current process. These additional metrics are flushed to the dispatcher as well as dumped to the log. See `conf.d/agent_metrics.yaml.default` for the default configuration, which can be extended as needed. Currently a setting under `process_metrics` is ignored if there is no corresponding `psutil.Process` method.

### DIFF
--- a/agent.py
+++ b/agent.py
@@ -24,6 +24,7 @@ import time
 # Custom modules
 from checks.collector import Collector
 from checks.check_status import CollectorStatus
+from utils.profile import pretty_statistics
 from config import (
     get_confd_path,
     get_config,
@@ -42,6 +43,7 @@ from util import (
 )
 from utils.pidfile import PidFile
 from utils.flare import configcheck, Flare
+from utils.profile import AgentProfiler
 
 # Constants
 PID_NAME = "dd-agent"
@@ -49,6 +51,8 @@ WATCHDOG_MULTIPLIER = 10
 RESTART_INTERVAL = 4 * 24 * 60 * 60  # Defaults to 4 days
 START_COMMANDS = ['start', 'restart', 'foreground']
 DD_AGENT_COMMANDS = ['check', 'flare', 'jmx']
+
+DEFAULT_COLLECTOR_PROFILE_INTERVAL = 20
 
 # Globals
 log = logging.getLogger('collector')
@@ -59,11 +63,12 @@ class Agent(Daemon):
     The agent class is a daemon that runs the collector in a background process.
     """
 
-    def __init__(self, pidfile, autorestart, start_event=True):
+    def __init__(self, pidfile, autorestart, start_event=True, in_developer_mode=False):
         Daemon.__init__(self, pidfile, autorestart=autorestart)
         self.run_forever = True
         self.collector = None
         self.start_event = start_event
+        self.in_developer_mode = in_developer_mode
 
     def _handle_sigterm(self, signum, frame):
         log.debug("Caught sigterm. Stopping run loop.")
@@ -110,6 +115,10 @@ class Agent(Daemon):
 
         self.collector = Collector(agentConfig, emitters, systemStats, hostname)
 
+        # In developer mode, the number of runs to be included in a single collector profile
+        collector_profile_interval = agentConfig.get('collector_profile_interval',
+                                                     DEFAULT_COLLECTOR_PROFILE_INTERVAL)
+
         # Configure the watchdog.
         check_frequency = int(agentConfig['check_freq'])
         watchdog = self._get_watchdog(check_frequency, agentConfig)
@@ -118,36 +127,30 @@ class Agent(Daemon):
         self.restart_interval = int(agentConfig.get('restart_interval', RESTART_INTERVAL))
         self.agent_start = time.time()
 
+        profiled = False
+        collector_profiled_runs = 0
+
         # Run the main loop.
         while self.run_forever:
-
-            # enable profiler if needed
-            profiled = False
-            if agentConfig.get('profile', False) and agentConfig.get('profile').lower() == 'yes':
+            # Setup profiling if necessary
+            if self.in_developer_mode and not profiled:
                 try:
-                    import cProfile
-                    profiler = cProfile.Profile()
+                    profiler = AgentProfiler()
+                    profiler.enable_profiling()
                     profiled = True
-                    profiler.enable()
-                    log.debug("Agent profiling is enabled")
-                except Exception:
-                    log.warn("Cannot enable profiler")
+                except Exception as e:
+                    log.warn("Cannot enable profiler: %s" % str(e))
 
             # Do the work.
             self.collector.run(checksd=checksd, start_event=self.start_event)
-
-            # disable profiler and printout stats to stdout
-            if agentConfig.get('profile', False) and agentConfig.get('profile').lower() == 'yes' and profiled:
-                try:
-                    profiler.disable()
-                    import pstats
-                    from cStringIO import StringIO
-                    s = StringIO()
-                    ps = pstats.Stats(profiler, stream=s).sort_stats("cumulative")
-                    ps.print_stats()
-                    log.debug(s.getvalue())
-                except Exception:
-                    log.warn("Cannot disable profiler")
+            if profiled:
+                if collector_profiled_runs >= collector_profile_interval:
+                    try:
+                        profiler.disable_profiling()
+                        profiled = False
+                        collector_profiled_runs = 0
+                    except Exception as e:
+                        log.warn("Cannot disable profiler: %s" % str(e))
 
             # Check if we should restart.
             if self.autorestart and self._should_restart():
@@ -158,8 +161,9 @@ class Agent(Daemon):
             if self.run_forever:
                 if watchdog:
                     watchdog.reset()
+                if profiled:
+                    collector_profiled_runs += 1
                 time.sleep(check_frequency)
-
         # Now clean-up.
         try:
             CollectorStatus.remove_latest_status()
@@ -212,7 +216,7 @@ def main():
     agentConfig = get_config(options=options)
     autorestart = agentConfig.get('autorestart', False)
     hostname = get_hostname(agentConfig)
-
+    in_developer_mode = agentConfig.get('developer_mode')
     COMMANDS_AGENT = [
         'start',
         'stop',
@@ -247,7 +251,7 @@ def main():
         deprecate_old_command_line_tools()
 
     if command in COMMANDS_AGENT:
-        agent = Agent(PidFile('dd-agent').get_path(), autorestart)
+        agent = Agent(PidFile('dd-agent').get_path(), autorestart, in_developer_mode=in_developer_mode)
 
     if command in START_COMMANDS:
         log.info('Agent version %s' % get_version())
@@ -301,17 +305,18 @@ def main():
             checks = load_check_directory(agentConfig, hostname)
             for check in checks['initialized_checks']:
                 if check.name == check_name:
-                    check.run()
-                    print check.get_metrics()
-                    print check.get_events()
-                    print check.get_service_checks()
+                    if in_developer_mode:
+                        check.run = AgentProfiler.wrap_profiling(check.run)
+
+                    cs = Collector.run_single_check(check, verbose=True)
+                    print CollectorStatus.render_check_status(cs)
+
                     if len(args) == 3 and args[2] == 'check_rate':
                         print "Running 2nd iteration to capture rate metrics"
                         time.sleep(1)
-                        check.run()
-                        print check.get_metrics()
-                        print check.get_events()
-                        print check.get_service_checks()
+                        cs = Collector.run_single_check(check, verbose=True)
+                        print CollectorStatus.render_check_status(cs)
+
                     check.stop()
 
     elif 'configcheck' == command or 'configtest' == command:
--- a/checks.d/agent_metrics.py
+++ b/checks.d/agent_metrics.py
@@ -1,0 +1,151 @@
+import os
+import threading
+
+# 3p
+try:
+    import psutil
+except ImportError:
+    psutil = None
+
+# project
+from checks import AgentCheck
+from checks.metric_types import MetricTypes
+from config import _is_affirmative
+
+MAX_THREADS_COUNT = 50
+MAX_COLLECTION_TIME = 30
+MAX_EMIT_TIME = 5
+MAX_CPU_PCT = 10
+
+
+class UnsupportedMetricType(Exception):
+    """
+    Raised by :class:`AgentMetrics` when a metric type outside outside of AgentMetrics.ALLOWED_METRIC_TYPES
+    is requested for measurement of a particular statistic
+    """
+    def __init__(self, metric_name, metric_type):
+        message = 'Unsupported Metric Type for {0} : {1}'.format(metric_name, metric_type)
+        Exception.__init__(self, message)
+
+class AgentMetrics(AgentCheck):
+    """
+    New-style version of `CollectorMetrics`
+    Gets information about agent performance on every collector loop
+    """
+
+    def __init__(self, *args, **kwargs):
+        AgentCheck.__init__(self, *args, **kwargs)
+        self._collector_payload = {}
+        self._metric_context = {}
+
+    def _psutil_config_to_stats(self, instance):
+        """
+        Reads `init_config` for `psutil` methods to call on the current process
+        Calls those methods and stores the raw output
+
+        :returns a dictionary of statistic_name: value
+        """
+        process_metrics = instance.get('process_metrics', self.init_config.get('process_metrics', None))
+        if not process_metrics:
+            self.log.error('No metrics configured for AgentMetrics check!')
+            return {}
+
+        methods, metric_types = zip(
+            *[(p['name'], p.get('type', MetricTypes.GAUGE))
+                for p in process_metrics if _is_affirmative(p.get('active'))]
+        )
+
+        names_to_metric_types = {}
+        for i, m in enumerate(methods):
+            names_to_metric_types[AgentCheck._get_statistic_name_from_method(m)] = metric_types[i]
+
+        stats = AgentCheck._collect_internal_stats(methods)
+        return stats, names_to_metric_types
+
+    def _send_single_metric(self, metric_name, metric_value, metric_type):
+        if metric_type == MetricTypes.GAUGE:
+            self.gauge(metric_name, metric_value)
+        elif metric_type == MetricTypes.RATE:
+            self.rate(metric_name, metric_value)
+        else:
+            raise UnsupportedMetricType(metric_name, metric_type)
+
+    def _register_psutil_metrics(self, stats, names_to_metric_types):
+        """
+        Saves sample metrics from psutil
+
+        :param stats: a dictionary that looks like:
+        {
+         'memory_info': OrderedDict([('rss', 24395776), ('vms', 144666624)]),
+         'io_counters': OrderedDict([('read_count', 4536),
+                                    ('write_count', 100),
+                                    ('read_bytes', 0),
+                                    ('write_bytes', 61440)])
+         ...
+         }
+
+         This creates a metric like `datadog.agent.collector.{key_1}.{key_2}` where key_1 is a top-level
+         key in `stats`, and key_2 is a nested key.
+         E.g. datadog.agent.collector.memory_info.rss
+        """
+
+        base_metric = 'datadog.agent.collector.{0}.{1}'
+        #TODO: May have to call self.normalize(metric_name) to get a compliant name
+        for k, v in stats.iteritems():
+            metric_type = names_to_metric_types[k]
+            if isinstance(v, dict):
+                for _k, _v in v.iteritems():
+                    full_metric_name = base_metric.format(k, _k)
+                    self._send_single_metric(full_metric_name, _v, metric_type)
+            else:
+                full_metric_name = 'datadog.agent.collector.{0}'.format(k)
+                self._send_single_metric(full_metric_name, v, metric_type)
+
+    def set_metric_context(self, payload, context):
+        self._collector_payload = payload
+        self._metric_context = context
+
+    def get_metric_context(self):
+        return self._collector_payload, self._metric_context
+
+    def check(self, instance):
+        if self.in_developer_mode:
+            stats, names_to_metric_types = self._psutil_config_to_stats(instance)
+            self._register_psutil_metrics(stats, names_to_metric_types)
+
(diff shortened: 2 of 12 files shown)


degemer:
Hey @talwai, I tested it a little and have some remarks (not concerning the code) about it, you probably already talked about it with @LeoCavaille, but I'm going to add my 2 cents.

I was expecting a lot more "by check" metrics, ie at the end of every run, display some stats for each check (and not only globally), for instance run time, memory consumption after/before, io read/write (basically the metrics you're already displaying for each run, and maybe tag them by check if you want to upload them to DD).
Also, about these metrics, could you display them in a nicer way ? Sure, once it's uploaded to DD, it should be really easy to analyze them, but a display in the logs would be a real plus.
Actually I was thinking something like the `info` display. :smile: (`collector` being the "global" metrics)
```bash
consul:
  run_time: 1.01s
  memory_before: 60785934
  memory_after: 60785935
  io read: 2
  ...
pgbouncer:
  run_time: 6.61s
  memory_before: 60785834
  memory_after: 60785840
  io read: 255
  ...
collector:
  run_time: 15.04s
  memory_before: 60785834
  memory_after: 60786001
  io read: 256
  ...
```

If you do something like this, I think that the profiler should be another option, not activated by default with the "developer mode", because we don't always need this level of details when debugging a check. And maybe also a profiler by check ? (not sure it's needed, just a suggestion)

Another thing: I don't think you should put `agent_metrics.py` in `checks.d`, because when you do this it becomes visible in 
> initialized checks.d checks: ['agent_metrics', 'network', 'pgbouncer', 'ntp', 'consul', 'nginx']

and a standard user shouldn't see this (event if it doesn't do anything).

The new agent_metrics are really nice! :+1: I think you could just prefix them by `datadog.agent.collector` instead of `datadog.agent`. (and as you wrote in your `TODOs`, some are probably rates and not gauges)

I just realized that we already have some agent_metrics here: https://github.com/DataDog/dd-agent/blob/master/checks/agent_metrics.py, you might want to put yours there too. (or maybe you split them on purpose ?)

talwai:
@degemer Thanks for the feedback!
> I was expecting a lot more "by check" metrics, ie at the end of every run, display some stats for each check (and not only globally), for instance run time, memory consumption after/before, io read/write 

I agree that we should be getting more data about the performance of individual checks, and not just the collector run as a whole. I can extend `wrap_profiling` to include some before and after `psutil` calls. Though I wonder if the extra system calls would hinder the performance of the check itself i.e. `check.run()` becomes `psutil.get_memory_info() ; check.run() ; psutil.get_memory_info()` for every check. 
> Also, about these metrics, could you display them in a nicer way ? 

Yep, the output is ugly. Will improve the formatting for the log dumps

> I think that the profiler should be another option, not activated by default with the "developer mode", because we don't always need this level of details when debugging a check. And maybe also a profiler by check ? (not sure it's needed, just a suggestion)

I get where you're coming from. In fact, I myself silenced the profiler dump when debugging the `AgentMetrics` check because it was mostly unhelpful. I wonder if that is the case in general though. Easy enough to make them two separate options, but would be interested in hearing other opinions.

> Another thing: I don't think you should put agent_metrics.py in checks.d, because when you do this it becomes visible in
initialized checks.d checks: ['agent_metrics', 'network', 'pgbouncer', 'ntp', 'consul', 'nginx']

The reason it's in there is: 
1. Because it appears that that's where all the new-style checks go 
2. So that it can be tested via `AgentCheckTest`. 
I realize that it is possible to test without the scaffolding, but my thought was that all tests for new-style checks should follow the same pattern. If the only issue with locating it in `checks.d/` is the log output, then I would prefer to just special-case it so it doesn't show up. Admittedly there may be other issues with having it in `checks.d/` that I'm not aware of.

> I just realized that we already have some agent_metrics here: https://github.com/DataDog/dd-agent/blob/master/checks/agent_metrics.py, you might want to put yours there too. (or maybe you split them on purpose ?)

Basically `checks.d/agent_metrics.py` is all of `checks/agent_metrics.py` and more. I opted against simply extending the class in place because it uses the old `Check` API, and if we plan to deprecate that API we should probably stop using it internally.

utils/profile.py
```diff
@@ -0,0 +1,44 @@
+#3p
+import psutil
```
LeoCavaille: You probably don't need psutil here!?

utils/profile.py
```diff
@@ -0,0 +1,44 @@
...
+
+import os
+import logging
+from config import _is_affirmative
```
LeoCavaille: This also can go away.

checks/__init__.py
```diff
@@ -531,8 +541,43 @@ def get_warnings(self):
         self.warnings = []
         return warnings
 
+    @staticmethod
+    def _collect_stats():
+        process = psutil.Process(os.getpid())
+        mem_info = process.get_memory_info()._asdict()
+        io_info = process.get_io_counters()._asdict()
```
LeoCavaille: This one is not available on OS X for instance :/

checks/__init__.py
```diff
@@ -560,6 +607,14 @@ def run(self):
...
                 )
             instance_statuses.append(instance_status)
+
+        if self.in_developer_mode:
+            try:
+                after = AgentCheck._collect_stats()
+                self.set_stats(before, after)
+            except Exception: # It's fine if we can't collect stats for the run
+                pass
+
```
LeoCavaille: Here I am not sure if for instance the CPU percentage sampling before and after would be useful (same for IO) because we would need to see it _during_ the call of `.check()`
For memory, we could see the leaks I guess.

conf.d/agent_metrics.yaml.default
```diff
@@ -0,0 +1,13 @@
...
+        get_io_counters: yes
+        get_num_threads: yes
+        get_connections: no
+    general:
+        foo: bar
+
+instances:
+  # No configuration is needed for this check.
+  # A single instance needs to be defined with any value.
+  - foo: bar
```
LeoCavaille: `[{}]`

checks.d/agent_metrics.py
```diff
@@ -0,0 +1,138 @@
+import psutil
+import os
+import threading
+from checks import AgentCheck
+from config import _is_affirmative
```
LeoCavaille: This should be ordered `#stdlib #3p #project`

checks.d/agent_metrics.py
```diff
@@ -0,0 +1,138 @@
...
+                method_key = method[4:] if method.startswith('get_') else method
+                try:
+                    raw_stats = getattr(current_process, method)()
+                    try:
+                        stats[method_key] = raw_stats._asdict()
+                    except AttributeError:
+                        if isinstance(raw_stats, int):
+                            stats[method_key] = raw_stats
+                        else:
+                            self.log.warn("Could not serialize output of {} to dict".format(method))
```
LeoCavaille: The `{}` syntax is not python 2.6 compatible... It should be named placeholders `{0}`.

LeoCavaille:
Nice! Good to see progress, this will prove super helpful to help the community design more performant checks. Made a few remarks just giving a first look.
Two ideas of stuff that it would be super cool to have:
* rather than just logging a dozen lines of pstats in logs at every run (and sometimes it might not be exactly what you want actually), could it be possible to enable profiling at a higher level for a while and accumulate the whole collector pstats?
* and optionally to be able to dump that to a file instead of logs so that you can use pstats for forensics later on offline.
What do you think?

checks/__init__.py
```diff
@@ -22,6 +22,7 @@
 
 # 3rd party
 import yaml
+import psutil
```
remh: psutil is actually not available on source install without a compiler :/ 
Can you try and catch an import error and disable the developer mode accordingly ?

talwai:
@LeoCavaille thanks for the review! I think both your ideas are good, if I understand correctly. I will look into moving the profiler higher in the call stack, and dumping cumulative statistics after longer intervals, rather than on every collector run. This should give us more useful higher-level insight than the current setup does. I will also add an option for dumping `pstats` to a file instead of / in addition to logging.

talwai:
The updated default functionality is as follows:

1. `agent.py` developer mode: Enables more granular checks in `checks.d/agent_metrics.py`. These checks are flushed to Datadog and also dumped to `log.info`. A `pstats` profile is dumped to the file `./collector-stats.dmp` after `COLLECTOR_PROFILE_INTERVAL` runs ( both the file name and the interval are candidates for moving to `agentConfig`). Additionally every check calls the static method `AgentCheck._collect_stats` before and after its run. This collects some `psutil` stats, the defaults of which are specified here: https://github.com/DataDog/dd-agent/blob/talwai/agent_dev_mode/checks/__init__.py#L37 (maybe also a candidate for living in `agentConfig`). These stats are prettified and dumped to `log.info`

2. `check` developer mode: This collects pstats output of the check run and dumps to `log.debug`. (Maybe it can dump to a file as well? Seems overkill). Additionally it calls `AgentCheck._collect_stats()` before and after its run and outputs data in the `agent.py info` format. e.g.: 
```
$ python agent.py check ntp --profile
2015-04-30 11:20:44,471 | INFO | dd.collector | config(config.py:922) | initialized checks.d checks: ['ntp', 'network']
2015-04-30 11:20:44,471 | INFO | dd.collector | config(config.py:923) | initialization failed checks.d checks: []
2015-04-30 11:20:44,471 | INFO | dd.collector | checks.collector(collector.py:435) | Running check ntp
Metrics:  [('ntp.offset', 1430407244, 0.02178668975830078, {'hostname': 'Aadityas-MacBook-Pro.local', 'type': 'gauge'})]
Events:  []
Service Checks:  [{'status': 0, 'tags': None, 'timestamp': 1430407244.5470452, 'check': 'ntp.in_sync', 'host_name': 'Aadityas-MacBook-Pro.local', 'message': None, 'id': 1}]
    ntp
    ---
      - instance #0 [OK] Last run duration: 0.0944309234619
      - Collected 1 metric, 0 events & 1 service check
      - Stats: 
            Memory Before (RSS): 18092032
            Memory After (RSS): 18108416
            Memory Before (VMS): 2534723584
            Memory After (VMS): 2534723584
```

checks/__init__.py
```diff
@@ -13,18 +14,30 @@
...
 from config import get_confd_path
 from checks import check_status
+from checks.utils import pretty_statistics
 
 # 3rd party
 import yaml
 
+try:
+    import psutil
+    PSUTIL_PRESENT = True
```
remh: You don't need this flag.
To test if psutil is here you can just do 
```python
if psutil is not None
```

remh: e.g. https://github.com/DataDog/dd-agent/blob/1bef95f968c1d884e6aeddc93a692d8a387535c8/checks/system/win32.py#L5-L8

agent.py
```diff
@@ -50,6 +52,11 @@
 START_COMMANDS = ['start', 'restart', 'foreground']
 DD_AGENT_COMMANDS = ['check', 'flare', 'jmx']
 
+PSTATS_LIMIT = 20
```
remh: Should those constants be in a different module ?

agent.py
```diff
@@ -134,19 +143,22 @@ def run(self, config=None):
...
-                    profiler.disable()
-                    import pstats
-                    from cStringIO import StringIO
-                    s = StringIO()
-                    ps = pstats.Stats(profiler, stream=s).sort_stats("cumulative")
-                    ps.print_stats()
-                    log.debug(s.getvalue())
-                except Exception:
-                    log.warn("Cannot disable profiler")
+            if profiled:
```
remh: Can we move this code to the new module ?

checks/__init__.py
```diff
@@ -265,6 +278,12 @@ def get_metrics(self, expire=True):
                 pass
         return metrics
 
+    def get_stats(self):
```
remh: Do we need it here? This class is a legacy class, not sure it's worth any further development.

checks/__init__.py
```diff
@@ -531,8 +553,61 @@ def get_warnings(self):
...
+            except psutil.AccessDenied:
+                if log is not None:
+                    log.warn("Cannot call psutil method {} : Access Denied".format(method))
+
+        return stats
+
+    def set_stats(self, before, after):
+        self.stats = {'before': before, 'after': after}
+
+    def get_stats(self):
```
remh: Watchout:
```python
get_stats
```
name 
is likely to be used inside checks.d.
It's already used by https://github.com/DataDog/dd-agent/blob/master/checks.d/rabbitmq.py#L162
And is likely used by some custom checks.d. 

You should use something different such as:
```python
def _get_internal_profiling_stats(...):
 ```

File: checks/__init__.py [hunk 14]
```diff
@@ -531,8 +553,61 @@ def get_warnings(self):
...
+        stats = self.stats
+        self.stats = None
+        return stats
+
     def run(self):
         """ Run all instances. """
+
+        # Store run statistics if needed
+        before, after = None, None
+        if self.in_developer_mode and self.name != 'agent_metrics':
```
remh:
```
agent_metrics
```
should be a constant

checks/__init__.py
```diff
@@ -560,6 +649,15 @@ def run(self):
...
                 )
             instance_statuses.append(instance_status)
+
+        if self.in_developer_mode and self.name != 'agent_metrics':
+            try:
+                after = AgentCheck._collect_stats()
+                self.set_stats(before, after)
+                log.info("\n \t %s %s" % (self.name, pretty_statistics(self.stats)))
+            except Exception: # It's fine if we can't collect stats for the run
+                pass
```
remh: we should probably log something in debug ?

checks/collector.py
```diff
@@ -390,6 +430,50 @@ def run(self, checksd=None, start_event=True):
...
+            current_service_checks = check.get_service_checks()
+
+            check_stats = check.get_stats()
+
+            # Save the status of the check.
+            metric_count = len(current_check_metrics)
+            event_count = len(current_check_events)
+            service_check_count = len(current_service_checks)
+
+            import pprint
```
remh: already imported

checks/utils.py
```diff
@@ -116,3 +116,24 @@ def tail(self, line_by_line=True, move_end=True):
...
+
+    before = stats.get('before')
+    after = stats.get('after')
+
+    mem_before = before.get('memory_info')
+    mem_after = after.get('memory_info')
+
+    if mem_before and mem_after:
+        return """
+            Memory Before (RSS): {0}
```
remh: This should probably display the difference after - before

config.py
```diff
@@ -56,12 +56,14 @@
     'nagios_perf_cfg'
     ]
 
-DEFAULT_CHECKS = ("network", "ntp")
```
remh: :+1:

remh:
Great work!
We should probably add the option in the datadog.conf by the way.

Also does it work properly on windows?

talwai:
Great question on Windows - I'm inclined to say that it works wherever `psutil` and `time.time()` work but will try it on a VM to be sure

talwai:
@remh Suggestions implemented. B2U

checks/collector.py
```diff
@@ -144,14 +149,26 @@ def run(self, checksd=None, start_event=True):
...
         service_checks = payload['service_checks']
-        
+        if checksd:
+            self.initialized_checks_d = checksd['initialized_checks'] # is of type {check_name: check}
+            self.init_failed_checks_d = checksd['init_failed_checks'] # is of type {check_name: {error, traceback}}
+
+            # Find the AgentMetrics check and pop it out
+            # This check must run at the end of the loop to collect info on agent performance
+            if not self._agent_metrics:
+                for check in self.initialized_checks_d[:]:
```
remh: you don't need the:
```python
[:]
```

agent.py
```diff
@@ -117,36 +123,30 @@ def run(self, config=None):
...
+                if collector_profiled_runs >= COLLECTOR_PROFILE_INTERVAL:
+                    try:
+                        profiler.disable_profiling()
+                        profiled = False
+                        collector_profiled_runs = 0
+                    except Exception:
```
remh: Any idea what kind of exception can be raised here ? 
If you just want to play it safe to make sure that the profiler doesn't crash the agent (which is a good thing) can you log the raised exception please ?

agent.py
```diff
@@ -50,6 +52,9 @@
 START_COMMANDS = ['start', 'restart', 'foreground']
 DD_AGENT_COMMANDS = ['check', 'flare', 'jmx']
 
+# How many runs to profile before dumping pstats
+COLLECTOR_PROFILE_INTERVAL = 1
```
remh: Should that be configurable ?

checks/__init__.py
```diff
@@ -286,6 +300,9 @@ def __init__(self, name, init_config, agentConfig, instances=None):
         self.name = name
         self.init_config = init_config or {}
         self.agentConfig = agentConfig
+        self.in_developer_mode = agentConfig.get('developer_mode') and psutil is not None
+        self.stats = None
```
remh: Same here, you should use a name that have less chances to conflict with existing checks

checks.d/agent_metrics.py
```diff
@@ -0,0 +1,151 @@
...
+    def set_metric_context(self, payload, context):
+        self._collector_payload = payload
+        self._metric_context = context
+
+    def get_metric_context(self):
+        return self._collector_payload, self._metric_context
+
+    def check(self, instance):
+        in_developer_mode = self.agentConfig.get('developer_mode', False)
+        if in_developer_mode and psutil is not None:
```
remh: This is redundant with your 
```python
self.in_developer_mode
```
https://github.com/DataDog/dd-agent/blob/talwai/agent_dev_mode/checks/__init__.py#L303

talwai:
@remh This works on windows

checks/__init__.py
```diff
@@ -564,6 +648,15 @@ def run(self):
                     tb=traceback.format_exc()
                 )
             instance_statuses.append(instance_status)
+
+        if self.in_developer_mode and self.name != 'agent_metrics':
```
remh: nitpick but you should use your 
```python
AGENT_METRICS_CHECK_NAME
```
constant

checks.d/agent_metrics.py
```diff
@@ -0,0 +1,150 @@
...
+         }
+
+         This creates a metric like `datadog.agent.collector.{key_1}.{key_2}` where key_1 is a top-level
+         key in `stats`, and key_2 is a nested key.
+         E.g. datadog.agent.collector.memory_info.rss
+        """
+
+        base_metric = 'datadog.agent.collector.{0}.{1}'
+        #TODO: May have to call self.normalize(metric_name) to get a compliant name
+        for k, v in stats.items():
```
remh: You should use an iterator
```
.iteritems()
```

checks.d/agent_metrics.py
```diff
@@ -0,0 +1,150 @@
...
+
+        :returns a dictionary of statistic_name: value
+        """
+        process_metrics = instance.get('process_metrics', self.init_config.get('process_metrics', None))
+        if not process_metrics:
+            self.log.error('No metrics configured for AgentMetrics check!')
+            return {}
+
+        methods, metric_types = zip(
+            *[(p['name'], p.get('type', 'gauge'))
```
remh: Nitpick but you should use constants for the metric types.

They are actually already existing:
https://github.com/DataDog/dd-agent/blob/master/checks/metric_types.py

checks/__init__.py
```diff
@@ -535,8 +552,61 @@ def get_warnings(self):
         self.warnings = []
         return warnings
 
+    @staticmethod
+    def _get_statistic_name_from_method(method_name):
+        return method_name[4:] if method_name.startswith('get_') else method_name
+
+    @staticmethod
+    def _collect_stats(methods=None, log=None):
```
remh: You should find a better name:
https://github.com/DataDog/dd-agent/blob/master/checks.d/postgres.py#L351
https://github.com/DataDog/dd-agent/blob/master/checks.d/pgbouncer.py#L69

What about 
```python
_collect_internal_stats
```
?

checks/__init__.py
```diff
@@ -535,8 +552,61 @@ def get_warnings(self):
...
+            stat_name = AgentCheck._get_statistic_name_from_method(method)
+            try:
+                raw_stats = getattr(current_process, method)()
+                try:
+                    stats[stat_name] = raw_stats._asdict()
+                except AttributeError:
+                    if isinstance(raw_stats, numbers.Number):
+                        stats[stat_name] = raw_stats
+                    else:
+                        if log is not None:
```
remh: When would log be None ?

We should always have a logger set in this code path.

remh:
Looks great beside the last comments!

It should be ready to merge once those are fixed

config.py
```diff
@@ -57,12 +57,14 @@
...
 ]
 
-DEFAULT_CHECKS = ("network", "ntp")
 LEGACY_DATADOG_URLS = [
     "app.datadoghq.com",
     "app.datad0g.com",
 ]
 
+#Checks whose log output to suppress, unless explicitly asked for
+HIDDEN_CHECKS = ['agent_metrics']
```
remh: You should use the constant you defined here.

tests/checks/mock/test_agent_metrics.py
```diff
@@ -0,0 +1,125 @@
...
+
+MOCK_NAMES_TO_METRIC_TYPES = {
+    'memory_info': 'gauge',
+    'cpu_times': 'gauge'
+}
+
+
+class AgentMetricsTestCase(AgentCheckTest):
+
+    CHECK_NAME = 'agent_metrics'
```
remh: Same here

checks/__init__.py
```diff
@@ -13,18 +14,30 @@
...
+    import psutil
+except ImportError:
+    psutil = None
+
 log = logging.getLogger(__name__)
 
+# Default methods run when collecting info about the agent in developer mode
+DEFAULT_PSUTIL_METHODS = ['get_memory_info', 'get_io_counters']
+
+AGENT_METRICS_CHECK_NAME = 'agent_metrics'
```
remh: You shouldn't define it twice

PR #1581, opened 2015-04-29
[ci] :racehorse: rake ci:run by splitting it

degemer:
`rake ci:run` is really slow (6min on Travis, but up to 15min locally), so nobody runs it before pushing.

This PR splits the default task into 3 tasks:
- `checks_mock` which runs all mocked tests for checks,
- `core_integration` which runs all integration tests of the agent core, (which takes a lot of time locally because of external calls & `time.sleep`)
- and `default`, which only runs core unit tests and `pylint` on changed files.

Pros:
* it takes 15s to run `rake ci:run` locally.
* nicer separation of tasks ?

Cons:
* Probably some overhead on Travis.

It has only been tested locally, so let's see what happens on Travis.


ci/default.rb
```diff
@@ -40,11 +40,24 @@
...
 
     task :lint do
-      sh %(find . -name '*.py' -not \\( -path '*.cache*' -or -path '*embedded*' -or -path '*venv*' \\) | xargs -n 1 pylint --rcfile=./.pylintrc)
+      output = %x[git status --porcelain]
+      paths = output.lines.collect{ |p|
+        p.strip.split ' '
+      }.select{ |stat_path|
+        status = stat_path.first
+        path = stat_path.last
+        path.end_with? ".py" and status != 'D'
```
LeoCavaille: It's a good idea but I think it should end up in a different rake task, people have different workflows. For instance in my case `git status --porcelain` would probably return nothing in 99% cases because my changes are always committed in a WIP commit on a branch and I carry non-committed files around.
Can you do maybe a `rake lint:changed` with this behavior and keep the `rake:lint`?

tests/core/test_common.py
```diff
@@ -300,6 +300,3 @@ def test_dupe_tags(self):
         self.assertEquals(len(self.aggr.metrics), 1, self.aggr.metrics)
         metric = self.aggr.metrics.values()[0]
         self.assertEquals(metric.value, 2)
-
-if __name__ == '__main__':
-    unittest.main()
```
LeoCavaille: nice cleanup!

tests/core/test_dogstatsd.py
```diff
@@ -1,12 +1,17 @@
-# -*- coding: utf-8 -*-
+ # -*- coding: utf-8 -*-
```
LeoCavaille: nitpicky: leading space?

tests/core/test_dogstatsd.py
```diff
@@ -1,12 +1,17 @@
...
+# stdlib
 import random
 import time
 
+# 3p
 import unittest
```
LeoCavaille: This is stdlib

tests/core/test_flare.py
```diff
@@ -1,6 +1,10 @@
-import unittest
+# stdlib
 import os.path
 import mock
+from nose.plugins.attrib import attr
```
LeoCavaille: You're missing 3p

tests/core/test_watchdog.py
```diff
@@ -1,16 +1,23 @@
...
-import subprocess
+# stdlib
 import os
-import sys
 from random import random, randrange
-import urllib as url
+import subprocess
+import sys
 import time
+import psutil
```
LeoCavaille: psutil is 3p

ci/default.rb
degemer: I'm not sure to understand: if `git status --porcelain` returns nothing (or there is no `*.py` files, the old `rake:lint` task will run (`if paths.empty?`). (see for instance on Travis)
I really wanted to have this behavior to have a fast `rake ci:run` when working on something (ie right now, it takes ~15s when a `.py` file is changed, versus more than 2min because of pylint (which is is really slooooow)).

degemer:
Thanks for the review @LeoCavaille !
I updated my code as per your comments (except `rake lint`), and since it's a :sunny: day, I also removed all the `unitest.main()`. :tada:

degemer:
I `rubocop -a`'d the new `ci/` files.

degemer:
I changed `lint`: it is still the same old command by default, but I added a `lint:changed`.

PR #1582, opened 2015-04-29
Update php fpm check to allow ping reply to be configured.

squaresurf:


LeoCavaille:
Cool thanks @squaresurf for your contribution.
Just a suggestion: could you add something in the integrations tests to test this new option, for instance set it to `blah` and test that the `can_ping` service check returns effectively a CRITICAL status?

squaresurf:
I've added a test for a bad ping reply. @LeoCavaille would you take a look and make sure I did it correctly? I've not used this test framework before.

LeoCavaille:
LGTM thanks again :+1:

squaresurf:
@LeoCavaille do you know why the test fails? It doesn't seem to be related to my PR.

LeoCavaille:
No the PHP-FPM test succeeds which is what matters here. We're currently fixing the CI @degemer is working on it, basically we need to have a different setup for contributors and external forks because they can't use our S3 cache due to credentials issues.
Closing this PR, merged upstream your changes in a rebased commit 552b5d1924ddbcb33862f26dbb781d73b4ef4d6a.
Thanks.

squaresurf:
Thank you!

PR #1586, opened 2015-04-30
[flare] add stderr of commands

degemer:
Also:
- check if files are readable before trying to add them to the
tarfile.
- remove whitespaces from input before checking for yes/no.

### DIFF
--- a/utils/flare.py
+++ b/utils/flare.py
@@ -1,4 +1,5 @@
 import atexit
+import cStringIO as StringIO
 import glob
 import logging
 import os.path
@@ -28,7 +29,7 @@ from util import (
 import requests
 
 # Globals
-log = logging.getLogger('flare')
+log = logging.getLogger(__name__)
 
 
 def configcheck():
@@ -88,8 +89,8 @@ class Flare(object):
     def check_user_rights():
         if Platform.is_unix() and not os.geteuid() == 0:
             log.warning("You are not root, some information won't be collected")
-            choice = raw_input('Are you sure you want to continue [y/N]? ').lower()
-            if choice not in ['yes', 'y']:
+            choice = raw_input('Are you sure you want to continue [y/N]? ')
+            if choice.strip().lower() not in ['yes', 'y']:
                 print 'Aborting'
                 sys.exit(1)
             else:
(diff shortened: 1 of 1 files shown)


utils/flare.py
```diff
@@ -88,8 +89,8 @@ def __init__(self, cmdline=False, case_id=None):
     def check_user_rights():
         if Platform.is_unix() and not os.geteuid() == 0:
             log.warning("You are not root, some information won't be collected")
-            choice = raw_input('Are you sure you want to continue [y/N]? ').lower()
-            if choice not in ['yes', 'y']:
+            choice = raw_input('Are you sure you want to continue [y/N]? ')
+            if "".join(choice.lower().split()) not in ['yes', 'y']:
```
LeoCavaille: What do you think of `choice.strip().lower()` ?

utils/flare.py
```diff
@@ -169,40 +171,52 @@ def _add_logs_tar(self):
         self._add_log_file_tar(self._dogstatsd_log)
         self._add_log_file_tar(self._jmxfetch_log)
         self._add_log_file_tar(
-            "{0}/*supervisord.log*".format(os.path.dirname(self._collector_log))
+            "{0}/*supervisord.log".format(os.path.dirname(self._collector_log))
```
LeoCavaille: :interrobang:

utils/flare.py
degemer: I chose this version for this :hammer: :
```bash
In [1]: ' y e s'.strip().lower()
Out[1]: 'y e s'

In [2]: ''.join(' y e s'.lower().split())
Out[2]: 'yes'
```

But I agree, it doesn't really make sense, I will change both of them.

utils/flare.py
degemer: It's not needed, since it's appended here: https://github.com/DataDog/dd-agent/blob/43180b827f48ec359d600eadf96ed1a74daa6838/utils/flare.py#L178

utils/flare.py
LeoCavaille: I'd say if someone typed `y e s` he probably had one too many :beers: and need to hold off sending his logs :wink:

utils/flare.py
LeoCavaille: :+1:

degemer:
Thanks for the review and the fix @LeoCavaille ! (tested and it works great)
Updated the `choice.strip().lower()`, will merge if Travis is :ok: .

degemer:
Travis :white_check_mark:

PR #1592, opened 2015-05-04
[kafka_consumer] add timeout for zk/kafka connections

### LINKED ISSUE #1589: [kafka_consumer] Kafka client takes 2 minutes to time out
If the KafkaClient can not connect to kafka, it takes 2 minutes to time out. The code at https://github.com/DataDog/dd-agent/blob/5.2.x/checks.d/kafka_consumer.py#L59-L72 is using the default timeout for kafka-client, which is 120 seconds: https://github.com/mumrah/kafka-python/blob/92aa7e94288cbfc4aed0dfbd52021d21694bced4/kafka/conn.py#L14

During the time that it's trying to connect to kafka, no data is reported by the agent.

LeoCavaille:
Because the default timeouts are pretty high in the kafka-python and
kazoo library (resp. 120s and 10s), we put them at a much lower
threshold (5s) for the default cases and offer a way to override in the
YAML init_config section.

Fixes #1589.

### DIFF
--- a/checks.d/kafka_consumer.py
+++ b/checks.d/kafka_consumer.py
@@ -1,19 +1,29 @@
 # stdlib
 from collections import defaultdict
-import random
 
-# project
-from checks import AgentCheck
-
-# 3rd party
+# 3p
 from kafka.client import KafkaClient
 from kafka.common import OffsetRequest
 from kazoo.client import KazooClient
 from kazoo.exceptions import NoNodeError
 
+# project
+from checks import AgentCheck
+
+DEFAULT_KAFKA_TIMEOUT = 5
+DEFAULT_ZK_TIMEOUT = 5
+
+
 class KafkaCheck(AgentCheck):
 
     SOURCE_TYPE_NAME = 'kafka'
+
+    def __init__(self, name, init_config, agentConfig, instances=None):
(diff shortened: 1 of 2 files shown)


clutchski:
nice thanks1

checks.d/kafka_consumer.py
```diff
@@ -1,20 +1,29 @@
...
 class KafkaCheck(AgentCheck):
 
     SOURCE_TYPE_NAME = 'kafka'
 
+    def __init__(self, name, init_config, agentConfig, instances=None):
+        self.zk_timeout = int(
+            init_config.get('zk_timeout', DEFAULT_ZK_TIMEOUT))
+        self.kafka_timeout = int(
+            init_config.get('kafka_timeout', DEFAULT_KAFKA_TIMEOUT))
+
```
degemer: You forgot the `AgentCheck.__init__(self, name, init_config, agentConfig, instances=instances)`.

LeoCavaille: Wow good catch! :+1:

degemer:
Nice improvement ! :+1:
Too bad we don't have any test for it. :disappointed: 
I just checked that it fails properly, we will have to really test it with the nightlies.

PR #1593, opened 2015-05-05
[gohai] Writing gohai errors to collector's log

elafarge:
From now on, everything gohai prints on its stderr (that is to say the
logs) is written to the collector.log file. It makes it easier for us to
pinpoint gohai related errors such as the "No IPV6" bug we just found.

### DIFF
--- a/checks/collector.py
+++ b/checks/collector.py
@@ -450,10 +450,12 @@ class Collector(object):
                     command = "gohai"
                 else:
                     command = "gohai\gohai.exe"
-                gohai_metadata = subprocess.Popen(
-                    [command], stdout=subprocess.PIPE
-                ).communicate()[0]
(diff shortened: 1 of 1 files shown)


remh:
Are there a lot of errors usually ? Could it become too spammy ?

elafarge:
@remh : As far as I could test, there aren't. On a fully working Gohai installation there is absolutely no log. There was one for this org but since not having an IPv6 interface is no longer considered to be an error, it's now fine.

In addition to that, GoHai is not run at every iteration of the collector's loop, so I don't think it could be a problem.

elafarge:
My point was just to have the errors in a file so that if an org has an unidentified issue and provides us with his collector.log, we'll be able to tell that it's in GoHai at one glance.

remh:
LGTM

PR #1595, opened 2015-05-05
[process] prettify & cache AccessDenied failures

LeoCavaille:
Refactor a bit this check, two noticeable changes:
1. on Windows especially, we get randomly some `AccessDenied` exceptions
when trying to look at some processes. Because we iterate on all the
processes at every run, it would trigger as many failures *every run*.
Once we get this error and `ignore_denied_access` is set to true, we'll
not look at it again until DEFAULT_AD_CACHE_DURATION or
`access_denied_cache_duration` in the init config.
2. instead of doing some sketchy try/catch over the code, all the
psutil-related code that actually gets metric values is wrapped into
`psutil_wrapper` which knows how to handle exceptions and known
failures.

Last remark: before we were setting some metrics to 0 even if we had no
values parsed, this is not the case anymore.

### DIFF
--- a/checks.d/process.py
+++ b/checks.d/process.py
@@ -1,4 +1,8 @@
-# 3rd party
+# stdlib
+from collections import defaultdict
+import time
+
+# 3p
 import psutil
 
 # project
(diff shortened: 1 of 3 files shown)


remh:
Nice cleanup!

We should probably cache the pid list too, i'm not sure it's useful to refresh it at every iteration. 
What do you think ?

According to the dev mode ( :+1: @talwai ) this is by far the most intensive operation of the check.

We could maybe cache it for a few iterations ?

LeoCavaille:
@remh it's a bit hard to make this design choice, where people could potentially put no-data monitors on top of those metrics. Or maybe hide it behind a feature flag?

remh:
+1 for the feature flag

LeoCavaille:
@remh implemented, let me know what you think?

checks.d/process.py
```diff
@@ -7,170 +11,213 @@
...
-    )
-
-    def find_pids(self, search_string, exact_match, ignore_denied_access):
+        # By default cache the PID list for a while
+        # Sometimes it's not wanted b/c it can mess with no-data monitoring
+        # This cache is indexed per instance
+        self.last_pid_cache_ts = {}
+        self.pid_cache = {}
+        self.pid_cache_duration = int(
+            init_config.get(
```
remh: shouldn't that be an instance parameter instead ?

LeoCavaille: I think in most cases, this shouldn't be changed so I don't think it's worth the trouble for an instance parameter.

Issue #568, opened 2013-07-05
Allow to exclude filesystems from metrics reporting


LotharSee:
Related to #537 but more generic.
We should be able to skip reporting metrics on certain disks (they can be not interesting because they are virtual, the disk of a guest VM, the swap ...). The main reason is that it avoid the user to filter them out on dashboards.

We can use a shared syntax with the network check which already allow to exclude network interface.
We can start with a list, but the support for pattern would be great too.

notjoe:
I am getting false alerts on tmpfs, which is not a real file system.

remh:
Thanks for the update @notjoe . Reopening this issue.

QuirkyMarc:
Excluding filesystems by type is needed to.

For instance, I have 100+ servers mounting the same NFS filesystem, I don't need to be alerted by the NFS server, and all 100+ clients, alerting about that same filesystem.

LeoCavaille:
It would be cool also to have an option to tag by filesystem type (that is interesting to filter out stuff also on graphs).
Nonetheless, it should be featured flagged off by default because it will recreate too many contexts otherwise.

PR #1596, opened 2015-05-06
[disk] AgentCheck for Disk metrics :tada:

degemer:
This commit is the first step towards the annihilation of the old `Check`
hidden in `checks/system/{unix,win32}.py`!
It's time to launch the final assault of `Check`. `AgentCheck` shall triumph!

Add a `checks.d/disk.py` checks, which uses psutil when available to get
disk metrics (same than the previous `disk` checks), and otherwise
get the same metrics with `df`.
It has a small integration test, just to check that metrics exist, and
more precise mock tests.
This new check is enabled by default, and the old `disk` checks are
deleted.

Fix https://github.com/DataDog/dd-agent/issues/568

### DIFF
--- a/checks.d/disk.py
+++ b/checks.d/disk.py
@@ -1,0 +1,232 @@
+# stdlib
+import os
+import re
+
+# 3p
+try:
+    import psutil
+except ImportError:
+    psutil = None
+
+# project
+from checks import AgentCheck
+from config import _is_affirmative
+from util import Platform
+import utils.subprocess_output
+
+
+class Disk(AgentCheck):
+    """ Collects metrics about the machine's disks. """
+    # -T for filesystem info
+    DF_COMMAND = ['df', '-T']
+    FAKE_DEVICES = ['udev', 'sysfs', 'rpc_pipefs', 'proc', 'devpts']
+    METRIC_DISK = 'system.disk.{0}'
+    METRIC_INODE = 'system.fs.inodes.{0}'
+
+    def __init__(self, name, init_config, agentConfig, instances=None):
+        if instances is not None and len(instances) > 1:
+            raise Exception("Disk check only supports one configured instance.")
+        AgentCheck.__init__(self, name, init_config,
+                            agentConfig, instances=instances)
+        # Get the configuration once for all
+        self._load_conf(instances[0])
+
+    def check(self, instance):
+        """Get disk space/inode stats"""
+        # Windows and Mac will always have psutil
+        # (we have packaged for both of them)
+        if self._psutil():
+            self.collect_metrics_psutil()
+        else:
+            self.collect_metrics_manually()
+
+    @classmethod
+    def _psutil(cls):
+        return psutil is not None
+
+    def _load_conf(self, instance):
+        self._excluded_filesystems = instance.get('excluded_filesystems', [])
+        self._excluded_disks = instance.get('excluded_disks', [])
+        self._tag_by_filesystem = _is_affirmative(
+            instance.get('tag_by_filesystem', False))
+        self._all_partitions = _is_affirmative(
+            instance.get('all_partitions', True))
+
+        # FIXME: 6.x, drop use_mount option in datadog.conf
+        self._load_legacy_option(instance, 'use_mount', False,
+                                 operation=_is_affirmative)
+        # FIXME: 6.x, drop device_blacklist_re option in datadog.conf
+        self._load_legacy_option(instance, 'excluded_disk_re', '^$',
+                                 legacy_name='device_blacklist_re',
+                                 operation=re.compile)
+
+    def _load_legacy_option(self, instance, option, default,
+                            legacy_name=None, operation=lambda l: l):
+        value = instance.get(option, default)
+        legacy_name = legacy_name or option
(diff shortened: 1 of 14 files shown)


checks.d/disk.py
```diff
@@ -0,0 +1,229 @@
...
+                previous = None
+        return devices
+
+    def _list_devices(self, df_output):
+        """
+        Given raw output for the df command, transform it into a normalized
+        list devices. A 'device' is a list with fields corresponding to the
+        output of df output on each platform.
+        """
+        all_devices = [l.strip().split() for l in df_output.split("\n")]
```
remh: I know it doesn't matter in this case as it's only for linux but you should use:
```python
df_output.splitlines()
```
instead. It's better practice as it should work on all platforms.

checks.d/disk.py
```diff
@@ -0,0 +1,229 @@
...
+        self._load_conf(instance)
+
+        # Windows and Mac will always have psutil
+        # (we have packaged for both of them)
+        if self._psutil():
+            self.collect_metrics_psutil()
+        else:
+            self.collect_metrics_manually()
+
+    def _psutil(self):
```
LeoCavaille: Should be a static method or a class method.

checks.d/disk.py
```diff
@@ -0,0 +1,229 @@
...
+        return psutil is not None
+
+    def _load_conf(self, instance):
+        self._excluded_filesystems = instance.get('excluded_filesystems', [])
+        self._excluded_disks = instance.get('excluded_disks', [])
+        self._tag_by_filesystem = _is_affirmative(
+            instance.get('tag_by_filesystem', ''))
+        self._all_partitions = _is_affirmative(
+            instance.get('all_partitions', 'yes'))
```
LeoCavaille: Just for clarity here I'd recommend using a boolean directly rather than a string for the default.

checks.d/disk.py
```diff
@@ -0,0 +1,229 @@
...
+
+    def _load_legacy_option(self, instance, option, default,
+                            legacy_name=None, operation=lambda l: l):
+        self_option = instance.get(option, default)
+        if legacy_name is None:
+            legacy_name = option
+        legacy_option = self.agentConfig.get(legacy_name)
+        if self_option == default and legacy_option is not None:
+            self_option = legacy_option
+        setattr(self, '_{0}'.format(option), operation(self_option))
```
LeoCavaille:
```python

Comment (LeoCavaille) [inline checks.d/disk.py]: Could be factorized a bit and also throw a deprecation warning?
```python
value = instance.get(option)
legacy_name = legacy_name or option
if value is None and legacy_name in self.agentConfig:
    self.log.warn(
        "Using `{0}` in datadog.conf has been deprecated in favor of `{1}` in disk.yaml"
        .format(legacy_name, option)
    )
    value = self.agentConfig(legacy_name)
setattr(self, '_{0}'.format(option), operation(value))
```

checks.d/disk.py
```diff
@@ -0,0 +1,229 @@
...
+
+    def collect_metrics_psutil(self):
+        self._valid_disks = {}
+        for part in psutil.disk_partitions(all=self._all_partitions):
+            # we check all exclude conditions
+            if self._exclude_disk_psutil(part):
+                continue
+            else:
```
LeoCavaille: Lose the `else:` just continue the loop for simplicity.

checks.d/disk.py
```diff
@@ -0,0 +1,229 @@
...
+                # For later, latency metrics
+                self._valid_disks[part.device] = [part.fstype, part.mountpoint]
```
LeoCavaille: Use a tuple and not a list for a fixed-size structure.

checks.d/disk.py
remh: (watchout not to spam the log with this deprecation warning )

checks.d/disk.py
```diff
@@ -0,0 +1,229 @@
...
+        metrics = {}
+        for name in ['total', 'used', 'free']:
+            # For legacy reasons,  the standard unit it kB
+            metrics[self.METRIC_DISK.format(name)] = getattr(usage, name) / 1024.0
+        # FIXME: 6.x, use percent, a lot more logical than in_use
+        metrics[self.METRIC_DISK.format('in_use')] = usage.percent / 100.0
+        if Platform.is_unix():
+            metrics.update(self._collect_inodes_metrics(part.mountpoint))
+
+        return metrics.iteritems()
```
LeoCavaille: I am not sure it's adding much here to return a generator, you should return a dictionary and iterate on it in the caller.
Otherwise it's confusing some `_collect_*` methods return dicts, other return dictionaries or call `self.gauge` directly.

degemer:
I addressed all your comments, moved `_load_conf` to `__init__` (it would be nice to  put everything in `init_config`, but we wouldn't have `instances`, and I guess the check would not run so...).
Also:
* `collect_*` actually send metrics (`self.gauge`),
* `_collect_*` returns dictionary of metrics -> value.

LeoCavaille:
Thanks for making the updates. Merging.

PR #1597, opened 2015-05-06
Add no start option to install agent script

hernit:
Add a no start service option to the installation script. We need this for prebaked vm images where we don't want the agent to start, but control the starting at later stage of the deployment.

PR #1600, opened 2015-05-06
[install script] add DD_INSTALL_ONLY option

elafarge:
Edit of this PR (https://github.com/DataDog/dd-agent/pull/1597) from an external contributor.

The script uses long option names so that it's easy to understand. Since the script is not only used by our one line installation command, I added a "usage" description displayed when the program is called with the --help option or with a wrong option.

### DIFF
--- a/packaging/datadog-agent/source/install_agent.sh
+++ b/packaging/datadog-agent/source/install_agent.sh
@@ -40,6 +40,12 @@ if [ -n "$DD_API_KEY" ]; then
     apikey=$DD_API_KEY
 fi
 
+if [ -n "$DD_INSTALL_ONLY" ]; then
+    no_start=true
+else
+    no_start=false
+fi
+
 if [ ! $apikey ]; then
     printf "\033[31mAPI key not available in DD_API_KEY environment variable.\033[0m\n"
     exit 1;
@@ -143,12 +149,25 @@ else
     $sudo_cmd sh -c "sed 's/api_key:.*/api_key: $apikey/' /etc/dd-agent/datadog.conf.example > /etc/dd-agent/datadog.conf"
 fi
 
-printf "\033[34m* Starting the Agent...\n\033[0m\n"
+restart_cmd="$sudo_cmd /etc/init.d/datadog-agent restart" 
 if command -v invoke-rc.d >/dev/null 2>&1; then
-    $sudo_cmd invoke-rc.d datadog-agent restart
-else
-    $sudo_cmd /etc/init.d/datadog-agent restart
+    restart_cmd="$sudo_cmd invoke-rc.d datadog-agent restart"
 fi
+
+if $no_start; then
+    printf "\033[34m 
+* DD_INSTALL_ONLY environment variable set: the newly installed version of the agent
+will not start by itself. You will have to do it manually using the following 
+command: 
+
+    $restart_cmd
+
+\033[0m\n"
+    exit
+fi
+
+printf "\033[34m* Starting the Agent...\n\033[0m\n"
+eval $restart_cmd
 
 # Wait for metrics to be submitted by the forwarder
 printf "\033[32m


degemer:
Quick remark: since we actually use this script with this one-liner `DD_API_KEY=tttt bash -c "$(curl -L https://raw.githubusercontent.com/DataDog/dd-agent/master/packaging/datadog-agent/source/install_agent.sh)"`, wouldn't it make more sense to enable this option via environment variable ? (`DD_NO_START` for instance)

elafarge:
It would work as well indeed but I don't really understand why you think it would be better than an option in that precise case (especially since it's a flag). Could you shed some light on it ?

packaging/datadog-agent/source/install_agent.sh
```diff
@@ -40,6 +40,12 @@ if [ -n "$DD_API_KEY" ]; then
     apikey=$DD_API_KEY
 fi
 
+if [ -n "$NO_START" ]; then
```
degemer: You should probably prefix it with `DD_`, just in case.
It's not really a no_start, since if it's already configured, it's automatically started after the install of the package, but I don't find anything better.

packaging/datadog-agent/source/install_agent.sh
```diff
@@ -143,13 +149,19 @@ else
...
+restart_cmd="$sudo_cmd /etc/init.d/datadog-agent restart" 
 if command -v invoke-rc.d >/dev/null 2>&1; then
-    $sudo_cmd invoke-rc.d datadog-agent restart
-else
-    $sudo_cmd /etc/init.d/datadog-agent restart
+    restart_cmd="$sudo_cmd invoke-rc.d datadog-agent restart"
 fi
 
+if $no_start; then
+    printf "\033[34m* DD_NO_START environment variable set : the newly installed version of the agent will not start by itself. You will have to do it manually using the following command : \n\n $restart_cmd \n  \n\033[0m\n"
```
degemer: Could you break it on multiple lines of max 80 chars ? (it's really better on terminal)
Could you also remove the space before the `:`: http://www.wikiwand.com/en/Colon_(punctuation)#/Spacing

elafarge: My mistakes. Doing that at once.

packaging/datadog-agent/source/install_agent.sh
```diff
@@ -40,6 +40,12 @@ if [ -n "$DD_API_KEY" ]; then
     apikey=$DD_API_KEY
 fi
 
+if [ -n "$DD_NO_START" ]; then
```
degemer: On second thought, you should probably call it `$DD_NO_RESTART`, because it's really a restart. But when set to true during an upgrade, it will restart anyway.
Even with `$DD_NO_RESTART`, I'm afraid that people may think that it actually always prevents the agent from restarting, even during upgrades...
`DD_NO_FIRST_RESTART` maybe ? That sounds ridiculous, do you have a better idea ?

packaging/datadog-agent/source/install_agent.sh
```diff
@@ -40,6 +40,12 @@ if [ -n "$DD_API_KEY" ]; then
     apikey=$DD_API_KEY
 fi
 
+if [ -n "$DD_INSTALL_ONLY" ] && $DD_INSTALL_ONLY ; then
```
degemer: I don't like the `& $DD_INSTALL_ONLY`, because if you put `DD_INSTALL_ONLY=True`, it won't work.
Probably just checking that the env var is here is enough. (or else checking that it's different than `false`)

packaging/datadog-agent/source/install_agent.sh
```diff
@@ -143,13 +149,26 @@ else
...
 if command -v invoke-rc.d >/dev/null 2>&1; then
-    $sudo_cmd invoke-rc.d datadog-agent restart
-else
-    $sudo_cmd /etc/init.d/datadog-agent restart
+    restart_cmd="$sudo_cmd invoke-rc.d datadog-agent restart"
 fi
 
+if $no_start; then
+    printf "\033[34m 
+*DD_INSTALL_ONLY environment variable set: the newly installed version of the agent
```
degemer: Last nitpick: could you put a space: `* DD_INSTALL_ONLY` ?

degemer:
Thanks for the improvement @elafarge! It should be good after the two last comments.

elafarge:
@degemer Sorry for the late reply. I'm going to ping dustin about that as soon as merged. I think it will just go to Zendesk. We just had one customer request for this feature so far so I don't think it's worth putting a notice about that under the one-line install command in datadoghq.com.

LeoCavaille:
Looking good, merging.

PR #1605, opened 2015-05-08
Change config.py to disable watchdog.

sjenriquez:


remh:
@sjenriquez I guess you found out that you can pass 
```
watchdog: no
```
in datadog.conf ?

sjenriquez:
Ahh, thank you, I didn't realize that. I was attempting to disable in config.py.

PR #1611, opened 2015-05-11
[core] service metadata collection

yannmh:
# Service metadata collection

## Update 05/18
`AgentCheck.service_metadata(metadata_name, value)` replaces `AgentCheck.svc_metadata(metadata_dict)` method.
Under the hood, metadata are stored in a list, and rolled-up to a dictionnary for every instance.
Usage:
```
self.service_metadata('foo', "bar")
```

### AgentCheck interface

Following methods were added to `AgentCheck` interface to support service metadata collection:
```python
self.service_metadata(name, value) # Check purpose - Send metadata
self.get_service_metadata() # Collector purpose - Return a list of metadata dictionaries saved by check -if any-
							# and clears them out of the instance's service_checks list
```

To send metadata from an Agent Check, decorate it with `self.svc_metadata(...)` statements.
```python
def	check(...)
	...
	# Send integration metadata
	self.service_metadata('foo', "bar")
	...
```

### Payload refactoring

Collector payload was re-factored behind `AgentPayload` interface: it offers the same single payload style interface but manages, distinguishes two payloads:
* A metadata payload related to hostname, system and services
* A data payload that contains metrics, events, service_checks and more

Each of these payloads *can be* automatically submitted to a specific endpoint.

Before
```
payload
+------------+
|'metrics'   |
+--+---------+
|'events'    |
+------------+
|'svc_cheks' |        submit
+--+---------+          ------->     /intake
|'gohai'     |
+------------+
|'systm_stts'|
+------------+
|'meta'      |
+------------+
|'agent_chks'|
+--+---------+
|'svc_meta'  |
+--+---------+
|............|
|            |
+            +
 ............
```

Now:
```
 payload
+---------------------------------------+
| payload_data           payload_meta   |
| +------------+         +------------+ |
| |'metrics'   |         |'systm_stts'| |
| +--+---------+         +------------+ |
| |'events'    |         |'meta'      | |
| +------------+         +------------+ |
| |'svc_cheks' |         |'agent_chks'| |
| +--+---------+         +--+---------+ |
| |............|         |'svc_meta'  | |
| |            |         +--+---------+ |
| +            +         |............| |
|  ............          |            | |
| +            +         |            | |
| +------------+         +------------+ |
+-+------------+---------+------------+-+

                       submit
                        + +
                        | |
                        | |
/intake/metrics  <------+ +------->  /intake/metadata
```

### Collector

Collector `run` method was reorganised to successively populate the payload by data type (it used to be more random).

Steps are:
1. Payload skeleton
	* `self._build_payload(...)`: only builds the payload skeleton, so it contains all of the generic payload data (no metadata).
2. Data payload
	* Metrics
	* Events
	* Service checks
3. Metadata payload
	* `self._populate_payload_metadata(...)`: periodically populates the payload with metadata related to the system, host, and/or checks.

### agent `check`command

Agent `check` command now displays collected metadata, in addition to metrics, events and service checks.

### agent `info` command

Set `display_service_metadata` to True in `datadog.conf` file.

```
...
display_service_metadata: True
...
```

A new **Service metadata** section is now displayed in the `info` command output whenever service metadata are collected.
screenshot 2015-05-12 10 05 35

###  Split data and metadata ?
Currently data and metadata from the agent are part of the same payload, which is processed by a single handler in Sobotka.
Metadata are extracted, directly written to Postgres when data are forwarded to Kafka.

As the volume of metadata is growing, it would be wiser to split data and metadata processing, so that postgres operations would not slow down the agent data pipeline.

First suggestion made was to have two different Sobotka handlers specific for data and metadata:
https://docs.google.com/document/d/10LAKQ1E_ez4jgTn_czFJwDkEIghGFUFsDF03LgY8QRE

### Include host tags in the data payload ?

Host tags are now listed in the metadata payload only.
As it is possibly sent in a different packet (to a different endpoint), it may lead to a 'no tag' situation when a data payload is received before any metadata's.

A workaround would be to add/duplicate host tags in the data payload, but it breaks the Sobotka handler logic described in the section above.

### Single payload/endpoint for metrics and metadata

Let's keep it simple for now, so we can test and play with metadata support internally with the 5.4 agent release.
As the volume of metadata is still low, I can set `AgentPayload` to send a single data+metadata payload to the current `/intake` endpoint.

## TODO
- [x] Fix flaky tests
- [x] Submit only one data+metadata payload to `/intake` endpoint (c.f. 'Single endpoint for metrics and metadata' section)
- [ ] Write a RFC on why and how we would split agent data and metadata

### DIFF
--- a/checks.d/redisdb.py
+++ b/checks.d/redisdb.py
@@ -334,7 +334,5 @@ class Redis(AgentCheck):
         self._check_slowlog(instance, custom_tags)
 
     def _collect_metadata(self, info):
-        metadata_dict = {}
         if info and 'redis_version' in info:
-            metadata_dict['version'] = info['redis_version']
-        self.svc_metadata(metadata_dict)
+            self.service_metadata('version', info['redis_version'])
--- a/checks.d/redisdb.py
+++ b/checks.d/redisdb.py
@@ -163,10 +163,12 @@ class Redis(AgentCheck):
         # Ping the database for info, and track the latency.
         # Process the service check: the check passes if we can connect to Redis
         start = time.time()
+        info = None
         try:
             info = conn.info()
             status = AgentCheck.OK
             self.service_check('redis.can_connect', status, tags=tags_to_add)
+            self._collect_metadata(info)
         except ValueError, e:
             status = AgentCheck.CRITICAL
             self.service_check('redis.can_connect', status, tags=tags_to_add)
@@ -271,7 +273,6 @@ class Redis(AgentCheck):
             self.service_check('redis.replication.master_link_status', status, tags=tags)
             self.gauge('redis.replication.master_link_down_since_seconds', down_seconds, tags=tags)
 
-
     def _check_slowlog(self, instance, custom_tags):
         """Retrieve length and entries from Redis' SLOWLOG
 
@@ -331,3 +332,9 @@ class Redis(AgentCheck):
 
         self._check_db(instance, custom_tags)
         self._check_slowlog(instance, custom_tags)
+
+    def _collect_metadata(self, info):
+        metadata_dict = {}
+        if info and 'redis_version' in info:
+            metadata_dict['version'] = info['redis_version']
+        self.svc_metadata(metadata_dict)
--- a/checks/__init__.py
+++ b/checks/__init__.py
@@ -49,6 +49,8 @@ class UnknownValue(CheckException): pass
 # and not this class. This class will be removed in a future version
 # of the agent.
 #==============================================================================
+
+
 class Check(object):
     """
     (Abstract) class for all checks with the ability to:
@@ -57,8 +59,6 @@ class Check(object):
     * only log error messages once (instead of each time they occur)
 
     """
-
-
     def __init__(self, logger):
         # where to store samples, indexed by metric_name
         # metric_name: {("sorted", "tags"): [(ts, value), (ts, value)],
@@ -66,7 +66,7 @@ class Check(object):
         #               None: [(ts, value), (ts, value)]}
         #                 untagged values are indexed by None
         self._sample_store = {}
-        self._counters = {} # metric_name: bool
+        self._counters = {}  # metric_name: bool
         self.logger = logger
         try:
             self.logger.addFilter(LaconicFilter())
@@ -316,7 +316,8 @@ class AgentCheck(object):
         self.warnings = []
         self.library_versions = None
         self.last_collection_time = defaultdict(int)
-        self.service_metadata = []
+        self._instance_metadata = []
+        self.svc_metadata = []
 
     def instance_count(self):
         """ Return the number of instances that are configured for this check. """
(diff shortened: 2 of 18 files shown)


checks/__init__.py
```diff
@@ -564,6 +586,14 @@ def run(self):
                     tb=traceback.format_exc()
                 )
             instance_statuses.append(instance_status)
+            # Generate empty metadata if nonexistent
+            service_metadata_count = len(self.service_metadata)
+            if service_metadata_count > i + 1:
```
remh: why does it matter ?

checks/__init__.py
```diff
@@ -502,6 +512,18 @@ def get_service_checks(self):
...
+    def get_service_metadata(self):
+        """
+        Return a list of the metadata dictionaries saved by the check -if any-
+        and clears them out of the instance's service_checks list
+
+        @return the list of metadata saved by this check
+        @rtype list of metadata dicts
+        """
+        service_metadata = self.service_metadata
+        self.service_metadata = []
```
remh: i think service_metadata should be a
```python
defaultdict(list)
```

use case: You have multiple versions of your application running on your instance

checks/__init__.py
yannmh: `self.service_metadata` is indexing integration metadata dictionnary by instance. 
This mechanism makes sure that the number of metadata dictionnaries generated matches the number of instances: if an instance does not produce any meta (purposely or because it's failing), we complete `self.service_metadata` adding some padding.

checks/__init__.py
yannmh: `self.service_metadata` is indexing metadata by instance. If we have multiple versions on an application running on different instances, we'll end up with `[{'version': 1.0}, {'version': 1.5}]`

checks.d/redisdb.py
```diff
@@ -329,3 +330,9 @@ def check(self, instance):
 
         self._check_db(instance, custom_tags)
         self._check_slowlog(instance, custom_tags)
+
+    def _collect_metadata(self, info):
+        metadata_dict = {}
+        if info and 'redis_version' in info:
+            metadata_dict['version'] = info['redis_version'].split('.')
```
remh: why not sending it as a string directly ? 
It probably better to have the same type for all metadata

checks/__init__.py
remh: I see, i don't really like the fact that it prevents from calling 
```python
self.svc_metadata
```
more than once per instance. This is bad pattern i think.

degemer:
Do you have a "standard" metadata payload ?
I noticed that on your `info` command screenshot, one of the version is a list of int, and the other one is a list of string.
I think it would be better to enforce a "standard" metadata payload (as we already had some weird type issues with the collector), or at least to add explicitly the advised types in the documentation.

remh:
Failing on Flake8, can you fix please ?

yannmh:
Test are passing. @remh Shall we merge it ?

PR #1612, opened 2015-05-11
[jmx] fix `jmx` commands false warning :bug:

yannmh:
**Bugfix** :bug:
`jmx` commands print a false warning, when it just ran fine.
Command outputs were ending with
```
Couldn't find any valid JMX configuration in your conf.d directory: /etc/dd-agent/conf.d
Have you enabled any JMX check ?
If you think it's not normal please get in touch with Datadog Support
```


remh:
looks good to me

PR #1614, opened 2015-05-12
[mesos] cast error message to string

degemer:
`AgentCheck.service_check` expect a string message. Otherwise it breaks the
emitter `json.dumps()`.


degemer:
@DorianZaccaria your new check has the same bug: https://github.com/DataDog/dd-agent/pull/1535/files#diff-a59500086f4aee21a686f3b35f12b189R166

LeoCavaille:
LGTM merging

LeoCavaille:
We can have a more generic casting somewhere else but this can be committed in another PR.

PR #1615, opened 2015-05-12
[Script] Fix package installation validation

remh:
Before that, if the forwarder was not in a ready state at the first
try, the installation script would fail directly without trying a few
more times as we used a trap on error and set -e.

This commit clears the trap and use set +e to allow a few failures for
the status verification.

### DIFF
--- a/packaging/datadog-agent/source/install_agent.sh
+++ b/packaging/datadog-agent/source/install_agent.sh
@@ -167,6 +167,14 @@ while [ "$c" -lt "30" ]; do
     c=$(($c+1))
 done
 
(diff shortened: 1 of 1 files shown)


remh:
@LeoCavaille can you review please ? it's making a few automated tests to fail.

remh:
going to refactor a bit to avoid the duplication of the trap restoration

LeoCavaille:
OK let me know when I can review.

remh:
Ready for review @LeoCavaille

LeoCavaille:
Thanks, looking good. Merging.

PR #1616, opened 2015-05-13
send fraction metrics as floats + cleaning

yannmh:
SQL fractional values represent ratios: it should be casted as `float` (rather than `int`) otherwise information is lost behind a rounded `0` int.

Default metrics impacted:
* `Buffer cache hit ratio` aka `sqlserver.buffer.cache_hit_ratio`

More information:
http://blogs.msdn.com/b/psssql/archive/2013/09/23/interpreting-the-counter-values-from-sys-dm-os-performance-counters.aspx

cc @simnyc


remh:
lgtm

PR #1617, opened 2015-05-13
[core] cast service_checks message to string

degemer:
As it might break the emitter if the message is not a string.
We cannot enforce a good use of service_checks in custom checks (and
also sometimes in our own checks,
see a874802ed52b0bdecc2340d5844795eff7c7d956)


remh:
looks good to me. let's merge once the CI is done.

LeoCavaille:
CI passed and looks good, merging now.

PR #1619, opened 2015-05-13
[jmx] run jmx as dd-agent user

yannmh:


degemer:
I canceled the build (afaik we don't test supervisor). The upgrade should be painless, logs are chowned, and jmxfetch is stopped.

remh:
LGTM

PR #1620, opened 2015-05-13
[core] run supervisor as dd-agent

degemer:
We don't need to run supervisor as root anymore.

Tested on:
- [x] debian 
- [ ] wheezy
- [ ] centos

Will launch a `dd-agent-testing` on them.

### DIFF
--- a/packaging/centos/datadog-agent.init
+++ b/packaging/centos/datadog-agent.init
@@ -102,7 +102,7 @@ start() {
     # no need to test for status before daemon,
     # the daemon function does the right thing
     echo -n "Starting Datadog Agent (using supervisord):"
-    daemon --pidfile=$SUPERVISOR_PIDFILE $SUPERVISORD_PATH -c $SUPERVISOR_CONF > /dev/null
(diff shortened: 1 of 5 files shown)


remh:
How does it play with the supervisor log file ?

degemer:
It won't be an issue, log files are chowned during the upgrade/install: https://github.com/DataDog/dd-agent-omnibus/blob/84a549d7c5dcdac7b59ccb067cdc23eede141110/package-scripts/datadog-agent/postinst#L52
Pidfile & sock are deleted when the agent is stopped, and the `/run` dir is also chowned: https://github.com/DataDog/dd-agent-omnibus/blob/84a549d7c5dcdac7b59ccb067cdc23eede141110/package-scripts/datadog-agent/postinst#L55

remh:
Let's test it once the migration to omnibus 4 is done.

yannmh:
This is planned for the 5.6.0

yannmh:
This also fix #1895

remh:
dd-agent-testing passes on this branch, merging.

PR #1628, opened 2015-05-18
[consul] add consul catalog check

talwai:
Adds a consul AgentCheck with the following metrics:

Number of Consul Agents in the Cluter
`consul.peers`: tagged by `consul_datacenter` and `mode` (`leader | follower`) 

Consul Catalog Nodes Up by Service
`consul.catalog.nodes_up`: tagged by `consul_datacenter` and `consul_service_id`

Consul Catalog Services Up by Node
`consul.catalog.services_up`: tagged by `consul_datacenter` and `consul_node_id`

This check also sends `consul.new_leader` events when a leader change is detected.

### DIFF
--- a/checks.d/consul.py
+++ b/checks.d/consul.py
@@ -31,7 +31,8 @@ class ConsulCheck(AgentCheck):
 
     def __init__(self, name, init_config, agentConfig, instances=None):
         AgentCheck.__init__(self, name, init_config, agentConfig, instances)
-
+        if instances is not None and len(instances) > 1:
+            raise Exception("Consul check only supports one configured instance.")
 
         self._local_config = None
         self._last_config_fetch_time = None
--- a/checks.d/consul.py
+++ b/checks.d/consul.py
@@ -1,6 +1,7 @@
 # stdlib
 from collections import defaultdict
 from datetime import datetime, timedelta
+from itertools import islice
 from urlparse import urljoin
 
 # project
@@ -20,6 +21,8 @@ class ConsulCheck(AgentCheck):
     SOURCE_TYPE_NAME = 'consul'
 
     MAX_CONFIG_TTL = 300 # seconds
+    MAX_SERVICES = 50 # cap on distinct Consul ServiceIDs to interrogate
+
     STATUS_SC = {
         'passing': AgentCheck.OK,
         'warning': AgentCheck.WARNING,
@@ -134,6 +137,21 @@ class ConsulCheck(AgentCheck):
     def get_services_on_node(self, instance, node):
         return self.consul_request(instance, '/v1/catalog/node/{0}'.format(node))
 
+    def _cull_services_list(self, services, service_whitelist):
+        if service_whitelist:
+            if len(service_whitelist) > self.MAX_SERVICES:
+                self.warning('More than %d services in whitelist. Service list will be truncated.' % self.MAX_SERVICES)
+
+            services = [s for s in services if s in service_whitelist][:self.MAX_SERVICES]
+        else:
+            if len(services) <= self.MAX_SERVICES:
+                self.warning('Consul service whitelist not defined. Agent will poll for all %d services found' % len(services))
+            else:
+                self.warning('Consul service whitelist not defined. Agent will poll for at most %d services' % self.MAX_SERVICES)
+                services = list(islice(services.iterkeys(), 0, self.MAX_SERVICES))
+
+        return services
+
     def check(self, instance):
         perform_new_leader_checks = instance.get('new_leader_checks',
                                                  self.init_config.get('new_leader_checks', False))
@@ -144,9 +162,6 @@ class ConsulCheck(AgentCheck):
             self.log.debug("This consul agent is not the cluster leader." +
                            "Skipping service and catalog checks for this instance")
             return
-
-        service_whitelist = instance.get('service_whitelist',
-                                         self.init_config.get('service_whitelist', []))
 
         service_check_tags = ['consul_url:{0}'.format(instance.get('url'))]
         perform_catalog_checks = instance.get('catalog_checks',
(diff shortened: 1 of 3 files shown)


checks.d/consul.py
```diff
@@ -0,0 +1,215 @@
...
+    CONSUL_CATALOG_CHECK = 'consul.catalog'
+    CONSUL_NODE_CHECK = 'consul.node'
+
+    def __init__(self, name, init_config, agentConfig, instances=None):
+        AgentCheck.__init__(self, name, init_config, agentConfig, instances)
+
+        self._last_known_leader = None
+
+    def consul_request(self, instance, endpoint):
+        url = "{0}{1}".format(instance.get('url'), endpoint)
```
remh: Nitpick but we should try to use urlparse.urljoin from now on to build urls.

checks.d/consul.py
```diff
@@ -0,0 +1,215 @@
...
+        except requests.exceptions.Timeout:
+            self.log.exception('Consul request to {0} timed out'.format(url))
+            raise
+
+        resp.raise_for_status()
+        return resp.json()
+
+    ### Consul Config Accessors
+    def _get_local_config(self, instance):
+        #TODO: Maybe store the config on `self` here so we don't have to keep retrieving it?
```
remh: We should implement your todo if it makes us save one http call per iteration. You could invalidate periodically, so store the timestamp it was last fetched and invalidate every 5min.

checks.d/consul.py
```diff
@@ -0,0 +1,215 @@
...
+            return
+
+        if leader != self._last_known_leader:
+            self.log.info(('Leader change from {0} to {1}. Sending new leader event').format(
+                self._last_known_leader, leader))
+
+            self.event({
+                "timestamp": int(time.time()),
+                "event_type": "consul.new_leader",
+                "api_key": self.agentConfig.get("api_key", ''),
```
remh: that's not needed:
https://github.com/DataDog/dd-agent/blob/master/checks/__init__.py#L464-L465

remh: You should pass a source type to your events though.

checks.d/consul.py
```diff
@@ -0,0 +1,215 @@
...
+            self.service_check(self.CONSUL_CHECK, AgentCheck.OK,
+                               tags=service_check_tags)
+
+        if perform_catalog_checks:
+            services = self.get_services_in_cluster(instance)
+            if service_whitelist:
+                services = [s for s in services if s in service_whitelist]
+            else:
+                #Default to polling all services
+                self.log.warn('Consul service whitelist not defined. Agent will poll for all %d services found' % len(services))
```
remh: Should we use self.warning instead to show that in the info page ?

remh: Also we should limit the number of services to pull, you don't want to pull 1000 services.

checks.d/consul.py
```diff
@@ -0,0 +1,215 @@
...
+        self._last_known_leader = leader
+
+    ### Consul Catalog Accessors
+    def get_services_in_cluster(self, instance):
+        return self.consul_request(instance, '/v1/catalog/services')
+
+    def get_nodes_in_cluster(self, instance):
+        return self.consul_request(instance, '/v1/catalog/nodes')
+
+    def get_nodes_with_service(self, instance, service, tag=None):
```
remh: It doesn't look like you are ever calling this function with "tag" set.

checks.d/consul.py
```diff
@@ -0,0 +1,215 @@
...
+                           len(nodes_with_service),
+                           tags=node_tags)
+
+                for n in nodes_with_service:
+                    node_id = n.get('Node') or None
+
+                    if not node_id:
+                        continue
+
+                    if node_id in nodes_to_services:
```
remh: if nodes_to_services was defined as 
```python
nodes_to_services = defaultdict(list)
```
You wouldn't need this extra logic

checks.d/consul.py
```diff
@@ -0,0 +1,215 @@
...
+                        nodes_to_services[node_id].append(service)
+                    else:
+                        nodes_to_services[node_id] = [service]
+
+            for node, services in nodes_to_services.items():
```
remh: With python2, you should use .iteritems() instead of .items() to avoid doing a copy of the dictionary.

checks.d/consul.py
```diff
@@ -0,0 +1,215 @@
...
+                                         self.init_config.get('service_whitelist', []))
+
+        service_check_tags = ['consul_url:{0}'.format(instance.get('url'))]
+        perform_catalog_checks = instance.get('catalog_checks',
+                                              self.init_config.get('catalog_checks'))
+
+        try:
+            health_state = self.consul_request(instance, '/v1/health/state/any')
+
+            STATUS_SC = {
```
remh: That should probably be defined as a module or class constant

checks.d/consul.py
```diff
@@ -0,0 +1,215 @@
...
+    CONSUL_CHECK = 'consul.up'
+    HEALTH_CHECK = 'consul.check'
+
+    CONSUL_CATALOG_CHECK = 'consul.catalog'
+    CONSUL_NODE_CHECK = 'consul.node'
+
+    def __init__(self, name, init_config, agentConfig, instances=None):
+        AgentCheck.__init__(self, name, init_config, agentConfig, instances)
+
+        self._last_known_leader = None
```
remh: This is not multi instance safe, but i guess, i don't see a case where one would be running multiple consul clusters. 
Can you check that similarly to what is done in the disk check ? https://github.com/DataDog/dd-agent/blob/master/checks.d/disk.py#L26-L27

remh:
It looks great besides the nitpick! Looking forward to release this!

talwai:
@remh - ping, does this look ready to merge?

remh:
http://zanyimages.com/Sports%20Lovers/Go%20for%20it%20!.jpg
image

PR #1630, opened 2015-05-18
[wmi] cast `tag_by` value as a string

yannmh:
As it could return an integer, for instance
```yaml
init_config:

instances:

#   Fetch metrics for a single running application
  - class: Win32_PerfFormattedData_PerfProc_Process
    metrics:
      - ...
    filters:
      - Name: appname
    tag_by: IDProcess
```

cc @simnyc


LeoCavaille:
:+1: thanks @yannmh

PR #1631, opened 2015-05-19
[sqlserver] properly close cursor, avoid leaks

LeoCavaille:
We saw memory leak issues with that check. It could have been caused by
two different things:

* The fact that we are not closing cursors properly, and therefore not
  freeing properly the recordsets loaded in them
* We instantiated wrongly two different cursors in the main check loop
  letting one leak...


hkaj:
lgtm

PR #1633, opened 2015-05-20
[Postgresql] logging bug for custom metrics

orenmazor:
This is actually a combination of two bugs:

1. If you use custom metrics and one of your queries has too many results, you will discover a typo where we attempt to log to 'self.warning', which is not a thing.

2. the brackets *on* that call to self.warning result in attempting to format the function result rather than the parameter.

### DIFF
--- a/checks.d/postgres.py
+++ b/checks.d/postgres.py
@@ -424,9 +424,9 @@ SELECT relname,
                     continue
 
                 if scope in custom_metrics and len(results) > MAX_CUSTOM_RESULTS:
-                    self.warning(
-                        "Query: {0} returned more than {1} results ({2}). Truncating").format(
-                        query, MAX_CUSTOM_RESULTS, len(results))
+                    log_func(
+                        "Query: {0} returned more than {1} results ({2}). Truncating".format(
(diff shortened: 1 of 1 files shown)


LeoCavaille:
Hey @orenmazor thanks a lot for contributing and catching that.
For 1., the `self.warning` method actually [exists](https://github.com/DataDog/dd-agent/blob/5.3.2/checks/__init__.py#L507-L510), it's meant to display the warning on the info page and in  [Datadog](https://app.datadoghq.com/infrastructure?filter=agent_issue).

For 2., you're completely right this is a typo. Also it looks like the CI is failing because you have a flake8 error. You can run `rake lint` (see also [how to setup a dev env](https://github.com/DataDog/dd-agent#setup-your-environment)) before pushing.

orenmazor:
Fixed! I didnt have flake8 built into vim in this env so I didn't catch that.

makes sense about self.warning. this was a flyby fix on one of my servers that I just caught. not super familiar with dd-agent.

LeoCavaille:
@orenmazor thanks, merged in a rebased version as 071d06ff0a85756552a6e227dc2cf32e6576f067 in #1634

PR #1636, opened 2015-05-21
[postgres/ci] delay PG init in CI

LeoCavaille:
Otherwise this causes psql to fail when we load data.


degemer:
:+1: , let's avoid these kind of errors ! https://travis-ci.org/DataDog/dd-agent/jobs/63087213#L232

PR #1638, opened 2015-05-21
[memory_snapshot] First pass

remh:
This is a quick check that will send an event with the list of processes that have the highest memory footprint (rss) when the overall memory usage is higher than X percent

It's not ready to be merged yet but it's here for discussion.

Still need to add a test, docs, and get some feedback and get a better rendering.

E.g. of event it's creating:

image

### DIFF
--- a/checks.d/memory_snapshot.py
+++ b/checks.d/memory_snapshot.py
@@ -1,0 +1,74 @@
+# stdlib
+import time
+
+# 3rd party
+try:
+    import psutil
+except ImportError:
+    psutil = None
+
+# project
+from checks import AgentCheck
+
+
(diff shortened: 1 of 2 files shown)


remh:
Not sure we want to merge that yet, as we should rework the process tree instead but it's here for discussion and for version control.

remh:
Another example https://app.datadoghq.com/event/event?id=2818088290886882367

VynceMontgomery:
on IRC it came up that this functionality, or something like it, but for CPU would also be very nice to have.

gmmeyer:
Closing in the integration sdk cleanup.

PR #1639, opened 2015-05-22
[jmxfetch] properly stop on windows :bug:

yannmh:
Enhance `ProcessWatchDog` win32/agent.py class to support 'cleanup'
methods. When set, a thread is attached to the process, waits for a
trigger and invoke 'cleanup'.

Alternative here: https://github.com/DataDog/dd-agent/pull/1643

### DIFF
--- a/win32/agent.py
+++ b/win32/agent.py
@@ -3,10 +3,11 @@ import logging
 import modules
 import multiprocessing
 from optparse import Values
-import servicemanager
 import sys
-import threading
 import time
+
+# 3rd
+import servicemanager
 import tornado.httpclient
 from win32.common import handle_exe_click
 import win32event
@@ -14,7 +15,8 @@ import win32evtlogutil
 import win32service
 import win32serviceutil
 
-# DD
+
+# datadog
 from checks.collector import Collector
 from config import (
     get_config,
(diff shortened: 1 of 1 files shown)


win32/agent.py
```diff
@@ -5,7 +5,6 @@
 from optparse import Values
 import servicemanager
 import sys
-import threading
```
remh: Can you reorganize the imports while you're at it please ?

remh:
If i read the code properly stop() is not called:
https://github.com/DataDog/dd-agent/blob/yann/fix-win-jmx-no-restart/win32/agent.py#L82-L89

So i don't think this fix works.

I think we should use a simpler solution and just send a suicide signal to jmxfetch through a temp file, it would be way easier to read and to implement and will also remove a layer of complexity as you won't need your new thread anymore.

yannmh:
Thanks for your feedback @remh. 
You're right it should be changed to `stop` instead: `ProcessWatchDog ` will then trigger `terminate` or a custom `cleanup` method depending on the situation.
https://github.com/DataDog/dd-agent/blob/yann/fix-win-jmx-no-restart/win32/agent.py#L184

I am writing the suggested alternative, i.e. communicating exit order through a file, so we can compare both solutions. I'll let you know when the PR is ready.

remh:
closing in favor of https://github.com/DataDog/dd-agent/pull/1643

PR #1640, opened 2015-05-26
[sqlserver] fix collector hangs when a connection is failing

yannmh:
SQLServer check hangs the collector when a connection is failing.
Default timeout is set to 30s, but it is raised multiple times for the same instance: after 120s, collector is restarted by the watchdog and loops back.
Solution: cache failing connections.

### DIFF
--- a/checks.d/sqlserver.py
+++ b/checks.d/sqlserver.py
@@ -44,6 +44,13 @@ VALUE_AND_BASE_QUERY = '''select cntr_value
                           order by cntr_type;'''
 
 
+class SQLConnectionError(Exception):
+    """
+    Exception raised for SQL instance connection issues
+    """
+    pass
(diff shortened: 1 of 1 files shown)


remh:
Does that mean that as soon as one connection fails, the check won't collect any metrics until a restart of the agent ?

yannmh:
I added a few changes. 
Failure connections are now cached at initialization only: in this case, the check would not collect any metrics until the agent is restarted. Please note that this is also the current behavior, as the list of metrics to collect is set at this time.

If the check initialization ran fine, but the connection is failing afterwards, we'll keep retrying at every check run: the downside is that it slows down the collector since it takes 30seconds to timeout so .

remh:
Looks good!

PR #1643, opened 2015-05-26
[jmxfetch][windows] write to file to exit JMXFetch

yannmh:
Order JMXFetch to exit, communicating through a file. Windows only.

Requires https://github.com/DataDog/jmxfetch/pull/58

Alternative: https://github.com/DataDog/dd-agent/pull/1639

### DIFF
--- a/jmxfetch.py
+++ b/jmxfetch.py
@@ -9,6 +9,7 @@ import time
 
 # datadog
 from util import get_os, yLoader, yDumper
+from utils.platform import Platform
 from config import get_config, get_confd_path, get_jmx_status_path, get_logging_config, \
     PathNotFound, DEFAULT_CHECK_FREQUENCY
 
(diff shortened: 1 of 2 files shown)


remh:
Nice, can we also clean the exit file when we start jmxfetch to make sure that jmxfetch starts even though the exit file didn't get cleaned properly before ?

yannmh:
Done, thanks @remh !

jmxfetch.py
```diff
@@ -230,6 +231,13 @@ def _start(self, path_to_java, java_run_opts, jmx_checks, command, reporter, too
                 command,  # Name of the command
             ]
 
+            if get_os() == 'windows':
```
remh:
```python
if Platform.is_windows()
```

win32/agent.py
```diff
@@ -282,10 +280,16 @@ def __init__(self, agentConfig, hostname):
...
             self.jmx_daemon.run()
 
-    def stop(self):
-        pass
+    def terminate(self):
+        """
+        Override `terminate` method to properly exit JMXFetch.
+        """
+        JMXFetch.write_exit_file()
+        self.join()
```
remh: Do we need that ? Can we just write the exit file and let jmxfetch closes whenever ? otherwise it might take up to 15 seconds to stop.

We just need to make sure to delete the exit file when starting the agent.

yannmh:
Brought the last changes suggested. Merging

PR #1644, opened 2015-05-27
[http_check] Fix status type err. in service check

elafarge:
PyYaml was returning an int for the expected status code when
"http_response_status_code: 200" (for instance) was used in the yaml
config while a Regex (and therefore an string) is expected. I added a
cast to avoid that from happening again. I also added some doc to tell 
our customers about the possibility to use a Regex for this parameter.


elafarge:
Thanks for the merge :)

PR #1645, opened 2015-05-27
[varnish] allow tags in varnish 3 xml style parsing

LeoCavaille:
The tags were just not passed at all in that case.
Also assert the tags in the CI, to avoid that kind of problems.


remh:
looks good to me!

PR #1649, opened 2015-05-29
[packaging] Fix install-script to work w/ python3

elafarge:
Our one-line install script wasn't compatible with Python3... we're
executing a single line of Python there and it contains an old-style
print which was making the install script crash. I just switched to the
new print/format syntax which should be compatible with any version of
Python.


remh:
Thanks, looks good to me!

elafarge:
Thanks for the merge :)

PR #1650, opened 2015-05-29
[Core] Proxy, fix the way we set the no proxy environment setting

### LINKED ISSUE #1594: [Core] Hardcoded no_proxy setting causes problems for EC2 metadata
Following this pull request: https://github.com/DataDog/dd-agent/pull/1518
`no_proxy` is now hardcoded to `127.0.0.1,localhost` in `dogstatsd.py` and `emitter.py`

Since this is set in the environment it affects fetching of EC2 metadata on AWS as that is fetched from `169.254.169.254`. If `http_proxy` is set, these requests will be sent through the proxy and return incorrect metadata.

This should be changed to respect addresses set in `no_proxy` and only append `127.0.0.1,localhost` if they are missing.

remh:
Fix #1594

Also:
* simplify the proxy parsing
* Move the code in a new utils module

### DIFF
--- a/utils/proxy.py
+++ b/utils/proxy.py
@@ -1,0 +1,66 @@
+import logging
+import os
+from urllib import getproxies
+from urlparse import urlparse
+
+log = logging.getLogger(__name__)
+
+def set_no_proxy_settings():
+    # Starting with Agent 5.0.0, there should always be a local forwarder
+    # running and all payloads should go through it. So we should make sure
+    # that we pass the no_proxy environment variable that will be used by requests
+    # See: https://github.com/kennethreitz/requests/pull/945
+    to_add = ["127.0.0.1", "localhost", "169.254.169.254"]
+    no_proxy = os.environ.get("no_proxy", "")
+    if not no_proxy.strip():
+        no_proxy = []
+    else:
+        no_proxy = no_proxy.split(',')
+
+    for host in to_add:
+        if host not in no_proxy:
+            no_proxy.append(host)
+
+    os.environ['no_proxy'] = ','.join(no_proxy)
+
+def get_proxy(agentConfig):
+    proxy_settings = {}
+
+    # First we read the proxy configuration from datadog.conf
+    proxy_host = agentConfig.get('proxy_host')
+    if proxy_host is not None:
+        proxy_settings['host'] = proxy_host
+        try:
+            proxy_settings['port'] = int(agentConfig.get('proxy_port', 3128))
+        except ValueError:
+            log.error('Proxy port must be an Integer. Defaulting it to 3128')
+            proxy_settings['port'] = 3128
+
+        proxy_settings['user'] = agentConfig.get('proxy_user')
+        proxy_settings['password'] = agentConfig.get('proxy_password')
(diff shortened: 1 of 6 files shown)


remh:
Looks like this one is GTG. Tests are passing despite what the travis -> github integration says.

utils/proxy.py
```diff
@@ -0,0 +1,66 @@
...
+        if host not in no_proxy:
+            no_proxy.append(host)
+
+    os.environ['no_proxy'] = ','.join(no_proxy)
+
+def get_proxy(agentConfig):
+    proxy_settings = {}
+
+    # First we read the proxy configuration from datadog.conf
+    proxy_host = agentConfig.get('proxy_host', None)
```
LeoCavaille: `.get(_, None)` from you @remh I am surprised :)

utils/proxy.py
```diff
@@ -0,0 +1,66 @@
...
+    if proxy_host is not None:
+        proxy_settings['host'] = proxy_host
+        try:
+            proxy_settings['port'] = int(agentConfig.get('proxy_port', 3128))
+        except ValueError:
+            log.error('Proxy port must be an Integer. Defaulting it to 3128')
+            proxy_settings['port'] = 3128
+
+        proxy_settings['user'] = agentConfig.get('proxy_user', None)
+        proxy_settings['password'] = agentConfig.get('proxy_password', None)
```
LeoCavaille: same here

utils/proxy.py
```diff
@@ -0,0 +1,66 @@
...
+        log.debug("Proxy Settings: %s:%s@%s:%s" % (proxy_settings['user'],
+            "*****", proxy_settings['host'], proxy_settings['port']))
```
LeoCavaille: nitpick: The "standard" way is to either pass the interpolation variables as the variadic args in log or use a format.
```python
log.debug("%s %s", string1, string2)
# or
log.debug("{0} {1}".format(string1, string2))
```

utils/proxy.py
remh: ah yeah, i copy pasted the old code around, will fix. Thanks

utils/proxy.py
```diff
@@ -0,0 +1,66 @@
...
+        return proxy_settings
+
+    # If no proxy configuration was specified in datadog.conf
+    # We try to read it from the system settings
+    try:
+        proxy = getproxies().get('https', None)
```
LeoCavaille: another occurrence here

utils/proxy.py
```diff
@@ -0,0 +1,66 @@
...
+            proxy_settings['user'] = parse.username
+            proxy_settings['password'] = parse.password
+
+            log.debug("Proxy Settings: %s:%s@%s:%s" % (proxy_settings['user'],
+                "*****", proxy_settings['host'], proxy_settings['port']))
+            return proxy_settings
+
+    except Exception, e:
+        log.debug("Error while trying to fetch proxy settings using urllib %s."
+            "Proxy is probably not set" % str(e))
```
LeoCavaille: same remark for the 2 log lines above.

LeoCavaille:
Except for the few nitpicks, LGTM. Thanks for isolating this in a new _ad hoc_ module.

PR #1652, opened 2015-06-01
[flake8-monday] enable F401 (unused import)

degemer:
And some cleaning.

Flake8 doesn't detect when a module gets reimported somewhere else, so I commented these lines with `# noqa`. (it should only be the case for `util.py`, we'll see if something is broken with the CI)


LeoCavaille:
:100: pretty impressed that you flaked8 some code before dawn.
This deserves a merge now, looking great :+1:

degemer:
:boom: :trophy:

PR #1654, opened 2015-06-01
Add ssl support for etcd plugin

KnownSubset:
Added in optional support etcd ssl configuration.
Users have to opt in to use the ssl monitoring.
Modified the example etcd.yaml.example to reflect the changes.

### DIFF
--- a/checks.d/etcd.py
+++ b/checks.d/etcd.py
@@ -70,6 +70,7 @@ class Etcd(AgentCheck):
         url = instance['url']
         instance_tags = instance.get('tags', [])
 
+        # Load the ssl configuration
         ssl_params = {
             'ssl': _is_affirmative(instance.get('ssl', False)),
             'ssl_keyfile': instance.get('ssl_keyfile', None),
@@ -157,7 +158,8 @@ class Etcd(AgentCheck):
     def _get_json(self, url, ssl_params, timeout):
         try:
             if ssl_params['ssl']:
-                r = requests.get(url, verify=ssl_params['ssl_cert_reqs'], cert=(ssl_params['ssl_certfile'], ssl_params['ssl_keyfile']), timeout=timeout, headers=headers(self.agentConfig))
+                certificate = (ssl_params['ssl_certfile'], ssl_params['ssl_keyfile']) if 'ssl_certfile' in ssl_params and 'ssl_keyfile' in ssl_params else None
+                r = requests.get(url, verify=ssl_params['ssl_cert_reqs'], cert=certificate, timeout=timeout, headers=headers(self.agentConfig))
             else:
                 r = requests.get(url, timeout=timeout, headers=headers(self.agentConfig))
         except requests.exceptions.Timeout:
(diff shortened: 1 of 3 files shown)


checks.d/etcd.py
```diff
@@ -68,6 +68,17 @@ def check(self, instance):
         # Load values from the instance config
         url = instance['url']
         instance_tags = instance.get('tags', [])
+
+        ssl_params = {
+            'ssl': instance.get('ssl', False),
```
LeoCavaille: You probably want to use `_is_affirmative` see https://github.com/DataDog/dd-agent/commit/0e862f5f7f16f5fc1466f4b8ebff1b4e44cac64e.

checks.d/etcd.py
```diff
@@ -68,6 +68,17 @@ def check(self, instance):
...
+            'ssl_keyfile': instance.get('ssl_keyfile', None),
+            'ssl_certfile': instance.get('ssl_certfile', None),
+            'ssl_cert_reqs':  instance.get('ssl_cert_reqs', True)
```
LeoCavaille: Same here.

checks.d/etcd.py
LeoCavaille: Also should SSL be enabled by default when using a `https` scheme? Or can we maybe get rid of this boolean and just check for that.

checks.d/etcd.py
```diff
@@ -133,18 +144,21 @@ def check(self, instance):
...
-        return self._get_json(url + "/v2/stats/leader", timeout)
+    def _get_leader_metrics(self, url, ssl_params, timeout):
+        return self._get_json(url + "/v2/stats/leader", ssl_params, timeout)
 
-    def _get_json(self, url, timeout):
+    def _get_json(self, url, ssl_params, timeout):
         try:
-            r = requests.get(url, timeout=timeout, headers=headers(self.agentConfig))
+            if ssl_params['ssl']:
+                r = requests.get(url, verify=ssl_params['ssl_cert_reqs'], cert=(ssl_params['ssl_certfile'], ssl_params['ssl_keyfile']), timeout=timeout, headers=headers(self.agentConfig))
```
LeoCavaille: If `ssl_certfile` was `None` before you removed it from the `ssl_params` dict so here you'd likely get a `KeyError`

LeoCavaille:
Thanks for working on the etcd check @KnownSubset!
I've made a few comments on your work. Also check the Travis failures, in the etcd check it looks like the tests are not passing anymore, if you want to run the tests locally I'd recommend:

1. https://github.com/DataDog/dd-agent#setup-your-environment
2. `rake lint`
3. `rake ci:run[etcd]`

checks.d/etcd.py
KnownSubset: We probably cannot use `_is_affirmative` for the `'ssl_cert_reqs'` as it can be used to pass in the location of the certificate authority certificate.  Unless you think we should just change it over to be explicitly for the CA file.

olivielpeau:
Thanks again @KnownSubset for contributing!

I've rebased your changes on the current master and changed a few things, see above for the updated PR that will be merged.

PR #1656, opened 2015-06-03
A tiny change to account for %guest cpu time.

rhormaza:
This metric is relevant when running the agent in a Linux host that runs a hypervisor such as KVM.

### DIFF
--- a/checks/system/unix.py
+++ b/checks/system/unix.py
@@ -586,8 +586,8 @@ class Cpu(Check):
         """Return an aggregate of CPU stats across all CPUs
         When figures are not available, False is sent back.
         """
-        def format_results(us, sy, wa, idle, st):
-            data = {'cpuUser': us, 'cpuSystem': sy, 'cpuWait': wa, 'cpuIdle': idle, 'cpuStolen': st}
+        def format_results(us, sy, wa, idle, st, guest=None):
+            data = {'cpuUser': us, 'cpuSystem': sy, 'cpuWait': wa, 'cpuIdle': idle, 'cpuStolen': st, 'cpuGuest'; guest}
             for key in data.keys():
                 if data[key] is None:
                     del data[key]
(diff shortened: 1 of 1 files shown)


degemer:
Hey @rhormaza, thanks for your interest !
Have you tried the `system_core` check ? It collects CPU information by core, and gives more details than the default system check (ie %guest is included). Metric prefix is `system.core.` (for instance `system.core.guest`, tagged by core).
Let us know if it's sufficient for you !

LeoCavaille:
I'd agree that it would be nice to enable %guest by default (in agent 6, or try to roll it out slowly). Otherwise this is the kind of graphs which show up on the default host dashboards (if you run KVM or xen, ...), the gap to 100% accounts for %guest.
screen shot 2015-06-29 at 3 50 27 pm

rhormaza:
Thanks for the hint @degemer  I was not aware of those `system_core` checks, I'll have a look. That said, I'd definitely recommend to consider adding %guest as default just to make the product better :) 
The graphs from our KVM hosts look exactly as the screenshot uploaded by Leo. IMHO, that shouldn't happen as it is misleading and inaccurate.

Thanks!

remh:
I'm OK to enable it by default if available, however this change itself won't be enough.

Metrics captured here require post processing in the server side unfortunately so this won't be enough.

We should probably take a quick pass and rewrite cpu metrics collection into a separate check. Maybe using psutil.

rhormaza:
Hi remh,

Yes, I did realize that when I put the changes in a few servers, the metric was presented just as `Guest`  which was different to the others. 
I'd image that the effort to change the backend wouldn't be that much as it is just adding another metric, however that is my assumption and it can be entirely wrong :)

olivielpeau:
Closing this PR as I've rebased it in #1845 . Thanks @rhormaza!

Issue #457, opened 2013-04-16
Use `_local` stats for elasticsearch


conorbranagan:
Instead of:
`curl localhost:9200/_nodes/stats?all`

We can use:
`curl localhost:9200/_nodes/_local/stats?all`

This means we can get rid of the hostname matching logic because we can assume the data is about just the local node.

Will have to check if this works in older versions of ES.

PR #1657, opened 2015-06-04
[elasticsearch] Remove hostname matching logic

olivielpeau:
The hostname matching is not needed anymore as:
- since PR #1181 we only ask the _local node for stats when `is_external` is set to `false`
- we don't match the hostname at all when `is_external` is `true`

Matching hostnames can also filter out legitimate data when the
local elasticsearch node reports a different hostname.

See also issue #457


checks.d/elastic.py
```diff
@@ -359,88 +356,15 @@ def _process_stats_data(self, nodes_url, data, stats_metrics, config):
...
-        ips = []
-        for iface in out.split("\n"):
-            iface = iface.strip()
-            if iface:
-                ips.append(iface.split(' ')[1].split('/')[0])
-
-        # Check the interface addresses against the primary address
-        return primary_addrs in ips
+            for metric in stats_metrics:
+                desc = stats_metrics[metric]
```
talwai: Nitpick but this loop can be simplified: `for metric, desc in stats_metrics.iteritems()`. This way is also usually faster than iterating and looking up by key

olivielpeau:
Thanks for the review @talwai ! I've changed the loop.

PR #1658, opened 2015-06-04
Modified yum install line to disable all other repos and only enable the DataDog repo

ip2k:
This fixes situations where a system might have broken repos, or repos that aren't available in specific environments. An example would be a company with an internal repo which is available during a build phase, but not available in a prod environment. I personally ran into this situation :)


degemer:
Thanks a lot @ip2k !
I rebased your PR to add the same ignore for `apt`: https://github.com/DataDog/dd-agent/pull/1695.
Our `master` branch is currently frozen, waiting for the `5.4` release, but once it's released we'll do some more tests (different versions of `yum`, just in case) and then merge your work.
Closing this PR.

PR #1664, opened 2015-06-08
[windows] record restarts on a specific timeframe

yannmh:
Drop the global count of restarts in favor of a timeframe based one.

### DIFF
--- a/win32/agent.py
+++ b/win32/agent.py
@@ -1,4 +1,5 @@
 # stdlib
+from collections import deque
 import logging
 import modules
 import multiprocessing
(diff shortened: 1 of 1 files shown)


win32/agent.py
```diff
@@ -145,14 +151,21 @@ def is_alive(self):
...
+            self._restarts.popleft()
+
+        return len(self._restarts) < self._MAX_RESTARTS
+
     def restart(self):
-        self._count_restarts += 1
-        if self._count_restarts >= self._MAX_RESTARTS:
+        if not self._can_restart():
             servicemanager.LogInfoMsg(
                 "%s reached the limit of restarts. Not restarting..." % self._name)
```
degemer: To ease a potential debug, could you add the number of restarts (or tries), max number of restarts, and maybe the time window ? 
So something like:
```
"{0} reached the limit of restarts ({1} tries during the last {3}s (max authorized: {2})). Not restarting...".format(self._name, len(self._restarts), self._restarts[0] - time.time()) ?
```

yannmh: Good idea, thanks.

win32/agent.py
```diff
@@ -127,10 +128,15 @@ class ProcessWatchDog(object):
...
+
     def __init__(self, name, process, max_restarts=5):
+        """
+        :param max_restarts: maximum number of restarts per _RESTART_TIMEFRAME timeframe.
+        """
         self._name = name
         self._process = process
-        self._count_restarts = 0
+        self._restarts = deque([])
         self._MAX_RESTARTS = max_restarts
```
degemer: I don't really like this way (not your PR, but close), having a "dynamic" constant. Maybe you could instead have a `DEFAULT_MAX_RESTARTS = 5`, `max_restarts=None`, and `self._max_restarts = max_restarts or self.DEFAULT_MAX_RESTARTS`: what do you think ?

degemer:
Nice improvement!
LGTM, if you could test a custom build before merging it, it would be perfect!

PR #1666, opened 2015-06-09
[disk] fix empty device regex exclusion issue

degemer:
When `device_blacklist_re` or `excluded_disk_re` were empty (let's say in
datadog.conf `device_blacklist_re: `), the disk check was failing, because all
disks were excluded.

This is fixed.

### DIFF
--- a/checks.d/disk.py
+++ b/checks.d/disk.py
@@ -70,7 +70,7 @@ class Disk(AgentCheck):
                 "Using `{0}` in datadog.conf has been deprecated"
                 " in favor of `{1}` in disk.yaml".format(legacy_name, option)
             )
-            value = self.agentConfig.get(legacy_name)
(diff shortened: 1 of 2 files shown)


checks.d/disk.py
```diff
@@ -70,7 +70,7 @@ def _load_legacy_option(self, instance, option, default,
                 "Using `{0}` in datadog.conf has been deprecated"
                 " in favor of `{1}` in disk.yaml".format(legacy_name, option)
             )
-            value = self.agentConfig.get(legacy_name)
+            value = self.agentConfig.get(legacy_name) or default
```
yannmh: or simply `value = self.agentConfig.get(legacy_name, default)`

yannmh: my bad it doesn't work

yannmh:
Good to go to me !  Thanks !

PR #1667, opened 2015-06-09
[CI] [Docker] Setup CircleCI builds for the Docker check.

hkaj:
This change allows the use CIrcleCI for integration tests.

To do so, simply add a setup file under `Flavored CircleCI jobs` in the `Rakefile` (**note**: Unlike Travis, CircleCI allows you to use `sudo`, so the installation step should be much simpler :tada: ), also add your new test name in the `CIRCLECI_FLAVORS` comma-separated list in `circle.yml`.

The command to run the tests is `rake ci:run_circle`. It's worth noting that the `tests/circleci.sh` script leverages all the containers that we have on CircleCI, and run the tests in parallel in them, so getting more horse power in CircleCI will mechanically speed up the test run.


circle.yml
```diff
@@ -0,0 +1,16 @@
...
+    SKIP_CLEANUP: true
+
+dependencies:
+  pre:
+    - pip install --use-wheel pyro4==4.35
+    - bundle install --system
+
+test:
+  override:
+    - rake ci:run[docker]
```
hkaj: forgot to remove the hardcoded part :neutral_face: I'll fix that asap.

remh:
Thjs PR is probably worth a description :)

tests/circleci.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+i=0
+tests=()
+exit_code=0
+IFS=',' read -ra flavors <<< "$CIRCLECI_FLAVORS"
+for flavor in $flavors
```
degemer: Why are you using a `bash` script when you have the power and the ease of `ruby` and `rake` ?
For instance https://github.com/degemer/dd-agent/blob/7b529d4fb89277226f21ac3edddcdec103f433a9/ci/resources/circleci.rb#L30-L44 (I was taking the flavors from the `.travis.yml` file, not wanted here, but I guess you could do the same with a `Rake::Task['ci:run_circle'].invoke(flavor)`)

circle.yml
```diff
@@ -0,0 +1,17 @@
+machine:
+  python:
+    version: 2.7.9
+  environment:
+    SKIP_CLEANUP: true
+    # comma separated list
+    CIRCLECI_FLAVORS: docker
+
+dependencies:
+  pre:
```
degemer: You probably want to override this phase, as otherwise `python setup.py install` and others are run, which could lead to unwanted effects.

circle.yml
```diff
@@ -0,0 +1,17 @@
...
+    - pip install --use-wheel pyro4==4.35
```
degemer: Why do you need specifically `pyro4==4.35` on CircleCI ?

Rakefile
```diff
@@ -113,6 +116,14 @@ namespace :ci do
     flavors = flavor.split(',')
     flavors.each { |f| Rake::Task["ci:#{f}:execute"].invoke }
   end
+
+  desc 'Run integration tests in CircleCI'
+  task :run_circle, :flavor do |_, args|
+    puts 'Assuming you are running these tests locally' unless ENV['TRAVIS']
+    flavor = args[:flavor] || ENV['CIRCLECI_FLAVOR'] || 'default'
+    flavors = flavor.split(',')
+    flavors.each { |f| Rake::Task["ci:#{f}:execute"].invoke }
```
degemer: It's not possible to test against different versions I guess then ?

ci/docker.rb
```diff
@@ -0,0 +1,55 @@
...
+      sh %(docker rm redis)
+      sh %(docker kill mongo)
+      sh %(docker rm mongo)
+    end
+
+    task :execute do
+      exception = nil
+      begin
+        %w(before_install install before_script
+           script before_cache cache).each do |t|
```
degemer: `before_cache` and `cache` are Travis-specific, they're not needed here.

degemer:
First test of the Docker check incoming ! :tada: 

Compared to Travis, with CircleCI we have access to `docker` and `sudo`, is that it ?
The only python version tested is `2.7`, it's still better than no tests at all, but did you find a way to test multiple versions @hkaj ?
Also if you want to change more stuff from the default build, you could create a `ci/circle.rb` (or `ci/resources/circle.rb`) and have the new `ci/flavors.rb` inherit from it.
CircleCI has an awesome way of displaying errors if you give it `junit` error files: https://github.com/degemer/dd-agent/commit/7b529d4fb89277226f21ac3edddcdec103f433a9#diff-131637f8162c817832b20b72b2ceb819R166

circle.yml
hkaj: indeed

circle.yml
hkaj: this was somehow needed because I didn't override this step, it's no longer the case now that I do.

PR #1571, opened 2015-04-22
[tests] split tests/ into two subdirectories

degemer:
Separate agent core tests and all integration tests. I think it looks nicer and cleaner.

It's a WIP for now. I can't/don't want to run all tests locally, so I will fix them once they fail on Travis.

Issue #1458, opened 2015-03-23
[tests] review of all our integration tests


degemer:
@hkaj :
- [ ] `btrfs.py`
- [x] `fluentd.py` - https://github.com/DataDog/dd-agent/pull/1574
- [x] ~~`lighttpd.py` - https://github.com/DataDog/dd-agent/pull/1473~~
- [ ] `mesos_slave.py`
- [ ] `network.py`
- [ ] `nginx.py`
- [ ] `redisdb.py`
- [ ] `sqlserver.py`
- [ ] `tcp_check.py`

@LeoCavaille :
- [ ] `cacti.py`
- [ ] `couchbase.py`
- [ ] `jenkins.py`
- [ ] `kafka_consumer.py`
- [ ] `mesos.py`
- [ ] `mongo.py`
- [ ] `ntp.py`
- [ ] `postfix.py`
- [x] ~~`rabbitmq.py` - https://github.com/DataDog/dd-agent/pull/1472~~
- [ ] `riakcs.py`
- [ ] `varnish.py`

@degemer :
- [x] ~~`couch.py` - https://github.com/DataDog/dd-agent/pull/1462~~
- [x] `go_expvar.py` - https://github.com/DataDog/dd-agent/pull/1491
- [x] ~~`haproxy.py` - https://github.com/DataDog/dd-agent/pull/1466~~
- [x] ~~`http_check.py` - https://github.com/DataDog/dd-agent/pull/1467~~
- [x] `postgres.py` - https://github.com/DataDog/dd-agent/pull/1415 (thanks @LeoCavaille)
- [x] ~~`process.py` - https://github.com/DataDog/dd-agent/pull/1470~~
- [x] `riak.py` - https://github.com/DataDog/dd-agent/pull/1499
- [x] ~~`system_core.py` - https://github.com/DataDog/dd-agent/pull/1485~~
- [ ] `vsphere.py` - it's a WON'T FIX for now
- [ ] `win32_event_log.py`

@remh :
- [x] `elastic.py` - https://github.com/DataDog/dd-agent/pull/1487
- [ ] `etcd.py`
- [x] ~~`gearmand.py`\- https://github.com/DataDog/dd-agent/pull/1476~~
- [ ] `gunicorn.py`
- [ ] `hdfs.py`
- [ ] `kyototycoon.py`
- [x] `mcache.py` - https://github.com/DataDog/dd-agent/pull/1490
- [ ] `ssh_check.py`
[...]

PR #1671, opened 2015-06-09
[core][windows] restore 'run_check' from 'checks'

yannmh:
PR #1571 introduced a bad regression: 'run_check' method was deleted
beside it is still used on Windows to debug.

### DIFF
--- a/tests/checks/common.py
+++ b/tests/checks/common.py
@@ -14,6 +14,7 @@ from pprint import pformat
 from checks import AgentCheck
 from config import get_checksd_path
 from util import get_os, get_hostname
+from utils.debug import get_check  # noqa -  FIXME 5.5.0 AgentCheck tests should not use this
 
 log = logging.getLogger('tests')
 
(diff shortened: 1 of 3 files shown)


utils/debug.py
```diff
@@ -0,0 +1,63 @@
...
+        check.check(instance)
+        if check.has_events():
+            print "Events:\n"
+            pprint(check.get_events(), indent=4)
+        print "Metrics:\n"
+        pprint(check.get_metrics(), indent=4)
+
+
+def get_check(name, config_str):
+    from checks import AgentCheck
```
degemer: Is there a reason why you import `AgentCheck` here instead of the top of the file ?

yannmh: to avoid cyclic dependencies, i.e.

1. `checks/__init__.py` imports `debug.py` to load
2. `debug.py` imports `checks/__init__.py` to load
3. `checks/__init__.py` imports `debug.py` ...

tests/checks/common.py
```diff
@@ -14,6 +14,7 @@
 from checks import AgentCheck
 from config import get_checksd_path
 from util import get_os, get_hostname
+from utils.debug import get_check  # noqa
```
degemer: Could you run a `sed` instead ? I think it's only used in old tests. Also, do you know the difference between `load_check` and `get_check` ? Because `load_check` seems to be used in `AgentCheckTest`.

utils/debug.py
degemer: Thanks for the explanation!

tests/checks/common.py
yannmh: This line will be gone with https://github.com/DataDog/dd-agent/issues/1458

PR #1673, opened 2015-06-09
[mysql] custom query metrics

garnermccloud:
Allow metrics to be generated from custom mysql queries.

Added logging, testing to #1274 and rebased.


checks.d/mysql.py
```diff
@@ -191,6 +193,15 @@ def _collect_metrics(self, host, db, tags, options):
                 "SHOW SLAVE STATUS", db, tags=tags
             )
 
+        # Collect custom query metrics
+        # Max of 20 queries allowed
+        if isinstance(queries, list):
+            for index, check in enumerate(queries):
+                if index == self.MAX_CUSTOM_QUERIES:
```
talwai: Instead of checking this condition everytime, maybe we can just trim the list to at most `self.MAX_CUSTOM_QUERIES` elems before iterating: `for index, check in enumerate(queries[:self.MAX_CUSTOM_QUERIES])` The warning can be logged just based on `len(queries)`

talwai:
Maybe there should be some notion of a timeout when executing custom queries. We're currently not putting any guards around how complex a query can be, and are at risk of having the agent block for indefinite periods if the query is long-running. I'd suggest we configure a `DEFAULT_TIMEOUT` and allow it to be overridden in `init_config` for the `mysql` check

talwai:
@garnermccloud any thoughts on the above suggestion?

garnermccloud:
Thanks for the comments @talwai - I'll update and rebase

yannmh:
Can you squash your commits @garnermccloud to clean up the git history ?

talwai:
Closing in favor of #1793

PR #1675, opened 2015-06-10
[haproxy] sc send proper hostname

degemer:
* Previously, the var `back_or_front` was not reset, so each loop could use the
  previous `back_or_front`.
* This `back_or_front` var is used to somehow determine the hostname for the
  service_check.
This problem is fixed.

Also:
* split mock and integration tests

Tests might fail.

### DIFF
--- a/checks.d/haproxy.py
+++ b/checks.d/haproxy.py
@@ -52,13 +52,13 @@ class HAProxy(AgentCheck):
         "eresp": ("rate", "errors.resp_rate"),
         "wretr": ("rate", "warnings.retr_rate"),
         "wredis": ("rate", "warnings.redis_rate"),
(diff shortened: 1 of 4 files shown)


degemer:
So, some remark about `back_or_front` (the core of the problem):
`back_or_front` is used even if the service if not a BACKEND or FRONTEND
it's visible in service_checks, but it also affects metrics nam.
This commit migth be relevant (in the history): https://github.com/DataDog/dd-agent/commit/8da8b1d18b77d1008c77df02c0bc5d87ce29767a

The conclusion: we're sending random metrics about services not BACKEND or FRONTEND
But are they useful ? What was done before ?

remh:
This is doing what is expected. Closing

PR #1676, opened 2015-06-10
[system_swap] add example conf file

talwai:
This example file was previously missing


yannmh:
Should this go with the 5.4 ?

hkaj:
@yannmh yes it should. Otherwise no one will know how to run the system_swap check.

talwai:
@yannmh ditto haissam - It was a mistake of mine to not include this in the original PR #1549

hkaj:
You can :ship: it, I'm testing the check in the meantime

PR #1679, opened 2015-06-10
[jmxfetch] move status & exit files to '/run'

yannmh:
Use '/run' for JMXFetch status & exit files instead of '/tmp/'. Move more functions to '/utils'.

### DIFF
--- a/utils/jmxfiles.py
+++ b/utils/jmxfiles.py
@@ -1,0 +1,109 @@
+# std
+import logging
+import os
+import tempfile
+import time
+
+# 3rd party
+import yaml
+
+# datadog
+from config import _windows_commondata_path
+from util import yDumper
+from utils.pidfile import PidFile
+from utils.platform import Platform
+
+
+log = logging.getLogger(__name__)
+
+
+class JMXFiles(object):
+    """
+    A small helper class for JMXFetch status & exit files.
+    """
+    _STATUS_FILE = 'jmx_status.yaml'
+    _PYTHON_STATUS_FILE = 'jmx_status_python.yaml'
+    _JMX_EXIT_FILE = 'jmxfetch_exit'
+
+    @classmethod
+    def _get_dir(cls):
+        if Platform.is_win32():
+            path = os.path.join(_windows_commondata_path(), 'Datadog')
+        elif os.path.isdir(PidFile.get_dir()):
+            path = PidFile.get_dir()
+        else:
+            path = tempfile.gettempdir()
+        return path
+
+    @classmethod
+    def _get_file_path(cls, file):
+        return os.path.join(cls._get_dir(), file)
+
+    @classmethod
+    def get_status_file_path(cls):
+        return cls._get_file_path(cls._STATUS_FILE)
+
+    @classmethod
+    def get_python_status_file_path(cls):
+        return cls._get_file_path(cls._PYTHON_STATUS_FILE)
+
+    @classmethod
+    def get_python_exit_file_path(cls):
+        return cls._get_file_path(cls._JMX_EXIT_FILE)
+
+    @classmethod
+    def write_status_file(cls, invalid_checks):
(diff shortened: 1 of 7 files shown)


win32/agent.py
```diff
@@ -298,14 +299,14 @@ def __init__(self, agentConfig, hostname):
 
     def run(self):
         if self.is_enabled:
-            JMXFetch.clean_exit_file()
+            JMXFiles.clean_exit_file()
             self.jmx_daemon.run()
 
     def terminate(self):
```
degemer: Is this method only called on Windows ?

degemer: :dizzy_face: OK, deleting this comment.

utils/jmxfiles.py
```diff
@@ -0,0 +1,106 @@
...
+        """
+        open(os.path.join(cls._get_dir(), cls._JMX_EXIT_FILE), 'a').close()
+
+    @classmethod
+    def clean_status_file(cls):
+        """
+        Removes JMX status files
+        """
+        try:
+            os.remove(os.path.join(cls._get_dir(), cls._STATUS_FILE))
```
degemer: Can you guarantee that if there is no `cls._STATUS_FILE` there won't be `cls._PYTHON_STATUS_FILE` ? Otherwise `cls._PYTHON_STATUS_FILE` may sometimes not be deleted.

degemer:
Just a small comment, it looks way more beauteous ! :tada:

utils/jmxfiles.py
```diff
@@ -0,0 +1,109 @@
...
+from config import _windows_commondata_path
+from util import yDumper
+from utils.pidfile import PidFile
+from utils.platform import Platform
+
+
+log = logging.getLogger(__name__)
+
+
+class JMXFiles(object):
```
elafarge: I don't get the point of writing a class with only class methods inside (at least not in Python :) ). Couldn't there just be static methods. Then if you import it this way : `from utils import jmxfiles` you could just call `jmxfiles.get_status_file_path()`. Wouldn't it make more sense.

utils/jmxfiles.py
```diff
@@ -0,0 +1,109 @@
...
+
+    @classmethod
+    def get_jmx_appnames(cls):
+        """
+        Retrieves the running JMX checks based on the {tmp}/jmx_status.yaml file
+        updated by JMXFetch (and the only communication channel between JMXFetch
+        and the collector since JMXFetch).
+        """
+        check_names = []
+        jmx_status_path = os.path.join(cls._get_dir(), cls._JMX_EXIT_FILE)
```
elafarge: Dear Yann,

Could you replace `cls._JMX_EXIT_FILE` with `cls._STATUS_FILE` please ? That'd be great :)

elafarge:
Just a slight mistake, as soon as fixed we should bee GTM :)

utils/jmxfiles.py
LeoCavaille: The only point that could be taken here, is that bundling the methods in a class is easier for imports and there is much less risk to override an existing method from another module.

elafarge: Yesah but if you use `from utils import jmxfiles` there can't be a conflict with any other module defining a method with the same name. Am I wrong ?

LeoCavaille: Indeed but you import everything class in that module (here I agree there is only one so it's ok).

elafarge: Under the hood, it works fine both ways anyway. I was under the influence of this video https://www.youtube.com/watch?v=o9pEzgHorH0 I watched a couple of weeks ago (same kind of ironic tone as the "Linux Sucks" conferences :) )

utils/jmxfiles.py
yannmh: Good catch, I fixed that

PR #1686, opened 2015-06-12
Add docker_image_tagless tag to docker check

veverjak:
Can be useful if you need to group images with same name but different tags


hkaj:
Thanks for contributing @veverjak!
Unfortunately we merged a similar feature a few days ago without checking in pending PRs if it was already there. You can find the tag under the name `image_repository`.

Thanks for taking the time to bring your improvement upstream though!

veverjak:
thanks for info, that is more elegant solution.

PR #1695, opened 2015-06-15
[packaging] do not fail if non-datadog repos are in a bad state

degemer:
Tested on CentOS 5 and Ubuntu 12 (needs to be run on dd-agent-testing).


degemer:
Thanks again @ip2k, it's now merged ! :tada:

ip2k:
Thanks much for doing it right! :)

PR #1700, opened 2015-06-16
[disk] disable all_partitions by default

degemer:
Otherwise a lot of useless/temp partitions appear.
See https://github.com/giampaolo/psutil/blob/58711c2db81d437fdcf5b8aedf68bfd00fc089fd/psutil/_pslinux.py#L651-L653


yannmh:
Looks good to me

PR #1710, opened 2015-06-17
[disk] backward compatibility with old Disk check

degemer:
The new disk check was ignoring by default tmpfs disks and some others,
breaking compatibility with the old check.

It is fixed by filtering on device name (exclusion on '' and 'none') and total
disk usage (!= 0), and setting all_partitions to True for psutil.
The new all_partitions option will send metrics from the non excluded disks
with 0 total disk usage.


remh:
LGTM!

PR #1720, opened 2015-06-22
[http_check] Improve SSL cert configuration & validation

olivielpeau:
Improves a few things:

- custom CA certificates are used in the `http.can_connect` check (in addition to the `http.ssl_cert` check)
- instance-level CA certificate files can be defined
- fixes flaky test

### DIFF
--- a/checks.d/http_check.py
+++ b/checks.d/http_check.py
@@ -167,7 +167,7 @@ class HTTPCheck(NetworkCheck):
                 ))
 
         if ssl_expire and urlparse(addr)[0] == "https":
-            status, msg = self.check_cert_expiration(instance)
+            status, msg = self.check_cert_expiration(instance, timeout)
             service_checks.append((
                 self.SC_SSL_CERT, status, msg
             ))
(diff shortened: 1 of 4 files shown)


checks.d/http_check.py
```diff
@@ -83,7 +84,7 @@ def _check(self, instance):
                 auth = (username, password)
 
             r = requests.get(addr, auth=auth, timeout=timeout, headers=headers,
-                             verify=not disable_ssl_validation)
+                             verify=False if disable_ssl_validation else instance_ca_certs)
```
remh: That seems weird, disable_ssl_validation is a boolean while instance_ca_certs is a string or a list of strings

olivielpeau: I agree that's a bit confusing but that's how the `requests` library works. From http://docs.python-requests.org/en/latest/user/advanced/#ssl-cert-verification:

>By default, verify is set to True. Requests can ignore verifying the SSL certificate if you set verify to False.
>You can pass verify the path to a CA_BUNDLE file with certificates of trusted CAs. This list of trusted CAs can also be specified through the REQUESTS_CA_BUNDLE environment variable.

`instance_ca_certs` should only be a single string though (not a list), ca_certs is plural because a single file usually contains multiple ca certificates. Do you think we should change the config parameter and variable names to make less confusing?

hkaj:
LGTM. The name `instance_ca_certs` is okay imo.

PR #1723, opened 2015-06-23
[elasticsearch] Add PShard statistic to ES check

elafarge:
Added statistics over primary shards only to our check. They're
basically retrieved under the `_stats` endpoint and are aggregated
metrics on the primary shards. It means that they don't take into
account metrics from replica shards. So for instance
`primaries.docs.count` will contain the total number of documents in the
cluster without replicas. Including replicas would be as simple as
summing the "standiard" document count metric over all nodes in the
cluster.


elafarge:
I didn't write tests for this check yet but it's planned :)

PR #1724, opened 2015-06-24
[elasticsearch] Add PShard statistic to ES check

elafarge:
Added statistics over primary shards only to our check. They're
basically retrieved under the `_stats` endpoint and are aggregated
metrics on the primary shards. It means that they don't take into
account metrics from replica shards. So for instance
`primaries.docs.count` will contain the total number of documents in the
cluster without replicas. Computing replica's would be as simple as
summing the "standard" document count metric over all nodes in the
cluster.

A test for those new metrics in particular has been added.

### DIFF
--- a/checks.d/elastic.py
+++ b/checks.d/elastic.py
@@ -18,6 +18,7 @@ class NodeNotFound(Exception):
 
 ESInstanceConfig = namedtuple(
     'ESInstanceConfig', [
+        'pshard_stats',
         'cluster_stats',
         'password',
         'service_check_tags',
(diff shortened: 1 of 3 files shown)


elafarge:
I just added some lines to delete all indices before the new test to make sure the document count will be the one I expect even though other tests add documents to our cluster (they don't at the moment but I thought they might do so in the future).

checks.d/elastic.py
```diff
@@ -304,7 +357,16 @@ def _define_params(self, version, cluster_stats):
 
         stats_metrics.update(additional_metrics)
 
-        return health_url, nodes_url, stats_url, pending_tasks_url, stats_metrics
+        # Version specific stats metrics about the primary shards
+        pshard_stats_metrics = dict(self.PRIMARY_SHARD_METRICS)
+
+        if version >= [1, 6, 0]:
+            additional_metrics = self.PRIMARY_SHARD_METRICS_POST_1_0
```
remh: is the minimum supported version for these metrics 1.6.0 (as per your if condition) or 1.0 (as per the constant name)

elafarge: No it should 1.0.0 actually, thanks for noticing it :)

elafarge:
I changed the version check to 1.0.0 instead of 1.6.0. The Travis CI build seems to be successful on elasticsearch flavors :)

elafarge:
Oh also I forgot but since @LeoCavaille deprecated the `is_external` option I changed it in the check's example configuration as well :)

PR #1726, opened 2015-06-25
[flare] Add JMXFetch-specific info

olivielpeau:
Adds in a `jmxinfo` directory:

- 2 JMXFetch status files: `jmx_status.yaml` and `jmx_status_python.yaml`
- JMXfetch commands output: `list_matching_attributes` and `list_everything`
- `java -version` output (using the configured java path) in a `java_version.log` file

@remh One thing that may not work properly in flare is the change I've made here https://github.com/DataDog/dd-agent/compare/olivielpeau/jmx-flare?expand=1#diff-64a477c26f008c29252aad6c1499c099R271 , which basically leaves file descriptors open (`close_fds=False`) when the JMXFetch subprocess is launched from the flare command. Do you know what issues can happen when the main process leaves the file descriptors open ?

cc @degemer @yannmh

### DIFF
--- a/agent.py
+++ b/agent.py
@@ -27,7 +27,6 @@ os.umask(022)
 from checks.check_status import CollectorStatus
 from checks.collector import Collector
 from config import (
-    get_confd_path,
     get_config,
     get_parsed_args,
     get_system_stats,
@@ -35,14 +34,13 @@ from config import (
 )
 from daemon import AgentSupervisor, Daemon
 from emitter import http_emitter
-from jmxfetch import JMX_LIST_COMMANDS, JMXFetch
 from util import (
     EC2,
     get_hostname,
-    get_os,
     Watchdog,
 )
 from utils.flare import configcheck, Flare
+from utils.jmx import jmx_command
 from utils.pidfile import PidFile
 from utils.profile import AgentProfiler
 
@@ -329,33 +327,7 @@ def main():
         configcheck()
 
     elif 'jmx' == command:
-        if len(args) < 2 or args[1] not in JMX_LIST_COMMANDS.keys():
-            print "#" * 80
-            print "JMX tool to be used to help configuring your JMX checks."
-            print "See http://docs.datadoghq.com/integrations/java/ for more information"
-            print "#" * 80
-            print "\n"
-            print "You have to specify one of the following commands:"
-            for command, desc in JMX_LIST_COMMANDS.iteritems():
-                print "      - %s [OPTIONAL: LIST OF CHECKS]: %s" % (command, desc)
-            print "Example: sudo /etc/init.d/datadog-agent jmx list_matching_attributes tomcat jmx solr"
-            print "\n"
-
-        else:
-            jmx_command = args[1]
-            checks_list = args[2:]
-            confd_directory = get_confd_path(get_os())
-
-            jmx_process = JMXFetch(confd_directory, agentConfig)
-            jmx_process.configure()
-            should_run = jmx_process.should_run()
-
-            if should_run:
-                jmx_process.run(jmx_command, checks_list, reporter="console")
-            else:
-                print "Couldn't find any valid JMX configuration in your conf.d directory: %s" % confd_directory
-                print "Have you enabled any JMX check ?"
-                print "If you think it's not normal please get in touch with Datadog Support"
+        jmx_command(args[1:], agentConfig)
 
     elif 'flare' == command:
         Flare.check_user_rights()
--- a/checks/check_status.py
+++ b/checks/check_status.py
@@ -21,7 +21,7 @@ import yaml
 import config
 from config import _is_affirmative, _windows_commondata_path, get_config
 from util import plural
-from utils.jmxfiles import JMXFiles
+from utils.jmx import JMXFiles
 from utils.ntp import get_ntp_args
 from utils.pidfile import PidFile
 from utils.platform import Platform
--- a/checks/check_status.py
+++ b/checks/check_status.py
@@ -88,7 +88,10 @@ def logger_info():
     if len(root_logger.handlers) > 0:
         for handler in root_logger.handlers:
             if isinstance(handler, logging.StreamHandler):
-                loggers.append(handler.stream.name)
+                try:
+                    loggers.append(handler.stream.name)
+                except AttributeError:
+                    loggers.append("unnamed stream")
             if isinstance(handler, logging.handlers.SysLogHandler):
                 if isinstance(handler.address, basestring):
                     loggers.append('syslog:%s' % handler.address)
--- a/jmxfetch.py
+++ b/jmxfetch.py
@@ -98,14 +98,14 @@ class JMXFetch(object):
         except ValueError:
             log.exception("Unable to register signal handlers.")
 
-    def configure(self, check_list=None):
+    def configure(self, checks_list=None):
         """
         Instantiate JMXFetch parameters, clean potential previous run leftovers.
         """
         JMXFiles.clean_status_file()
 
         self.jmx_checks, self.invalid_checks, self.java_bin_path, self.java_options, self.tools_jar_path = \
-            self.get_configuration(check_list)
+            self.get_configuration(self.confd_path, checks_list=checks_list)
 
     def should_run(self):
         """
@@ -113,7 +113,7 @@ class JMXFetch(object):
         """
         return self.jmx_checks is not None and self.jmx_checks != []
 
-    def run(self, command=None, check_list=None, reporter=None, redirect_std_streams=False):
+    def run(self, command=None, checks_list=None, reporter=None, redirect_std_streams=False):
         """
         Run JMXFetch
 
@@ -123,10 +123,10 @@ class JMXFetch(object):
         and sys.stderr.
         """
 
-        if check_list or self.jmx_checks is None:
-            # (Re)set/(re)configure JMXFetch parameters when `check_list` is specified or
+        if checks_list or self.jmx_checks is None:
+            # (Re)set/(re)configure JMXFetch parameters when `checks_list` is specified or
             # no configuration was found
-            self.configure(check_list)
+            self.configure(checks_list)
 
         try:
             command = command or JMX_COLLECT_COMMAND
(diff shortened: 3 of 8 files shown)


utils/flare.py
```diff
@@ -257,16 +295,26 @@ def _strip_password(self, file_path):
...
+    def _add_command_output_tar(self, name, command, logger_name=None):
+        temp_path = os.path.join(tempfile.gettempdir(), name.replace(os.sep, '__'))
         if os.path.exists(temp_path):
             os.remove(temp_path)
         backup_out, backup_err = sys.stdout, sys.stderr
         backup_handlers = logging.root.handlers[:]
         out, err = StringIO.StringIO(), StringIO.StringIO()
+        if logger_name:
+            # If specified, redirect logger output to `out`
+            logger = logging.getLogger(logger_name)
```
degemer: Since you send the output of jmxfetch to `stdout` (https://github.com/DataDog/dd-agent/pull/1726/files#diff-64a477c26f008c29252aad6c1499c099R283), do you still need this ? I guess it's for `jmxfetch.py`, what does it log ? (I have no idea, just asking)

olivielpeau: It's for `jmxfetch.py` indeed, it logs exceptions that occur, problems with configuration files, and  the exact command that is used to launch JMXFetch (which includes the java bin path and JVM options)

utils/flare.py
```diff
@@ -257,16 +295,26 @@ def _strip_password(self, file_path):
         return temp_path, password_found
 
     # Add output of the command to the tarfile
-    def _add_command_output_tar(self, name, command):
-        temp_path = os.path.join(tempfile.gettempdir(), name)
+    def _add_command_output_tar(self, name, command, logger_name=None):
+        temp_path = os.path.join(tempfile.gettempdir(), name.replace(os.sep, '__'))
```
degemer: I'm not a huge fan of this (especially just after a `os.path.join`, it's kind of redondant), maybe instead of using the `tempfile.gettempdir()` you could call https://docs.python.org/2/library/tempfile.html#tempfile.mkstemp ? Or add a `subdir=` option. Or even easier a `temp_name=` option.

jmxfetch.py
```diff
@@ -255,14 +264,27 @@ def _start(self, path_to_java, java_run_opts, jmx_checks, command, reporter, too
...
             self.jmx_process = jmx_process
 
             # Register SIGINT and SIGTERM signal handlers
             self.register_signal_handlers()
 
-            # Wait for JMXFetch to return
-            jmx_process.wait()
+            if redirect_std_streams:
+                # Wait for JMXFetch to return, and write out the stdout and stderr of JMXFetch to sys.stdout and sys.stderr
+                out, err = jmx_process.communicate()
```
degemer: It's not possible to redirect directly `Popen` to `stdout` ?

degemer: My bad, it's way easier with Python 2.7, but for `Popen` (Python 2.6...), it's not possible.

degemer:
Really nice, thanks !
Two remarks on the output of the `flare` command:
- jmx commands output is always collected even if no integration is setup (files are ~empty, but you still display _collecting java version_), I think it might mislead users, could you not collect it if no integration using `jmxfetch` is setup ? (`JMXFetch.should_run()` ?)
- when `LOG_LEVEL: DEBUG` (support recommends to do it before sending logs, so it's quite often), the ouptut is really verbose, maybe too much. Do you know if you could remove these debug statements (users won't have to debug `flare`) ?
screen shot 2015-06-29 at 12 20 51

agent.py
```diff
@@ -35,14 +34,13 @@
...
 from emitter import http_emitter
-from jmxfetch import JMXFetch, JMX_LIST_COMMANDS
 from util import (
     EC2,
     get_hostname,
-    get_os,
     Watchdog,
 )
 from utils.flare import configcheck, Flare
+from utils.jmx_command import jmx_command
```
yannmh: :+1: for the cleanup !

Maybe we should merge `utils.jmx_command` and `utils.jmxfiles` to have only a single JMX `util` package. What do you think ?

yannmh:
Another great feature, that will make support way easier. Thanks @olivielpeau.

Can you attach an example of a flare archive to have an idea of what it looks like ?

agent.py
olivielpeau: I agree, that makes a lot of sense, I'll regroup them in one `utils.jmx` module ;)

olivielpeau: Hmm actually I get circular imports when the two are merged together, as `jmxfetch` imports `utils.jmx` (for the `JMXFiles` class) that itself imports `jmxfetch` (for the `JMXFetch` class used by `jmx_command`).

So unless there's a better solution I'll leave the two modules separated for now...

olivielpeau:
Thanks a lot for your reviews @yannmh and @degemer!

I've added two commits that:

- capture the whole output of the commands run by flare (so that users only see the actual output from flare)
- only add jmx info if a valid jmx config file is present (as evaluated by `JMXFetch.should_run()`)

And here is an example flare archive: https://www.dropbox.com/s/orjuu2kbn9d1aol/datadog-agent-2015-06-30-15-03-16.tar.bz2?dl=0

agent.py
yannmh: Importing `JMXFetch` in the scope of the method only, would prevent this:
```python
def jmx_command()
  from jmxfetch import JMX_LIST_COMMANDS, JMXFetch
  ...
```

utils/flare.py
```diff
@@ -206,15 +211,55 @@ def _add_conf_tar(self):
...
+                    partial(self._jmx_command_call, command)
+                )
+
+            # java version
+            log.info("  * java -version output")
+            self._add_command_output_tar(
+                os.path.join('jmxinfo', 'java_version.log'),
+                # We use lambda so that JMXFetch.get_configuration is evaluated in _add_command_output_tar,
+                # which captures the logging output from JMXFetch
+                lambda: self._java_version(JMXFetch.get_configuration(get_confd_path())[2])
```
yannmh: Nitpicky: we are also adding the output of `JMXFetch.get_configuration(get_confd_path())` to `java_version.log`. Is it intended ?

```
>>>> STDOUT <<<<
tomcat check does not have a valid JMX configuration: You need to have at least one instance defined in the YAML file for this check
Popen(['java', '-version'], stderr = -2, stdout = -1) called
java version "1.7.0_51"
Java(TM) SE Runtime Environment (build 1.7.0_51-b13)
Java HotSpot(TM) 64-Bit Server VM (build 24.51-b03, mixed mode)

>>>> STDERR <<<<
```
Maybe we could drop it and add a `>>>> CMD <<<<` section to `_add_command_output_tar ` with the actual command printed (in this case it's interesting as it may contains a custom java path).

checks/check_status.py
```diff
@@ -88,7 +88,10 @@ def logger_info():
     if len(root_logger.handlers) > 0:
         for handler in root_logger.handlers:
             if isinstance(handler, logging.StreamHandler):
-                loggers.append(handler.stream.name)
+                try:
+                    loggers.append(handler.stream.name)
+                except AttributeError:
+                    loggers.append("unnamed stream")
             if isinstance(handler, logging.handlers.SysLogHandler):
```
yannmh: I don't understand why we need to do this :confounded:

yannmh:
Looks great @olivielpeau. Let's merge it whenever you want.

checks/check_status.py
olivielpeau: I documented it in the message of https://github.com/DataDog/dd-agent/commit/378fdaef8c96c05a18a110b29ec5744d5fa309c2:

```
In `check_status.py`, stop assuming that all streams have a `name` attribute
(for instance `StringIO.StringIO` instances do not have one).
```

Do you think it deserves a comment directly in the source code?

olivielpeau:
Thanks @yannmh for your very good points ;)

I've merged `jmx_command` and `JMXFiles` in a single `utils.jmx` module, fixed an issue that could make `flare` fail silently (see https://github.com/DataDog/dd-agent/commit/5c75b06e4e543f24f64a39919204efac0d6a0bea).

Now commands can also have an optional description. `jmxinfo/java_version.log` now looks like (note that the `Popen` line is only there when the log_level is `DEBUG`):
```
>>>> CMD <<<<
java -version
>>>> STDOUT <<<<
Popen(['java', '-version'], stderr = -2, stdout = -1) called
java version "1.7.0_51"
Java(TM) SE Runtime Environment (build 1.7.0_51-b13)
Java HotSpot(TM) 64-Bit Server VM (build 24.51-b03, mixed mode)

>>>> STDERR <<<<
```

utils/jmx.py
```diff
@@ -8,15 +8,48 @@
 import yaml
 
 # datadog
-from config import _windows_commondata_path
-from util import yDumper
+from config import _windows_commondata_path, get_confd_path
+from util import get_os, yDumper
```
degemer: Can you remove this import ? (`get_os`) If `get_confd_path` doesn't have a `get_os()`, it will call it itself, so it's no necessary.

utils/flare.py
```diff
@@ -279,6 +321,23 @@ def _add_command_output_tar(self, name, command):
...
+    def _capture_output(self, command):
+        backup_out, backup_err = sys.stdout, sys.stderr
+        out, err = StringIO.StringIO(), StringIO.StringIO()
+        backup_handlers = logging.root.handlers[:]
+        logging.root.handlers = [logging.StreamHandler(out)]
+        sys.stdout, sys.stderr = out, err
+        try:
+            return_value = command()
+        finally:
+            # Stop capturing in a `finally` block to let uncaught exceptions be printed out
```
degemer: Exceptions are printed in the user `stdout` or the `StringIO` ?

olivielpeau: In `stdout`,

After discussing it we should probably catch all exceptions and print them in the file that is added to the tar, I'll change that.

olivielpeau:
Thanks again @degemer! I've changed the `get_os` import and the handling of exceptions:

- exceptions from commands which output is printed to a file in the tar are printed in the file
- exceptions from commands which output is not printed at all are printed in `stdout`
- all `Exception`s raised in those commands are handled by flare

degemer:
_[flare] Fix typo_ is not going to work as a commit short desc.
Otherwise, really nice additions, when it's fixed, :shipit: !

olivielpeau:
Haha, I guess I got lazy on this one, I've fixed it, thanks!

Will merge as soon as the ci passes.

PR #1728, opened 2015-06-25
[packaging][osx] add install script

degemer:
Add the install script before the changes, #1361 will need a rebase.


degemer:
@elafarge I'll merge it as it is (it's working, already in use), if you have additional comments, I'll open a PR to address them.

elafarge:
I tested it on Monday evening, I just forgot to add comments here. My only remarque was about using `curl -o my-file xxx` instead of `curl xxx > my-file` but I guess it's not a big deal at all :) Good job !

PR #1733, opened 2015-06-26
[ci] fix two (often) false-negative tests

degemer:
- `test_min_collection_interval` in `core/test_common.py`: use `disk` check
  instead of `ntp` (no network call)
- `apache`: increase timeout before start


yannmh:
LGTM !

Issue #1439, opened 2015-03-16
[win32_event_log] Security events don't have the same "types" - specifying a type breaks the instance


samanthadrago:
A user had an issue setting up his windows event log integration to monitor security events. it looks like the issue is because he was trying to specify a "type", but the security events don't have the same types (warning, info, error) as other events (they have success or failure audits: http://support.microsoft.com/en-us/kb/308427). When he included the "type" parameter, the security events didn't get sent in to DD.

He suggested we update the win32_event_log.yaml to make mention of this.

Here is the case: https://datadog.desk.com/agent/case/24604

PR #1734, opened 2015-06-26
[event_viewer] Improve comments in example configuration

olivielpeau:
Linked to https://github.com/DataDog/documentation/pull/442 and https://github.com/DataDog/dogweb/pull/8622

Closes https://github.com/DataDog/dd-agent/issues/1439

### DIFF
--- a/conf.d/win32_event_log.yaml.example
+++ b/conf.d/win32_event_log.yaml.example
@@ -2,30 +2,63 @@ init_config:
 
 instances:
   # Each Event Log instance lets you define the type of events you want to
-  # match and how to tag those events. You can use the following filters:
-  #
-  # - log_file: Application, System, Setup, Security
-  #
-  # - source_name: Any available source name
-  #
-  # - type: Warning, Error, Information
-  #
-  # - user: Any valid user name
-  #
-  # - event_id: The Event ID can be found through http://www.eventid.net/ and
-  #   viewed in the window event viewer.
-  #
-  # - message_filters: A list of message filters, using % as a wildcard.
-  #   See http://msdn.microsoft.com/en-us/library/aa392263(v=vs.85).aspx for
-  #   more on the format for LIKE queries.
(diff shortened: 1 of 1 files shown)


degemer:
Since you're updating the example, could you also add these options https://github.com/DataDog/dd-agent/blob/master/checks.d/win32_event_log.py#L30-L32 (there is already `user`, but it is in the middle, not its right place). It shouldn't be used often, but it's still weird that it's not there.

One other thing: could you update the format to have something similar to  https://github.com/DataDog/dd-agent/blob/master/conf.d/disk.yaml.default, ie the description of the option and under an example ? (although we should keep both basic example)

olivielpeau:
Thanks @degemer for your review! I hadn't realized how outdated the config file was so I've almost completely rewritten it, could you have a look now?

degemer:
Way way way better ! :clap: 
Some last remarks:
- usually we write the default file so that users just have to uncomment lines and set the correct parameter: do you think you can do it ? (related to next remark)
- from the example yaml, all parameters are optional: does it mean that if I write
```
instances:
  - random: iamramdom
```
it will collect all events ? I don't think we want this anyway, do you know if there is a parameter always used ?
- I think we either dropped already `notify` or are deprecating it (it's not in the API docs), so could you remove it from the example ?

olivielpeau:
All the parameters are optional indeed, but we can probably set a default value on the log file that's captured (`System` could be a good candidate).

olivielpeau:
I've removed the `notify` section, but after some further digging I haven't been able to find a sane default on any of the parameters that would make the check capture only specific interesting events.

To avoid sending thousands of events in case an instance's filters aren't specific enough, we could:

- In the config file, set a fully-defined default instance that would capture `Critical` and `Error` on the `System` LogFile
- Limit the number of events that the integration captures.

I'm open to suggestions if you think there's a better solution :)

olivielpeau:
Discussed with @degemer, we'll leave an uncommented dash in the example yaml as it doesn't define a valid instance by itself (so the config file doesn't collect all the events as-is).

Merging

PR #1737, opened 2015-06-29
[GCE] Update hostname to be unique

### LINKED ISSUE #1736: [GCE] Non unique hostname
The agent currently uses the instance name as the hostname of GCE instances.
If one has several projects with instances having the same name, this hostname isn't unique, thus the tags and metrics not matching the expected values.
A unique hostname would be either what's called `internal-hostname` or `name-project`

tmichelet:
fixes #1736 
- update the hostname to `host_metadata['instance']['hostname']` so it's unique
- drop some tags that we don't want to collect
- add `"instance_name.project_id"` as a host alias

### DIFF
--- a/datadog.conf.example
+++ b/datadog.conf.example
@@ -43,6 +43,9 @@ api_key:
 # for more information
 # collect_instance_metadata: yes
 
+# use unique hostname for GCE hosts. see #1736
+gce_updated_hostname: yes
+
 # Set the threshold for accepting points to allow anything
 # with recent_point_threshold seconds
 # Defaults to 30 seconds if no value is provided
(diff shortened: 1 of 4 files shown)


datadog.conf.example
```diff
@@ -43,6 +43,9 @@ api_key:
 # for more information
 # collect_instance_metadata: yes
 
+# use unique hostname for GCE hosts. see #1736
```
tmichelet: let's put PR link

remh: Ping me when this is done

tmichelet: first pass is done, see https://github.com/DataDog/dd-agent/wiki/Hostname-on-Google-Cloud-Engine

tmichelet:
@remh link updated, good to merge?

PR #1740, opened 2015-06-30
Add default host tags in agent config for dogstatsd

ermandoser:


LeoCavaille:
Hi @ermandoser, thanks for contributing. I am actually wondering what's the usecase hiding behind this change, because this should be already supported by the backend.
If you set the `tags` attribute in the general `datadog.conf` file, these tags are stored for the host in our backend and later on if you report a metric tagged with this host (which comes by default in dogstatsd) the tags will be expanded in the backend.

ermandoser:
I've been talking to your support.
It seems like it's supported on the backend as you said, and we overriding
the host tag caused mismatching.

We'll investigate on our side, so you can close the pull request as it's
not needed.

On Tuesday, June 30, 2015, Léo Cavaillé <notifications@github.com> wrote:

> Hi @ermandoser <https://github.com/ermandoser>, thanks for contributing.
> I am actually wondering what's the usecase hiding behind this change,
> because this should be already supported by the backend.
> If you set the tags attribute in the general datadog.conf file, these
> tags are stored for the host in our backend and later on if you report a
> metric tagged with this host (which comes by default in dogstatsd) the tags
> will be expanded in the backend.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/DataDog/dd-agent/pull/1740#issuecomment-117243701>.
>


-- 
Erman Doser.

LeoCavaille:
Ok cool! Don't hesitate to reopen a separate issue if this mismatch cannot be solved on your side.

PR #1743, opened 2015-06-30
Add some documentation about running individual unittests

sebest:


remh:
Thanks a lot @sebest !

PR #1744, opened 2015-06-30
[docker] add some logging around API response parsing

talwai:


checks.d/docker.py
```diff
@@ -384,15 +384,19 @@ def _get_json(self, uri, params=None, multi=False):
...
 
         if not response:
             return []
 
-        return json.loads(response)
+        try:
+            return json.loads(response)
+        except Exception as e:
+            self.log.error('Failed to parse Docker API response: %s', response)
+            raise
```
remh: Not sure we should raise here, it will interrupt the whole execution of the check although only the "events" feature of the check would be broken.

PR #1745, opened 2015-06-30
[etcd] Add full SSL support

olivielpeau:
Supersedes #1654.

Based on @KnownSubset's work, with a few additional changes:
- remove the `ssl` parameter (using SSL or not depends on the url scheme)
- split `ssl_cert_reqs` in 2 parameters
- reformat the example config file

### DIFF
--- a/checks.d/etcd.py
+++ b/checks.d/etcd.py
@@ -3,6 +3,7 @@ import requests
 
 # project
 from checks import AgentCheck
+from config import _is_affirmative
 from util import headers
 
 
@@ -71,9 +72,10 @@ class Etcd(AgentCheck):
 
         # Load the ssl configuration
         ssl_params = {
-            'ssl_keyfile': instance.get('ssl_keyfile', None),
-            'ssl_certfile': instance.get('ssl_certfile', None),
-            'ssl_cert_reqs':  instance.get('ssl_cert_reqs', True)
+            'ssl_keyfile': instance.get('ssl_keyfile'),
+            'ssl_certfile': instance.get('ssl_certfile'),
+            'ssl_cert_validation': _is_affirmative(instance.get('ssl_cert_validation', True)),
+            'ssl_ca_certs': instance.get('ssl_ca_certs'),
         }
 
         for key, param in ssl_params.items():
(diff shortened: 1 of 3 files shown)


checks.d/etcd.py
```diff
@@ -68,6 +69,18 @@ def check(self, instance):
         # Load values from the instance config
         url = instance['url']
         instance_tags = instance.get('tags', [])
+
+        # Load the ssl configuration
+        ssl_params = {
+            'ssl_keyfile': instance.get('ssl_keyfile'),
+            'ssl_certfile': instance.get('ssl_certfile'),
+            'ssl_cert_validation': _is_affirmative(instance.get('ssl_cert_validation', True)),
```
degemer: By default ssl_cert_validation is enabled: it could be a breaking change for people using the current check with an invalid cert.

checks.d/etcd.py
```diff
@@ -133,18 +146,20 @@ def check(self, instance):
...
-    def _get_leader_metrics(self, url, timeout):
-        return self._get_json(url + "/v2/stats/leader", timeout)
+    def _get_leader_metrics(self, url, ssl_params, timeout):
+        return self._get_json(url + "/v2/stats/leader", ssl_params, timeout)
 
-    def _get_json(self, url, timeout):
+    def _get_json(self, url, ssl_params, timeout):
         try:
-            r = requests.get(url, timeout=timeout, headers=headers(self.agentConfig))
+            certificate = (ssl_params['ssl_certfile'], ssl_params['ssl_keyfile']) if 'ssl_certfile' in ssl_params and 'ssl_keyfile' in ssl_params else None
```
remh: nitpick but you should use ternary conditions only when it makes the code easier to read.

In that case as it makes a pretty long line i would simply go with:
```python
certificate = None
if  'ssl_certfile' in ssl_params and 'ssl_keyfile' in ssl_params:
    certificate = (ssl_params['ssl_certfile'], ssl_params['ssl_keyfile'])
```

checks.d/etcd.py
olivielpeau: Actually the default behaviour of the `requests` library is to validate certs. See its [docs] (http://docs.python-requests.org/en/latest/user/advanced/#ssl-cert-verification):

> By default, verify is set to True.

So the check already validates certs, BC is preserved here.

degemer: :trophy:

olivielpeau:
Thanks @degemer and @remh for your reviews, I've changed the ternary condition to make it more readable.

I'm going to merge once the ci passes.

PR #1746, opened 2015-07-01
[core] use real path for PidFile

degemer:
It was previously a relative path, it's clearer this way.


yannmh:
Looks good to me !

PR #1750, opened 2015-07-06
[ec2] Add trailing slash to metadata lists endpoints

olivielpeau:
Thanks to @dspangen for the initial fix.

I've also added the trailing slash to the `public-keys` endpoint as it might break in the same way in the future.

### DIFF
--- a/util.py
+++ b/util.py
@@ -332,8 +332,8 @@ class EC2(object):
             pass
 
         try:
-            iam_role = urllib2.urlopen(EC2.METADATA_URL_BASE + "/iam/security-credentials").read().strip()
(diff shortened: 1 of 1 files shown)


dspangen:
Is this going to merge soon? This is preventing datadog agent from restarting on any of our ec2 instances.

olivielpeau:
@dspangen This fix will be merged soon and shipped with v5.5.0 of the agent. The issue shouldn't prevent the agent from restarting though (currently the agent should start fine but with these metadata tags missing).

If you have any trouble with the agent not restarting properly feel free to contact support.

PR #1755, opened 2015-07-08
[redis] handle exception when CONFIG disabled

degemer:
On AWS Elasticache, Redis config command is disabled by default. We only use it
to get the max_slow_entries (and we have a default), so now it tries to get it
from the config and catch potential errors, defaulting to
DEFAULT_MAX_SLOW_ENTRIES.


remh:
:+1:

PR #1756, opened 2015-07-09
[mysql] rename max connections for clarity

hutchic:
calling Max_used_connections just max_connections leads to confusion as it sounds like it would be the my.cnf max_connections variable (IMO)

### DIFF
--- a/checks.d/mysql.py
+++ b/checks.d/mysql.py
@@ -17,7 +17,7 @@ RATE = "rate"
 
 STATUS_VARS = {
     'Connections': ('mysql.net.connections', RATE),
-    'Max_used_connections': ('mysql.net.max_connections', GAUGE),
+    'Max_used_connections': ('mysql.net.max_used_connections', GAUGE),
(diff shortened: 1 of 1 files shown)


yannmh:
Thanks a lot for your contribution @hutchic, and sorry for getting back to you so late.

I had a look into your changes, and while I agree that the current metric name is quite confusing, it's difficult to back-pedal: renaming a metric introduces backward incompatibilities like drops in graphs or dashboard/monitor metric definitions getting obsolete ...

We'll keep the ticket pending, so we can assess a potential solution for this type of issues. Thanks again.

olivielpeau:
Moving this to the 6.0 milestone as it's a BC breaking change that would be difficult to include in a minor version.

hutchic:
closed?

masci:
@hutchic I owe you an explanation here, sorry.
The next major version of the agent will be a complete rewrite, we're going to open source the new repo soon and we'll take into consideration things from milestone `6.0.0` there. Feel free to reach out if you need any other detail.

PR #1757, opened 2015-07-09
[docker] catch error when parsing json response

talwai:
When Docker Remote API `/events` returns invalid JSON, just default to an empty list


hkaj:
go for it

Issue #1748, opened 2015-07-02
[docker] cgroups not found on Kubernetes


degemer:
We search for containers files in /sys/fs/cgroup/{memory,cpuacct}/{lxc,docker,system.slide}/{docker_id}. Otherwise it seems that Kubernetes (at least for this case https://datadog.zendesk.com/agent/tickets/28574) uses /sys/fs/cgroup/{memory,cpuacct}/{docker_id}.

A quick fix: https://gist.github.com/degemer/54ddc50e62cade7f3162#file-docker-py-L418-L420

What do you think @hkaj ?

remh:
Don't think it's related to kubernetes but more likely to the distribution used.

degemer:
According to the info, it's happening on `Linux-3.16.0-0.bpo.4-amd64-x86_64-with-debian-8.1`.

hkaj:
It seems to be a kubernetes thing. I run debian with the same kernel and don't have the issue.
Your fix looks good @degemer, wanna open a PR? :)

PR #1759, opened 2015-07-09
[docker] Kubernetes support (new cgroups path)

degemer:
On Kubernetes, container files are directly at the top level of cgroups,
instead of being in a directory (`docker` or `lxc`).

Fixes https://github.com/DataDog/dd-agent/issues/1748.

### DIFF
--- a/checks.d/docker.py
+++ b/checks.d/docker.py
@@ -408,13 +408,14 @@ class Docker(AgentCheck):
 
     # Cgroups
 
-    def _find_cgroup_filename_pattern(self):
+    def _find_cgroup_filename_pattern(self, container_id):
         if self._mountpoints:
             # We try with different cgroups so that it works even if only one is properly working
             for mountpoint in self._mountpoints.values():
                 stat_file_path_lxc = os.path.join(mountpoint, "lxc")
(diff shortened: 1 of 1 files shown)


checks.d/docker.py
```diff
@@ -422,6 +422,9 @@ def _find_cgroup_filename_pattern(self):
                     return os.path.join('%(mountpoint)s/docker/%(id)s/%(file)s')
                 elif os.path.exists(stat_file_path_coreos):
                     return os.path.join('%(mountpoint)s/system.slice/docker-%(id)s.scope/%(file)s')
+                # Kubernetes, we can't test the existence of the path without the id
+                else:
```
hkaj: Not sure we want an else here since it acts as a `default`. An `elif` could be interesting too don't you think? This way if the path is not found because of a problem, we'll get an exception here instead of later in the execution.

degemer: Indeed, it was a quick fix. :disappointed:  
I updated the PR, passing a `container_id` to `_find_cgroup_filename_pattern` to be able to test the path.

hkaj:
lgtm

degemer:
Merging then.

Issue #1749, opened 2015-07-03
[couch] support _stats server-only check


ferrouswheel:
In restrictive environments it may not be possible to let datadog agent access database statistics.

CouchDB's server `/_stats/` at the root is public, so it'd be nice if it was possible to retrieve these stats without returning an error when the databases themselves would return "401 Unauthorized".

There is `db_whitelist`, but that requires at least one db that can be accessed. Just specifying the key with no value isn't recognised by the integration check.

I have worked around this by creating an empty database that can be accessed, and adding that to the whitelist, but it's sub-optimal.

remh:
Thanks @ferrouswheel , it's definitely a legitimate use case. We'll work on it.
Thanks for the feedback!

irabinovitch:
@remh Added tests for whitelist and blacklist functionality.

yannmh:
Fixed by #1760. Thanks @irabinovitch !

PR #1760, opened 2015-07-09
[couchdb] Improve error handling and add Blacklist support.

irabinovitch:
- fail gracefully when one or more individual dbs are not readable by the configured user.
- allow blacklisting of specific databases via db_blacklist in couch.yaml
- append to blacklist at run time if unable to query a given db's stats

### DIFF
--- a/checks.d/couch.py
+++ b/checks.d/couch.py
@@ -19,6 +19,10 @@ class CouchDb(AgentCheck):
     SOURCE_TYPE_NAME = 'couchdb'
     TIMEOUT = 5
 
+    def __init__(self, name, init_config, agentConfig, instances=None):
+        AgentCheck.__init__(self, name, init_config, agentConfig, instances)
+        self.db_blacklist = {}
+
     def _create_metric(self, data, tags=None):
         overall_stats = data.get('stats', {})
         for key, stats in overall_stats.items():
@@ -42,8 +46,10 @@ class CouchDb(AgentCheck):
         auth = None
         if 'user' in instance and 'password' in instance:
             auth = (instance['user'], instance['password'])
-
-        r = requests.get(url, auth=auth, headers=headers(self.agentConfig),
+        # Override Accept request header so that failures are not redirected to the Futon web-ui
+        request_headers = headers(self.agentConfig)
+        request_headers['Accept'] = 'text/json'
+        r = requests.get(url, auth=auth, headers=request_headers,
                          timeout=int(instance.get('timeout', self.TIMEOUT)))
         r.raise_for_status()
         return r.json()
@@ -98,8 +104,10 @@ class CouchDb(AgentCheck):
 
         # Get the list of whitelisted databases.
         db_whitelist = instance.get('db_whitelist')
+        self.db_blacklist.setdefault(server,[])
+        self.db_blacklist[server].extend(instance.get('db_blacklist',[]))
         whitelist = set(db_whitelist) if db_whitelist else None
-        databases = set(self._get_stats(url, instance))
+        databases = set(self._get_stats(url, instance)) - set(self.db_blacklist[server])
         databases = databases.intersection(whitelist) if whitelist else databases
 
         if len(databases) > self.MAX_DB:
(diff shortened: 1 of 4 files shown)


checks.d/couch.py
```diff
@@ -18,6 +18,7 @@ class CouchDb(AgentCheck):
     SERVICE_CHECK_NAME = 'couchdb.can_connect'
     SOURCE_TYPE_NAME = 'couchdb'
     TIMEOUT = 5
+    db_blacklist = []
```
remh: So, this won't play well in the (unlikely) case where you are monitoring multiple instances of couchdb on a single server, and if they have some database names in common.

If you want to blacklist one of these db in one of this instance, you'll end up blacklisting it for all instances.

So, the best (but kind of ugly) solution is to use a dictionary instead of a list whose keys are unique keys of an instance (e.g. the "server" parameter of the configuration).

Let me know if you have more questions

checks.d/couch.py
```diff
@@ -42,8 +43,10 @@ def _get_stats(self, url, instance):
         auth = None
         if 'user' in instance and 'password' in instance:
             auth = (instance['user'], instance['password'])
-
-        r = requests.get(url, auth=auth, headers=headers(self.agentConfig),
+        # Override Accept request header so that failures are not redirected to the Futon web-ui
+        request_headers=headers(self.agentConfig)
```
remh: Nitpick but as per PEP8, there should be a space in both side of the "=" sign.

checks.d/couch.py
```diff
@@ -98,19 +101,23 @@ def get_data(self, server, instance):
...
         db_whitelist = instance.get('db_whitelist')
+        self.db_blacklist.extend(instance.get('db_blacklist',[]))
         whitelist = set(db_whitelist) if db_whitelist else None
-        databases = set(self._get_stats(url, instance))
+        databases = set(self._get_stats(url, instance)) - set(self.db_blacklist)
         databases = databases.intersection(whitelist) if whitelist else databases
 
         if len(databases) > self.MAX_DB:
-            self.warning('Too many databases, only the first %s will be checked.' % self.MAX_DB)
+            self.log.warning('Too many databases, only the first %s will be checked.' % self.MAX_DB)
```
remh: Nope we wan't to keep using self.warning here. It will display the warning in the info page

checks.d/couch.py
```diff
@@ -98,19 +101,23 @@ def get_data(self, server, instance):
...
             databases = list(databases)[:self.MAX_DB]
 
         for dbName in databases:
             url = urljoin(server, dbName)
-
-            db_stats = self._get_stats(url, instance)
+            try:
+                db_stats = self._get_stats(url, instance)
+            except requests.exceptions.HTTPError as e:
```
remh: Is it possible to only catch an "Unauthorized" error ? I guess it returns a 403 if you are not authorized ? This way we won't blacklist databases in case of a transient issue.

checks.d/couch.py
```diff
@@ -98,19 +101,23 @@ def get_data(self, server, instance):
...
+                couchdb['databases'][dbName] = None
+                self.db_blacklist.append(dbName)
+                self.log.warning('Database %s is not readable by the configured user. It will be added to the blacklist. Please restart then agent to clear.' % dbName)
```
remh: We should likely use self.warning here as well.

checks.d/couch.py
```diff
@@ -98,19 +101,23 @@ def get_data(self, server, instance):
...
             databases = list(databases)[:self.MAX_DB]
 
         for dbName in databases:
             url = urljoin(server, dbName)
-
-            db_stats = self._get_stats(url, instance)
+            try:
+                db_stats = self._get_stats(url, instance)
+            except requests.exceptions.HTTPError as e:
+                couchdb['databases'][dbName] = None
```
remh: Maybe we should do:
```python
if dbName in couchdb['databases']:
    del couchdb['databases'][dbName]
```
so that we don't store useless data in the dictionary ?

remh:
Nice first pass! I made a few comments (mostly nitpicks).

Do you think it's possible to extend the tests to test this scenario as well ?

irabinovitch:
@remh thanks for the feedback. Yes, we should be able to extend the tests. I'll take a stab at it tomorrow.

remh:
This fixes #1749

checks.d/couch.py
irabinovitch: Addressed in  22d0a3.

checks.d/couch.py
irabinovitch: We can but that would have a similar effect to the comments you left on line 120. Intermittent failures would result in the DB no longer being monitored.

checks.d/couch.py
remh: :+1:

checks.d/couch.py
```diff
@@ -108,9 +113,13 @@ def get_data(self, server, instance):
...
+                if e.response.status_code == 403:
+                    self.db_blacklist[server].append(dbName)
+                    self.warning('Database %s is not readable by the configured user. It will be added to the blacklist. Please restart then agent to clear.' % dbName)
```
remh: typo, should be:
```Please restart THE agent to clear```

irabinovitch: Fixed.

irabinovitch:
@remh @JohnLZeller OK. All tests are passing and I believe this is complete. Let me know if you have any further comments you'd like me to address.

JohnLZeller:
Thanks @irabinovitch! I'll take a look here shortly

JohnLZeller:
lgtm, good work :) :+1:

PR #1761, opened 2015-07-09
[checks] Handle requests raised timeouts in http_check

hkaj:
This fixes a bug reported by a customer where the service check wasn't set to DOWN since the exception was caught by the last, general except.


remh:
Nice,

Can we refactor this part of the code ? 
https://github.com/DataDog/dd-agent/blob/haissam/http_check_exception/checks.d/http_check.py#L88-L116

Looks like we could catch these 3 exceptions at once.

hkaj:
thanks for the suggestion @remh, it's done

PR #1770, opened 2015-07-14
gather stats from zookeeper mntr command

jpittis:
With zookeeper `3.4.0`, the `mntr` 4 letter admin command was added.
It provides zookeeper stats in key value form and is designed to be parsed.
Future stats will be added to the `mntr` command output specifically for this reason.

#### First Change
This change updates `checks.d/zk.py` to first parse the original `stat` command. If the zookeeper version supports the `mntr` command, it parses it too.

Here is an example output of the `mntr` command.
```
zk_version  3.4.5-cdh4.4.0--1, built on 09/04/2013 01:46 GMT
zk_avg_latency  0
zk_max_latency  0
zk_min_latency  0
zk_packets_received 4
zk_packets_sent 3
zk_num_alive_connections    1
zk_outstanding_requests 0
zk_server_state standalone
zk_znode_count  4
zk_watch_count  0
zk_ephemerals_count 0
zk_approximate_data_size    27
zk_open_file_descriptor_count   29
zk_max_file_descriptor_count    4096
```

#### Second Change
This change adds `report_instance_mode` which when called, sends stats regarding the zookeeper status.

A zookeeper node can be one of a `leader`, `follower`, `observer` or `standalone`. In addition, this check will report a mode of:
- `down` when the check cannot connect to zookeeper
- `inactive` when a node looses connection with the cluster
- `unknown` when the check fails in an unexpected way

This data is reported firstly in a set `zookeeper.instances` as well as 7 gauges that can be set to `0` or `1` named `zookeeper.instances.<mode>`.

When a node shifts say from `leader` to `down`, it's `zookeeper.instances.leader` value will be set to `0` and it's `zookeeper.instances.down` value will be set to `1`. This means you can rely on these values to notify you right when a node goes down.

The new mode system facilitates creating dashes and alerts for the change and distribution state in the cluster.

This patch is designed to be backwards compatible.


lgtml:
:heart:

jpittis:
Tests are passing locally but failing in travis. Everything will be green soon. :tm:

checks.d/zk.py
```diff
@@ -1,4 +1,25 @@
...
+As of zookeeper 3.4.0, the `mntr` admin command is provided for easy parsing of zookeeper stats.
+This check first parses the `stat` admin command for a version number.
+If the zookeeper version supports `mntr`, it is also parsed.
+
+Duplicate information is being reported by both `mntr` and `stat` to keep backwards compatability.
+Example:
+    `stat` reports: zookeeper.latency.avg
+    `mntr` reports: zookeeper.avg.latency
+If available, make use of the data reported by `mntr` not `stat`.
+The dublicate `stat` reports are only kept for backward compatability.
```
talwai: typo duplicate

checks.d/zk.py
```diff
@@ -96,6 +156,50 @@ def check(self, instance):
...
+            'standalone': 0,
+            'down': 0,
+            'inactive': 0,
+            'unknown': 0
+        }
+
+        if mode in gauges:
+            gauges[mode] = 1
+        else:
+            gauges['unknown'] = 1
```
talwai: Maybe cleaner to rely on a class level variable `ZookeeperCheck.STATUS_TYPES` and then building the `gauges` dictionary from it. What do you think?

checks.d/zk.py
```diff
@@ -210,4 +317,34 @@ def parse_stat(self, buf):
...
+        first = buf.readline() # first is version string or error
+        if first == 'This ZooKeeper instance is not currently serving requests':
+            return (None, 'inactive')
+
+        metrics = {}
+
+        for line in buf:
+            data = line.split()
+            if len(data) == 2:
+                name = data[0].replace('zk', 'zookeeper').replace('_', '.')
```
talwai: Tiny nitpick, but would we want to be careful about inadvertently replacing `zk` from the middle of the string? Also can you split out this logic into a `_normalize_metric_label` function so it's easier to test and play around with?

talwai:
@jpittis Thanks for your contribution! This looks like a very useful addition. I made a few comments on your PR, let me know when you've had the chance to look at them

checks.d/zk.py
jpittis: Your suggestion sounds much clearer.

jpittis:
Thanks @talwai for the swift reply and comments. I believe that I've improved the areas in question but am less fluent in Python so more review would be a pleasure.

I've made the changes in a separate commit which can be rebased at a later time. Please alert me if there is any way I can make this process easier for you.

*(the failing test seems unrelated to zookeeper and to this change)*

checks.d/zk.py
```diff
@@ -51,8 +105,12 @@ def check(self, instance):
         tags = instance.get('tags', [])
         cx_args = (host, port, timeout)
         sc_tags = ["host:{0}".format(host), "port:{0}".format(port)]
+        hostname = socket.gethostname()
```
talwai: The preferred way to do this in the agent is via `get_hostname` in the `util` module. See [the mongo check](https://github.com/DataDog/dd-agent/blob/7312a61a6d359e461c3481e24bd98ae6e489306a/checks.d/mongo.py#L10) for an example

checks.d/zk.py
```diff
@@ -68,22 +128,32 @@ def check(self, instance):
...
         # Read metrics from the `stat` output.
         try:
             stat_out = self._send_command('stat', *cx_args)
         except ZKConnectionFailure:
             self.increment('zookeeper.timeouts')
+            self.report_instance_mode(hostname, 'down', tags)
+            raise
+        except:
+            e = sys.exc_info()[1]
+            print >> sys.stderr, "Error: %s" % e
```
talwai: You should perhaps catch the base Exception with `except Exception as e` and then log it via `self.warning`. That way it shows up with the `dd-agent info` command.

checks.d/zk.py
```diff
@@ -96,6 +166,44 @@ def check(self, instance):
...
                                    tags=sc_tags)
 
+        if zk_version and LooseVersion(zk_version) > LooseVersion("3.4.0"):
+            try:
+                mntr_out = self._send_command('mntr', *cx_args)
+            except ZKConnectionFailure:
+                self.increment('zookeeper.timeouts')
+                self.report_instance_mode(hostname, 'down', tags)
+                raise
+            except:
```
talwai: Ditto my comment on Line 144

checks.d/zk.py
```diff
@@ -96,6 +166,44 @@ def check(self, instance):
...
+                raise
+            else:
+                metrics, mode = self.parse_mntr(mntr_out)
+                mode_tag = "mode:%s" % mode
+                if mode != 'inactive':
+                    for name in metrics:
+                        self.gauge(name, metrics[name], tags=tags + [mode_tag])
+
+
+    def report_instance_mode(self, hostname, mode, tags):
```
talwai: Now that I understand a little better the intent of `report_instance_mode` it seems that the method of passing in `hostname` (both your current one, and my suggestion) won't really cut it. Since the check is running off a single agent, even when there are multiple ZK instances, the hostnames passed in to this function will be identical every time. Is that we want? The `zookeeper.instances` set will only ever have a size of 1. Instead the `hostname` passed in to this function should be derived from the instance configuration (maybe just `instance.get('host')`)

checks.d/zk.py
```diff
@@ -96,6 +166,44 @@ def check(self, instance):
...
+            gauges[key] = 0
+
+        if mode in gauges:
+            gauges[mode] = 1
+        else:
+            gauges['unknown'] = 1
+
+        for k in gauges:
+            gauge_name = 'zookeeper.instances.%s' % k
+            self.gauge(gauge_name, gauges[k])
```
talwai: If you don't give this gauge additional context specific to the instance it will be overwritten by the last gauge to register. I'd recommend adding a `zk_host:<hostname>` tag.

checks.d/zk.py
```diff
@@ -210,4 +321,40 @@ def parse_stat(self, buf):
...
+            return (None, 'inactive')
+
+        metrics = {}
+
+        for line in buf:
+            try:
+                key, value = line.split()
+                name = self._normalize_metric_label(key)
+                metrics[name] = value
+            except:
```
talwai: Ditto my comment on Line 144

talwai:
Thanks for the updates @jpittis . Took a second pass at it and made a few more comments, let me know if they make sense

talwai:
@jpittis ping - any thoughts on this?

jpittis:
Sorry @talwai. I'll post my thoughts and changes today. Thanks for your patience.

checks.d/zk.py
jpittis: As far as I understand, the `host:<hostname>` is automatically added to the gauge?
We have a version of this code running which reports `zookeeper.instances.<mode>` and `zookeeper.instances` correctly.

I'm not sure how clear I've been: Each zookeeper node reports all 7 of the `zookeeper.instances.*` gauges. For each node, 6 of them will be toggled to `0` and one of them will be toggled to `1`.

I'm not entirely sure what issue you're speaking of as this code functions as I expect it to. Sorry to be a bother, my knowledge of datadog is on the side of practical rather than thorough. :)

jpittis:
@talwai: I made some of the changes you suggested. See my comment on adding additional context. I believe there's some kind of misunderstanding between us. Knowing me, I'm probably causing it. :)

checks.d/zk.py
talwai: @jpittis I wonder if the fact the code functions as you expect is a result of you happening to run the zookeeper instances on the same host as the agent. The `host: <hostname>` tag added by the agent is inferred from the machine the agent runs on. But the way the check is set up right now, the agent need not be monitoring `localhost` but can in fact be monitoring remote hosts with different hostnames. This is why the tag added by the agent can be inaccurate unless you explicitly add additional context. I see your point about the gauges registering correctly, but I wonder whether the check will be more useful in the datadog backend if some explicit tagging were included. I'm going to set up a ZK cluster of my own to play around with your check - hopefully that will get us on the same page :)

talwai:
Thanks @jpittis - as I noted I'm going to take a day or so to play around with your check on a ZK cluster so that I can be sure we're on the same page. I'll ping you again soon!

jpittis:
@talwai: Haven't heard back from you in a while, hope all is well! If there is any chance we could work together to get this done by the end of this week, I'm headed into the wilderness this weekend so I won't have access to the internet!

talwai:
Hi @jpittis , sincere apologies for the delay on this. Unfortunately we have frozen our feature set for the upcoming `5.5.0` release, but we have these updates slated for `5.6.0`! I got tied up earlier, but I can take another pass at this over the next few days and will get back to you next week. Enjoy your time in the wilderness!

jpittis:
@talwai I've been ignoring this patch for a long while. I'd like to see it through to approval. I'm not sure if you're still maintaining this project. I'm looking into the host tag stuff you mentioned in the previous comment. Hope to hear from you soon.

talwai:
@jpittis again sorry for the delay. I can confirm that this patch will make it into 5.7.0, after some minor cosmetic cleanup on our end. I'll send a more detailed update soon

checks.d/zk.py
jpittis: I see what you're meaning and I have yet to confirm your concern with a test.
Do note that the [current master `zk.py`](https://github.com/DataDog/dd-agent/blob/ab469ec2eed4c89daf17ad0d429a94c421185619/checks.d/zk.py#L86) does not seem to add extra host tags to the gauge calls. In the code linked above, neither `tags` or `new-tags` include a host tag. `new-tags` is set [here](https://github.com/DataDog/dd-agent/blob/ab469ec2eed4c89daf17ad0d429a94c421185619/checks.d/zk.py#L207) with no host tag. `tags` is set [here](https://github.com/DataDog/dd-agent/blob/ab469ec2eed4c89daf17ad0d429a94c421185619/checks.d/zk.py#L51) which does not specify a host tag either.

`tags` is configured from the config file and so the gauge in the update does require that additional context which I havn't added. (which I will add right away)

jpittis:
Thanks @talwai.

checks.d/zk.py
talwai: @jpittis if you haven't started work around this already, i wouldn't worry about adding the additional context. After playing around with the check in a zk environment, I think that the user-configured tags in the `instance` level config should suffice for adding necessary contexts (in fact this is probably a better option since the check's operators have complete control over what the additional context looks like). I think your change looks good as it stands

talwai:
Added a rebased version of this #2156 ; we will look to merge this in to master today

talwai:
These changes, along with some others were merged in #2156 . Thanks for the contribution @jpittis !

jpittis:
Thanks @talwai for your support!

Issue #1496, opened 2015-03-31
[postgres] make the relation metrics aware of schemas


clofresh:
Currently the postgres check can collect relation metrics from pg_stat_user_tables and pg_stat_user_indexes, but they don't specify the schema, so it could mixing up be metrics from tables with the same name in different schemas. I'd like to update the check to tag the relation metrics by schema as well as support specifying the schema in config.

So if you have the following db:

```
create schema prod;
create table prod.foo(id int);

create schema staging;
create table staging.foo(id int);
```

And you're running the old check:

```
    relations:
        - foo
```

You'd submit two rows of data, but the same set of tags, so one of the metrics would win.

In the new version, you'd get two rows of data, but one tagged with `schema:prod` and `schema:staging`.

You'd also be able to specify the schema:
```
    relations:
        - prod.foo
```
Which would only query for the prod schema's foo table.

yannmh:
Thanks for the detailed explanations @clofresh.

We didn't have time to do it for the 5.4.0 agent release. I am pushing back to 5.5.0.

yannmh:
Fixed with https://github.com/DataDog/dd-agent/pull/1771.

Thanks @kostas0

PR #1771, opened 2015-07-15
[postgres] add support for relation schemas

kostas0:
This change populates an additional tag for postgres relations containing the schema name. With this separate tag we can disambiguate metrics coming from same name tables but in separate schemas (e.g. staging vs production).

In addition we can now limit publishing to specific relation schemas via configuration.

Original issue #1496.


remh:
Thanks Kostas ! 
Can you also update the postgres.yaml.example to explain how to use this feature and what parameters to use ?

kostas0:
Added config syntax example.

conf.d/postgres.yaml.example
```diff
@@ -18,6 +18,14 @@ instances:
 #    relations:
 #      - my_table
 #      - my_other_table
+#
+# By default all schemas are included. To track relations from specific schemas only, 
+# use the following syntax:
+#
+#    relations:
+#      - relation_name: another_table
+#        schemas: public prod
```
remh: schemas should be a yaml list, instead of a space separated list

checks.d/postgres.py
```diff
@@ -354,6 +358,25 @@ def _get_replication_metrics(self, key, db):
             metrics = self.replication_metrics.get(key)
         return metrics
 
+    def _build_relations_config(self, yc):
+        """Builds a dictionary from relations configuration while maintaining compatibility
+        """
+        config = {}
+        for x in yc:
```
remh: You should use more explicit variable names

checks.d/postgres.py
```diff
@@ -354,6 +358,25 @@ def _get_replication_metrics(self, key, db):
...
+                if isinstance(x, str):
+                    config[x] = {'relation_name': x, 'schemas': []}
+                elif isinstance(x, dict):
+                    name = x['relation_name']
+                    config[name] = {}
+                    config[name]['schemas'] = x['schemas'].split()
+                    config[name]['relation_name'] = name
+                else:
+                    self.log.warn('Unhandled relations config type: %s' % str(x))
+            except:
```
remh: bare:
```python
except:
```
should almost always be avoided as they would catch  SystemExit, KeyboardInterrupt, and GeneratorExit as well which you don't want. See https://docs.python.org/2/howto/doanddont.html#except for more details.

You should catch the specific exceptions that can be expected, or if they are too many, you should do a:
```python
except Exception:
```

checks.d/postgres.py
```diff
@@ -354,6 +358,25 @@ def _get_replication_metrics(self, key, db):
...
+                pass
```
remh: Should we log something here ?

checks.d/postgres.py
```diff
@@ -461,7 +497,7 @@ def _collect_stats(self, key, db, instance_tags, relations, custom_metrics):
                     else:
                         tags = [t for t in instance_tags]
 
-                    tags += ["%s:%s" % (d[0][1], d[1]) for d in zip(desc, row[:len(desc)])]
+                    tags += [("%s:%s" % (k,v)) for (k,v) in desc_map.items()]
```
remh: You should use .iteritems() instead of .items() in python2

kostas0:
Thanks for the suggestions, now updated, please take a look.

remh:
Looks good to me!
 Thanks!

PR #1772, opened 2015-07-15
[checks.d][process] correctly handle disappearing pids

### LINKED ISSUE #1721: [process] Correctly handle psutil.NoProcessFound when getting process state
In situations where a process disappears after we discover its PID, we should be skipping over the subsequent interrogation of the process [here](https://github.com/DataDog/dd-agent/blob/cc5a67bb47110ef145e58561ab4471b946354742/checks.d/process.py#L173-L181). Currently the check is throwing errors like:
```
  File "/opt/datadog-agent/agent/checks.d/process.py", line 181, in get_process_state
    meminfo = self.psutil_wrapper(p, 'memory_info', ['rss', 'vms'])
UnboundLocalError: local variable 'p' referenced before assignment
```

talwai:
fixes #1721

### DIFF
--- a/checks.d/process.py
+++ b/checks.d/process.py
@@ -175,8 +175,9 @@ class ProcessCheck(AgentCheck):
             # Skip processes dead in the meantime
             except psutil.NoSuchProcess:
                 self.warning('Process %s disappeared while scanning' % pid)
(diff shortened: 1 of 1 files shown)


remh:
Looks good to me. Thanks!

PR #1773, opened 2015-07-16
[windows] add flare function

degemer:
Add a Windows flare function inside the GUI, using only popups (because it's
the easiest way to do this).

### DIFF
--- a/utils/flare.py
+++ b/utils/flare.py
@@ -140,28 +140,39 @@ class Flare(object):
         self._tar.close()
 
     # Upload the tar file
-    def upload(self):
+    def upload(self, email=None):
         self._check_size()
 
         if self._cmdline:
             self._ask_for_confirmation()
 
-        email = self._ask_for_email()
+        if not email:
+            email = self._ask_for_email()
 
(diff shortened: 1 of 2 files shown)


utils/flare.py
```diff
@@ -360,8 +371,8 @@ def _strip_password(self, file_path):
     # Add output of the command to the tarfile
     def _add_command_output_tar(self, name, command, command_desc=None):
         out, err, _ = self._capture_output(command, print_exc_to_stderr=False)
-        _, temp_path = tempfile.mkstemp(prefix='dd')
-        with open(temp_path, 'w') as temp_file:
+        fh, temp_path = tempfile.mkstemp(prefix='dd')
+        with os.fdopen(fh, 'w') as temp_file:
```
olivielpeau: :+1: 
We should probably check those lines too as they also use `tempfile.mkstemp`:

- https://github.com/DataDog/dd-agent/blob/quentin/flare-windows/utils/flare.py#L332
- https://github.com/DataDog/dd-agent/blob/quentin/flare-windows/utils/flare.py#L356

olivielpeau:
LGTM! :clap:

PR #1778, opened 2015-07-20
Add warning to 'status' and failover to 'info'

JohnLZeller:
On a fresh CoreOS install using the dockerized agent (https://registry.hub.docker.com/u/datadog/docker-dd-agent/) running service datadog-agent status in the container will return a misleading error message of "refused connection" followed by "Datadog Agent (supervisor) is NOT running all child processes".

The connection refusal is a known bug in the Linux kernel, introduced in 3.18 with the addition of overlayfs. This bug will show up in any system using kernel 3.18, which hardlinks unix domain sockets on overlayfs. The result is that calling supervisorctl will cause it to refuse the connection. For the record, tcp sockets are unaffected.

This patch adds a warning to the call to 'status' by first checking if the Linux kernel is version 3.18 or higher. If it is, a warning is presented, and we automatically call 'info' instead. If the version is less than 3.18, then functionality continues as normal.

### DIFF
--- a/packaging/centos/datadog-agent.init
+++ b/packaging/centos/datadog-agent.init
@@ -167,13 +167,17 @@ case "$1" in
         restart
         ;;
     status)
+        # Note: sh does not support arrays
         # Check for kernel version 3.18+ - overlayfs has known bug affecting unix domain sockets
-        IFS='.' read -ra VERS <<< "$( uname -r )"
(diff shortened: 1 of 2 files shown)


JohnLZeller:
Are we already testing datadpg-agent commands somewhere? I can't seem to find them anywhere in `tests/core`

hkaj:
LGTM

PR #1779, opened 2015-07-21
Install apt-transport-https

darron:
In preparation to serve packages through HTTPS - we need to install `apt-transport-https` so that `apt` can download through HTTPS.

From what I understand, YUM needs no additional packages - this is only for Debian and Ubuntu.


remh:
This change also needs to be done for chef/puppet/ansible/salt recipes as well as step by step instructions.

darron:
There's already a PR in for the Chef Recipe:

https://github.com/DataDog/chef-datadog/pull/219

Since it's not going to be a requirement and will be optional - it's not strictly required.

Puppet / Ansible / Salt recipes can be done no problem.

I couldn't find the location to change the step-by-step instructions - where are they located?

remh:
@darron https://github.com/DataDog/dogweb/blob/prod/dogweb/templates/macros/agent.mako#L74

PR #1781, opened 2015-07-21
Allow for more complex object hiearchy in debug.get_check

jshum:
In the original version of get_check(name,config_str),
if the class specified by name was not a direct
subclass of AgentCheck, the get_check function would return the
direct subclass of AgentCheck. This verison takes into account the object
hierarchy and returns the lowest class on the object hierarchy instead.
Examples for this are HTTPCheck(NetworkCheck(AgentCheck))

### DIFF
--- a/utils/debug.py
+++ b/utils/debug.py
@@ -47,10 +47,17 @@ def get_check(name, config_str):
     check_module = __import__(name)
     check_class = None
     classes = inspect.getmembers(check_module, inspect.isclass)
+
+    max_hierarchy_len = 0
     for name, clsmember in classes:
         if AgentCheck in clsmember.__bases__:
             check_class = clsmember
             break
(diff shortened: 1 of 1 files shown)


jshum:
>>> b = '''
... init_config:
...
... instances:
...   - name: My first service
...     url: http://some.url.example.com
...     timeout: 1
... '''
>>> from utils import debug
>>> debug.get_check('http_check',b)
(<checks.network_checks.NetworkCheck object at 0x7fd71f0cd990>, [{'url': 'http://some.url.example.com', 'name': 'My first service', 'timeout': 1}])

After the change,

>>> debug.get_check('http_check',b)
(<http_check.HTTPCheck object at 0x7f6404e97c90>, [{'url': 'http://some.url.example.com', 'name': 'My first service', 'timeout': 1}])

PR #1784, opened 2015-07-22
Recalculate disk.in_use to make consistent with df's Use% metric

JohnLZeller:
As it turns out psutil is actually UNDER counting disk.in_use because it is INCLUDING reserved space.
The solution, to bring this back to what df shows, is to calculate the disk usage ourselves by doing ( disk.used / (disk.used + disk.free) ), which will completely ignore the root space.

I separated 'system.disk.used' and 'system.disk.free' into different variables for readability.

Further Proof:
The calculation by psutil of disk.total is (disk.free + disk.used + <root_reserved_space>). You can prove that disk.total is including root space by adding disk.free + disk.used. The number you'll come up with is exactly <root_reserved_space> less than disk.total.

The calculation by psutil of disk.in_use is ( disk.used / disk.total ), pushing the percentage down because the disk.total includes the <root_reserved_space>.

### DIFF
--- a/checks.d/disk.py
+++ b/checks.d/disk.py
@@ -96,10 +96,21 @@ class Disk(AgentCheck):
             tags = [part.fstype] if self._tag_by_filesystem else []
             device_name = part.mountpoint if self._use_mount else part.device
 
+            # Note: psutil (0.3.0 to at least 3.1.1) calculates in_use as (used / total)
+            #       The problem here is that total includes reserved space the user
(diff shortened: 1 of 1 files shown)


elafarge:
That's a very good workaround indeed :) Simple and efficient. Good job.

Before we merge this to `master`, @remh I'd love to have your opinion on how to release this. As a bugfix ? For 5.5 ?

remh:
Looks good to me!

Not sure if we are going to do a bugfix release for this or not yet.

PR #1786, opened 2015-07-22
Force exclusion of CDROM (iso9660) from disk check

JohnLZeller:
Currently the agent disk check is including the monitoring of mounted cd-rom images. This isn't good as these are read-only and already have 100% usage. This is a threat for triggers false positives.

Solution is to forcibly exclude CDROM (iso9660) filesystems from the disk check.

### DIFF
--- a/checks.d/disk.py
+++ b/checks.d/disk.py
@@ -52,6 +52,9 @@ class Disk(AgentCheck):
         self._all_partitions = _is_affirmative(
             instance.get('all_partitions', False))
 
+        # Force exclusion of CDROM (iso9660) from disk check
+        self._excluded_filesystems.append('iso9660')
+
         # FIXME: 6.x, drop use_mount option in datadog.conf
         self._load_legacy_option(instance, 'use_mount', False,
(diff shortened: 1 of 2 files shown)


remh:
Nice but as far as i can see iso9660 is for Windows only. Did you try this on  a linux host ? 

(https://en.wikipedia.org/wiki/ISO_9660)

remh:
Also it seems that for windows, the CD rom case is already covered:
https://github.com/DataDog/dd-agent/blob/zeller/exclude-cdrom/checks.d/disk.py#L117-L119

JohnLZeller:
Bah, I'll take another look at this.

JohnLZeller:
Looks like iso9660 is for both Windows and Linux. I tested on Linux, and sure enough the fstype is iso9660. Testing now between this change, and how we did things with 5.3.0

JohnLZeller:
Agent 5.3.0 reports disk usage for read-only disks. This PR will exclude mounted cdrom disks, but preserves the behavior of agent 5.3.0 reporting disk usage of read-only disks.

PR #1788, opened 2015-07-23
[Docker] Add ECS tags for docker check

hkaj:
documentation: http://docs-staging.datadoghq.com/integrations/ecs/

### DIFF
--- a/checks.d/docker.py
+++ b/checks.d/docker.py
@@ -3,6 +3,7 @@ from collections import defaultdict
 import httplib
 import os
 import re
+import requests
 import socket
 import time
 import urllib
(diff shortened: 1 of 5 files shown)


checks.d/docker.py
```diff
@@ -16,6 +17,8 @@
 
 EVENT_TYPE = SOURCE_TYPE_NAME = 'docker'
 
+ECS_PORT = '51678/tcp'
```
remh: Do we need to hardcode this ? There is no way to get it from the docker api directly ?

hkaj: the response looks like this: ```"Ports": {
            "51678/tcp": [
                {
                    "HostIp": "127.0.0.1",
                    "HostPort": "51678"
                }
            ]
        }```
We can either hardcode the port, or choose arbitrary the first port in `ecs_config.get("Ports", {}).keys()`. In both case it may change in the future (the port can change, there can be another port open too). Should I pick the other solution?

remh:
Closing in favor of #1908

PR #1791, opened 2015-07-27
Have the mongo check be configurable for dbStats only

mmmmmm2:
This fixes the issue of having multiple databases will get the serverStatus data and report the values multiple times to Datadog. Also if the server has auth enabled, serverStatus will fail on non admin databases that are added to the check, when using the recommended 'clusterMonitor' role

### DIFF
--- a/checks.d/mongo.py
+++ b/checks.d/mongo.py
@@ -172,6 +172,8 @@ class MongoDb(AgentCheck):
             'ssl_ca_certs': instance.get('ssl_ca_certs', None)
         }
 
+        db_stats_only = instance.get('db_stats_only', False)
+
         for key, param in ssl_params.items():
             if param is None:
                 del ssl_params[key]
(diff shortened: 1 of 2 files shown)


remh:
Thanks a lot @mathrawka . We'll review that soon for our 5.6.0 release.

remh:
@yannmh can you have a look into this one while you're at it please ?

yannmh:
Thanks @mathrawka. 

Your solution looks good to me. However, it requires users to manually  set the `db_stats_only ` options through their different instances, while the check should be able to do the job itself. 
At this point, I think it would be easier if we just re-write the MongoDB check to be 'mutiple databases' friendly. 

This could be done for instance by having a `dbname` list option in the YAML configuration file, i.e.
```yaml
  - server: mongodb://user:pass@host:port
    dbname:
      - foo
      - bar
``` 

What do you think ?

irabinovitch:
@mmmmmm2  Thank you for your contribution! We look forward to your feedback on @yannmh 's comments.

irabinovitch:
@mmmmmm2  We'd love to hear your thoughts so we can include this in the 5.8 release.

olivielpeau:
The merge window for 5.8 is closing. Moving to 5.9

remh:
Closing for now. We will reopen if needed.

PR #1808, opened 2015-08-01
[mesos] Support Mesos version 0.23.0

uzyexe:
### Problem

datadog-agent 5.4.3 is not compatible with mesos version 0.23.0.
I got the error below...
```
Datadog's mesos integration is reporting:
Instance #0[ERROR]:"'failed_tasks'"
```

### Pull Request

/master/stats.json has been migrated to /metrics /snapshot from mesos 0.23.0,
and the following state attr keys has obsolete from mesos 0.23.0.
```
deactivated_slaves
failed_tasks
finished_tasks
killed_tasks
lost_tasks
staged_tasks
started_tasks
```

### DIFF
--- a/checks.d/mesos.py
+++ b/checks.d/mesos.py
@@ -22,6 +22,8 @@ class Mesos(AgentCheck):
         default_timeout = self.init_config.get('default_timeout', 5)
         timeout = float(instance.get('timeout', default_timeout))
 
+        self.version = self.get_version(url, timeout)
+
         response = self.get_master_roles(url, timeout)
         if response is not None:
             for role in response['roles']:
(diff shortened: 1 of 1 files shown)


yannmh:
Thanks for reporting @uzyexe. @DorianZaccaria can you take a pass on this issue ?

yannmh:
@uzyexe,
We are actually deprecating this generic Mesosphere check in favor of the Mesosphere master and slave specific checks introduced in the 5.4.0 release, which support Mesos version 0.23.0 . We'll add a note about it in the 5.5.0, and it'll likely be removed in a future version of the Datadog Agent. 
While we don't bring any more update to this check, we are very grateful for your time and your contribution. Thank you.

PR #1642, opened 2015-05-26
[HTTP/TCP Check] Add support for custom tags 

remh:
And get rid of legacy ch…eck names

The old check names are currently transformed in the backend

cc: @conorbranagan

PR #1811, opened 2015-08-03
[network] normalize `instance_name` tags

### LINKED ISSUE #1796: Network check instance tag mismatch in 5.4
Instance tags used to be a part of the check name and hence normalized on the agent before sending to DD. However, since the new change in 5.4 where instance is separated as tag it doesn't get normalized on the agent. The problem is that there is a difference between the agent's normalize and sobotka's tag normalize and hence there are cases when instance tag names don't match up. 

eg: `instance:tag-tag`  in 5.3 would get normalized to `instance:tag_tag` in 5.3 but remains `instance:tag-tag` in 5.4 

I've created a fix (https://github.com/DataDog/dogweb/pull/9031) in sobotka for now but this might be better fixed on the agent itself for tcp and http checks.

yannmh:
### Context
Instance name used to be a part of the check name and hence normalized
on the agent before being sent to Datadog.
As part of changes #1642 which introduced support for custom tags,
`instance_name` is now sent as a proper tag. 

### Changes
Normalize it, to avoid mismatch and backward incompatiblity. Fix #1796.
Add tests for `instance_name` tag normalization.
Move `core/test_service_checks.py` tests to `checks/integration/test_http_check.py` and `checks/integration/test_tcp_check.py`.


remh:
Nice, can you add a test for that please ?

yannmh:
Done

PR #1812, opened 2015-08-04
[http_check] Add quote to http copy for clarity

Viq111:
Per customer request:
> The `content_match` text is not picked out of the body of the email, so it’s not clear.

[Related card](https://trello.com/c/4TXcnGc8/314-copy-update-for-http-tcp-check-message)


remh:
Thanks!

Issue #1819, opened 2015-08-07
[RabbitMQ] Error when an absolute queue name is used


MartinDatadog:
When adding a queue in the rabbitmq.yaml with its absolute name i.e. `vhost/queue_name`, the datadog-agent runs into the following error:

Traceback (most recent call last): 
File "/opt/datadog-agent/agent/checks/__init__.py", line 661, in run 
self.check(copy.deepcopy(instance)) 
File "/opt/datadog-agent/agent/checks.d/rabbitmq.py", line 139, in check 
QUEUE_TYPE], specified[QUEUE_TYPE], auth=auth) 
File "/opt/datadog-agent/agent/checks.d/rabbitmq.py", line 214, in get_stats 
explicit_filters.remove(name) 
ValueError: list.remove(x): x not in list

Reported in ticket: https://datadog.zendesk.com/agent/tickets/29961

yannmh:
Fixed with #1820. Thanks @MartinDatadog !

PR #1820, opened 2015-08-07
[RabbitMQ] Fix variable for absolute queue name situations

MartinDatadog:
Fixing issue #1819
where exception `ValueError: list.remove(x): x not in list` is raised
any time there is a queue with an absolute name (format `vhost/queue_name`)
in the rabbitmq.yaml


remh:
Good catch! Thanks!

PR #1824, opened 2015-08-10
[vmware] Adding network.received and network.transmitted to basic_metrics.py

JohnLZeller:
The vsphere dashboard is currently attempting to use these two metrics, but they are not included in the basic metrics collected. This should add those metrics, without the need to enable all metrics (read: a TON of metrics).

The vsphere dashboard is also trying to display disk.usage and mem.usage, which are also not basic metrics, but those are not being added at this time, because their compatibility is unknown as shown [here](https://github.com/DataDog/dd-agent/blob/master/checks/libs/vmware/all_metrics.py#L1215) and [here](https://github.com/DataDog/dd-agent/blob/master/checks/libs/vmware/all_metrics.py#L699)

### DIFF
--- a/checks/libs/vmware/basic_metrics.py
+++ b/checks/libs/vmware/basic_metrics.py
@@ -89,4 +89,16 @@ BASIC_METRICS = {
         'rollup'       : 'average',
         'entity'       : ['VirtualMachine', 'HostSystem', 'ResourcePool']
     },
+    'network.received': {
+        's_type'       : 'rate',
+        'unit'         : 'kiloBytesPerSecond',
+        'rollup'       : 'average',
+        'entity'       : ['VirtualMachine', 'HostSystem']
+    },
+    'network.transmitted': {
+        's_type'       : 'rate',
+        'unit'         : 'kiloBytesPerSecond',
+        'rollup'       : 'average',
+        'entity'       : ['VirtualMachine', 'HostSystem']
+    },
 }


LeoCavaille:
LGTM. The only thing that bothers me is that we'll report these as gauges, might not be ideal because we cannot switch from the rate value to a count because it won't show the proper type.
But that's also true if someone enables `all_metrics` so it would need a deeper refactoring I guess, can you assess if that's true and file an issue if needed?

JohnLZeller:
Will do

JohnLZeller:
@LeoCavaille: So, it looks like [we do currently report these as gauges:](https://github.com/DataDog/dd-agent/blob/master/checks.d/vsphere.py#L751-L756)

What's the ideal way to report these, in your opinion? 
And, do you think that this PR could be merged before changing the vsphere check to use something other than gauge?

LeoCavaille:
I'd advise to report it correctly from the beginning, we know that this
change will be backwards incompatible and hard to make if people are using
the metric.
@remh can weigh in maybe.

Also it might already be the case, some rates/counters might be wrongly
reported. It does not seem too hard to read the metric type from the VMware
metrics module and call self.gauge or self.rate, etc. accordingly.

-- 
Léo

JohnLZeller:
Upon looking even further at this, it's definitely a candidate for its' own ticket in my opinion, since it not only affects these 2 metrics, but also the rest listed in [all_metrics](https://github.com/DataDog/dd-agent/blob/zeller/vsphere-basic/checks/libs/vmware/all_metrics.py).

JohnLZeller:
(internal) Here is the corresponding ticket on dogweb: https://github.com/DataDog/dogweb/pull/9194

JohnLZeller:
@remh could you weigh in?

JohnLZeller:
I *think* that this is the right way to handle the different metric types, but I will need to test it on a vsphere setup.

https://github.com/DataDog/dd-agent/compare/zeller/vsphere-metrics

@LeoCavaille Am I right in thinking that types absolute and delta are okay to be reported as a gauge, leaving rate to be reported as... rate? :)

JohnLZeller:
Blocked by https://github.com/DataDog/dd-agent/pull/2115

PR #1827, opened 2015-08-11
[consul] some tag reorganization

talwai:


talwai:
failures are flaky tests. Merging this

PR #1829, opened 2015-08-12
[network] fix flaky tests + better exception log

yannmh:
Fix flaky tests and enhance content exception messages.


benmccann:
Cool. It passed both the network tests. Failed on zookeeper, so I wonder if the zookeeper test is flaky as well

yannmh:
You're right, it looks like more tests would need a review  :smirk:. Merging this one, as a first step.

PR #1831, opened 2015-08-13
[docker] Add support for Docker 1.8

remh:
Docker 1.8 adds cgroup bind mount by default ( See
https://github.com/docker/docker/blob/v1.8.0/CHANGELOG.md#runtime )

Which confuses the docker check when running in a container. This is a
tiny fix to make the docker check prefers mountpoint starting with
“/host”.

Also, use "with" syntax to open files.

### DIFF
--- a/checks.d/docker.py
+++ b/checks.d/docker.py
@@ -442,13 +442,9 @@ class Docker(AgentCheck):
     def _find_cgroup(self, hierarchy, docker_root):
         """Finds the mount point for a specified cgroup hierarchy. Works with
         old style and new style mounts."""
-        fp = None
-        try:
-            fp = open(os.path.join(docker_root, "/proc/mounts"))
+        with open(os.path.join(docker_root, "/proc/mounts"), 'r') as fp:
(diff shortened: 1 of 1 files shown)


checks.d/docker.py
```diff
@@ -456,20 +452,24 @@ def _find_cgroup(self, hierarchy, docker_root):
...
+
 
     def _parse_cgroup_file(self, stat_file):
         """Parses a cgroup pseudo file for key/values."""
-        fp = None
         self.log.debug("Opening cgroup file: %s" % stat_file)
         try:
-            fp = open(stat_file)
-            return dict(map(lambda x: x.split(), fp.read().splitlines()))
+            with open(stat_file, 'r') as fp:
```
yannmh: `'r'` is the default mode in [open](https://docs.python.org/2/library/functions.html#open) method. Nitpick, since it's specified here but not on :445.

yannmh:
LGTM :cactus:

checks.d/docker.py
remh: switched line 445 as well, it's better to be explicit :)

PR #1832, opened 2015-08-13
[mongo] add tcmalloc metrics

benmccann:
You will need MongoDB 3.2 or newer to benefit from this PR since that's when the tcmalloc metrics start being returned. This PR works just fine on older versions of MongoDB, but you won't get these new metrics.

### DIFF
--- a/checks.d/mongo.py
+++ b/checks.d/mongo.py
@@ -17,12 +17,12 @@ class MongoDb(AgentCheck):
     SOURCE_TYPE_NAME = 'mongodb'
 
     GAUGES = [
-        # L21-23,25,68-74 Deprecated as if V 3.0.0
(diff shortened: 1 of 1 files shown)


remh:
Thanks a lot @benmccann we'll review this for our 5.6.0 release!

benmccann:
Thanks. Would we be able to get this in now that 5.5.0 has been released?

yannmh:
Hi @benmccann,

Sincere apologies for the delay on this. 
I rebased your work to sync with the current state of the check and added TCMalloc metrics collection behind an option. The idea is to keep the metric list as succinct as possible but allows 'advanced user' to enhance it by turning on some options.
Changes are available in this PR #1979 and contains your commit. In the meantime, I am closing this one.

Please let me know what you think about it :smile:

PR #1833, opened 2015-08-14
[osx][install_script] Remove confusing sudo prompt

elafarge:
We use to display a prompt when asking for a sudo password which turns
out to be useful if the user's sudo password is cached because recently
used in the same terminal... that's really confusing and in addition, it
leads customers to type their password in clear on a terminal, which
isn't really cool.

This has been removed and we now rely on the system prompt only when the
sudo password is needed :)

We keep on notifying the user he MAY be asked for his sudo password
since OSX just shows `password`, it made sense to precise that it
actually means the SUDO password and not the one on Datadog or whatever.

[skip ci]

### DIFF
--- a/packaging/osx/install.sh
+++ b/packaging/osx/install.sh
@@ -45,15 +45,16 @@ if [ ! $apikey ]; then
 fi
 
 # Install the agent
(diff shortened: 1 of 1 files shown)


yannmh:
:+1: 

Thanks for adding more captions during the installation process.

PR #1838, opened 2015-08-18
[wmi] do not raise on null values

yannmh:
Sinces #1800, `val` variable is unreferenced when WMI metric is null.
Thus, an exception is raised when we try to submit it to Datadog.
Ignore null values, and continue, so it does not raise.


JohnLZeller:
LGTM, assuming all checks pass

PR #1839, opened 2015-08-18
[btrfs] track usage based on used bytes instead of free bytes

pbitty:
In other datadog stats (eg. `system.fs.inodes.in_use`, `system.disk.in_use`), 0 means no usage, and 1 means full usage.

This change makes `btrfs.disk.btrfs.usage` report values in the same way: 1 = 100% usage.


remh:
Thanks a lot. It was definitely a bug.

PR #1840, opened 2015-08-18
[snmp] add debug level logs

yannmh:
SNMP check lacks of logs, thus, it is quite difficult to troubleshoot.
Adding debug level logs to fix that.


olivielpeau:
LGTM ! :+1: 
The build failure is unrelated.

PR #1843, opened 2015-08-19
[mesos_slave] Do not fail if no cluster name

DorianZaccaria:
If the cluster doesn't have a name the check shouldn't fail.

### DIFF
--- a/checks.d/mesos_slave.py
+++ b/checks.d/mesos_slave.py
@@ -135,8 +135,8 @@ class MesosSlave(AgentCheck):
             if state_metrics is not None:
                 self.version = map(int, state_metrics['version'].split('.'))
                 master_state = self._get_state('http://' + state_metrics['master_hostname'] + ':5050', timeout)
-                if master_state is not None and 'cluster' in master_state:
-                    self.cluster_name = master_state['cluster']
+                if master_state is not None:
+                    self.cluster_name = master_state.get('cluster')
 
         return state_metrics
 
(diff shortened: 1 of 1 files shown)


checks.d/mesos_slave.py
```diff
@@ -135,7 +135,7 @@ def _get_constant_attributes(self, url, timeout):
             if state_metrics is not None:
                 self.version = map(int, state_metrics['version'].split('.'))
                 master_state = self._get_state('http://' + state_metrics['master_hostname'] + ':5050', timeout)
-                if master_state is not None:
+                if master_state is not None and 'cluster' in master_state:
                     self.cluster_name = master_state['cluster']
```
yannmh:
```python
if master_state is not None and 'cluster' in master_state:
    self.cluster_name = master_state['cluster']
```
could be simplified:
```python
if master_state :
    self.cluster_name = master_state.get('cluster')
```

I am pretty sure we can drop the `if` statement too, and have simply:
```python
 self.cluster_name = master_state.get('cluster') 
```
as `master_state` could not be null at this point.

yannmh:
LGTM :cactus:. Thanks for tackling this issue.

I just made a tiny/nitpick comment.

Feel free to merge it.

checks.d/mesos_slave.py
DorianZaccaria: What if the `get master state` fail?

PR #1845, opened 2015-08-19
Account for %guest cpu time

olivielpeau:
Rebased https://github.com/DataDog/dd-agent/pull/1656 on current master

Thanks to @rhormaza for contributing!

### DIFF
--- a/checks/system/unix.py
+++ b/checks/system/unix.py
@@ -587,8 +587,8 @@ class Cpu(Check):
         """Return an aggregate of CPU stats across all CPUs
         When figures are not available, False is sent back.
         """
-        def format_results(us, sy, wa, idle, st):
-            data = {'cpuUser': us, 'cpuSystem': sy, 'cpuWait': wa, 'cpuIdle': idle, 'cpuStolen': st}
+        def format_results(us, sy, wa, idle, st, guest=None):
+            data = {'cpuUser': us, 'cpuSystem': sy, 'cpuWait': wa, 'cpuIdle': idle, 'cpuStolen': st, 'cpuGuest': guest}
             for key in data.keys():
                 if data[key] is None:
                     del data[key]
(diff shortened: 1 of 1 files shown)


checks/system/unix.py
```diff
@@ -587,8 +587,8 @@ def check(self, agentConfig):
         """Return an aggregate of CPU stats across all CPUs
         When figures are not available, False is sent back.
         """
-        def format_results(us, sy, wa, idle, st):
-            data = {'cpuUser': us, 'cpuSystem': sy, 'cpuWait': wa, 'cpuIdle': idle, 'cpuStolen': st}
+        def format_results(us, sy, wa, idle, st, guest=None):
+            data = {'cpuUser': us, 'cpuSystem': sy, 'cpuWait': wa, 'cpuIdle': idle, 'cpuStolen': st, 'cpuGuest': guest}
             for key in data.keys():
                 if data[key] is None:
                     del data[key]
```
yannmh: Unrelated to your changes, and nitpick, but this can be simplified:

```python
data = dict((k,v) for k,v in data.iteritems() if v)
```
(Python 2.6+ compatible)

yannmh:
Thanks for the addition @olivielpeau. While you are on it, can you reword the git commit title to match our guidelines please ?

checks/system/unix.py
olivielpeau: We do want to keep `0` values, so I guess it'd rather be:
```python
data = dict((k,v) for k,v in data.iteritems() if v is not None)
```

right ?

Also I personally prefer keeping `del` as it doesn't recreate a new dict but just deletes the entry in the existing dict. But the dict is very small here so I guess it makes no difference.

olivielpeau:
@yannmh Thanks for the review! I've reworded the commit message and addressed your comment, could you take one last look at my changes please? :smiley:

olivielpeau:
CI passed, merging!

PR #1847, opened 2015-08-20
Windows process management refactoring

elafarge:
As part as the "Use Omnibus4 for our Windows build", this contains a big refactoring of the Windows Service layer on the agent. 

Basically we now have a `real` supervisord-like program (it doesn't read the `supervisord.conf` but I'm working on it on my free time) that logs things, that can be launched from the Python command line (without having to be declared as a Windows service). It launches real commands just like `supervisord` would do in `psutil` subprocesses (it's a wrapper around the `subprocess` class providing some additional utilities).

It also mean that the agent's components (`dogstatsd`, the `collector` and the `forwarder`) had to be slightly modified to be Windows compatible. They can now be launched independently from the command line, which (along with the agent's Python `venv` being completely accessible thanks to the new packaging of the agent) makes hacking on the agent on Windows a lot easier. 

Finally the "real" windows service resides in a independant layer now: `win32/ddagent_service.py`. It's python right now and built into an exe with py2exe (so it's shipped with it's own python installation, around 10 megs :s) but it could be rewritten in any compiled language (Go ? C ?) to be made lighter. It basically aims at starting our supervisor and killing it when asked to do so (right now it also kills all the children from that supervisor, I couldn't find a cleaner way to make our supervisor exit so far (and I've been trying hard). There might be an alternative involving windows events... or make the supervisor listening to a port on which it can receive instruction from the `ddagent_service.py`.


setup.py
```diff
@@ -25,62 +24,26 @@
...
-                    continue
-                # we skip psycopg2 now because don't want to install PG
-                # on windows
-                if 'psycopg2' in line:
-                    continue
-                install_requires.append(line)
 
     # windows-specific deps
-    install_requires.append('pywin32==217')
+    install_requires.append('pywin32==219')
```
remh: I think there was a reason we were using 217.
@LeoCavaille do you remember ?

agent.py
```diff
@@ -92,11 +92,12 @@ def run(self, config=None):
         # Gracefully exit on sigterm.
         signal.signal(signal.SIGTERM, self._handle_sigterm)
 
-        # A SIGUSR1 signals an exit with an autorestart
-        signal.signal(signal.SIGUSR1, self._handle_sigusr1)
+        if os.name != 'nt':
```
remh: You should use the Platform util class

util.py
```diff
@@ -422,49 +422,97 @@ def get_instance_id(agentConfig):
...
+if os.name == 'nt':
+    import psutil
+    import threading
+
+    class Watchdog(threading.Thread):
+        """ Simple watchdog for Windows (relies on psutil) """
+        def __init__(self, duration, max_mem_mb = None):
+            self._duration = int(duration)
+
+            if max_mem_mb is not None:
```
remh: Let's remove this. It's not used, even less on windows.

util.py
```diff
@@ -422,49 +422,97 @@ def get_instance_id(agentConfig):
...
+            with self.tlock:
+                self.expire_at = time.time() + self._duration
+
+        def watch(self):
+            while True:
+                if time.time() > self.expire_at:
+                    self.self_destruct()
+                time.sleep(self._duration/20)
+
+else:
```
remh: You shouldn't do that.
You should have two classes:
- WatchdogPosix
- WatchdogWindows

And choose the right one when instantiating the watchdog

requirements.txt
```diff
@@ -19,6 +19,7 @@ requests==2.6.0
 simplejson==3.6.5
 supervisor==3.1.3
 tornado==3.2.2
+psutil==2.1.1
```
remh: it shouldn't be in requirements.txt

remh:
@degemer are you basing your work on this PR ?

degemer:
@remh Indeed, I assumed it was needed for Omnibus.

elafarge:
Yes indeed, the idea was to have a process management that would look like the one on Linux to make the transition to Omnibus much easier as far as I remember. 

The main goal was to split the Windows Service layer and the process management layer in two parts so that the former gets packaged with py2exe and launches the latter just like InitD/SystemD would launch supervisord on Linux. Let me know if you have any questions @degemer  ;-)

olivielpeau:
Superseded by #2417

PR #1848, opened 2015-08-20
[zk] new `zookeeper.bytes_received/sent*` rates 

### LINKED ISSUE #1729: [zookeeper] metrics should be reported as counts not gauges
For instance:
* `zookeeper.bytes_received` is a monotonically increasing gauge since ZK start
* `zookeeper.bytes_sent`

yannmh:
Switch `zookeeper.bytes_received`, `zookeeper.bytes_sent` from `gauge` to `count` type. 
**WARNING**: potential backward incompatibility issues.
Fix #1729.

@remh can you take a pass on it ? I am also wondering if we should use `rate` instead of a `count`, as both are valid.

### DIFF
--- a/checks.d/zk.py
+++ b/checks.d/zk.py
@@ -24,9 +24,10 @@ Tested with Zookeeper versions 3.0.0 to 3.4.5
 
 '''
 # stdlib
+from collections import namedtuple
+from StringIO import StringIO
 import re
 import socket
-from StringIO import StringIO
 import struct
 
 # project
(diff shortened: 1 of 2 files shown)


yannmh:
cc @LeoCavaille who opened the original issue :key:

checks.d/zk.py
```diff
@@ -38,6 +39,15 @@ class ZKConnectionFailure(Exception):
     pass
 
 
+class ZKMetric(namedtuple('ZKMetric', ['name', 'value', 'm_type'])):
```
remh: This is a bit weird. You are using named tuples but you are not using the fact that attributes are named. You should just use regular tuples IMO.

yannmh:
__Updated__
Switch `zookeeper.bytes_received`, `zookeeper.bytes_sent` from count to rate type. Rename both metrics accordingly.

Make `ZKMetric` inherits from tuple rather than namedtuple.

yannmh:
I renamed `zookeeper.bytes_(received/sent)` to `zookeeper.bytes_(received/sent)_per_second` to be consistent with other integrations.
Howver it would trigger `No Data` alerts on some monitors. I reckon we should maybe keep the old `zookeeper.bytes_(received/sent)` gauges, and create the new metrics separately.

remh:
@yannmh can you rebase this PR and keep the old metrics as well ?

PR #1807, opened 2015-07-31
[Docker] New check

hkaj:

PR #1851, opened 2015-08-20
[docker] `docker.containers.count` metric

### LINKED ISSUE #1648: Send docker.containers.stopped as 0 when it is 0 instead of skipping key.
Currently the dd-agent seems to count stopped containers and send to `docker.containers.stopped`. If there are no stopped containers it seems to skip sending the key at all.

The problem with that arises when I try to monitor the key as in our system there should never be any container that is down for longer than a few seconds for upgrades and such.

However datadog does not seem to count a missing value as 0. So if a container is stopped at 1:00am for 1 minute for upgrade and then another is stopped at 10:00am for a minute then when graphing it in datadog it will show that 1 container has been down continuously for the last 10 hours which is simply not the case.

That means that making a monitor that alerts me when *`docker.containers.stopped` is above or equal '1' on average for the last 10 minutes* will trigger for any stopped container immediately instead of when it has been down for 10 minutes.

yannmh:
Built on top of #1807.
Add a `docker.containers.count` metric to count the number of docker containers.

Fix #1648 as we can now monitor `docker.containers.count - docker.containers.running`. 

Assigning you @remh for the review :mag:

### DIFF
--- a/checks.d/docker_daemon.py
+++ b/checks.d/docker_daemon.py
@@ -45,8 +45,8 @@ CGROUP_METRICS = [
 
 TAG_EXTRACTORS = {
     "docker_image": lambda c: c["Image"],
-    "image_name": lambda c: c["Image"].split(':', 1)[0],
-    "image_tag": lambda c: c["Image"].split(':', 1)[1],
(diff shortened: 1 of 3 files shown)


remh:
Superseded by https://github.com/DataDog/dd-agent/pull/1908

PR #1852, opened 2015-08-20
[logging] Fix logs duplication in jmxfetch.log

olivielpeau:
Replaces #1850 , which was very "hacky".

The collector logs were being duplicated to `jmxfetch.log` because `agent.py` imports `jmxfetch.py`, which causes the `initialize_logging` method to be called twice.

This fixes the issue by only initializing logging in `jmxfetch.py` when it's the main module (and not imported from another module).


yannmh:
Looks good to me !

PR #1853, opened 2015-08-21
[linuxvm] Add a new check for LinuxVm metrics

DorianZaccaria:
Aggregates metrics available in /proc/vmstat and /proc


irabinovitch:
Hey @DorianZaccaria  Looks like we might have some failing tests here.

irabinovitch:
@DorianZaccaria  do we still want to keep this PR open?

checks.d/linuxvm.py
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env python
```
remh: You don't need this line

checks.d/linuxvm.py
```diff
@@ -0,0 +1,49 @@
...
+
+"""DataDog agent checks for system.vm proc info."""
+
+from checks import AgentCheck
+
+
+def valid_instance(instance):
+    return "proc" in instance or "vmstat" in instance
```
remh: this helper seems a bit overkill :)

checks.d/linuxvm.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+class LinuxVm(AgentCheck):
+    VM_BASE = "system.vm"
+
+    def check(self, instance):
+        if not valid_instance(instance):
+            return
```
remh: You should raise an Exception here

checks.d/linuxvm.py
```diff
@@ -0,0 +1,49 @@
...
+
+        proc = instance.get("proc", {})
```
remh: you don't need to set default values as you checked before that these keys were in the instance dict.

checks.d/linuxvm.py
```diff
@@ -0,0 +1,49 @@
...
+
+        if proc:
+            for gauge in proc.get("gauges", []):
+                self.proc(self.gauge, gauge)
+
+            for counter in proc.get("counts", []):
+                self.proc(self.count, counter)
+
+        if vmstat:
+            vmstats = {x: int(y) for x, y in map(str.split, open("/proc/vmstat").readlines())}
```
remh: python2.6 doesn't support dictionary comprehension unfortunately :(

remh: also it's good practice to always open files in a context manager:
```python
with open('/proc/vmstat') as f:
     content = vm.readlines()
```
this way you make sure the file handler will be properly closed.

checks.d/linuxvm.py
```diff
@@ -0,0 +1,49 @@
...
+            print vmstats
```
remh: should be removed.

checks.d/linuxvm.py
```diff
@@ -0,0 +1,49 @@
...
+            for gauge in vmstat.get("gauges", []):
+                self.vm(self.gauge, gauge, vmstats)
+
+            for counter in vmstat.get("counts", []):
+                self.vm(self.count, counter, vmstats)
+
+    def proc(self, metfunc, path):
```
remh: you should use better names for your methods (usually use "actions")

checks.d/linuxvm.py
```diff
@@ -0,0 +1,49 @@
...
+        try:
+            value = float(open(path).read().strip())
```
remh: you're not closing your file.

checks.d/linuxvm.py
```diff
@@ -0,0 +1,49 @@
...
+            name = path.rstrip('/').split('/')[-1]
```
remh:
```python
import ntpath
_, name = ntpath.split(path)
```

checks.d/linuxvm.py
```diff
@@ -0,0 +1,49 @@
...
+            metfunc("%s.%s" % (self.VM_BASE, name), value)
+        except Exception as e:
+            self.log.error(e)
+
+    def vm(self, metfunc, stat, stats):
```
remh: same here, it's not really understandable

checks.d/linuxvm.py
```diff
@@ -0,0 +1,49 @@
...
+        if stat not in stats:
+            return
```
remh: should we log an error or something here ?

irabinovitch:
@DorianZaccaria  any thoughts/feedbacks on Remi's comments?

irabinovitch:
@DorianZaccaria  Ping?

DorianZaccaria:
Not really generic, it needs a bit of work. So closing for now. We might want to revisit our approach to get these metrics.

irabinovitch:
Thanks for the update @DorianZaccaria

PR #1854, opened 2015-08-24
[snmp] do not infinite loop on `error_status`

yannmh:
Fix SNMP check infinite loop when SNMP command returns an
`error_status`. Update `first_oid` reference before `continue`
statements are called.


elafarge:
:sheep: it !

Issue #1156, opened 2014-10-14
Should not be using netstat to gather socket connections in network.py


tliakos:
Should be using 'ss' as opposed to netstat. I have instances that are running with socket connections > 20000 - Netstat will either timeout or topple over before the metric is sent. Can we discuss changing this to ss since it gets its metrics directly from the kernel space as opposed to /proc (netstat) which is heavy.

(line 141) netstat = subprocess.Popen(["netstat", "-n", "-u", "-t", "-a"],
stdout=subprocess.PIPE,
close_fds=True).communicate()[0]

joelvanvelden:
:+1: Netstat is evil - it will throw out information to ensure output is formatted nicely - surely the worst design goal ever. Aparantly this is not considered a bug!

https://bugzilla.redhat.com/show_bug.cgi?id=755905

>this is expected behavior.
>Netstat produces output formatted to approx. 80 characters wide.
>To achieve this width, columns are truncated.

yannmh:
Fixed with #1859.

PR #1859, opened 2015-08-25
[network] Use `ss` instead of `netstat` on linux

olivielpeau:
`netstat` is deprecated on many distributions, `ss` is its replacement and is present by default in the `PATH` on most of the distributions that I've tested (CentOS >= 6, Ubuntu >= 12.04, but not on CentOS 5).

We still fallback to `netstat` if `ss` is not installed.

Addresses https://github.com/DataDog/dd-agent/issues/1156

I've also checked that `subprocess.Popen` doesn't hang with `stdout=PIPE`: as long as we use it with `communicate()` (which we do here) we should be fine.

### DIFF
--- a/checks.d/network.py
+++ b/checks.d/network.py
@@ -8,6 +8,7 @@ import subprocess
 # project
 from checks import AgentCheck
 from utils.platform import Platform
+from utils.subprocess_output import get_subprocess_output
 
 BSD_TCP_METRICS = [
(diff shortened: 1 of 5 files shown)


remh:
Nice. Can you write a test that checks that the metrics collected are equivalent ?

remh:
Also can you switch to this helper ? https://github.com/DataDog/dd-agent/blob/master/utils/subprocess_output.py

It will be easier to debug and is safer to use

hkaj:
LGTM

olivielpeau:
Thanks @hkaj and @remh for your reviews, merging.

PR #1864, opened 2015-08-26
[checks.d][openstack] add monitoring for Openstack Nova and Neutron

talwai:
The OpenStack `AgentCheck` is intended to run besides individual hypervisors. It can be scoped to any set of projects living on that host via instance-level config, 

At the hypervisor-level it collects:

- [Hypervisor Uptime and Statistics](http://developer.openstack.org/api-ref-compute-v2.1.html#os-hypervisors-v2.1) 

At the project-level it collects:
- [Diagnostics for guest servers within that project](http://developer.openstack.org/api-ref-compute-v2.1.html#diagnostics-v2.1)
- [Absolute limits for the project](http://developer.openstack.org/api-ref-compute-v2.1.html#limits-v2.1)

Additionally it sends service checks to register the UP/DOWN state of networks discovered by the agent, 
the locally running hypervisor, and the outward-facing (`public || internal`) API services of Nova, Neutron and Keystone

### Authentication
While the check will run without issues as an `admin` user, it is recommended to configure a read-only `datadog` user and configure the check with the corresponding user/password. Instructions on setting up the datadog user + role , as well as the changes required to the `policy.json` file can be found [here](https://gist.github.com/talwai/b8698e061795cec9f263#configure-a-datadog-user)

### A note on compatibility
Authentication is performed via the password method, and requires Identity API v3
Nova API v2 and v2.1 are supported, with minor additional configuration necessary for v2

### DIFF
--- a/checks.d/openstack.py
+++ b/checks.d/openstack.py
@@ -1,7 +1,22 @@
-import json
-import requests
+# stdlib
+from datetime import datetime, timedelta
+from urlparse import urljoin
 
+# project
 from checks import AgentCheck
+from util import get_hostname
+
+# 3p
+import requests
+import simplejson as json
+
+SOURCE_TYPE = 'openstack'
+
+DEFAULT_KEYSTONE_API_VERSION = 'v3'
+DEFAULT_NOVA_API_VERSION = 'v2.1'
+DEFAULT_NEUTRON_API_VERSION = 'v2.0'
+
+DEFAULT_API_REQUEST_TIMEOUT = 5 # seconds
 
 NOVA_HYPERVISOR_METRICS = [
     'current_workload',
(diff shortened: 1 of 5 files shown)


checks.d/openstack.py
```diff
@@ -0,0 +1,421 @@
...
+
+        if not server_ids:
+            self.warning("Your check is not configured to monitor any servers.\n" +
+                         "Please list `server_ids` under your init_config in openstack.yaml")
+
+        for sid in server_ids:
+            self.get_stats_for_single_server(sid)
+
+    def get_all_server_ids(self):
+        url = '{0}/servers'.format(self._nova_url)
```
mtougeron: Needs to have have the option to have `?all_tenants=1` in the URL or else it will only get all servers for the current project/tenant. Also, FYI, if the auth scope is not 'admin' then it will still limit the results to the current tenant (no errors).

checks.d/openstack.py
```diff
@@ -0,0 +1,827 @@
+# stdlib
+from datetime import datetime, timedelta
+import json
```
remh: use simplejson instead

checks.d/openstack.py
```diff
@@ -0,0 +1,827 @@
...
+        Parse authorization scope out of init_config
+
+        To guarantee a uniquely identifiable scope, expects either:
+        {'project': {'name': 'my_project', 'domain': {'id': 'my_domain_id'}}}
+        OR
+        {'project': {'id': 'my_project_id'}}
+        """
+        auth_scope = instance_config.get('auth_scope')
+        if not auth_scope or not auth_scope.get('project'):
+            raise IncompleteAuthScope
```
remh: As per PEP8 you should raise an instance of the class, not the class directly

conf.d/openstack.yaml.example
```diff
@@ -0,0 +1,67 @@
+init_config:
+      # All directives prefixed with a '#' sign are optional and will default to sane values when omitted
+
+      # Where your identity server lives (please omit the trailing slash). Note that the server must support Identity API v3
+      keystone_server_url: "https://my-keystone-server.com:<port>"
```
remh: I think all of these settings should be at the instance level

checks.d/openstack.py
```diff
@@ -0,0 +1,827 @@
...
+        # It is still unclear when this happens, but for now the user can configure
+        # `append_tenant_id` to manually add this suffix for downstream requests
+        if instance_config.get("append_tenant_id", False):
+            t_id = auth_scope["project"].get("id")
+
+            assert t_id and t_id not in service_catalog.nova_endpoint,\
+                """Incorrect use of append_tenant_id, please inspect the service catalog response of your Identity server.
+                   You may need to disable this flag if your Nova service url contains the tenant_id already"""
+
+            service_catalog.nova_endpoint += "/{0}".format(t_id)
```
remh: You should use urljoin instead. It's more fault tolerant.

checks.d/openstack.py
```diff
@@ -0,0 +1,827 @@
...
+        identity = {
+            "methods": ['password'],
+            "password": {"user": user}
+        }
+        return identity
+
+    @classmethod
+    def request_auth_token(cls, auth_scope, identity, keystone_server_url, ssl_verify):
+        payload = {"auth": {"scope": auth_scope, "identity": identity}}
+        auth_url = "{0}/{1}/auth/tokens".format(keystone_server_url, DEFAULT_KEYSTONE_API_VERSION)
```
remh: same here, let's use urljoin instead

checks.d/openstack.py
```diff
@@ -0,0 +1,827 @@
...
+        self.neutron_endpoint = neutron_endpoint
+
+    @classmethod
+    def from_auth_response(cls, json_response, nova_api_version):
+        return cls(
+            nova_endpoint=cls.get_nova_endpoint(json_response, nova_api_version),
+            neutron_endpoint=cls.get_neutron_endpoint(json_response)
+        )
+
+        pass
```
remh: not needed

checks.d/openstack.py
```diff
@@ -0,0 +1,827 @@
...
+
+        return None
+
+
+
+    def get_my_hostname(self):
+        """
+        Returns a best guess for the hostname registered with OpenStack for this host
+        """
+        return self.init_config.get("os_host") or socket.gethostname()
```
remh: would the agent hostname be a better guess than `socket.gethostname()` ?

checks.d/openstack.py
```diff
@@ -0,0 +1,827 @@
...
+            "domain_id": project_auth_scope.domain_id
+        }
+
+        url = "{0}/{1}/{2}".format(self.keystone_server_url, DEFAULT_KEYSTONE_API_VERSION, "projects")
+        headers = {'X-Auth-Token': self.get_auth_token(instance)}
+
+        projects = []
+        try:
+            resp = self._make_request_with_auth_fallback(url, headers, params=filter_params)
+            project_details = resp.json()
```
remh: Looks like you're always calling `.json` everytime you call `_make_request_with_auth_fallback` so it should likely be part of the method.

checks.d/openstack.py
```diff
@@ -0,0 +1,827 @@
...
+                            {'project': {'id': 'my_project_id'}}
+                             """)
+            elif isinstance(e, IncompleteIdentity):
+                self.warning("Please specify the user via the `user` variable in your init_config.\n" +
+                             "This is the user you would use to authenticate with Keystone v3 via password auth.\n" +
+                             "The user should look like: {'password': 'my_password', 'name': 'my_name', 'domain': {'id': 'my_domain_id'}}")
+            else:
+                self.warning("Configuration Incomplete! Check your openstack.yaml file")
+
+        else:
```
remh: I think this overall logic will result in cases where not all 3 service checks will send a result.
e.g. if KeystoneUnreachable is reached, only the IDENTIFY_API service check will send a status.

checks.d/openstack.py
```diff
@@ -0,0 +1,827 @@
...
+            # All is good, send the word up to DD
+            self.service_check(self.NETWORK_API_SC, AgentCheck.OK, tags=["keystone_server:%s" % self.init_config.get("keystone_server_url")])
+            self.service_check(self.COMPUTE_API_SC, AgentCheck.OK, tags=["keystone_server:%s" % self.init_config.get("keystone_server_url")])
+            self.service_check(self.IDENTITY_API_SC, AgentCheck.OK, tags=["server:%s" % self.init_config.get("keystone_server_url")])
```
remh: Shouldn't it be keystone_server_url instead of server here ?

remh:
Also looks like metric names are not properly normalized:
e.g. `openstack.nova.limits.maxTotalKeypairs`

this should be camel_cased

conf.d/openstack.yaml.example
talwai: So with the recommendation of having 1 agent configured per physical host, we make the pretty important assumption that there will be exactly 1 hypervisor and 1 keystone server per physical host. It made sense to me to include the project-level config under `instances`, because one physical host can have many projects, but include the "global" configuration in `init_config` since they will be unique to the physical host, and won't change instance-by-instance. Let me know if you disagree

checks.d/openstack.py
```diff
@@ -0,0 +1,850 @@
+# stdlib
+from datetime import datetime, timedelta
+from urlparse import urljoin
+import simplejson as json
```
remh: nitpick but this is 3rd party

remh:
Great stuff! Easy to read and pretty nice features! :+1:

talwai:
CI has been blocked for a while, going to go ahead and merge

PR #1866, opened 2015-08-26
[http][doc] Add docs about the SSL cert. check

elafarge:
The documentation wasn't totally clear about how to monitor the SSL
cert. check in the backend, some customers were wondering why the
expiration warnings weren't triggering on the `http.can_connect` service
check. It will hopefully make more sense now :)

[skip ci]

### DIFF
--- a/conf.d/http_check.yaml.example
+++ b/conf.d/http_check.yaml.example
@@ -77,6 +77,11 @@ instances:
     # The SSL certificate will always be validated for this additional
     # service check regardless of the value of disable_ssl_validation
     #
+    # Be aware that the warning/errors will not be issued to the traditionnal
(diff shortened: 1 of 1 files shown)


conf.d/http_check.yaml.example
```diff
@@ -77,6 +77,11 @@ instances:
     # The SSL certificate will always be validated for this additional
     # service check regardless of the value of disable_ssl_validation
     #
```
yannmh: > create a service check
>be validated for this additional service check regardless of the value of disable_ssl_validation	service check

It sounds quite clear enough to me.

elafarge: Yeah, to me as well but it seems that it wasn't for everybody... I'll delete that and we'll reconsider that decision if support gets more tickets about that part of the check ;-)

PR #1881, opened 2015-09-01
[packaging][rpm] Check the RPM signature

elafarge:
This enables GPG signature check when installing RPM packages from our
one line install script. The key is also downloaded (just the first time
therefore it's rather DNS-spoofing proof).
We'll have to switch all these URLs to HTTPS as soon as it's enabled on
the `yum.datadoghq.com` domain.

[skip ci]

### DIFF
--- a/packaging/datadog-agent/source/install_agent.sh
+++ b/packaging/datadog-agent/source/install_agent.sh
@@ -98,7 +98,7 @@ if [ $OS = "RedHat" ]; then
     else
         ARCHI="x86_64"
     fi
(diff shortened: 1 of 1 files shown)


olivielpeau:
LGTM :+1: Thanks @elafarge !

elafarge:
Well, let's wait for 5.5 before merging anything :)

remh:
So, i'm not sure we should host the public key over plain http. We should wait until https is enabled or use a different location.

irabinovitch:
HTTPS for the public repos is now available.

olivielpeau:
HTTPS has been enabled for a while on our public repos, merging this!

PR #1888, opened 2015-09-03
[core] log collector runs's exceptions

yannmh:
* Create a decorator that catches any exceptions thrown by the decorated
  function and logs them along with a traceback.
* Use it with `Collector.run` method to help troubleshooting.
* Avoid code duplication between Windows and other environments.

### DIFF
--- a/checks/collector.py
+++ b/checks/collector.py
@@ -30,6 +30,7 @@ from util import (
     get_uuid,
     Timer,
 )
+from utils.debug import log_exceptions
 from utils.jmx import JMXFiles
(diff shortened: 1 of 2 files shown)


talwai:
:+1: do we need to add it win32/agent.py as well?

yannmh:
Nop it's applied to the method `Collector.run` whether it's called from `agent.py` or `win32/agent.py`

talwai:
:ship: it

remh:
:+1:

PR #1889, opened 2015-09-03
[wmi_check] Adding TypeError handling

JohnLZeller:


JohnLZeller:
float(None) -> TypeError    # happens when property exists, and is null
float(“”) or float(“dffdsf”) -> ValueError    # happens when property exists, and is not a digit
getattr(“IDoNotExist”) -> AttributeError   # happens when property does not exist

yannmh:
:+1:

Thanks @JohnLZeller

PR #1890, opened 2015-09-03
[redis] `keyspace_hits` metrics: 'gauge' → 'count'

### LINKED ISSUE #1887: Wrong metric type in redis check
the `keyspace_hits` metrics in the redis check are sent as GAUGEs, but should be COUNTERS as their value is always increasing.
screen shot 2015-09-03 at 10 41 01 am
cc @yannmh

yannmh:
**Or #1891**

Switch `redis.stats.keyspace_*` metrics from gauge to
count type.
WARNING: potential backward incompatibility issues.

Fix #1887


remh:
Let's use a rate instead.

PR #1891, opened 2015-09-03
[redis] `keyspace_hits` metrics: 'gauge' → 'rate'

### LINKED ISSUE #1887: Wrong metric type in redis check

yannmh:
**Or #1890**

Switch `redis.stats.keyspace_*` metrics from gauge to
rate type.
WARNING: potential backward incompatibility issues.

Fix #1887


checks.d/redisdb.py
```diff
@@ -205,10 +212,17 @@ def _check_db(self, instance, custom_tags=None):
                     self.gauge(metric, val, tags=db_tags)
 
         # Save a subset of db-wide statistics
-        for k in set(info).intersection(self.GAUGE_KEYS):
-            self.gauge(self.GAUGE_KEYS[k], info[k], tags=tags)
-        for k in set(info).intersection(self.RATE_KEYS):
-            self.rate(self.RATE_KEYS[k], info[k], tags=tags)
+        for info_name, value in info.iteritems():
```
remh: This refactoring isn't improving the readability IMO :/ we should likely keep the previous version.

PR #1892, opened 2015-09-03
[core] Replace subprocess.Popen with safer get_subprocess_output

JohnLZeller:
Issue came up when investigating hostdown issues.
https://datadog.zendesk.com/agent/tickets/30510

It seems as though, periodically, a subprocess call hangs and ultimately causes the watchdog to reset the agent. During this period of time, there is then a host down event, followed by a host recovered event posted to the customers events page. We noticed that it seems as though the hanging happens in various checks (not the same for every host down event). Offending checks have included the customers own custom `iostat` check, as well as our own `postfix` check, and even `resources/processes.py`.

The theory is that this is caused by a hanging call to `subprocess.Popen`. There is a known issue where the use of a `subprocess.PIPE` for stdout (or even stderr) could cause the process to hang once the `subprocess.PIPE` fills up. The solution to this is to use `get_subprocess_output()` in our own `utils/subprocess_output.py`, which uses a file for stdout and stderr, removing such a low memory capacity issue.

To avoid further chasing this problem around the agent, this PR attempts to replace the use of `subprocess.Popen` anywhere it shows up in the agent with `get_subprocess_output()`. There are a few instances where further work will be needed to refactor the code in a clean way, and for those instances, I have added a `FIXME` note.

### DIFF
--- a/checks.d/disk.py
+++ b/checks.d/disk.py
@@ -12,7 +12,7 @@ except ImportError:
 from checks import AgentCheck
 from config import _is_affirmative
 from util import Platform
-import utils.subprocess_output
+from utils.subprocess_output import get_subprocess_output
 
 
 class Disk(AgentCheck):
@@ -176,9 +176,7 @@ class Disk(AgentCheck):
 
     # no psutil, let's use df
     def collect_metrics_manually(self):
-        df_out = utils.subprocess_output.get_subprocess_output(
-            self.DF_COMMAND + ['-k'], self.log
-        )
+        df_out, _, _ = get_subprocess_output(self.DF_COMMAND + ['-k'], self.log)
         self.log.debug(df_out)
         for device in self._list_devices(df_out):
             self.log.debug("Passed: {0}".format(device))
--- a/checks.d/mysql.py
+++ b/checks.d/mysql.py
@@ -1,16 +1,16 @@
 # stdlib
-import subprocess
 import os
 import sys
 import re
 import traceback
 
+# 3p
+import pymysql
+
 # project
 from checks import AgentCheck
 from utils.platform import Platform
-
-# 3rd party
-import pymysql
+from utils.subprocess_output import get_subprocess_output
 
 GAUGE = "gauge"
 RATE = "rate"
@@ -385,11 +385,10 @@ class MySql(AgentCheck):
         if pid is None:
             try:
                 if sys.platform.startswith("linux"):
-                    ps = subprocess.Popen(['ps', '-C', 'mysqld', '-o', 'pid'],
-                                          stdout=subprocess.PIPE, close_fds=True).communicate()[0]
-                    pslines = ps.strip().split('\n')
+                    ps, _, _ = get_subprocess_output(['ps', '-C', 'mysqld', '-o', 'pid'], self.log)
+                    pslines = ps.strip().splitlines()
                     # First line is header, second line is mysql pid
-                    if len(pslines) == 2 and pslines[1] != '':
+                    if len(pslines) == 2:
                         pid = int(pslines[1])
             except Exception:
                 self.log.exception("Error while fetching mysql pid from ps")
--- a/checks.d/postfix.py
+++ b/checks.d/postfix.py
@@ -3,6 +3,7 @@ import os
 
 # project
 from checks import AgentCheck
+from utils.subprocess_output import get_subprocess_output
 
 class PostfixCheck(AgentCheck):
     """This check provides metrics on the number of messages in a given postfix queue
@@ -55,8 +56,8 @@ class PostfixCheck(AgentCheck):
                 # can dd-agent user run sudo?
                 test_sudo = os.system('setsid sudo -l < /dev/null')
                 if test_sudo == 0:
-                    count = os.popen('sudo find %s -type f | wc -l' % queue_path)
-                    count = count.readlines()[0].strip()
+                    output, _, _ = get_subprocess_output(['sudo', 'find', queue_path, '-type', 'f'], self.log)
+                    count = len(output.splitlines())
                 else:
                     raise Exception('The dd-agent user does not have sudo access')
 
--- a/checks.d/varnish.py
+++ b/checks.d/varnish.py
@@ -1,11 +1,11 @@
 # stdlib
 from collections import defaultdict
 import re
-import subprocess
 import xml.parsers.expat # python 2.4 compatible
 
 # project
 from checks import AgentCheck
+from utils.subprocess_output import get_subprocess_output
 
 
 class BackendStatus(object):
@@ -88,7 +88,7 @@ class Varnish(AgentCheck):
         else:
             tags += [u'varnish_name:default']
 
-        output = self._get_varnishstat_output(cmd)
+        output, _, _ = get_subprocess_output(cmd, self.log)
 
         self._parse_varnishstat(output, use_xml, tags)
 
@@ -97,24 +97,13 @@ class Varnish(AgentCheck):
         if varnishadm_path:
             secretfile_path = instance.get('secretfile', '/etc/varnish/secret')
             cmd = ['sudo', varnishadm_path, '-S', secretfile_path, 'debug.health']
-            proc = subprocess.Popen(cmd, stdout=subprocess.PIPE)
-            output, _ = proc.communicate()
+            output, _, _ = get_subprocess_output(cmd, self.log)
             if output:
                 self._parse_varnishadm(output)
 
-    def _get_varnishstat_output(self, cmd):
-        proc = subprocess.Popen(cmd, stdout=subprocess.PIPE,
-                                stderr=subprocess.PIPE)
-        output, error = proc.communicate()
-        if error and len(error) > 0:
-            self.log.error(error)
-        return output
-
     def _get_version_info(self, varnishstat_path):
         # Get the varnish version from varnishstat
-        output, error = subprocess.Popen([varnishstat_path, "-V"],
-                                         stdout=subprocess.PIPE,
-                                         stderr=subprocess.PIPE).communicate()
+        output, error, _ = get_subprocess_output([varnishstat_path, "-V"], self.log)
 
         # Assumptions regarding varnish's version
         use_xml = True
--- a/checks/collector.py
+++ b/checks/collector.py
@@ -32,7 +32,7 @@ from util import (
 from utils.debug import log_exceptions
 from utils.jmx import JMXFiles
 from utils.platform import Platform
-from utils.subprocess_output import subprocess
+from utils.subprocess_output import get_subprocess_output
 
 log = logging.getLogger(__name__)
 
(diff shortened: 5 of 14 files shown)


JohnLZeller:
I have tested these on personal_chef, and fixed any issues that came up.

checks.d/postfix.py
```diff
@@ -55,8 +57,13 @@ def _get_queue_count(self, directory, queues, tags):
                 # can dd-agent user run sudo?
                 test_sudo = os.system('setsid sudo -l < /dev/null')
                 if test_sudo == 0:
-                    count = os.popen('sudo find %s -type f | wc -l' % queue_path)
-                    count = count.readlines()[0].strip()
+                    find = get_subprocess_output(['sudo', 'find', queue_path, '-type', 'f'], self.log)
```
remh: the second command is
```
wc -l
```
so you can change this whole block with:
```python
count = len(get_subprocess_output(['sudo', 'find', queue_path, '-type', 'f'], self.log).splitlines())
```

JohnLZeller:
Reference material: http://thraxil.org/users/anders/posts/2008/03/13/Subprocess-Hanging-PIPE-is-your-enemy/

checks.d/postfix.py
JohnLZeller: aha, clever :) I missed it! *shakes fist at the sky*

config.py
```diff
@@ -581,15 +583,20 @@ def get_system_stats():
     platf = sys.platform
 
     if Platform.is_linux(platf):
-        grep = subprocess.Popen(['grep', 'model name', '/proc/cpuinfo'], stdout=subprocess.PIPE, close_fds=True)
-        wc = subprocess.Popen(['wc', '-l'], stdin=grep.stdout, stdout=subprocess.PIPE, close_fds=True)
-        systemStats['cpuCores'] = int(wc.communicate()[0])
+        grep = get_subprocess_output(['grep', 'model name', '/proc/cpuinfo'], log)
+        # Must use tempfiles to redirect stdout to second command
+        tempgrep = tempfile.TemporaryFile()
```
remh: same thing here

utils/subprocess_output.py
```diff
@@ -8,16 +8,23 @@
...
     if an error occurs.
     """
+
+    # Use tempfile, allowing a larger amount of memory. The subprocess.Popen
+    # docs warn that the data read is buffered in memory. They suggest not to
+    # use subprocess.PIPE if the data size is large or unlimited.
     with tempfile.TemporaryFile('rw') as stdout_f:
-        proc = subprocess.Popen(command, close_fds=True,
-                                stdout=stdout_f, stderr=subprocess.PIPE)
+        stderr_f = tempfile.TemporaryFile('rw')
```
remh: with statements are your friend :)

JohnLZeller:
@remh incorporated all your changes :)

Sorry I sorta jumped the :gun: on squashing the commits.
Here are the changes:
[postfix.py Line 59](https://github.com/DataDog/dd-agent/pull/1892/files#diff-7bb12651742be50bad442a75e07fd0f8R59)
[config.py Line 585](https://github.com/DataDog/dd-agent/pull/1892/files#diff-fdf313ed2303059c44a9ebfffbef8eebR585)
[subprocess_output.py Line 21](https://github.com/DataDog/dd-agent/pull/1892/files#diff-463543c96bdaf50254bcaf9eb154ed3eR21)

All changes have been tested! :+1:

checks.d/mysql.py
```diff
@@ -385,8 +385,7 @@ def _get_server_pid(self, db):
         if pid is None:
             try:
                 if sys.platform.startswith("linux"):
-                    ps = subprocess.Popen(['ps', '-C', 'mysqld', '-o', 'pid'],
-                                          stdout=subprocess.PIPE, close_fds=True).communicate()[0]
+                    ps = get_subprocess_output(['ps', '-C', 'mysqld', '-o', 'pid'], self.log)
```
JohnLZeller: Use splitlines()

utils/subprocess_output.py
```diff
@@ -8,22 +8,29 @@
...
-                                stdout=stdout_f, stderr=subprocess.PIPE)
-        proc.wait()
-        err = proc.stderr.read()
-        if err:
-            log.debug("Error while running {0} : {1}".format(" ".join(command),
-                                                             err))
-
-        stdout_f.seek(0)
-        output = stdout_f.read()
+        with tempfile.TemporaryFile('rw') as stderr_f:
```
remh: http://stackoverflow.com/questions/3024925/python-create-a-with-block-on-several-context-managers

JohnLZeller: aha, I didn't find this post. thanks!

remh:
@JohnLZeller can you fix the conflicts please ?

checks.d/varnish.py
```diff
@@ -97,21 +98,13 @@ def check(self, instance):
...
-        proc = subprocess.Popen(cmd, stdout=subprocess.PIPE,
-                                stderr=subprocess.PIPE)
-        output, error = proc.communicate()
-        if error and len(error) > 0:
-            self.log.error(error)
-        return output
-
     def _get_version_info(self, varnishstat_path):
         # Get the varnish version from varnishstat
+        # FIXME: Use get_subprocess_output() instead of subprocess.Popen
```
remh: Can you do it while we're at it ?

checks/collector.py
```diff
@@ -642,6 +642,7 @@ def _populate_payload_metadata(self, payload, check_statuses, start_event=True):
                     command = "gohai"
                 else:
                     command = "gohai\gohai.exe"
+                # FIXME: Use get_subprocess_output() instead of subprocess.Popen
```
remh: Same here

jmxfetch.py
```diff
@@ -275,6 +275,7 @@ def _start(self, path_to_java, java_run_opts, jmx_checks, command, reporter, too
             log.info("Running %s" % " ".join(subprocess_args))
 
             # Launch JMXfetch subprocess
+            # FIXME: Use get_subprocess_output() instead of subprocess.Popen
```
remh: Same here

JohnLZeller: This one is gonna take some more work in the morning. This function returns the return code of the process, making the return of get_subprocess_output inadequate, unless the caller of this function is okay with either a 0 or 1 return code, representing either an error string that is either empty or not.

JohnLZeller:
Fixed a couple of the FIXME's. Needs a little more love, and some testing.

remh:
Thanks @JohnLZeller can you fix the conflicts please ?

JohnLZeller:
Okay, fixed the other places where there were existing FIXME comments for subprocess, as well as merged master to catch this branch up, and resolved conflicts.

Had to make some changes to the way the `get_subprocess_output()` function works, in order to allow it to work for all the cases we needed it. I did not however end up using it for `jmxfetch.py` because it would have made `get_subprocess_output()` too messy. Instead I implemented a safer way of calling subprocess for `jmxfetch.py`.

JohnLZeller:
Could use some feedback.

@olivielpeau would you mind reviewing half? And @yannmh could you review the other half?

checks.d/disk.py
```diff
@@ -176,9 +176,7 @@ def collect_latency_metrics(self):
 
     # no psutil, let's use df
     def collect_metrics_manually(self):
-        df_out = utils.subprocess_output.get_subprocess_output(
-            self.DF_COMMAND + ['-k'], self.log
-        )
+        df_out, err, rtcode = get_subprocess_output(self.DF_COMMAND + ['-k'], self.log)
```
olivielpeau: Just a general comment when using function that return multiple values: you can discard the values that you don't use with `_`. It's just a convention but makes it clear that you're not using the value afterwards. For instance here you can use:

```python
df_out, _, _ = get_subprocess_output(self.DF_COMMAND + ['-k'], self.log)
```

checks.d/mysql.py
```diff
@@ -385,9 +385,8 @@ def _get_server_pid(self, db):
         if pid is None:
             try:
                 if sys.platform.startswith("linux"):
-                    ps = subprocess.Popen(['ps', '-C', 'mysqld', '-o', 'pid'],
-                                          stdout=subprocess.PIPE, close_fds=True).communicate()[0]
-                    pslines = ps.strip().split('\n')
+                    ps, err, rtcode = get_subprocess_output(['ps', '-C', 'mysqld', '-o', 'pid'], self.log)
+                    pslines = ps.strip().splitlines()
                     # First line is header, second line is mysql pid
                     if len(pslines) == 2 and pslines[1] != '':
```
olivielpeau: Nitpick: Now that we're using `splitlines()` instead of `split("\n")` we don't need to test `pslines[1] != ''` anymore, right? (as `splitlines()` removes the last line when it's empty) :)

utils/subprocess_output.py
```diff
@@ -8,23 +9,29 @@
 
 
 # FIXME: python 2.7 has a far better way to do this
-def get_subprocess_output(command, log):
+def get_subprocess_output(command, log, shell=None, stdin=None):
```
olivielpeau: Probably has no impact at all, but the default value of `shell` in `subprocess.Popen` is `False`, not `None` (see https://docs.python.org/2/library/subprocess.html#subprocess.Popen).

jmxfetch.py
```diff
@@ -274,26 +276,27 @@ def _start(self, path_to_java, java_run_opts, jmx_checks, command, reporter, too
...
-                sys.stdout.write(out)
-                sys.stderr.write(err)
-            else:
-                # Wait for JMXFetch to return
-                jmx_process.wait()
+            # Launch JMXfetch subprocess manually, w/o get_subprocess_output(), since it's a special case
+            with nested(tempfile.TemporaryFile('rw'), tempfile.TemporaryFile('rw')) as (stdout_f, stderr_f):
+                jmx_process = subprocess.Popen(
+                    subprocess_args,
+                    close_fds=not redirect_std_streams,  # set to True instead of False when the streams are redirected for WIN compatibility
```
olivielpeau: Comes from one of my commits actually but the comment here is wrong, it's not `set to True when the streams are redirected` but rather `only set to True when the streams are not redirected`. Could you correct it please? :)

olivielpeau:
Great job, looks way cleaner now!

Looks good to me apart from my comments :+1:

JohnLZeller:
Thanks @olivielpeau :)

Made the changes you mentioned, thanks for the good feedback!

remh:
@JohnLZeller can you squash your commits please ?

JohnLZeller:
@remh will do

JohnLZeller:
@remh okay, I have squashed and rebased to matched up with the current state of master

miketheman:
I'm very interested in trying out some of these methods in a new check, where might I find a recent build that includes this code?

miketheman:
Reason i ask is that http://apt.datadoghq.com/dists/nightly/ has datadog-agent_5.5.0.git.6.ff669c3-1_amd64.deb - which is [back on Sept 17](https://github.com/DataDog/dd-agent/commit/ff669c3).

remh:
http://apt.datad0g.com/dists/nightly/

miketheman:
Thanks @remh. I must have missed that when looking.

PR #1900, opened 2015-09-07
[consul] Send more than the last service check

jgmchan:
At the moment, the consul check will only send the status of the last
service check in `/api/health/state/any`.

This pull request will send the status of all the service checks it can
find.


yannmh:
Thanks a lot for your contribution @jgmchan !
We'll review shorty and be in touch with any feedback.

talwai:
This was a bug, thanks for catching it @jgmchan! Merging

PR #1902, opened 2015-09-09
[pgbouncer] Fix check

remh:
Previously it was assuming that python dicts are ordered which is not
the case. So collected metrics were wrong.

We can’t use python OrderedDict as it was only introduced in python 2.7
and we want to support python 2.6
So here is an implementation using lists instead.


olivielpeau:
LGTM!
CI failure is unrelated.

Issue #1742, opened 2015-06-30
[docker] add labels as tags


tj:
Any interest in supporting a "tags" (or similar) label, or simply grabbing all labels for container metrics? For our case we'd love to have `LABEL program="program-name"` for example and filter that way, even when the `image_repository` tag is available we may have several instances of an image with different configurations so that one will be helpful but it would still be great to have per-container tagging.

cheers

remh:
That's a reasonable idea, it should probably be opt-in though as labels may be containing sensitive data. What do you think @tj ?

remh:
Implemented in https://github.com/DataDog/dd-agent/pull/1824

jippi:
Hi,

I don't see this being resolved in #1824 - do datadog automatically extract all docker labels as tags now ?

remh:
@jippi Hi, no it doesn't. You have to whitelist the labels that you want to be extracted:
https://github.com/DataDog/dd-agent/blob/master/conf.d/docker_daemon.yaml.example#L129-L132

Issue #1739, opened 2015-06-29
[docker] add memory usable


tj:
And/or % in use.

With ECS specifically there's no way to configure swap at the moment so we need to have a good idea of when containers are about to cap out.

dkastner:
It looks like we can get `hierarchical_memory_limit` and 
`hierarchical_memsw_limit1`  from `CGROUP_METRICS` https://github.com/DataDog/dd-agent/blob/master/checks.d/docker.py#L19

Can we just collect those raw values and calculate a percentage when graphing or monitoring from datadog, or does the custom check need to calculate a percentage of memory available and log it as a percentage metric?

remh:
Implemented in https://github.com/DataDog/dd-agent/commit/96c0305c13fdc8dfa004e551466bb1638873f71e

dkastner:
:beers:

PR #1908, opened 2015-09-11
[Docker check] V2

### LINKED ISSUE #1299: [docker] TLS support in docker connection / docker-py
The recommended way of running docker is with TLS support:

    export DOCKER_HOST=tcp://foobar.com:2376
    export DOCKER_TLS_VERIFY=1
    export DOCKER_CERT_PATH=/etc/tls/docker

It seems that there is no way for dd-agent to pick up those env variables, the urllib call therefore fails to establish a connection.

remh:
This is a new version of the docker check, it's called docker_daemon (name can be changed).
The previous check "docker" is now deprecated and won't receive further support.

In terms of features, this adds:
- Support for TLS connections to the daemon
- New metrics: 
    - Network metrics
    - Memory limits
    - Container size (rootfs)
    - Image size
- Support for labels (convert them into tags). Off by default, uses a list of labels that should be converted.
- Support for ECS tags: task name and task version

Backward incompatible changes:

    - docker.disk.size metric is renamed to docker.container.size_rw
    - Old optional metrics: https://github.com/DataDog/dd-agent/blob/5.4.x/checks.d/docker.py#L29-L38 Are not collected anymore
    - Old old tags are not supported anymore (e.g. `name` instead of container_name)

fix #1299 #1648 #1739 #1742


checks.d/docker_daemon.py
```diff
@@ -0,0 +1,640 @@
...
+        "file": 'blkio.throttle.io_service_bytes',
+        "metrics": {
+            "io_read": ("docker.io.read_bytes", RATE),
+            "io_write": ("docker.io.write_bytes", RATE),
+        },
+    },
+]
+
+DEFAULT_CONTAINER_TAGS = [
+    "docker_image",
```
LotharSee: Do we want to have `docker_image` as a default tag?
We wanted to replace it with the combo `image_name` / `image_tag` (same cardinality but much more flexible), and kept `docker_image` just for compatibility (as said in the config file).
Did we change our mind?

remh: I think it makes sense to keep it so we don't break everything (dashboards monitors) and as you said it's the same cardinality as image_name / image_tag so it's not creating more contexts, i think it's a good compromise.

hkaj: We had customer requests for that though. Some people want to be able to filter on image_name because they run several tags of an image.

checks.d/docker_daemon.py
```diff
@@ -0,0 +1,640 @@
...
+            self.tag_names = {
+                CONTAINER: instance.get("container_tags", DEFAULT_CONTAINER_TAGS),
+                PERFORMANCE: instance.get("performance_tags", DEFAULT_PERFORMANCE_TAGS),
+                IMAGE: instance.get('image_tags', DEFAULT_IMAGE_TAGS)
+
+            }
+
+            # Set filtering settings
+            if not instance.get("exclude"):
+                self._filtering_enabled = False
```
talwai: Since it seems `include` is meaningless without `exclude`, should we show a warning for the case `instance.get('include') and not instance.get('exclude')`? I can see that being a common mistake

checks.d/docker_daemon.py
```diff
@@ -0,0 +1,640 @@
...
+        self._report_performance_metrics(containers_by_id)
+
+        if self.collect_container_size:
+            self._report_container_size(containers_by_id)
+
+        # Send events from Docker API
+        if self.collect_events:
+            self._process_events(containers_by_id)
+
+    def _count_and_weight_images(self):
```
talwai: typo: s/weight/weigh

checks.d/docker_daemon.py
```diff
@@ -0,0 +1,640 @@
...
+                if re.match(rule, tag):
+                    return True
+        return False
+
+    def _is_container_excluded(self, container):
+        """Check if a container is excluded according to the filter rules.
+
+        Requires _filter_containers to run first.
+        """
+        return container.get('_is_filtered', False)
```
talwai: I'm not sure I like that `_get_and_count_containers` is opaquely mutating the container dict in a way that subsequent functions rely on (e.g. `_report_container_size`). Can we have `_is_container_excluded` actually call `_filter_containers` if the key is missing, and hence always return an accurate value? It would make debugging checks a lot easier imo, not to mention that it makes the function name actually truthful

checks.d/docker_daemon.py
```diff
@@ -0,0 +1,640 @@
...
+                            self.log.debug("Couldn't compute {0}, some keys were missing.".format(mname))
+                            continue
+                        value = fct(*values)
+                        if value is not None:
+                            metric_func(self, mname, value, tags=tags)
+
+
+
+        except MountException as ex:
+            if self.cgroup_listing_retries > 3:
```
talwai: The threshold of 3 seems a good candidate for a class variable

checks.d/docker_daemon.py
```diff
@@ -0,0 +1,640 @@
...
+                raise ex
+            else:
+                self.warning("Couldn't find the cgroup files. Skipping the CGROUP_METRICS for now."
+                             "Will retry a few times before failing.")
```
talwai: We should replace "a few" with the actual number of additional retries we're going to do. No reason to be vague when we know the exact value

conf.d/docker_daemon.yaml.example
```diff
@@ -0,0 +1,131 @@
...
+    # exclude: [".*"]
+    # include: ["docker_image:ubuntu", "docker_image:debian"]
+    #
+    # include all, except ubuntu and debian.
+    # exclude: ["docker_image:ubuntu", "docker_image:debian"]
+    # include: []
+    #
+    # Default: include all
+    # exclude: []
+    # include: []
```
talwai: I think it's a bit weird that when `exclude = []` it implies exclude **none**, but when `include = []` it implies include **all**. I know that our code actually defaults include to `[]` when it is absent so that it is falsy. But I wonder if the doc needs to leak this implementation detail. Probably just a matter of opinion

conf.d/docker_daemon.yaml.example
```diff
@@ -0,0 +1,130 @@
...
+instances:
+  - ## Daemon and system configuration
+    ##
+
+    # URL of the Docker daemon socket to reach the Docker API. HTTP/HTTPS also works.
+    # Warning: if that's a non-local daemon, we won't be able to collect performance metrics.
+    #
+    url: "unix://var/run/docker.sock"
+
+    # Cgroup root (HIDDEN OPTION?)
```
hkaj: We should remove it since we auto-detect the path. If a new platform with a different mount point arises we'll add its path in dockerutil.py so that it's supported out of the box anyway.

conf.d/docker_daemon.yaml.example
```diff
@@ -0,0 +1,130 @@
...
+    # Default: None (auto-detected)
+    #
+    # cgroup_path:
+
+    # Cgroup layout (HIDDEN OPTION?)
+    # How cgroup files are organized inside the cgroup root.
+    # Default: None (auto-detected)
+    # Example: "/%(cgroup_root)s/docker/%(container_id)s/%(file)s"
+    #
+    # cgroup_layout: None
```
hkaj: same here

tests/checks/integration/test_docker_daemon.py
```diff
@@ -0,0 +1,439 @@
...
+                "exclude": ["docker_image:nginx"],
+                "collect_image_size": True,
+            },
+            ],
+        }
+
+        self.run_check_twice(config, force_reload=True)
+
+        for mname, tags in expected_metrics:
+            self.assertMetric(mname, tags=tags, count=1, at_least=1)
```
hkaj: shouldn't we check explicitly that the excluded metrics are not there?

tests/checks/integration/test_docker_daemon.py
```diff
@@ -0,0 +1,439 @@
...
+                "url": "unix://var/run/docker.sock",
+                "collect_container_size": True,
+                "collect_image_size": True,
+            },
+            ],
+        }
+
+        self.run_check(config, force_reload=True)
+        for mname, tags in expected_metrics:
+            self.assertMetric(mname, tags=tags, count=1, at_least=1)
```
hkaj: no coverage_report? :(

remh: We can't do coverage report as the test can't run on travis or circleci. I run the test locally on a machine that have more containers than the one that are spawned.

tests/checks/integration/test_docker_daemon.py
remh: done

talwai:
:+1: :confetti_ball:

conf.d/docker.yaml.example
```diff
@@ -67,6 +67,6 @@ instances:
     # collect_all_metrics: false
 
     # Collect images stats
-    # Number of available active images and intermediate images as gauges. Default: true.
+    # Number of available active images and intermediate images as gauges. Default: false.
```
LotharSee: It looks like the default in the code is still `True`.

conf.d/docker_daemon.yaml.example
```diff
@@ -0,0 +1,115 @@
...
+    # ensure that `docker ps -a -q` run fast before enabling it.
+    # Defaults to false.
+    #
+    # collect_container_size: false
+
+    # Collect images stats
+    # Number of available active images and intermediate images as gauges.
+    # Defaults to true.
+    #
+    collect_images_stats: false
```
LotharSee: Since we start from fresh, let's make it a real default value (default to `False` in the code + in this file).

conf.d/docker_daemon.yaml.example
LotharSee: The point about this option was to allow users on a new platform/with a new cgroup pattern to configure the Agent, without waiting for a new Agent release.
That's pretty minor, maybe not worth spending the time to add this feature,

checks.d/docker_daemon.py
```diff
@@ -0,0 +1,666 @@
...
+    def _crawl_container_pids(self, container_dict):
+        """Crawl `/proc` to find container PIDs and add them to `containers_by_id`."""
+        proc_path = os.path.join(self._docker_root, 'proc')
+        pid_dirs = [_dir for _dir in os.listdir(proc_path) if _dir.isdigit()]
+
+        if len(pid_dirs) == 0:
+            self.warning("Unable to find any pid directory in {0}. "
+                "If you are running the agent in a container, make sure to "
+                'share the volume properly: "/proc:/host/proc:ro". '
+                "Network metrics will be missing".format(proc_path))
```
LotharSee: Maybe link directly to the (up-to-date) documentation directly?

PR #1910, opened 2015-09-11
[dogstream] Add debug logs

JohnLZeller:
Currently dogstream does not print debug logs about events


JohnLZeller:
Removed the addition of `check_freq`, as it's not just for dogstream. There are multiple other parts of the agent, including the watchdog, which search for this config. From what I can see the default for `check_freq` is 15, which is in seconds, NOT minutes.

PR #1912, opened 2015-09-14
[darwin][iostat] Fix bad conversion

### LINKED ISSUE #1718: Incorrect conversion of system.io.bytes_per_s metric
Incorrect conversion between megabytes and bytes [HERE](https://github.com/DataDog/dd-agent/blob/09249df7a895c534e6150ca066f544ba72606354/checks/system/unix.py#L75).

Also in the test file: tests/checks/mock/test_sysstat.py

```
'system.io.bytes_per_s': mb_s * 10**6
```
Should be
```
'system.io.bytes_per_s': mb_s * 2**20
```

I'm assuming here that we always mean Mebibytes when we say Megabytes. When metrics metadata is fully implemented the plan is to completely shift to the unambiguous kibi, mebi, gibi byte prefixes and no longer refer to the ambiguous kilo, mega, giga scientific prefixes for byte units.

remh:
Fix #1718

### DIFF
--- a/checks/system/unix.py
+++ b/checks/system/unix.py
@@ -72,7 +72,7 @@ class IO(Check):
         for idx, disk in enumerate(disks):
             kb_t, tps, mb_s = map(float, lastline[(3 * idx):(3 * idx) + 3])  # 3 cols at a time
             io[disk] = {
-                'system.io.bytes_per_s': mb_s * 10**6,
+                'system.io.bytes_per_s': mb_s * 2**20,
             }
         return io
 


checks/system/unix.py
```diff
@@ -72,7 +72,7 @@ def _parse_darwin(self, output):
         for idx, disk in enumerate(disks):
             kb_t, tps, mb_s = map(float, lastline[(3 * idx):(3 * idx) + 3])  # 3 cols at a time
             io[disk] = {
-                'system.io.bytes_per_s': mb_s * 10**6,
+                'system.io.bytes_per_s': mb_s * 2**20,
```
LeoCavaille: `mb_s << 20` ?

remh: mb_s is a float, so that wouldn't work but yeah i agree it would be better otherwise.

PR #1914, opened 2015-09-14
[nginx] Log exception message

### LINKED ISSUE #1813: [nginx] No debug logging when doing Pokémon exception handling
The [nginx code](https://github.com/DataDog/dd-agent/blob/49952b4429b125619bc8d9f51bb6564e7c0d2e12/checks.d/nginx.py#L45-L50) follows a [pattern of catching all exceptions](http://blog.codinghorror.com/new-programming-jargon/) when attempting to submit a metric, which makes sense due to wanting to continue without error. Unfortunately, it additionally eats the real error message and displays a super generic message in the logs. This makes it very difficult to actually debug issues.

I ended up manually modifying the code on my local install to figure out that my `nginx.yaml` had a syntax error in it. It would have been more obvious if the true exception (`unhashable type: 'dict'`) had been printed out. (Additionally, there should probably be error checking when tags are loaded instead of reported.)

I'm willing to fix either or both of these issues, but would like advice on how you would like to see them done (e.g. log at a different level? Log to stderr?). Thanks!

remh:
Fix #1813

### DIFF
--- a/checks.d/nginx.py
+++ b/checks.d/nginx.py
@@ -49,8 +49,8 @@ class Nginx(AgentCheck):
                 name, value, tags, metric_type = row
                 func = funcs[metric_type]
                 func(name, value, tags)
-            except Exception:
-                self.log.error(u'Could not submit metric: %s' % repr(row))
+            except Exception, e:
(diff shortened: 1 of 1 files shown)


yannmh:
Only a few flake8 offenses to fix and it's good to go

PR #1917, opened 2015-09-15
[wmi] new AgentCheck for WMI metrics

yannmh:
Introducing a new AgentCheck for WMI metrics:
`wmi_alternative_check.py`.

Contrary to the previous `wmi_check.py` AgentCheck, it does not use
`wmi` Python package but directly relies on `win32com.client`: as a
result, queries are about 10 times faster !

It uses the same YAML configuration template file as the previous check,
but does not support the following options yet:
* `constant_tags`
* `tag_queries`

### DIFF
--- a/checks.d/wmi_alternative_check.py
+++ b/checks.d/wmi_alternative_check.py
@@ -1,0 +1,222 @@
+# stdlib
+from collections import namedtuple
+
+# 3rd
+import pywintypes
+from win32com.client import Dispatch
+
+# project
+from checks import AgentCheck
+
+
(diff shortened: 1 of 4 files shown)


TheCloudlessSky:
:+1:

I'm just starting out with Datadog and have been in contact with support that we've run into performance problems on Windows just with the *default checks* (since they use the `wmi` Python package). Hopefully this can be ported to the default checks too!

yannmh:
Closing in favor of https://github.com/DataDog/dd-agent/pull/2011

PR #1924, opened 2015-09-16
[Windows] Have a single instance of Agent Manager

elafarge:
When the GUI is started:
    - It checks for a Pidfile left by a previously running (and potentially still running) GUI
    - If such a Pidfile was found and if the Pid inside is file is indeed the Pid of a running instance of `agent-manager.exe`, then let's kill that old process
    - Let's write the Pid of the current process to that pidfile

Nothing else was changed :)

### DIFF
--- a/win32/gui.py
+++ b/win32/gui.py
@@ -92,8 +92,16 @@ AGENT_UNKNOWN = 4
 # Windows management
 # Import Windows stuff only on Windows
 if Platform.is_windows():
+    import win32api
+    import win32con
+    import win32process
     import win32serviceutil
     import win32service
+
+    # project
+    from utils.pidfile import PidFile
+    from utils.process import pid_exists
+
     WIN_STATUS_TO_AGENT = {
         win32service.SERVICE_RUNNING: AGENT_RUNNING,
         win32service.SERVICE_START_PENDING: AGENT_START_PENDING,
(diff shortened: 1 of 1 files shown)


win32/gui.py
```diff
@@ -791,7 +799,53 @@ def info_popup(message, parent=None):
     QMessageBox.information(parent, 'Message', message, QMessageBox.Ok)
 
 
+def kill_old_process():
+    """ Kills or brings to the foreground (if possible) any other instance of this program. It
+    avoids multiple icons in the Tray on Windows. On OSX, we don't have to do anything: icons
+    don't get duplicated. """
+    # On Windows, if a window is already opened, let's just bring it on the foreground
+    if Platform.is_windows():
```
hkaj: Nitpicking but you could check that before calling the function. That would avoid opening a new context for nothing if the platform is not Windows (and wrapping the whole thing in a big if).

win32/gui.py
```diff
@@ -791,7 +799,53 @@ def info_popup(message, parent=None):
...
+        pidfile = PidFile('agent-manager-gui').get_path()
+
+        old_pid = None
+        try:
+            pf = file(pidfile, 'r')
+            old_pid = int(pf.read().strip())
+            pf.close()
+        except IOError:
+            pass
+        except ValueError:
```
hkaj: or:
```python
except (IOError, ValueError):
    pass
```

hkaj:
Great success @elafarge! Except for a few nitpick I don't have much to say. Feel free to merge once you've considered the 2 comments.

elafarge:
Thanks @hkaj :) I took your remarks into account. I'm right now building a Windows version of the Agent to run some tests. If it still works as expected, I'll merge that once and for all ;-)

elafarge:
@yannmh I was wrong, since we're using the 5.5.x branch of the Agent, this is not gonna be shipped with the 5.5.0 by default. It can be an interesting set commits to cherrypick for the next bugfix release though ;-)

tmichelet:
@elafarge don't forget to delete branches once merged

win32/gui.py
```diff
@@ -791,7 +799,52 @@ def info_popup(message, parent=None):
     QMessageBox.information(parent, 'Message', message, QMessageBox.Ok)
 
 
+def kill_old_process():
+    """ Kills or brings to the foreground (if possible) any other instance of this program. It
```
yannmh: > or brings to the foreground (if possible) any other instance of this program

It looks like you actually always kill the previous instance in fact.

win32/gui.py
yannmh: You are testing `Platform.is_windows()` two times.

yannmh:
I know it has been merged, but I think killing the process rather than bringing the UI back is dangerous: for instance, any change made to a configuration file without saving would be lost.

elafarge:
Oh yeah, actually it doesn't kill the process if the Window is opened, at least not on Windows server 2012: if an application is in the taskbar, starting it again from the launch menu will just bring the window to the front. I don't know if that behaviour is the same on Windows 7 likes (ie Windows server 2008).

elafarge:
@yannmh ^^

Issue #1660, opened 2015-06-05
[Process check] CPU Times require an interval


remh:
psutil lets the possibility to not specify an interval for the process cpu time:
http://pythonhosted.org/psutil/#psutil.Process.cpu_times

However there are a few issues:
- We assert that the cpu_check_interval so we can't set it to None: https://github.com/DataDog/dd-agent/blob/cc5a67bb47110ef145e58561ab4471b946354742/checks.d/process.py#L233

- The way psutil works, is that it will save the previous cpu times in the Process object, however, we recreate a new object at each iteration: https://github.com/DataDog/dd-agent/blob/cc5a67bb47110ef145e58561ab4471b946354742/checks.d/process.py#L174 which means that we lose these values

As a result the only way to get cpu metrics for a process is to set a valid cpu_interval, which will be blocking, and if it's too low, it will be pretty irrelevant.

We shouldn't recreate a Process object at every iteration if the pid of the process we want to monitor hasn't changed. It would be the simplest way to fix this issue.

remh:
@olivielpeau

olivielpeau:
Fixed by #1928

PR #1928, opened 2015-09-18
[process] Improve sampling of cpu.pct

olivielpeau:
See https://github.com/DataDog/dd-agent/issues/1660

Improvements:
- Non-blocking call to `cpu_percent` after the first sample
- After the first sample, retrieves an averaged value since the last call (instead of an averaged value over the fraction of a second defined by `cpu_check_interval`)

The first sample of `cpu.pct` on every process still uses a blocking call so that the check can retrieve a value as soon as it finds new processes. Setting the `cpu_check_interval` to `0` removes that first blocking call.

Also fixed the handling of the cache of AccessDenied PIDs (it was previously refreshed at every run of every instance).

### DIFF
--- a/checks.d/process.py
+++ b/checks.d/process.py
@@ -60,6 +60,9 @@ class ProcessCheck(AgentCheck):
             )
         )
 
+        # Process cache, indexed by instance
+        self.process_cache = {}
+
     def should_refresh_ad_cache(self):
         now = time.time()
         return now - self.last_ad_cache_ts > self.access_denied_cache_duration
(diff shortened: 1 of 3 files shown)


yannmh:
> Setting the cpu_check_interval to 0 removes that first blocking call.

As we discussed, I think that's the right thing to do.

olivielpeau:
Thanks @yannmh for the review!

I've implemented the changes you suggested (see commit message of my last commit: da212b8).

yannmh:
Nice improvement. Looks good to merge to me ! :ship:

olivielpeau:
Thanks again for the thorough review ;)

I've rebased the PR and changed the Exception type tested in `test_ad_cache` (per your comment), I'll merge once the CI passes.

olivielpeau:
The CI is failing on unrelated tests, merging.

remh:
:+1:

PR #1929, opened 2015-09-18
OpenStack integration

mtougeron:
Hi @talwai, I was talking with @2bethere and @irabinovitch about using the OpenStack integration here at Lithium. I needed to make a few changes to your script in order to get it to work. Here's a PR that will merge my changes into your branch. I'm not sure if that's something you want or not so feel free to reject it. I just wanted to share up my changes. This branch is based on PR #1864 

In addition to the policy changes listed https://gist.github.com/talwai/7445f16c2584b8caffe2 I also modified `compute_extension:v3:os-hypervisors`, `compute_extension:v3:os-server-diagnostics`, `compute_extension:availability_zone:detail`, `compute_extension:v3:os-availability-zone:detail`.

Key changes:
- Can, assuming permissions allow, dynamically get the list of hypervisors and networks to monitor
- Better handling of the uptime api calls for when a hypervisor's nova-compute service is offline.
- Nova pre v2.1 does not support the 'state' field in the hypervisor data so I'm falling back to check uptime.
- Some versions of the Keystone api requires the content-type application/json header to be sent.
- In the service catalog, use the 'public' urls instead of just the first one.

### DIFF
--- a/checks.d/openstack.py
+++ b/checks.d/openstack.py
@@ -159,7 +159,11 @@ class OpenstackCheck(AgentCheck):
             return None
 
     def get_network_stats(self):
-        network_ids = self.init_config.get('network_ids', [])
+        if self.init_config.get('check_all_networks', False):
+            network_ids = self.get_all_network_ids()
+        else:
+            network_ids = self.init_config.get('network_ids', [])
+
         if not network_ids:
             self.warning("Your check is not configured to monitor any networks.\n" +
                          "Please list `network_ids` under your init_config")
@@ -167,6 +171,20 @@ class OpenstackCheck(AgentCheck):
         for nid in network_ids:
             self.get_stats_for_single_network(nid)
 
+    def get_all_network_ids(self):
+        url = '{0}/{1}/networks'.format(self._neutron_url, self.DEFAULT_NEUTRON_API_VERSION)
+        headers = {'X-Auth-Token': self._auth_token}
+
+        network_ids = []
+        try:
+            resp = self._make_request_with_auth_fallback(url, headers)
+            net_details = resp.json()
+            for network in net_details['networks']:
+                network_ids.append(network['id'])
+        except:
+            self.warning('Unable to get the list of all network ids.')
+
+        return network_ids
 
     def get_stats_for_single_network(self, network_id):
         url = '{0}/{1}/networks/{2}'.format(self._neutron_url, self.DEFAULT_NEUTRON_API_VERSION, network_id)
--- a/checks.d/openstack.py
+++ b/checks.d/openstack.py
@@ -233,7 +233,10 @@ class OpenstackCheck(AgentCheck):
             return None
 
     def get_hypervisor_stats(self):
-        hypervisors = self.init_config.get('hypervisor_ids', [])
+        if self.init_config.get('check_all_hypervisors', False):
+            hypervisors = self.get_all_hypervisor_ids()
+        else:
+            hypervisors = self.init_config.get('hypervisor_ids', [])
 
         if not hypervisors:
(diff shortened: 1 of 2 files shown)


talwai:
Thanks for the updates @mtougeron ! At first glance, these look like very compelling (and necessary) improvements to the check as it stands. Will give this a more thorough review in the next day or so

mtougeron:
Sounds good. We've been running it in production since last night and so far all is well. :)

mtougeron:
@talwai I did have one question about the stats, the `openstack.nova.hypervisor.vcpus` is the # of free vCPUs for that hypervisor. I know that name matches what comes back from the API call but I wonder if perhaps it should be renamed to `openstack.nova.hypervisor.vcpus_free` to avoid confusion? I'm kind of on the fence with that one.

checks.d/openstack.py
```diff
@@ -206,12 +244,34 @@ def get_hypervisor_stats(self):
...
-            }
+            stats[hyp] = {}
+            try:
+                stats[hyp]['payload'] = self.get_stats_for_single_hypervisor(hyp)
+            except:
+                self.warning('Unable to get stats for hypervisor {0}.'.format(hyp))
+
+            try:
+                stats[hyp]['uptime'] = self.get_uptime_for_single_hypervisor(hyp)
+            except:
```
talwai: Do we know enough about the failure modes of these functions to catch named `Exception`s here? My guess is we don't, in which case maybe it's prudent to log the stringified `Exception` in the warning message so that we know what went wrong.

Ditto for the other bare `except`s in this file.

checks.d/openstack.py
```diff
@@ -225,9 +285,30 @@ def get_stats_for_single_hypervisor(self, hyp_id):
...
-        if hyp['state'] != self.HYPERVISOR_STATE_UP:
+        hyp_state = hyp.get('state', None)
+        if hyp_state is None:
+            try:
+                # Fall back for pre Nova v2.1 to the uptime response
+                uptime = self.get_uptime_for_single_hypervisor(hyp_id)
+                if uptime.get('uptime_sec', 0) > 0:
+                    hyp_state = self.HYPERVISOR_STATE_UP
+                else:
+                    hyp_state = 'down'
```
talwai: Would be nice to have `self.HYPERVISOR_STATE_DOWN` here for consistency

checks.d/openstack.py
```diff
@@ -167,7 +192,11 @@ def get_stats_for_single_network(self, network_id):
         resp = self._make_request_with_auth_fallback(url, headers)
 
         net_details = resp.json()
-        service_check_tags = ['network:{0}'.format(network_id)]
+        service_check_tags = [
+            'network:{0}'.format(network_id),
+            'network_name:{0}'.format(net_details.get('network', {}).get('name')),
+            'tenant_id:{0}'.format(net_details.get('network', {}).get('tenant_id'))
```
talwai: If `network_name` or `tenant_id` is missing a significant amount of time, we should maybe check beforehand that the metadata we need exists, rather than adding tags like`network_name:None`.

talwai:
@mtougeron looks (and works) great! Pointed out a few minor things in the code, let me know what you think. After that, happy to merge this in to my branch and continue development from there

mtougeron:
Great, thanks! I'll address these this morning.

checks.d/openstack.py
mtougeron: I can't remember why I did it this way. I may have just copy/pasted from a different location. I'll clean it up.

mtougeron:
Pushed the updates based on your PR feedback.

talwai:
Great, thanks for the quick response! Merging this.

PR #1932, opened 2015-09-18
[docker] Fix the ecs tags fetching

hkaj:


checks.d/docker_daemon.py
```diff
@@ -364,8 +364,8 @@ def refresh_ecs_tags(self):
         ecs_tags = {}
         if net_conf:
             net_conf = net_conf[0] if isinstance(net_conf, list) else net_conf
-            ip, port = net_conf.get('HostIp'), net_conf.get('HostPort')
-            tasks = requests.get('http://%s:%s' % (ip, port)).json()
+            ip, port = ecs_config['NetworkSettings']['IPAddress'], net_conf.get('HostPort')
```
remh: You don't want use the HostPort here as it's, well, the host port :)
You want to use the container port instead which is in the key of this dictionary:
```python
{u'51678/tcp': [{u'HostPort': u'51678', u'HostIp': u'127.0.0.1'}]}
```

remh:
Thanks for fixing this, added a small comment

PR #1936, opened 2015-09-21
[couchbase] Only send selected bucket-level metrics

olivielpeau:
Trims down the number of metrics retrieved per bucket from ~250 to ~90.

Very simple approach for now. I've selected the metrics based on the little information that is available in the docs (http://developer.couchbase.com/documentation/server/4.0/monitoring/monitoring-rest.html), and on the metrics that are currently being used on monitors/screenboards/dashboards.

We could also limit the number of buckets on which stats are collected (and allow listing these buckets in the yaml file), but given that this check is not used that much I'm not sure it's worth it.

### DIFF
--- a/checks.d/couchbase.py
+++ b/checks.d/couchbase.py
@@ -19,6 +19,86 @@ class Couchbase(AgentCheck):
     """
     SERVICE_CHECK_NAME = 'couchbase.can_connect'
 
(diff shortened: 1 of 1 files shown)


olivielpeau:
Thanks @yannmh for the mistake you pointed out, I've fixed it ;)

PR #1941, opened 2015-09-22
Measure user functions in postgres

alq666:
In order to track PL/pgSQL functions (in particular), we extract metrics from `pg_stat_user_functions`.

# Metrics

All tagged by `schema` and `function`.

* `postgresql.function.calls`: number of calls per second of a given function
* `postgresql.function.total_time`: ratio of time spent executing a given function (and its sub-functions) per second elapsed
* `postgresql.function.self_time`: same ratio but excludes sub-function calls.

See it in action [here](https://dd.datad0g.com/dash/5334/plsql?live=true&page=0&is_auto=false&from_ts=1442944113379&to_ts=1442947713379&tile_size=m)

# Questions

The "time" metrics are unit-less. Since postgres collects these metrics as `COUNTER`s, I am turning them into rates. Ideally we would send this as they are and have the backend compute the rate. This would through the REST API, but not dogstatsD.

Suggestions to make this more obvious are welcome.


remh:
@alq666 to make the metric times more relevant we could compute the average time per request.

e.g.: postgresql.function.avg_totaltime = (new_total_time - old_total_time) / (new_calls - old_calls)

It would require a bit of refactoring though as the way we are defining scopes doesn't support custom computation.

alq666:
@remh we can achieve the same with `postgresql.function.total_time / postgresql.function.calls`, under the assumption that both metrics are always collected at the same time.

isaacdd:
@alq666 I think you want to use a query like this one to query the stats, because `pg_stat_user_functions` does not have a way distinguish between overloaded functions.

```
with overloaded_funcs as (
    select funcname
      from pg_stat_user_functions s
  group by s.funcname
    having count(*) > 1
)
select s.schemaname,
       case when o.funcname is null then p.proname
            else p.proname || '_' || array_to_string(p.proargnames, ',')
        end funcname,
        %s
    from pg_proc p
    join pg_stat_user_functions s
      on p.oid = s.funcid
    left join overloaded_funcs o
      on o.funcname = s.funcname;
```

alq666:
Merged @isaacdd's suggestion.

remh:
superseded by #2164

PR #1943, opened 2015-09-22
[nginx] Fix 'json' content_type support :bug:

yannmh:
Use JSON parser to extract Nginx metrics when `content_type` *starts
with* (rather than *equals*) `application/json`.
Also add DEBUG level logs to help troubleshooting.

Thanks @tmichelet for investigating.


tmichelet:
:+1:

PR #1947, opened 2015-09-23
[process][doc] Add comment on cpu.pct for short-lived processes

olivielpeau:
PR #1928 improved the sampling of `cpu.pct` but its value
is not accurate for short-lived processes (it was even less
accurate before that PR but it's still worth mentioning).

### DIFF
--- a/conf.d/process.yaml.example
+++ b/conf.d/process.yaml.example
@@ -5,6 +5,9 @@ init_config:
   # pid_cache_duration: 120
 
 instances:
+# The `system.processes.cpu.pct` metric sent by this check is only accurate for processes that live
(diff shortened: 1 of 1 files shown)


yannmh:
I know that's what we currently do for most of our checks, but I am not a huge fan of adding notes and comments in the YAML configuration file. They looks better short and concise :art:.

Maybe, in the future, we could add this type of note (related to a specific metric) in the new 'Metrics' section of an integration tile
k1jonb71lu2imyfooae8iffq_wditel_vulgw_ixasy

olivielpeau:
I couldn't agree more. We should find a way to give the same kind of info for checks that don't have an integration tile (or is that already planned?).

PR #1948, opened 2015-09-24
[activemq_xml] limiting queues in config should work

joelvanvelden:
Specifying queue names in config will now correctly limit metrics to
only those queues/topics/subscribers.
If you have lots of queues this is a way to monitor a subset of them.
Also needed if over the 300 max metrics threshold.

### DIFF
--- a/checks.d/activemq_xml.py
+++ b/checks.d/activemq_xml.py
@@ -68,6 +68,9 @@ class ActiveMQXML(AgentCheck):
     def _process_data(self, data, el_type, tags, max_elements, detailed_elements):
         root = ElementTree.fromstring(data)
         elements = [e for e in root.findall(el_type) if e.get("name")]
+	# if list provided in config, only send those metrics
+	if detailed_elements:
+		elements = [e for e in elements if e.get('name') in detailed_elements]
         count = len(elements)
 
         if count > max_elements:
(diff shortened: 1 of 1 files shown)


yannmh:
Thanks for your contribution @joelvanvelden.

 We'll review your PR very soon, so it can get released with the 5.6.0.

checks.d/activemq_xml.py
```diff
@@ -68,6 +68,9 @@ def _fetch_data(self, base_url, xml_url, username, password):
     def _process_data(self, data, el_type, tags, max_elements, detailed_elements):
         root = ElementTree.fromstring(data)
         elements = [e for e in root.findall(el_type) if e.get("name")]
```
remh: I think this bit should be rewritten in 
```python
    if detailed_elements:
        elements = [e for e in root.findall(el_type) if e.get("name") in detailed_elements]
    else:
         elements = [e for e in root.findall(el_type) if e.get("name")]
```
It would be more efficient.

checks.d/activemq_xml.py
```diff
@@ -98,6 +99,9 @@ def _process_data(self, data, el_type, tags, max_elements, detailed_elements):
     def _process_subscriber_data(self, data, tags, max_subscribers, detailed_subscribers):
         root = ElementTree.fromstring(data)
         subscribers = [s for s in root.findall("subscriber") if s.get("clientId")]
```
remh: Same here:
```python
if detailed_subscribers:
    subscribers = [s for s in root.findall("subscriber") if s.get("clientId") in detailed_subscribers]
else:
    subscribers = [s for s in root.findall("subscriber") if s.get("clientId")]
```

remh:
Thanks for this contribution @joelvanvelden ! 
It looks good besides the 2 nitpicks mentioned above!

joelvanvelden:
@remh, i've added your suggestions.

PR #1949, opened 2015-09-25
Add a command "predicate" to allow runtime blackholing of metrics.

gphat:
# What's this PR do?

Adds a configuration option `flush_metrics_predicate` that controls if the forwarder flushes or "blackholes" metrics, dropping them permanently.

# Motivation

Our hosts have an on-box sentinel that controls if they are "live" or not. This liveness is used for health checks and other things. The semantics are such that if the command returns a non-zero status, we don't want to collect metrics for this host.  Assuming that the monitors in question do not alert on "No Data" then we've effectively muted the host and removed it's weirdo metrics.

# Notes

* I had to add a few methods to the `TransactionManager` to facilitate testing.
* It's possible that our liveness utility may occasionally want to send metrics anyway in some circumstances and that this won't ultimately be what we use but it felt like a somewhat useful feature.
* The semantics of this option and the methods (false versus true, non-zero output, etc) are totally changeable if you folks see it another way.
* Since the execution of the check is a blocking operation for the `TransactionManager` I made notes in the config file warning of slow checks.
* I'm not a Pythonista so improvements to the execution code are welcome if I've done poorly. :)

### DIFF
--- a/datadog.conf.example
+++ b/datadog.conf.example
@@ -93,6 +93,13 @@ use_mount: no
 # histogram_aggregates: max, median, avg, count
 # histogram_percentiles: 0.95
 
+# Use a local command as a predicate to determine if metrics should be sent
+# or just quietly dropped. This is useful for organizations with on-box "liveness"
+# controls to determine if hosts are up and worth monitoring. If the executed
+# command returns a non-zero status then metrics will be quietly dropped. This
+# check will be performed at each flush, so it is best to be a fast check!
+# flush_metrics_predicate: test -f /var/run/liveness-check
+
 # ========================================================================== #
 # DogStatsd configuration                                                    #
 # ========================================================================== #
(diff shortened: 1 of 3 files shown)


alq666:
@gphat won't comment on the substance of this particular one (others will) but your PRs consistently rank high in attention to detail and making me want to read them ;)

gphat:
lol @alq666, thanks for the kind words! Just trying to be thorough and give the reviewers as much info as I can! :wave:

remh:
Thanks a lot @gphat . As @alq666 , i love the details you brought to this PR. :clap: 

Regarding the idea itself, i understand the use case, but would it make sense to use a cron job or maybe a long living script that would control (stop / start ) the agent depending on the liveness check ? 

Another benefit of that approach is that the agent would effectively be stopped and won't consume any additional resources.

gphat:
Hey @remh!

Actually, yes. I think that's probably a better solution and saves you having to maintain the feature. Thanks for the feedback!

PR #1950, opened 2015-09-25
Add Nagios Plugin Runner… Plugin

gphat:
# What's this PR do?

Adds a `nagios_runner` plugin which allows you to execute Nagios plugins — or any command that follows [Nagios' Plugin API](https://assets.nagios.com/downloads/nagioscore/docs/nagioscore/3/en/pluginapi.html) and output Service Checks

# Motivation

Stripe has a significant number of in-house Nagios plugins that test non-trivial system things. There are also capabilities of Nagios plugins that do not exist in the Datadog agent. Rather than invest the time to port all of these plugins to Datadog and have to maintain both for a time, it's much easier to reuse them with the Datadog agent as Service Checks!

# Notes

* This does **not** convert Nagios plugin output in to metrics. I saw the very outdated *bernard* branch tried to do that. It's not necessary in my world to do that, as we use the plugins in a Service Check way.
* This might not be useful for everyone, but it definitely is for us. I'm also happy to make it a separate repo that people can grab and install themselves.
* I am not tied to naming of the plugin or it's configuration. Happy to change these.
* Service checks don't seem to be documented anywhere plainly. There are mentioned all over and clearly a first-class thing, but I can't find a good page for them and how to use them semantically. An area for improvement?

### DIFF
--- a/checks.d/nagios_runner.py
+++ b/checks.d/nagios_runner.py
@@ -1,0 +1,53 @@
+import subprocess
+
+from checks import AgentCheck
+
+# Runs arbitrary commands, but expects the exit code of the executed command
+# to adhere to the Nagios Plugin API:
+# https://assets.nagios.com/downloads/nagioscore/docs/nagioscore/3/en/pluginapi.html
+# Config file looks like:
+#
+# init_config:
+# # Not required for this check
+#
+# instances:
+#     - name: "some.check.name1"
+#       command: "/path/to/command with args"
+#     - name: "some.check.name2"
+#       command: "/path/to/command with args2"
+class NagiosRunner(AgentCheck):
+    def __init__(self, name, init_config, agentConfig):
+        AgentCheck.__init__(self, name, init_config, agentConfig)
+        self.last_ts = {}
+
+    def check(self, instance):
+
+        cmd = instance.get('command')
+        name = instance.get('name')
+
+        status = AgentCheck.UNKNOWN
+        try:
+            output = subprocess.check_output(cmd, stderr=subprocess.STDOUT, shell=True)
+            status = AgentCheck.OK
+            # If we get here it's because the exit code was 0
+            self.log.debug("Got OK {0}".format(name))
+        except subprocess.CalledProcessError as e:
+            # This is thrown if return code is != 0
+            ret = e.returncode
+            self.log.debug("Got NOK {0}: {1}".format(name, ret))
+            if ret == 1:
+                status = AgentCheck.WARNING
+                output = e.output
+            elif ret == 2:
+                status = AgentCheck.CRITICAL
+                output = e.output
+            else:
+                status = AgentCheck.UNKNOWN
+                output = e.output
+
+        self.service_check(
+            name,
+            status,
+            message = output,
+            tags = []
+        )
--- a/conf.d/nagios_runner.yaml.example
+++ b/conf.d/nagios_runner.yaml.example
@@ -1,0 +1,7 @@
+init_config:
+
+instances:
+  # - name: i.am.a.service.check.name
+  #   command: /usr/bin/foobar -A baz -C gorch
+  # - name: nagios.check.swap
+  #   command: /usr/lib/nagios/plugins/check_swap -w 524288000 -c 104857600 


alq666:
@gphat We secretly played with this a few years back https://github.com/DataDog/dd-agent/wiki/Bernard but I don't think it's seen any real maintenance since.

gphat:
@alq666 Indeed! I saw it in my searching and poking around. I needed something simple, so where we go. As stated, it might not be the solution you want in the official agent. I got no problems running this as a vendored thing. I just wanted to offer it up!

gphat:
Ping!

alq666:
@gphat triaging

gphat:
Was just a friendly ping. Nothing too hot. :)

checks.d/nagios_runner.py
```diff
@@ -0,0 +1,52 @@
...
+#
+# instances:
+#     - name: "some.check.name1"
+#       command: "/path/to/command with args"
+#     - name: "some.check.name2"
+#       command: "/path/to/command with args2"
+class NagiosRunner(AgentCheck):
+  def __init__(self, name, init_config, agentConfig):
+    AgentCheck.__init__(self, name, init_config, agentConfig)
+    self.last_ts = {}
```
LeoCavaille: I believe this is not used? Maybe we can get rid of this constructor then.

checks.d/nagios_runner.py
```diff
@@ -0,0 +1,52 @@
...
+
+  def check(self, instance):
+
+    cmd = instance.get('command')
+    name = instance.get('name')
+
+    status = AgentCheck.UNKNOWN
+    try:
+      output = subprocess.check_output(cmd, stderr=subprocess.STDOUT, shell=True)
```
LeoCavaille: I wonder if `check_output` has the same problem we saw with pipes, if it's the case we implemented that workaround f298832b2cfdcb8e5846d50b239d83bdd07a5469 which is now in the `utils.subprocess_output` module.

LeoCavaille:
Being able to run a one-off nagios script as a service check is interesting, I agree that the simpler it is the better, going down the rabbit hole and trying to reimplement a nagios interface in Python is not really desirable :wink: . Thanks for sharing!

checks.d/nagios_runner.py
gphat: Interesting! I defer to your judgement since that's clearly battle-won. :)

gphat:
You're quite welcome. I'll be sure and send along any changes we run in to between now and 5.6.0.

irabinovitch:
@gphat  Quick note, the test failures seem to mostly be style/spacing related. Should be quick to fix / identify by running flake8 locally.

gphat:
Will do. Thanks Ilan!

gphat:
All fixed up @irabinovitch!

gphat:
Hrm. The one failure in travis-ci doesn't look to be related…?

remh:
Thanks @gphat we need to spend some more time to discuss about this one. Postponing to 5.7.0

gphat:
Okey doke. I suspected from jump this might not be useful to include. Happy to set up a separate repo if this is not a thing you want!

remh:
@gphat what do you think of using the regular nagios integration combined with events monitors ? 
https://app.datadoghq.com/monitors#create/event

It seems to me that it would some the same use case.

gphat:
That would require running Nagios, yeah? Part of the reason I want to do it this way is to avoid using Nagios at all. I'm only interfacing with the Nagios plugins because we have a lot of existing plugins that I don't want to have to port wholesale. This seems like a valuable way for customers to transition as well as giving you a large corpus of plugins you can lean on to add functionality. (Although I get it not being a permanent fix.)

gphat:
Also I'm pretty happy to set up a separate repo of plugins stripe has found useful that might not be part of the core DD package, if that's preferable. I just want to always try and offer this work to you folks both to keep from deviating and to pay it forward!

remh:
Thanks a lot @gphat 
We really love all the work you put in your contributions.
That being said we won't be able to merge this PR as even though it's a good fit for this use case, this won't scale well with dozens of Nagios checks (especially if they take a few seconds to run).

So i'm going to close this pull request but you are more than welcome to host this on a different repo or on a github gist and we would be happy to list it with the list of community contributions http://docs.datadoghq.com/libraries/

irabinovitch:
@gphat  When you do post your check, please do let us know where it ends up (ilan _at_ datadoghq). I'd be interested in seeing if we can either spotlight it on our blog and/or in the newsletter. We can also list it on the libraries page Remi mentioned.

gphat:
Sounds great! I assumed we'd arrive at this juncture eventually. Thanks for considering it!

PR #1951, opened 2015-09-25
[util] Incorporating security-groups into tags collected from ec2

JohnLZeller:
Looks like the agent pulls AWS host tags, but doesn't pull `security-groups`. Additionally, it doesn't look like there is a good way to grab security group ids from EC2, however we do have access to the security group names. The fix to apply here is simply to make sure we add that to the tags collected as `security-group-name:<name>`.

### DIFF
--- a/datadog.conf.example
+++ b/datadog.conf.example
@@ -33,6 +33,9 @@ api_key:
 # Collect AWS EC2 custom tags as agent tags
 # collect_ec2_tags: no
 
+# Incorporate security-groups into tags collected from AWS EC2
+# collect_security_groups: no
+
 # Enable Agent Developer Mode
 # Agent Developer Mode collects and sends more fine-grained metrics about agent and check performance
 # developer_mode: no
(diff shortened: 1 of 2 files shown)


JohnLZeller:
@tmichelet any opinion on whether or not it's a good idea to add this here?

tmichelet:
I think that makes sense for this use case

tmichelet:
reading [the doc](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html), it seems like from within a VPC, when people update a security group this metadata won't be updated:
>>>
The names of the security groups applied to the instance.
>>>
After launch, you can only changes the security groups of instances running in a VPC. Such changes are reflected here and in network/interfaces/macs/mac/security-groups.

Is that a use case we want to handle? if not, :+1: good to merge, otherwise we'd need some more work

tmichelet:
Also, this will churn all our contexts, is this something we want or do we want to have an option for that?

JohnLZeller:
Updated this PR by adding `if EC2.metadata.get('security-groups'):` to ensure we catch a situation where metadata does not contain `security-groups`

remh:
@tmichelet is this ready to be merged ? Should we add another feature flag in datadog.conf to avoid churning existing contexts ?

tmichelet:
@remh cc @JohnLZeller yes I believe we should add another feature flag in `datadog.conf`.

JohnLZeller:
@tmichelet how's this?

tmichelet:
:+1: good to merge

PR #1959, opened 2015-09-28
[system] Filter SunOS memory_cap kstats by module

pfmooney:
On large SunOS systems, the kstat population is often quite large.  Using `kstat -c <class>` requires a full scan of the metric list and can be slow/expensive.  Filtering by `module` allows kstat consumers to use the kstat index (which includes `module`, `name` and `instance`, but not `class`). Using this filter strategy for the `zone_memory_cap` check results in a speedup of 1-2 orders of magnitude.


remh:
Thanks a lot @pfmooney !

Issue #1855, opened 2015-08-24
[mongo] Connection DB should be separate from monitored DB(s)


asiebert:
Good day, 

we recently tried to integrate the agent with our dedicated mongo clusters, so we could correlate their metrics with ours more easily. We're using dedicated cluster / replica sets, which allows us to group multiple DBs on each cluster/host.

The integration documentation (visible within Datadog's integration panel) recommends to create a read-only admin user, or an admin user with the `clusterMonitor` role for Mongodb 3.0 (_as a side note, this is also available in mongo 2.6_).
Here comes the issue: the service check will report `dbstats()` on the database configured, by default `admin`. I don't think anybody really cares about such metrics (collections, dataSize, fileSize...) for this DB , while those would be really useful on "real", business databases. In this scenario, only the cluster-wide metrics (from `serverStatus()`, `replSetGetStatus()`) are useful.

Of course, we could create a read-only user for each DB, and duplicate the mongo URI on the `mongo.yaml`, but it's extra cumbersome maintenance/setup, and would mean running the cluster-wide monitoring commands for each specific DB several times in a row (even though they're on the same host). It also defeats the purpose of using a [`clusterMonitor`](http://docs.mongodb.org/manual/reference/built-in-roles/#clusterMonitor) user role, which grants permission for `dbStats()` "on _all_ databases in the cluster".
For context on the authentication, note that the DB currently listed in the URI is the ["authentication database"](http://docs.mongodb.org/manual/core/authentication/#client-users) for the user, which is always going to be `admin` for admin users. Hence the need to decouple the authentication part and monitored objects.

---
I see two ways to address this. Both involve keeping the existing `server` connection URI in the config, and keeping the existing behavior for cluster-wide metric collection (as the DB used doesn't matter).
#### Option A
Add a new `server.dbs` setting, containing the list of DBs to monitor. The service checks uses the connection URI to log in, and loop over the `dbs` list to collect and report `dbstats()` metrics.
If no `dbs` value is configured, the service check keeps the same behavior as currently (monitor the DB to log into).

#### Option B
If we're an admin (especially with `clusterMonitor` privileges), we can just use the [`listDatabases`](http://docs.mongodb.org/manual/reference/privilege-actions/#authr.listDatabases) command and loop/report for each of the existing DBs.
An explicit setting to control whether to use that new behavior or the current one would most likely be needed.

----

In both options, each DB can be included as a tag when reporting its metrics.

[...]

PR #1961, opened 2015-10-01
[mongo] Connection DB should be separate from monitored DB(s)

truthbk:
This fix attempts to address what is described here: https://github.com/DataDog/dd-agent/issues/1855

Two main things have been done:
 - We now collect mongodb dbstats for each DB running on the cluster/host and return them with an associated tag cluster:db:foo where foo would be the database name.
  - We now also correctly deal with replica sets. While getting the integration tests to work I found an issue with the way we were attempting to collect replica set stats. These commands can only be made against the 'admin' db (which always exists I believe), so we now issue replicaset commands against it. 

Also, since pymongo.Connection is deprecated, I changed this over to use MongoClient which is the preferred way to do so. Finally, the api_key was causing issues when creating an event when dealing with some of the replicaset data and creating events - older integration tests were not picking up on this because replicaset's were essentially not being handled originally.

Integration tests are good. Local tests against docker are good.

Some refactoring may be in order. Also note my python game definitely need polishing - so any comments are welcome.

### DIFF
--- a/checks.d/mongo.py
+++ b/checks.d/mongo.py
@@ -1,6 +1,5 @@
 # stdlib
 import time
-import types
 
 # 3p
 import pymongo
@@ -168,7 +167,7 @@ class MongoDb(AgentCheck):
         self.event({
             'timestamp': int(time.time()),
             'event_type': 'Mongo',
-            'api_key': agentConfig['api_key'],
+            'api_key': agentConfig.get('api_key', ''),
             'msg_title': msg_title,
             'msg_text': msg,
             'host': hostname
@@ -176,7 +175,8 @@ class MongoDb(AgentCheck):
 
     def check(self, instance):
         """
-        Returns a dictionary that looks a lot like what's sent back by db.serverStatus()
+        Returns a dictionary that looks a lot like what's sent back by
+        db.serverStatus()
         """
         if 'server' not in instance:
             raise Exception("Missing 'server' in mongo config")
@@ -200,7 +200,7 @@ class MongoDb(AgentCheck):
         username = parsed.get('username')
         password = parsed.get('password')
         db_name = parsed.get('database')
-        clean_server_name = server.replace(password, "*" * 5) if password is not None else server
+        clean_server_name = server.replace(password, "*"*5) if password is not None else server
 
         tags = instance.get('tags', [])
         tags.append('server:%s' % clean_server_name)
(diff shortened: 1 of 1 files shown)


checks.d/mongo.py
```diff
@@ -164,11 +164,15 @@ def get_state_description(state):
         hostname = get_hostname(agentConfig)
         msg_title = "%s is %s" % (clean_server_name, status)
         msg = "MongoDB %s just reported as %s" % (clean_server_name, status)
+        api_key = ""
+
+        if 'api_key' in agentConfig:
+            api_key = agentConfig['api_key']
```
hkaj: you can replace these 3 lines with 
`api_key = agentConfig.get('api_key', '')`

checks.d/mongo.py
```diff
@@ -164,11 +164,15 @@ def get_state_description(state):
...
 
         self.event({
             'timestamp': int(time.time()),
             'event_type': 'Mongo',
-            'api_key': agentConfig['api_key'],
+            'api_key': api_key,
```
hkaj: you can replace the modifications above with
`'api_key': agentConfig.get('api_key', ''),`
here

checks.d/mongo.py
```diff
@@ -252,17 +258,35 @@ def check(self, instance):
             raise Exception(status['errmsg'].__str__())
 
         status['stats'] = db.command('dbstats')
+        dbstats = {}
+        dbstats[db_name] = {'stats': status['stats']}
+
+        tags = list(set(tags))
```
hkaj: It's already done L213, do we need to do it again?

checks.d/mongo.py
```diff
@@ -252,17 +258,35 @@ def check(self, instance):
...
 
+                #need a new connection to deal with replica sets
+                setname = replSet.get('set')
+                cli = pymongo.mongo_client.MongoClient(server, socketTimeoutMS=timeout,
+                    replicaset=setname, read_preference=pymongo.ReadPreference.NEAREST,
+                    **ssl_params)
+                db = cli[db_name]
+
+                if do_auth:
+                    if not db.authenticate(username, password):
```
hkaj: `if do_auth and not db.authenticate`

checks.d/mongo.py
```diff
@@ -326,3 +360,27 @@ def check(self, instance):
...
+                value = dbstats[s]
+                dbtag = s #key is db name, so we use that as the tag
+                try:
+                    for c in m.split('.'):
+                        value = value[c]
+                except KeyError:
+                    continue
+
+                # value is now status[x][y][z]
+                assert type(value) in (types.IntType, types.LongType, types.FloatType)
```
hkaj: `isinstance(value, (int, long, float))` would be more pythonic

checks.d/mongo.py
```diff
@@ -200,7 +201,10 @@ def check(self, instance):
         username = parsed.get('username')
         password = parsed.get('password')
         db_name = parsed.get('database')
-        clean_server_name = server.replace(password, "*" * 5) if password is not None else server
+        clean_server_name = server.replace(
+            password,
+            "*" *
+            5) if password is not None else server
```
hkaj: the previous version is more readable. As a general rule, as long as the line is fully visible on GitHub you can leave it uncut.

checks.d/mongo.py
```diff
@@ -209,7 +213,8 @@ def check(self, instance):
         tags = list(set(tags))
 
         if not db_name:
-            self.log.info('No MongoDB database found in URI. Defaulting to admin.')
+            self.log.info(
+                'No MongoDB database found in URI. Defaulting to admin.')
```
hkaj: same here

checks.d/mongo.py
```diff
@@ -227,42 +232,86 @@ def check(self, instance):
 
         do_auth = True
         if username is None or password is None:
-            self.log.debug("Mongo: cannot extract username and password from config %s" % server)
+            self.log.debug(
+                "Mongo: cannot extract username and password from config %s" %
+                server)
```
hkaj: ditto

checks.d/mongo.py
```diff
@@ -304,19 +357,31 @@ def check(self, instance):
...
+                        value = value[c]
+                except KeyError:
+                    continue
 
             # value is now status[x][y][z]
-            assert type(value) in (types.IntType, types.LongType, types.FloatType)
+            assert type(value) in (
+                types.IntType,
+                types.LongType,
+                types.FloatType)
```
hkaj: You should use something more pythonic and raise a more descriptive error. Something like:
```python
if not isinstance(value, (int, long, float)):
    raise TypeError('{0} value is {1}, it should be an int, a float or a long instead.'.format(m, type(value)))
```

checks.d/mongo.py
```diff
@@ -326,3 +391,29 @@ def check(self, instance):
             if m in self.RATES:
                 m = self.normalize(m.lower(), 'mongodb') + "ps"
                 self.rate(m, value, tags=tags)
+
+        stat_metrics = filter(lambda x: x.startswith('stats.'), self.METRICS)
+        for s in dbstats:
```
hkaj: `for s, value in dbstats.iteritems():`
`iteritems()` is a generator that yields 2-tupes representing the key and the value. This way s is the key and value is the value

checks.d/mongo.py
```diff
@@ -326,3 +391,29 @@ def check(self, instance):
...
+            for m in stat_metrics:
+                value = dbstats[s]
```
hkaj: not needed with the previous comment

checks.d/mongo.py
```diff
@@ -326,3 +391,29 @@ def check(self, instance):
...
+                dbtag = s  # key is db name, so we use that as the tag
+                try:
+                    for c in m.split('.'):
+                        value = value[c]
```
hkaj: if i understand correctly, c is either `stats` or the metric name. If that is the case you can simply do `value = value[m.split('.')[1]]` and drop the try/except.

checks.d/mongo.py
```diff
@@ -326,3 +391,29 @@ def check(self, instance):
...
+                except KeyError:
+                    continue
+
+                # value is now status[x][y][z]
+                assert type(value) in (
+                    types.IntType,
+                    types.LongType,
+                    types.FloatType)
```
hkaj: same as the comment L384, `if not isinstance(...): raise TypeError` would be better.

checks.d/mongo.py
truthbk: Yeah we need to do it again, in fact we need to do it for every db in the mongo cluster, stats reported are specific to the db you call them on.

checks.d/mongo.py
truthbk: I cleaned it up a bit following your advice, but I need the try...except: because apparently not all metrics are always reported by mongo. At least nsSizeMB isn't :(, I'll keep the try...except block in for now, if that's OK.

hkaj:
LGTM

hkaj:
Don't forget to remove your branch from Github once you're all set :)

PR #1962, opened 2015-10-01
[checks.d][openstack] Try to get the list of aggregates and the associated hypervisors. 

mtougeron:
Hi @talwai, here's the update I was talking about for the Aggregate & Availability Zone I emailed you about yesterday.
 
This will allow us to map the aggregate and AZ info onto the hypervisor as a tag. Depending on the permissions granted to the user account doing the checks, the api call for the aggregates may fail. The check will gracefully ignore the aggregate & az tags if that is the case.

Need to update the nova policy.json to allow the role access to this api call.

```
    "compute_extension:aggregates": "rule:admin_api or role:datadog_monitoring",
    "compute_extension:v3:os-aggregates:index": "rule:admin_api or role:datadog_monitoring",
    "compute_extension:v3:os-aggregates:show": "rule:admin_api or role:datadog_monitoring",
```

/cc @marktward

### DIFF
--- a/checks.d/openstack.py
+++ b/checks.d/openstack.py
@@ -69,6 +69,8 @@ class OpenstackCheck(AgentCheck):
         self._nova_url = None
         self._neutron_url = None
 
+        self._aggregate_list = None
+
         #Store the tenant id of the datadog user.
 
     def _make_request_with_auth_fallback(self, url, headers=None):
@@ -257,6 +259,8 @@ class OpenstackCheck(AgentCheck):
             self.warning("Your check is not configured to monitor any hypervisors.\n" +
                          "Please list `hypervisor_ids` under your init_config")
 
+        self._aggregate_list = self.get_all_aggregate_hypervisors()
+
         stats = {}
         for hyp in hypervisors:
             stats[hyp] = {}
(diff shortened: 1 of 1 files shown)


checks.d/openstack.py
```diff
@@ -319,6 +343,11 @@ def get_stats_for_single_hypervisor(self, hyp_id):
             'hypervisor_id:{0}'.format(hyp['id']),
             'virt_type:{0}'.format(hyp['hypervisor_type'])
         ]
+        if hyp['hypervisor_hostname'] in self._aggregate_list:
+            service_check_tags.append('aggregate:{0}'.format(self._aggregate_list[hyp['hypervisor_hostname']]['aggregate']))
+            # Need to check if there is a value for availability_zone because it is possible to have an aggregate without an AZ
+            if self._aggregate_list[hyp['hypervisor_hostname']]['availability_zone']:
```
talwai: In the `aggregate without AZ` scenario, are we guaranteed the `availability_zone` key will still exist? If not, the check should be `if self._aggregate_list[hyp['hypervisor_hostname']].get('availability_zone')`.

checks.d/openstack.py
```diff
@@ -384,6 +413,11 @@ def nova_stats_to_metrics(self, stats):
                 'hypervisor_id:{0}'.format(payload['id']),
                 'virt_type:{0}'.format(payload['hypervisor_type'])
             ]
+            if payload['hypervisor_hostname'] in self._aggregate_list:
+                tags.append('aggregate:{0}'.format(self._aggregate_list[payload['hypervisor_hostname']]['aggregate']))
+                # Need to check if there is a value for availability_zone because it is possible to have an aggregate without an AZ
+                if self._aggregate_list[payload['hypervisor_hostname']]['availability_zone']:
```
talwai: Ditto above

checks.d/openstack.py
mtougeron: The key will exist but it will either be an empty string or None. I didn't want to send the empty tag so I put this conditional around it.

mtougeron: I believe, behind the scenes, OpenStack will assign a server in that Aggregate to a default AZ but it doesn't expose that info to this API call.

talwai:
looks good @mtougeron . Thanks!

PR #1965, opened 2015-10-05
Create a container_name extractor to always extract the real container name

hkaj:
... not the first one.

### DIFF
--- a/checks.d/docker_daemon.py
+++ b/checks.d/docker_daemon.py
@@ -96,12 +96,23 @@ def image_tag_extractor(c, key):
     return None
 
 
+def container_name_extractor(c):
+    # we sort the list to make sure that a docker API update introducing
+    # new names with a single "/" won't make us report dups.
+    names = sorted(c.get('Names', []))
+    for name in names:
+        # the leading "/" is legit, if there's another one it means the name is actually an alias
+        if name.count('/') <= 1:
+            return [str(name).lstrip('/')]
+    return [c.get('Id')[:11]]
+
+
 TAG_EXTRACTORS = {
     "docker_image": lambda c: [c["Image"]],
     "image_name": lambda c: image_tag_extractor(c, 0),
     "image_tag": lambda c: image_tag_extractor(c, 1),
     "container_command": lambda c: [c["Command"]],
-    "container_name": lambda c: [str(c['Names'][0]).lstrip("/")] if c.get("Names") else [c['Id'][:11]],
+    "container_name": container_name_extractor,
 }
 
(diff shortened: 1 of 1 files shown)


hkaj:
I have been able to test the fix thanks to the affected VM @sjenriquez sent me, the duplicate containers disappeared from the metric explorer so it's ready to be reviewed.

LotharSee:
Are we sure that this is going to be enough to ensure the consistency? Is the "Names" field from the API always sorted in the same order?

checks.d/docker_daemon.py
```diff
@@ -96,12 +96,21 @@ def image_tag_extractor(c, key):
...
+    return [c.get('Id')[:11]]
+
+
 TAG_EXTRACTORS = {
     "docker_image": lambda c: [c["Image"]],
     "image_name": lambda c: image_tag_extractor(c, 0),
     "image_tag": lambda c: image_tag_extractor(c, 1),
     "container_command": lambda c: [c["Command"]],
-    "container_name": lambda c: [str(c['Names'][0]).lstrip("/")] if c.get("Names") else [c['Id'][:11]],
+    "container_name": lambda c: container_name_extractor(c),
```
LotharSee: You can put directly `"container_name": container_name_extractor`

checks.d/docker_daemon.py
```diff
@@ -96,12 +96,21 @@ def image_tag_extractor(c, key):
     return None
 
 
+def container_name_extractor(c):
```
LotharSee: I'd expect it to return directly a single name, instead of a list. So that we can `container_name_extractor(container)` instead of `container_name_extractor(container)[0]`.

checks.d/docker_daemon.py
```diff
@@ -96,12 +96,21 @@ def image_tag_extractor(c, key):
...
+    names = c.get('Names', [])
+    for name in names:
+        # the leading "/" is legit, if there's another one it means the name is actually an alias
+        if name.count('/') <= 1:
+            return [str(name).lstrip('/')]
+    return [c.get('Id')[:11]]
+
+
 TAG_EXTRACTORS = {
```
LotharSee: Just a question: why are all these extractors returning a list of 1 element, instead of the element itself?

hkaj: do they though? Reading `image_tag_extractor` it looks like it could return a list of more than one element.

LotharSee: Right!

LotharSee:
:ok_hand:

PR #1966, opened 2015-10-05
[checks.d][openstack] Get the limits (quotas) for each project (tenant)

mtougeron:
This update will query keystone to get the list of available projects (tenants) and then iterate through the list and get the limits (quotas) for each. Alternatively you can specify the project_ids you want to check if you only want to check certain ones. If for some reason it isn't able to get the limits for a project it will simply move on to the next one in the list.

If you want to get the list of all projects, you will need to update the keystone policy.json to allow the role access to this api call.

```
"identity:list_projects": "rule:admin_required or role:datadog_monitoring",
"identity:list_user_projects": "rule:admin_or_owner or role:datadog_monitoring",
```

I'm not entirely sure, but I think that the nova policy.json may also need to be updated for access to the admin limits too.

```
"compute_extension:used_limits_for_admin": "rule:admin_api or role:datadog_monitoring",
```

@talwai

### DIFF
--- a/checks.d/openstack.py
+++ b/checks.d/openstack.py
@@ -37,6 +37,31 @@ NOVA_SERVER_METRICS = [
 
 NOVA_SERVER_INTERFACE_SEGMENTS = ['_rx', '_tx']
 
+PROJECT_METRICS = [
+    "maxImageMeta",
+    "maxPersonality",
+    "maxPersonalitySize",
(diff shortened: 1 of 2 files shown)


mtougeron:
@talwai should I close this PR?

talwai:
@mtougeron yes, please :). Your changes have been incorporated in #2001

Issue #1957, opened 2015-09-28
Test status is "ok" with command errors


xiaclo:
From ```rake ci:run```:

```
...
tests.core.test_check_status.test_check_status_fail ... ok
tests.core.test_check_status.test_check_status_pass ... ok
tests.core.test_check_status.test_persistence ... ok
tests.core.test_check_status.test_persistence_fail ... ok
test_dupe_tags (tests.core.test_common.TestAggregator) ... ok
testEdgeCases (tests.core.test_common.TestCore) ... ok
Tests that the app tags are sent if specified so ... Usage: nosetests [options]

nosetests: error: no such option: -s
ok
test_collector (tests.core.test_common.TestCore) ... Usage: nosetests [options]

nosetests: error: no such option: -s
ok
test_counter (tests.core.test_common.TestCore) ... ok
test_gauge (tests.core.test_common.TestCore) ... ok
test_get_proxy (tests.core.test_common.TestCore) ... ok
test_min_collection_interval (tests.core.test_common.TestCore) ... ok
test_name (tests.core.test_common.TestCore) ... ok
Starting with Agent 5.0.0, there should always be a local forwarder ... ok
...
```

I doubt "no such option: -s" qualifies as the tests running "ok"

This is using nose==1.3.4

remh:
@xiaclo Thanks for the feedback! the `-s` parameter is used to tell nose to get into pdb when a breakpoint is hit. This is not required for the test suite to be executed. So it looks like the tests are passing properly.

LeoCavaille:
This isn't supposed to show up in the build logs, so I would be for investigating it a bit, maybe we are skipping/missing tests because of that?
example: https://travis-ci.org/DataDog/dd-agent/jobs/82912178#L263

The `-s` flag is passed here
https://github.com/DataDog/dd-agent/blob/master/ci/common.rb#L202

PR #1971, opened 2015-10-08
[tests] remove misleading nosetests error message

degemer:
Problem
-------
When running `rake ci:run` we can see `nosetests: error: no such option:
-s` for some specific tests.

It's caused by calls to `get_config` (from config.py) with `parse_args=True`,
which calls `OptionParser`, which tries to parse the `nosetests` command line,
and fails because `dd-agent` doesn't have a `-s` option. (but has a `-A` and
`-v`). Source of the error: https://github.com/DataDog/dd-agent/blob/4907d1178a0eb38558cfa2cc644be769835a7b92/config.py#L88

Solution
--------
One of the concerned tests was directly calling `get_config`, the
`parse_args=False` arg was added.

The others are using a `Collector` which calls `get_hostname` (itself calling
`get_config` with `parse_args=True`) in `_get_hostname_metadata`:
https://github.com/DataDog/dd-agent/blob/4907d1178a0eb38558cfa2cc644be769835a7b92/checks/collector.py#L754

However the `Collector` already stores the agent config in `self.agentConfig`,
so this is used to avoid the useless `get_config` (not called if a config is
provided).

This fixes https://github.com/DataDog/dd-agent/issues/1957.

### DIFF
--- a/checks/collector.py
+++ b/checks/collector.py
@@ -25,7 +25,6 @@ import modules
 from util import (
     EC2,
     GCE,
-    get_hostname,
     get_os,
     get_uuid,
     Timer,
@@ -751,7 +750,7 @@ class Collector(object):
         except Exception:
             pass
 
-        metadata["hostname"] = get_hostname()
+        metadata["hostname"] = self.hostname
 
         # Add cloud provider aliases
         host_aliases = GCE.get_host_aliases(self.agentConfig)
--- a/tests/core/test_config.py
+++ b/tests/core/test_config.py
@@ -19,7 +19,8 @@ class TestConfig(unittest.TestCase):
         """Leading whitespace confuse ConfigParser
         """
         agentConfig = get_config(cfg_path=os.path.join(os.path.dirname(os.path.realpath(__file__)),
-                                                       'fixtures', 'badconfig.conf'))
+                                                       'fixtures', 'badconfig.conf'),
+                                 parse_args=False)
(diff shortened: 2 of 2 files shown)


LeoCavaille:
Nice! Thanks for digging into it. :rocket: :rocket: :guitar: :star2: 
:+1: 

Next in line, fixing the couchdb flavor on Travis :stuck_out_tongue_winking_eye: ?

checks/collector.py
```diff
@@ -751,7 +751,7 @@ def _get_hostname_metadata(self):
         except Exception:
             pass
 
-        metadata["hostname"] = get_hostname()
+        metadata["hostname"] = get_hostname(config=self.agentConfig)
```
LeoCavaille: re-reading this, it looks like that it's exactly the same call that we do before when launching the agent, see https://github.com/DataDog/dd-agent/blob/quentin/fix-nosetests-output/agent.py#L109
maybe use this instead of another call?

degemer: Wow, nice catch, `self.hostname` seems to come from this [commit](https://github.com/DataDog/dd-agent/commit/766ee8f8b495d6d3f61da1fcc5123ad2c413de39), it's probably just an omission. Fixing it.

degemer:
Failed again on the couch test... :crying_cat_face: 
Have you tried ~~rebooting~~ emptying the cache ?

LeoCavaille:
Not yet. Looking good!

PR #1973, opened 2015-10-12
[windows] log + timeout WMI queries after 10 seconds

yannmh:
## Context
On Windows, the collector sometimes spends an abnormally high amount of time collecting system metrics, causing the agent to restart.
System metric are collected through WMI, using the `wmi` Python package. It is unfortunately known to cause us a lot of issues... In particular, most of the WMI queries, here, use Win32_PerfFormattedData_* which are slow !

### `log_subprocess` → `logged`
Make `log_subprocess` decorator more generic:
 * Rename it `logged`
 * Move it to `utils/debug`

### WMI activity logging
Log DEBUG WMI queries among Windows system check.

###  Decorator to timeout a function
Create a decorator to timeout a function. Decorated method calls are executed in a separate new thread with a specified timeout.
Also check if a thread for the same function already exists before creating a new one.

__Note__: Compatible with Windows (thread based).

#### Timeout WMI queries after 10 seconds
Timeout WMI queries after 10 seconds among Windows system check.

## Milestone

Seems like a good 5.5.2 candidate.

### DIFF
--- a/checks/system/win32.py
+++ b/checks/system/win32.py
@@ -1,6 +1,7 @@
 # datadog
 from checks import Check
 from utils.debug import logged
+from utils.timeout import timeout, TimeoutException
 
 # 3rd party
 try:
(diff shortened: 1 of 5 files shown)


yannmh:
Closing in favor of https://github.com/DataDog/dd-agent/pull/2011

PR #1974, opened 2015-10-12
Add UDP metrics.

gphat:
# What's this PR do?

Adds UDP metrics from `/proc/net/snmp` in addition to the existing TCP metrics.

# Motivation

We have some things using UDP and they are having trouble keeping up. Our most reliable indicator of this problem are dropped UDP due to full buffers. This adds that functionality to the agent so that we can see it in charts!

# Notes

I basically lifted the existing TCP code and made few modifications to have different variables for TCP versus UDP. If you'd prefer this refactored some for lessening code duplication, I'm open to refactoring with guidance.

### DIFF
--- a/checks.d/network.py
+++ b/checks.d/network.py
@@ -223,10 +223,15 @@ class Network(AgentCheck):
                 proc.close()
 
             tcp_lines = [line for line in lines if line.startswith('Tcp:')]
-            column_names = tcp_lines[0].strip().split()
-            values = tcp_lines[1].strip().split()
+            udp_lines = [line for line in lines if line.startswith('Udp:')]
 
-            tcp_metrics = dict(zip(column_names,values))
+            tcp_column_names = tcp_lines[0].strip().split()
+            tcp_values = tcp_lines[1].strip().split()
+            tcp_metrics = dict(zip(tcp_column_names, tcp_values))
+
+            udp_column_names = udp_lines[0].strip().split()
+            udp_values = udp_lines[1].strip().split()
+            udp_metrics = dict(zip(udp_column_names, udp_values))
 
             # line start indicating what kind of metrics we're looking at
             assert(tcp_metrics['Tcp:'] == 'Tcp:')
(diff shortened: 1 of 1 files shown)


remh:
Thanks @gphat we'll review this soon

checks.d/network.py
```diff
@@ -239,6 +244,20 @@ def _check_linux(self, instance):
 
             for key, metric in tcp_metrics_name.iteritems():
                 self.rate(metric, self._parse_value(tcp_metrics[key]))
+
+            asert(udp_metrics['Udp:'] == 'Udp:')
```
hkaj: small typo, `asert` --> `assert`

checks.d/network.py
```diff
@@ -223,10 +223,15 @@ def _check_linux(self, instance):
...
-            values = tcp_lines[1].strip().split()
+            udp_lines = [line for line in lines if line.startswith('Udp:')]
 
-            tcp_metrics = dict(zip(column_names,values))
+            tcp_column_names = tcp_lines[0].strip().split()
+            tcp_values = tcp_lines[1].strip().split()
+            tcp_metrics = dict(zip(tcp_column_names, tcp_values))
+
+            udp_column_names = udp_lines[0].strip().split()
+            udp_values = udp_lines[1].stripe().split()
```
hkaj: `stripe` --> `strip`

checks.d/network.py
```diff
@@ -239,6 +244,20 @@ def _check_linux(self, instance):
 
             for key, metric in tcp_metrics_name.iteritems():
                 self.rate(metric, self._parse_value(tcp_metrics[key]))
+
+            asert(udp_metrics['Udp:'] == 'Udp:')
+
+            udp_metrics_name = {
+                'InDatagrams': 'system.net.udp.in_datadgrams',
```
hkaj: datadgrams --> datagrams

hkaj:
Hi @gphat, thanks for the contribution! A few typos to fix, but apart from that it looks good to me.

gphat:
Thanks @hkaj! Gonna do some testing here to verify things work before I say this is ready again. Sorry for not testing this.

checks.d/network.py
alq666: @gphat muscle memory, eh ;)

gphat:
Ok, I verified this works as expected locally. Sorry again, @hkaj!

hkaj:
No worries. Thanks for the contribution, UDP metrics will join their TCP counterpart in 5.6.0! :confetti_ball:

gphat:
:tada: Yeah @alq666 I guess I type "stripe" a lot. :sweat_smile:

olivielpeau:
FYI, the `RcvbufErrors` and `SndbufErrors` metrics do not exist on CentOS 5, which makes the network check fail on that OS (for reference see https://www.mail-archive.com/haifux@haifux.org/msg03802.html).

A simple test on the existence of the metrics [here](https://github.com/DataDog/dd-agent/blob/ae2d52d7ede2c7034c07203eaebd77b372e88095/checks.d/network.py#L259) could be enough.

I won't be super available for the next couple of days, could you have a further look at this @hkaj? :)

gphat:
Added a test, good catch!

hkaj:
Good catch indeed! I checked on Ubuntu and CentOS 7 but didn't test on them oldies. https://github.com/DataDog/dd-agent/pull/1986 should fix that.

PR #1975, opened 2015-10-14
Enable http_check to function with weak ciphers on SSL/TLS.

truthbk:
With this PR we enable the RSA+RC4+MD5 cipher which is considered weak. We can add further weak ciphers if needed. Disabled by default, the user must specify `weakciphers: true` in their http_check.yaml for this to have any real effect - otherwise weak ciphers are disabled.

Session object has nearly identical request method.

Default value for weakcipher is false.

Adding debug logging for weak ciphers.

Importing datetime differently to urllib3 code.

Added Transport adapter for weak cyphers, requests lib sessions required too.

Session object has nearly identical request method.

Default value for weakcipher is false.

Adding debug logging for weak ciphers.

Importing datetime differently to urllib3 code.

Changing cipherlist for weak ciphers, now it works.

### DIFF
--- a/checks.d/http_check.py
+++ b/checks.d/http_check.py
@@ -109,11 +109,11 @@ class WeakCiphersPoolManager(urllib3.poolmanager.PoolManager):
             return WeakCiphersHTTPSConnectionPool(host, port, **(self.connection_pool_kw))
         return super(WeakCiphersPoolManager, self)._new_pool(scheme, host, port)
 
+
 class WeakCiphersAdapter(HTTPAdapter):
     """"Transport adapter" that allows us to use TLS_RSA_WITH_RC4_128_MD5."""
 
-    def init_poolmanager(self, connections, maxsize, block=False,
-                            **pool_kwargs):
+    def init_poolmanager(self, connections, maxsize, block=False, **pool_kwargs):
         # Rewrite of the
         # requests.adapters.HTTPAdapter.init_poolmanager method
         # to use WeakCiphersPoolManager instead of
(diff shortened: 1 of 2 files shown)


checks.d/http_check.py
```diff
@@ -5,18 +5,127 @@
...
+            self._tunnel()
+            # Mark this connection as not reusable
+            self.auto_open = 0
+
+            # Override the host with the one we're requesting data from.
+            hostname = self._tunnel_host
+
+        is_time_off = datetime.today().date() < urllib3.connection.RECENT_DATE
+        if is_time_off:
+            warnings.warn((
```
remh: We can assume that the system time is correct. If it's not, there are way more things that would be broken

truthbk: This was actually part of the "parent" connect() method in urllib3, I had to overwrite it because I had to change the way we called `ssl_wrap_socket` and add the `ciphers` parameter to the call - the rest of the method is pretty much verbatim.

checks.d/http_check.py
```diff
@@ -5,18 +5,127 @@
...
+    def connect(self):
+        # Add certificate verification
+        conn = self._new_conn()
+
+        resolved_cert_reqs = ssl_.resolve_cert_reqs(self.cert_reqs)
+        resolved_ssl_version = ssl_.resolve_ssl_version(self.ssl_version)
+
+        hostname = self.host
+        if getattr(self, '_tunnel_host', None):
+            # _tunnel_host was added in Python 2.6.3
```
remh: You can assume that python >= 2.6.3

checks.d/http_check.py
remh: Still we can remove it. It pollutes the code for a case that will cause more issues than this ssl connection not working if it happens and that will be caught in other parts of the agent.

PR #1978, opened 2015-10-15
Add StatsD integration.

gphat:
# What's this PR do?

Adds a StatsD integration for collecting metrics and service checks using [StatsD's admin interface](https://github.com/etsy/statsd/blob/master/docs/admin_interface.md).

# Motivation

Even though Datadog provides DogStatsD some organizations may choose to run StatsD for one reason or another. They may also be migrating from an existing StatsD setup. Or maybe they are monitoring someone else's stuff that uses StatsD.

We're in the second group, migrating from an existing StatsD setup. Having the ability to get service checks and metrics from our existing StatsD gives us both monitoring of current infra until it's gone and a way to measure our progress in moving "legacy" clients when that comes to pass.

# Notes

* I can imagine this seeming odd since DogStatsD, but I hope I've justified it well.
* I'm running this in our QA environment and it works fine.
* As I often say, I'm not a native Pythoner so helping me with style or idiomatic shit is welcome.
* Suggestions for improvement are also welcome. I don't write a lot of socket code. :)
* I used two different service check names for failure and for "health". Is this ok?
* I used `gauge` for all the things, even though `uptime` and `messages.bad_lines_seen` will only increase unless StatsD restarts. Which metric type is appropriate here?

Thanks for taking a look!

### DIFF
--- a/checks.d/statsd.py
+++ b/checks.d/statsd.py
@@ -1,0 +1,80 @@
+# stdlib
+import re
+import socket
+from StringIO import StringIO
+
+# project
+from checks import AgentCheck
+
+SERVICE_CHECK_NAME = "statsd.can_connect"
+SERVICE_CHECK_NAME_HEALTH = "statsd.is_up"
+
+ENDER = re.compile("^(END|health: up|health: down)\n$", re.MULTILINE)
+BAD_ENDER = re.compile("^ERROR\n$", re.MULTILINE)
+
+class StatsCheck(AgentCheck):
+    def check(self, instance):
+        host = instance.get("host", "localhost")
+        port = instance.get("port", 8126)
+        tags = instance.get("tags", [])
+        tags = ["host:{0}".format(host), "port:{0}".format(port)] + tags
+
+        # Is it up?
+        health = self._send_command(host, port, "health", tags).getvalue().strip()
+        if health == "health: up":
+            self.service_check(
+                SERVICE_CHECK_NAME_HEALTH, AgentCheck.OK, tags
+            )
+        else:
+            self.service_check(
+                SERVICE_CHECK_NAME_HEALTH, AgentCheck.CRITICAL, tags
+            )
+
+        # Get general stats
+        stats = self._send_command(host, port, "stats", tags)
+        stats.seek(0)
+        for l in stats.readlines():
+            parts = l.strip().split(":")
+            if len(parts) == 2:
+                # Uptime isn't a gauge. Since we have only one exception, this
+                # seems fine. If we make more a lookup table might be best.
+                if parts[0] == "bad_lines_seen":
+                    self.monotonic_count("statsd.{0}".format(parts[0]), float(parts[1]), tags=tags)
+                else:
+                    self.gauge("statsd.{0}".format(parts[0]), float(parts[1]), tags=tags)
+
+        counters = len(self._send_command(host, port, "counters", tags).getvalue().splitlines()) - 1
+        self.gauge("statsd.counters.count", counters, tags=tags)
+
+        gauges = len(self._send_command(host, port, "gauges", tags).getvalue().splitlines()) - 1
+        self.gauge("statsd.gauges.count", gauges, tags=tags)
+
+        timers = len(self._send_command(host, port, "timers", tags).getvalue().splitlines()) - 1
+        self.gauge("statsd.timers.count", timers, tags=tags)
+
+    def _send_command(self, host, port, command, tags):
+        try:
+            s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+            s.connect((host, port))
+
+            s.sendall("{0}\n".format(command))
+
+            buf = StringIO()
+
+            chunk = s.recv(1024)
+            buf.write(chunk)
+            while chunk:
+                if ENDER.search(chunk):
+                    break
+                if BAD_ENDER.search(chunk):
+                    raise Exception("Got an error issuing command: {0}".format(command))
+                chunk = s.recv(1024)
+                buf.write(chunk)
+            return buf
+        except Exception as e:
+            self.service_check(
+                SERVICE_CHECK_NAME, AgentCheck.CRITICAL, tags
+            )
+            raise Exception("Failed connection {0}".format(str(e)))
+        finally:
+            s.close()
--- a/conf.d/statsd.yaml.example
+++ b/conf.d/statsd.yaml.example
@@ -1,0 +1,8 @@
+init_config:
+
+instances:
+  - host: localhost
+    port: 8126
+    # tags:
+    #  - optional_tag1
+    #  - optional_tag2


gphat:
Note that your guidelines state that all CI tests must pass, but it seems like for all my PRs at least one Travis check fails for reasons that aren't mine. I assume that's ok, but others might get confused!

checks.d/statsd.py
```diff
@@ -0,0 +1,87 @@
...
+import socket
+import sys
+from StringIO import StringIO
+
+# project
+from checks import AgentCheck
+
+SERVICE_CHECK_NAME = "statsd.can_connect"
+SERVICE_CHECK_NAME_HEALTH = "statsd.is_up"
+INFO_METRICS = [
```
remh: Looks like you are not using this list anywhere

gphat: Thanks, good catch! Removed.

checks.d/statsd.py
```diff
@@ -0,0 +1,78 @@
...
+
+class StatsCheck(AgentCheck):
+    def check(self, instance):
+        host = instance.get("host", "localhost")
+        port = instance.get("port", 8126)
+        tags = instance.get('tags', [])
+
+        # Is it up?
+        health = self._send_command(host, port, "health").getvalue().strip()
+        if health.strip() == "health: up":
```
remh: it was already stripped line 23

checks.d/statsd.py
```diff
@@ -0,0 +1,78 @@
...
+            self.service_check(
+                SERVICE_CHECK_NAME_HEALTH, AgentCheck.OK, []
```
remh: can you tag by host and port ? just in case someone has multiple statsd servers

checks.d/statsd.py
```diff
@@ -0,0 +1,78 @@
...
+            )
+        else:
+            self.service_check(
+                SERVICE_CHECK_NAME_HEALTH, AgentCheck.CRITICAL, []
```
remh: same here

remh: also should the check returns here ? it statsd is not up it's unlikely that we'll be able to collect metrics. What do you think ?

checks.d/statsd.py
```diff
@@ -0,0 +1,78 @@
...
+            )
+
+        # Get general stats
+        stats = self._send_command(host, port, "stats")
+        stats.seek(0)
+        for l in stats.readlines():
+            parts = str.split(l.strip(), ":")
```
remh:
```python
parts = l.strip().split(":")
```
is more pythonic

checks.d/statsd.py
```diff
@@ -0,0 +1,78 @@
...
+            if(len(parts) == 2):
```
remh: Parenthesis are not needed in python :)

checks.d/statsd.py
```diff
@@ -0,0 +1,78 @@
...
+                self.gauge("statsd.{0}".format(parts[0]), parts[1].strip(), tags=tags)
```
remh: it should be:
```python
float(parts[1])
```
instead of
```python
parts[1].strip()
```

checks.d/statsd.py
remh: host and port should actually be added to your `tags` variable

checks.d/statsd.py
```diff
@@ -0,0 +1,78 @@
...
+
+        counters = len(str.split(self._send_command(host, port, "counters").getvalue(), '\n')) - 1
```
remh:
```python
counters = len(self._send_command(host, port, "counters").getvalue().splitlines()) - 1
```

checks.d/statsd.py
```diff
@@ -0,0 +1,78 @@
...
+        self.gauge("statsd.counters.count", counters, tags=tags)
+
+        gauges = len(str.split(self._send_command(host, port, "gauges").getvalue(), '\n')) - 1
```
remh: same here, you should use splitlines

checks.d/statsd.py
```diff
@@ -0,0 +1,78 @@
...
+        self.gauge("statsd.gauges.count", gauges, tags=tags)
+
+        timers = len(str.split(self._send_command(host, port, "timers").getvalue(), '\n')) - 1
```
remh: same here

checks.d/statsd.py
```diff
@@ -0,0 +1,78 @@
...
+            buf = StringIO()
+
+            chunk = s.recv(1024)
+            buf.write(chunk)
+            while chunk:
+                if ENDER.search(chunk):
+                    break
+                if BAD_ENDER.search(chunk):
+                    self.service_check(
+                        SERVICE_CHECK_NAME, AgentCheck.CRITICAL, []
```
remh: if this happens, you'll send twice the same service check (here and on line 73)

checks.d/statsd.py
```diff
@@ -0,0 +1,78 @@
...
+                    )
+                    raise Exception("Got an error issuing command: {0}".format(command))
+                chunk = s.recv(1024)
+                buf.write(chunk)
+            return buf
+        except Exception as e:
+            self.service_check(
+                SERVICE_CHECK_NAME, AgentCheck.CRITICAL, []
```
remh: It should be tagged as well

remh:
Thanks a lot Cory, this will be pretty useful!

remh:
Regarding your questions:

`I used two different service check names for failure and for "health". Is this ok?`

Yes, this is fine.

`I used gauge for all the things, even though uptime and messages.bad_lines_seen will only increase unless StatsD restarts. Which metric type is appropriate here?`

Yeah uptime should be reported as a gauge,
bad_lines_seen should likely be reported as a monotonic_count though.

checks.d/statsd.py
gphat: I've got `_send_command` doing a `raise` if there are any failures, so the whole thing will error out. Do I need to do more / am I misunderstanding? :)

iow, there's nothing to check because `_send_command` shouldn't return (without an exception) unless it worked.

gphat:
Ugh, sorry for all the problems @remh. I'm apparently sloppy!

I've factored in all your feedback and verified that things still work locally before sending this along. Lemme know if you find anything else!

remh:
@gphat Thanks for implementing the changes ?
uptime should be as a gauge as by my previous comment, there is no great other metric type for it i think.
bad_lines_seen should be the monotonic_count though.

gphat:
Oops! I misread the comment @remh. Fixed!

gphat:
Added tags to the service checks @remh!

gphat:
Friendly ping to make sure you don't need anything else from me here!

remh:
Thanks a lot @gphat can you please squash your commits into one ? I'll merge once it's done!

gphat:
What!? You don't want all my trash commits!? :)

Will do. Looks like I branched off the UDP branch or something. Will rebase.

gphat:
Yeah, ok I am terrible at Git and have borked this. The diff and stuff is right, but the merge is not gonna be clean. Evidently I don't know how to rebase correctly.

How would you like me to proceed? :cry:

gphat:
(I think my fuckup was branching from the UDP feature branch which got merged and now does weird things when I try and `git rebase -i master`)

checks.d/statsd.py
```diff
@@ -0,0 +1,81 @@
+# stdlib
+import re
+import socket
+import sys
```
irabinovitch: ./checks.d/statsd.py:4:1: F401 'sys' imported but unused

checks.d/statsd.py
```diff
@@ -0,0 +1,81 @@
...
+                if ENDER.search(chunk):
+                    break
+                if BAD_ENDER.search(chunk):
+                    raise Exception("Got an error issuing command: {0}".format(command))
+                chunk = s.recv(1024)
+                buf.write(chunk)
+            return buf
+        except Exception as e:
+            self.service_check(
+                SERVICE_CHECK_NAME, AgentCheck.CRITICAL, tags
```
irabinovitch: ./checks.d/statsd.py:77:58: F821 undefined name 'tags'

gphat: Fixed by passing tags in. Thanks and sorry for missing that. :(

gphat:
Squashed!

remh:
Thanks @gphat !

gphat:
Woot. Thank **you**!

PR #1979, opened 2015-10-16
[mongo] add tcmalloc metrics

yannmh:
Rebased of #1832 on master, thanks a lot @benmccann !

My two cents: 
Split the metric list in multiple parts:
* `COMMON_METRICS`: default metrics to collect
* `V2_ONLY_METRICS`: metrics available in MongoDB 2.x only
* `TCMALLOC_METRICS`: TCMalloc memory allocator metrics

It's easier to assess the check coverage in our CI, and gives more
control on the level of metric collection.

Create a `collect_tcmalloc_metrics` option (default to False).

### DIFF
--- a/checks.d/mongo.py
+++ b/checks.d/mongo.py
@@ -16,12 +16,12 @@ class MongoDb(AgentCheck):
     SOURCE_TYPE_NAME = 'mongodb'
 
     GAUGES = [
-        # L21-23,25,68-74 Deprecated as if V 3.0.0
+        # L21-24 removed as of V 3.0.0
         "indexCounters.btree.missRatio",
         "indexCounters.missRatio",
         "globalLock.ratio",
-        "globalLock.totalTime",
         "globalLock.lockTime",
+        "globalLock.totalTime",
         "globalLock.currentQueue.total",
         "globalLock.currentQueue.readers",
         "globalLock.currentQueue.writers",
@@ -40,7 +40,6 @@ class MongoDb(AgentCheck):
         "cursors.timedOut",
         "uptime",
 
-
         "stats.collections",
         "stats.objects",
         "stats.avgObjSize",
@@ -52,8 +51,6 @@ class MongoDb(AgentCheck):
         "stats.fileSize",
         "stats.nsSizeMB",
 
-
-
         "replSet.health",
         "replSet.state",
         "replSet.replicationLag",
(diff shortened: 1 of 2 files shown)


checks.d/mongo.py
```diff
@@ -6,124 +6,132 @@
...
+        "stats.indexes": GAUGE,
+        "stats.indexSize": GAUGE,
+        "stats.nsSizeMB": GAUGE,
+        "stats.numExtents": GAUGE,
+        "stats.objects": GAUGE,
+        "stats.storageSize": GAUGE,
+        "uptime": GAUGE,
+    }
+
+    V2_ONLY_METRICS = {
```
benmccann: I don't think this is the entire list of metrics available only in mongo 2.x. E.g. I don't think that Mongo 3.x will return "cursors.totalOpen" or "cursors.timedOut"

checks.d/mongo.py
```diff
@@ -200,11 +243,20 @@ def check(self, instance):
...
         db_name = parsed.get('database')
-        clean_server_name = server.replace(password, "*"*5) if password is not None else server
+        clean_server_name = server.replace(password, "*" * 5) if password is not None else server
 
         tags = instance.get('tags', [])
         tags.append('server:%s' % clean_server_name)
 
+        # Get the list of metrics to collect
+        collect_tcmalloc_metrics = _is_affirmative(
+            instance.get('collect_tcmalloc_metrics', False)
```
benmccann: If we're going to add an option for collect_tcmalloc_metrics then we can use this to decide whether to request tcmalloc stats on older versions of MongoDB as well. They're starting to return them in 3.1.8 and later by default, but you can request them on older versions of MongoDB by passing `tcmalloc:true` when calling serverStats on line 316.

https://jira.mongodb.org/browse/SERVER-19941

checks.d/mongo.py
yannmh: I just ran some tests against a MongoDB 2.6 and 3.0 instances, the `V2_ONLY_METRICS ` seens exhaustive.
`cursors.totalOpen` or `cursors.timedOut` are still part of MongoDB 3.0 as stated in the [MongoDB 3.0 documentation] (https://docs.mongodb.org/manual/reference/command/serverStatus/#serverStatus.cursors.totalOpen)

yannmh:
Thanks for your feedback @benmccann. I adressed your comments and rebased the PR.

It's ready to be merged now :ship:

checks.d/mongo.py
yannmh: That's good to know, thanks @benmccann. Added the change to my commit.

PR #1982, opened 2015-10-16
[mcache] report `listen_disabled_num`

### LINKED ISSUE #1977: [mcache] report `listen_disable_num`
I propose to report `listen_disable_num` value from Memcached statistics. It's of much value when diagnosing connection errors, etc. [memcached Wiki](https://code.google.com/p/memcached/wiki/Timeouts) recommends consulting this value at the first when troubleshooting timeouts. So it would make sense to monitor the increase of such events.

alaz:
[closes #1977]

### DIFF
--- a/checks.d/mcache.py
+++ b/checks.d/mcache.py
@@ -46,6 +46,8 @@ from checks import AgentCheck
 #                            use for storage.
 # threads           32u      Number of worker threads requested.
 #                            (see doc/threads.txt)
+# listen_disabled_num 32u    How many times the server has reached maxconns
+#                            (see https://code.google.com/p/memcached/wiki/Timeouts)
 #     >>> mc.get_stats()
 # [('127.0.0.1:11211 (1)', {'pid': '2301', 'total_items': '2',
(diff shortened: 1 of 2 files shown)


checks.d/mcache.py
```diff
@@ -77,7 +79,8 @@ class Memcache(AgentCheck):
         "curr_connections",
         "connection_structures",
         "threads",
-        "pointer_size"
+        "pointer_size",
+        "listen_disabled_num"
```
hkaj: as far as I understand, `listen_disabled_num` is an uncapped counter. Reporting values that could potentially grow to infinity is not good practice. It would be more relevant to collect it as a rate (difference between the previous value and the current one). Rates also allow you to draw this kind of graphs:

screen shot 2015-10-23 at 1 59 15 pm

Instead of a simple line growing slowly:
image

It also makes alerting easier (setting a threshold value wouldn't be as effective on a gauge as on a rate).

hkaj:
Hi @alaz, thanks for taking care of this :tada: just made a short comment, let me know what you think.

alaz:
Hi,

Oh yes, good point. I shall re-submit the patch.

Thanks!

PR #1983, opened 2015-10-16
[core] Hot reloading of integration configs

JohnLZeller:
This adds the ability to send a SIGUSR2 signal to the collector and successfully have it reload the `conf.d` integration configs.

Also note, this does not yet work on windows.

### DIFF
--- a/agent.py
+++ b/agent.py
@@ -68,8 +68,14 @@ class Agent(Daemon):
         self.collector = None
         self.start_event = start_event
         self.in_developer_mode = in_developer_mode
+        self._agentConfig = {}
+        self._checksd = []
+        self.collector_profile_interval = DEFAULT_COLLECTOR_PROFILE_INTERVAL
+        self.check_frequency = None
+        self.configs_reloaded = False
 
     def _handle_sigterm(self, signum, frame):
+        """Handles SIGTERM and SIGINT, which gracefully stops the agent."""
         log.debug("Caught sigterm. Stopping run loop.")
         self.run_forever = False
 
@@ -78,8 +84,31 @@ class Agent(Daemon):
         log.debug("Collector is stopped.")
 
     def _handle_sigusr1(self, signum, frame):
+        """Handles SIGUSR1, which signals an exit with an autorestart."""
         self._handle_sigterm(signum, frame)
         self._do_restart()
+
+    def _handle_sighup(self, signum, frame):
+        """Handles SIGHUP, which signals a configuration reload."""
+        log.info("SIGHUP caught!")
+        self.reload_configs()
+        self.configs_reloaded = True
+
+    def reload_configs(self):
+        """Reloads the agent configuration and checksd configurations."""
+        log.info("Attempting a configuration reload...")
+
+        # Reload checksd configs
+        hostname = get_hostname(self._agentConfig)
+        self._checksd = load_check_directory(self._agentConfig, hostname)
+
+        # Logging
+        num_checks = len(self._checksd['initialized_checks'])
+        if num_checks > 0:
+            log.info("Successfully reloaded {num_checks} checks".
+                     format(num_checks=num_checks))
+        else:
+            log.info("No checksd configs found")
 
     @classmethod
     def info(cls, verbose=None):
@@ -98,6 +127,9 @@ class Agent(Daemon):
         # Handle Keyboard Interrupt
         signal.signal(signal.SIGINT, self._handle_sigterm)
 
+        # A SIGHUP signals a configuration reload
+        signal.signal(signal.SIGHUP, self._handle_sighup)
+
         # Save the agent start-up stats.
         CollectorStatus().persist()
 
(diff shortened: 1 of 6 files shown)


JohnLZeller:
Ready for a first round of quick feedback @hkaj 
Still going to add the functions for the init.d script so it's callable. I will push those in a sec

packaging/debian/datadog-agent.init
```diff
@@ -151,6 +152,11 @@ case "$1" in
         exit $?
         ;;
 
+    reload)
+        kill -12 `cat $COLLECTOR_PIDFILE`
```
remh: Watchout SIGUSR2 is not always 12, this is platform dependant. You should use `kill -SIGUSR2 instead`

checks/collector.py
```diff
@@ -253,6 +253,8 @@ def run(self, checksd=None, start_event=True):
         """
         Collect data from each check and submit their data.
         """
+        log.info("Start of Collector.run()")
```
remh: We don't want these extra log lines at every iteration of the collection

JohnLZeller: I only left in logging to make it easier for reviewers to do testing, I will remove them in the final review.

packaging/debian/datadog-agent.init
JohnLZeller: yeah, missed that :) Thanks

agent.py
```diff
@@ -255,6 +292,7 @@ def main():
     if command in START_COMMANDS:
         log.info('Agent version %s' % get_version())
 
+    # MAYBE: Add reload command.. problem is that when called this way, agent.collector is None
```
LeoCavaille: I think it's a no, we were trying to move away from these ad hoc commands, basically separate the management of the processes from the processes themselves.

agent.py
```diff
@@ -131,6 +162,10 @@ def run(self, config=None):
 
         # Run the main loop.
         while self.run_forever:
+            log.info("Found {0} initialized checks".format(len(self._checksd['initialized_checks'])))
+            for che in self._checksd['initialized_checks']:
+                log.info("\t{0}".format(che))
```
LeoCavaille: Logging this way is a bit of an anti-pattern, usually prefer:
```python
log.info("Found %d checks", len(checks))
```

agent.py
```diff
@@ -78,9 +83,32 @@ def _handle_sigterm(self, signum, frame):
...
+        # Reload checksd configs
+        hostname = get_hostname(self._agentConfig)
+        self._checksd = load_check_directory(self._agentConfig, hostname)
+
+        # Logging
+        num_checks = len(self._checksd['initialized_checks'])
+        if num_checks > 0:
+            log.info("SIGUSR2: Successfully reloaded {num_checks} checks".
+                     format(num_checks=num_checks))
+            for che in self._checksd['initialized_checks']:
```
LeoCavaille: nitpick: when there is no real need to shorten the variable names, there is no need :) always prefer clarity to brevity if the performance is the same

agent.py
```diff
@@ -78,9 +83,32 @@ def _handle_sigterm(self, signum, frame):
         log.debug("Collector is stopped.")
 
     def _handle_sigusr1(self, signum, frame):
+        """Handles SIGUSR1, which signals an exit with an autorestart."""
         self._handle_sigterm(signum, frame)
         self._do_restart()
 
+    def _handle_sigusr2(self, signum, frame):
```
LeoCavaille: I guess I would expect `SIGHUP` to implement such behavior in lieu of `SIGUSR2`.

LeoCavaille:
Nice! Was not assigned on this one, but I was very excited to see this dream is finally real so I dropped a few remarks. :rocket:

hkaj:
Great work @JohnLZeller ! Tested it on Ubuntu and CentOS, worked like a charm.
I don't have anything to add to the comments from Léo and Rémi except for the logging. Some `log.info` should be kept to ease troubleshooting in the future, but it should be short (no loop over the checks), yet helpful. Maybe a message when entering the signal handler, and another one just after reloading the configs.

packaging/debian/datadog-agent.init
remh: as leo pointed out using SIGHUP has actually became a de facto standard, we should use it.

JohnLZeller:
Thanks for the feedback everyone! :dancers:

What do we think about using `SIGUSR2` vs `SIGHUP`? I'm reluctant to using `SIGHUP` only because it seems more common than `SIGUSR2`.

packaging/centos/datadog-agent.init
```diff
@@ -150,6 +151,14 @@ info() {
     exit $(($FORWARDER_RETURN+$COLLECTOR_RETURN+DOGSTATSD_RETURN))
 }
 
+reload() {
+    kill -HUP `cat $COLLECTOR_PIDFILE` 2>/dev/null
+    if [ $? -ne 0 ]; then
+        kill -SIGHUP `cat $COLLECTOR_PIDFILE`
```
hkaj: Not the cleanest code we have, but that will work on platforms that only have one or the other. :+1:

JohnLZeller:
@hkaj Using `-1` instead of `-SIGHUP` or `-HUP`, because it seems that 1 is very consistently the HUP signal, and having to silence errors to try both is just too sloppy.

Lemme know what you think.

hkaj:
@JohnLZeller actually we should use `-HUP`. We ship a `kill` command with the Linux agent, and it's based on `procps`, so `kill` will be consistent everywhere for the agent. Also mac os supports `-HUP` so we don't to make an exception there.
https://github.com/DataDog/dd-agent-omnibus/blob/master/config/projects/datadog-agent.rb#L105-L106

JohnLZeller:
@hkaj changed to `HUP` and squashed commits. g2g :)

hkaj:
Awesome, let's do it then!

remh:
Watchout though, the kill command you call in the init script is not the one we ship unless you change the PATH.

JohnLZeller:
Got it. Using the embedded kill commands in all scipts

hkaj:
good catch @remh, lgtm now.

PR #1984, opened 2015-10-19
[docker_daemon] Do not run check if init failed

remh:


remh:
@hkaj can you review please ?

hkaj:
Good idea. Most complaints we receive about docker are simply due to an incomplete initialization of the check. :ship: it

PR #1985, opened 2015-10-19
[status page] Make status page use user defined ntp settings

remh:


truthbk:
Looks good to me! Should we log what ntp server instance is actually being used given we always take the first one?

remh:
It's already logged when doing the actual ntp calls. Thanks for the review @truthbk !

PR #1986, opened 2015-10-20
[check] [network] Test if UDP metrics exist before trying to report them

hkaj:
As @olivielpeau [mentioned](https://github.com/DataDog/dd-agent/pull/1974#issuecomment-149275279) on https://github.com/DataDog/dd-agent/pull/1974, some of these can be missing in old Linux versions.
We want to make sure we found them before submitting them.


remh:
:+1:

PR #1988, opened 2015-10-20
[core] Normalize exit code of fallback `status` command

### LINKED ISSUE #1976: 3.18 kernel detection breaks ansible service hooks
a5dda4a19a1a6c71d1ceecf7a665c1ea7d928a4d by @JohnLZeller appears to be solving a problem related to overlayfs and the kernel.

However not everyone on 3.18 is using overlayfs.  This check is creating a new problem.

The ansible-datadog role, https://github.com/DataDog/ansible-datadog uses the ansible core module for service management here https://github.com/DataDog/ansible-datadog/blob/master/tasks/main.yml#L20

This commit breaks it, since the service module expects an output it doesn't get on the `info` call.

I'm using 14.04.3 LTS Ubuntu, nothing to do with CoreOS and OverlayFS

olivielpeau:
The exit code of the `status` command has to be one of 0, 1, 2, 3, 4,
69 to be correctly understood by Ansible, but was returning `253`
when the agent was not running on linux kernel versions 3.18+

Normalize the exit code to 0 or 1 to avoid the issue.

See https://github.com/ansible/ansible-modules-core/blob/cbbb20b6bf0d82fdc92f064c9e4d18c04f203da6/system/service.py#L590

Fixes #1976

More generally we could try to stick more closely to the [LSB specs](http://refspecs.linuxbase.org/LSB_4.1.0/LSB-Core-generic/LSB-Core-generic/iniscrptact.html) for the exit code of the `status` command.

### DIFF
--- a/packaging/centos/datadog-agent.init
+++ b/packaging/centos/datadog-agent.init
@@ -178,6 +178,9 @@ case "$1" in
             echo "${RED}Warning: Known bug in Linux Kernel 3.18+ causes 'status' to fail.${NC}"
             echo "Calling 'info', instead..."
             service datadog-agent info
+            if [ $? -ne 0 ]; then
+                exit 1
+            fi
         else
             check_status
         fi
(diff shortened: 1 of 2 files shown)


remh:
:+1:

PR #1826, opened 2015-08-11
Add support for MemAvailable for newer Linux kernels.

jraede:
See discussion at https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/commit/?id=34e431b0ae398fc54ea69ff85ec700722c9da773 - MemAvailable is preferable to manually calculating using MemFree, Buffers, Cached, Shared, etc.

PR #1993, opened 2015-10-22
Adding support for MemAvailable

truthbk:
Based on: https://github.com/DataDog/dd-agent/pull/1826 - bringing in @jraede changes and making sure it's only reported if the kernel actually has it in `/proc/meminfo`

### DIFF
--- a/checks/system/unix.py
+++ b/checks/system/unix.py
@@ -303,6 +303,7 @@ class Memory(Check):
             # $ cat /proc/meminfo
             # MemTotal:        7995360 kB
             # MemFree:         1045120 kB
+            # MemAvailable:    1253920 kB
             # Buffers:          226284 kB
             # Cached:           775516 kB
             # SwapCached:       248868 kB
@@ -361,6 +362,7 @@ class Memory(Check):
             try:
                 memData['physTotal'] = int(meminfo.get('MemTotal', 0)) / 1024
                 memData['physFree'] = int(meminfo.get('MemFree', 0)) / 1024
+                memData['physAvailable'] = int(meminfo.get('MemAvailable', 0)) / 1024
                 memData['physBuffers'] = int(meminfo.get('Buffers', 0)) / 1024
                 memData['physCached'] = int(meminfo.get('Cached', 0)) / 1024
                 memData['physShared'] = int(meminfo.get('Shmem', 0)) / 1024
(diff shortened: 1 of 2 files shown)


remh:
This will need some changes in the backend as well. 
@olivielpeau i think you've done similar things in the past when you add guest cpu metrics. Can you work with @truthbk on that please ?

truthbk:
I was adding the MemAvailable stuff to the backend, but it kind of feels like overkill. Maybe we should just report MemAvailable as MemUsable when its available, and compute usable when it's not available (aka. like it's done now). Thoughts?

truthbk:
It doesn't make sense to report usable and available. We just use available if the kernel has it, or compute usable if it's not.

checks/system/unix.py
```diff
@@ -361,13 +365,17 @@ def check(self, agentConfig):
             try:
                 memData['physTotal'] = int(meminfo.get('MemTotal', 0)) / 1024
                 memData['physFree'] = int(meminfo.get('MemFree', 0)) / 1024
+                if 'MemAvailable' in meminfo:
+                    memData['physUsable'] = int(meminfo.get('MemAvailable', 0)) / 1024
+                else:
+                    # Usable is relative since cached and buffers are actually used to speed things up.
+                    memData['physUsable'] = memData['physFree'] + memData['physBuffers'] + memData['physCached']
```
talwai: doesn't look like `physBuffers` and `physCached` are populated until L374-375 . Maybe move this block of code below that?

truthbk: Thanks for spotting that! Careless.

talwai:
:+1:

remh:
@talwai @truthbk Did you check if those metrics were roughly equivalent ? We need to make sure that they'll be concistent with the previous value.

truthbk:
@remh the thing is MemAvailable was introduced precisely because it's not always so straight-forward to estimate the amount of memory available to the system: https://access.redhat.com/solutions/776393. As it turns out the typical way of doing `free + cached + buffers` didn't always reflect the actual memory available.

JohnLZeller:
@remh can +1 what @truthbk is saying here. Looks like the MemAvailable number is lower than our calculation (ex `memavail: 529  VS  calculated: 663`), meaning that our manual calculation is actually off a bit.

PR #1997, opened 2015-10-23
[sqlserver] Close database connections

JohnLZeller:
If we leave database connections open it can cause the SQL Server Agent to hang when attempting to stop.

### DIFF
--- a/checks.d/sqlserver.py
+++ b/checks.d/sqlserver.py
@@ -88,6 +88,8 @@ class SQLServer(AgentCheck):
                 self.log.exception("Skipping SQL Server instance")
                 continue
 
+        self.close_db_connections()
+
     def _make_metric_list_to_collect(self, instance, custom_metrics):
         """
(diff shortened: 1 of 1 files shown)


checks.d/sqlserver.py
```diff
@@ -190,6 +192,22 @@ def _conn_string(self, instance):
             conn_str += 'Integrated Security=SSPI;'
         return conn_str
 
+    def _conn_key_to_string(self, conn_key):
```
hkaj: can't we just add that to `_conn_string`? We could change it to `_conn_string(self, **kwargs)` and get the DB info with either `self._get_access_info(instance)` or `conn_key` depending on what's in `**kwargs`.

If we can avoid duplicating code that would be great!

JohnLZeller:
@hkaj I chose to modify the _conn_string function to explicitly refer to the 2 possibilities instead. Let me know what you think

hkaj:
lgtm

PR #2000, opened 2015-10-26
[core][transaction] Use utcnow instead of now

remh:
If the server is configured in a local timezone that follows DST, the
forwarder can get stuck if we use .now instead of .utcnow as it will
think that there is already a transaction in progress.


yannmh:
Good to go !

LeoCavaille:
#2000 :fireworks:

PR #2001, opened 2015-10-26
[integration] [openstack] Add project limit collection and implement agent scoping

talwai:
This PR is a rebased and updated version of @mtougeron's PR #1966 
It also incorporates a few other miscellaneous changes and fixes (apologies to reviewers for shoddy commits)

It adds the following:
- Collection of project-level metrics related to limits and quotas, courtesy @mtougeron
- Caching of host aggregates, which change infrequently, in between check runs. This allows us to query 
for `os-aggregates` only every `CACHE_TTL` seconds, rather than at the standard interval of 15s
- Proper failover of endpoints in the service catalog - we now correctly accept `internal` endpoints for 
Nova and Neutron when `public` endpoints are not present
- **IMPT - a new default** for the metric collection - local-only scoping. This is to handle the tricky case of ensuring that no two agents running in the same cluster accidentally duplicate the same metrics. This default ensures that the agent will collect metrics local to its `(physical_host, hypervisor, project)` along with guest servers contained in that scope. Any external servers / hypervisors / projects previously captured by the `check_all_*` directives will cease to be monitored. This is to ensure a manageable workload for the agent and allows it to run with fewer global permissions. The prior behavior can still be maintained by setting `local_only: false` in the `init_config` for this check.
- Also ensures that metrics from local-only servers / hypervisors as collected in the default scenario, have host-level tags propagated to them to allow for better aggregation on the backend

cc @mtougeron , let me know what you think!

### DIFF
--- a/checks.d/openstack.py
+++ b/checks.d/openstack.py
@@ -388,23 +388,31 @@ class OpenStackCheck(AgentCheck):
                 self.warning('Unable to get stats for hypervisor {0}: {1}'.format(hyp, str(e)))
 
     def get_all_hypervisor_ids(self, filter_by_host=None):
-        url = '{0}/os-hypervisors'.format(self._nova_url)
-        headers = {'X-Auth-Token': self._auth_token}
+        nova_version = self.init_config.get("nova_api_version", self.DEFAULT_NOVA_API_VERSION)
+        if nova_version == "v2.1":
+            url = '{0}/os-hypervisors'.format(self._nova_url)
+            headers = {'X-Auth-Token': self._auth_token}
 
-        hypervisor_ids = []
-        try:
-            resp = self._make_request_with_auth_fallback(url, headers, verify=self._ssl_verify)
-            hv_list = resp.json()
-            for hv in hv_list['hypervisors']:
-                if filter_by_host and hv['hypervisor_hostname'] == filter_by_host:
-                    # Assume one-one relationship between hypervisor and host, return the 1st found
-                    return [hv['id']]
+            hypervisor_ids = []
+            try:
+                resp = self._make_request_with_auth_fallback(url, headers, verify=self._ssl_verify)
+                hv_list = resp.json()
+                for hv in hv_list['hypervisors']:
+                    if filter_by_host and hv['hypervisor_hostname'] == filter_by_host:
+                        # Assume one-one relationship between hypervisor and host, return the 1st found
+                        return [hv['id']]
 
-                hypervisor_ids.append(hv['id'])
-        except Exception as e:
-            self.warning('Unable to get the list of all hypervisors: {0}'.format(str(e)))
+                    hypervisor_ids.append(hv['id'])
+            except Exception as e:
+                self.warning('Unable to get the list of all hypervisors: {0}'.format(str(e)))
 
-        return hypervisor_ids
+            return hypervisor_ids
+        else:
+            if not self.init_config.get("hypervisor_ids"):
+                self.warning("Nova API v2 requires admin privileges to index hypervisors." +\
+                             "Please specify the hypervisor you wish to monitor under the `hypervisor_ids` section")
+                return []
+            return self.init_config.get("hypervisor_ids")
 
     def get_all_aggregate_hypervisors(self):
         url = '{0}/os-aggregates'.format(self._nova_url)
@@ -631,8 +639,6 @@ class OpenStackCheck(AgentCheck):
                     hyp = self.get_local_hypervisor()
                     project = self.get_scoped_project()
 
-                    aggregate = self.get_aggregates_for_local_host()
-
                     # Restrict monitoring to hyp and servers, don't do anything else
                     servers = self.get_servers_managed_by_hypervisor()
                     host_tags = self._get_tags_for_host()
@@ -641,7 +647,9 @@ class OpenStackCheck(AgentCheck):
                         server_tags = host_tags + ['host:%s' % sid]
                         self.get_stats_for_single_server(sid, tags=server_tags)
 
-                    self.get_stats_for_single_hypervisor(hyp, host_tags=host_tags)
+                    if hyp:
+                        self.get_stats_for_single_hypervisor(hyp, host_tags=host_tags)
+
                     if project:
                         self.get_stats_for_single_project(project)
                 else:
@@ -666,7 +674,10 @@ class OpenStackCheck(AgentCheck):
         # Look up hypervisors available filtered by my hostname
         host = self.get_my_hostname()
         hyp = self.get_all_hypervisor_ids(filter_by_host=host)
-        return hyp[0]
+        if hyp:
+            return hyp[0]
+        else:
+            return None
 
     def get_scoped_project(self):
         """
@@ -680,8 +691,8 @@ class OpenStackCheck(AgentCheck):
             self._get_auth_scope()
 
         filter_params = {
-                "name": self._project_name,
-                "domain_id": self._domain_id
+            "name": self._project_name,
+            "domain_id": self._domain_id
         }
 
 
@@ -702,10 +713,6 @@ class OpenStackCheck(AgentCheck):
     def get_my_hostname(self):
         return socket.gethostname()
 
-    def get_aggregates_for_local_host(self):
-        aggregate_list = self._get_and_set_aggregate_list()
-        return aggregate_list[self.get_my_hostname()]
-
     def get_servers_managed_by_hypervisor(self):
         return self.get_all_server_ids(filter_by_host=self.get_my_hostname())
 
@@ -718,6 +725,8 @@ class OpenStackCheck(AgentCheck):
             # Need to check if there is a value for availability_zone because it is possible to have an aggregate without an AZ
             if self._aggregate_list[hostname]['availability_zone']:
                 tags.append('availability_zone:{0}'.format(self._aggregate_list[hostname]['availability_zone']))
+        else:
+            self.log.info('Unable to find hostname {0} in aggregate list. Assuming this host is unaggregated'.format(hostname))
 
         return tags
     ####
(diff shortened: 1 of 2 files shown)


checks.d/openstack.py
```diff
@@ -512,3 +650,57 @@ def check(self, instance):
...
+
+
+    #### Local Info accessors
+    def get_local_hypervisor(self):
+        # Look up hypervisors available filtered by my hostname
+        host = self.get_my_hostname()
+        hyp = self.get_all_hypervisor_ids(filter_by_host=host)
+        return hyp[0]
+
+    def get_local_project(self):
```
mtougeron: I'm not sure I understand this method. Projects by definition are global and not specific to a hypervisor.

checks.d/openstack.py
```diff
@@ -512,3 +650,57 @@ def check(self, instance):
...
+            self.warning('Unable to get the list of all project ids: {0}'.format(str(e)))
+
+        return projects[0]
+
+    def get_my_hostname(self):
+        return socket.gethostname()
+
+    def get_aggregates_for_local_host(self):
+        aggregate_list = self._get_and_set_aggregate_list()
+        return aggregate_list[self.get_my_hostname()]
```
mtougeron: It is unlikely, but possible, for the hostname to not appear in the aggregate list. e.g., openstack thinks the HV name is `node-24` but the fqdn in datadog is `node-24.my-company.com`

```
        if self.get_my_hostname() in aggregate_list:
            return aggregate_list[self.get_my_hostname()]
        else:
            self.warning('Unable to find hostname {0} in aggregate list'.format(self.get_my_hostname()))
            return {}
```

checks.d/openstack.py
```diff
@@ -501,10 +616,33 @@ def check(self, instance):
...
+
+                    aggregate = self.get_aggregates_for_local_host()
+
+                    # Restrict monitoring to hyp and servers, don't do anything else
+                    servers = self.get_servers_managed_by_hypervisor()
+                    host_tags = self._get_tags_for_host()
+
+                    for sid in servers:
+                        server_tags = host_tags + ['host:%s' % sid]
+                        self.get_stats_for_single_server(sid, tags=server_tags)
```
mtougeron: Need to try/except this block in case there is an error getting stats for a single server. (Not sure why but one of our instances is failing this call. I think it is in a corrupt state). Same for in the `get_server_stats` method.

checks.d/openstack.py
```diff
@@ -427,20 +499,21 @@ def get_server_stats(self):
         for sid in server_ids:
             self.get_stats_for_single_server(sid)
 
-    def get_all_server_ids(self):
-        url = '{0}/servers'.format(self._nova_url)
+    def get_all_server_ids(self, filter_by_host=None):
+        url = '{0}/servers{1}'.format(self._nova_url, "?host=%s" % filter_by_host if filter_by_host else '')
```
mtougeron: If you are getting a list of all instances, you will need to have `all_tenants=1` in the url

checks.d/openstack.py
```diff
@@ -512,3 +650,57 @@ def check(self, instance):
...
+        return projects[0]
+
+    def get_my_hostname(self):
+        return socket.gethostname()
+
+    def get_aggregates_for_local_host(self):
+        aggregate_list = self._get_and_set_aggregate_list()
+        return aggregate_list[self.get_my_hostname()]
+
+    def get_servers_managed_by_hypervisor(self):
```
mtougeron: Instead of calling the get_all_server_ids method and filtering by the host, I'd recommend calling the hypervisor servers endpoint. It is *significantly* faster.

`url = '{0}/os-hypervisors/{1}/servers'.format(self._nova_url, hypervisor_id)`

conf.d/openstack.yaml.example
```diff
@@ -62,6 +62,15 @@ init_config:
       # need to set to False when using self-signed certs
       # ssl_verify: true
 
+      #IDs of Projects (aka Tenants) to monitor
+      project_ids:
+          - 2f60042a0d464023905aaebf23a350ff
```
mtougeron: This one is my bad. That should probably be commented out like all the other examples.

checks.d/openstack.py
```diff
@@ -501,10 +616,33 @@ def check(self, instance):
...
+                    project = self.get_local_project()
+
+                    aggregate = self.get_aggregates_for_local_host()
+
+                    # Restrict monitoring to hyp and servers, don't do anything else
+                    servers = self.get_servers_managed_by_hypervisor()
+                    host_tags = self._get_tags_for_host()
+
+                    for sid in servers:
+                        server_tags = host_tags + ['host:%s' % sid]
```
mtougeron: If you are setting the `host` tag here to be the instance id, should do the same when when it isn't in _local_only mode. Can handle both situations if the tag is appended in the `get_stats_for_single_server` method.

checks.d/openstack.py
mtougeron: I think you'll still need to filter the list because `/os-hypervisors/node-1/servers` will return servers with node-1*. e.g., node-1, node-10, node-11, etc.

checks.d/openstack.py
talwai: Agreed, it is poorly named. What it means to do is "get the project that this check is scoped to", which it knows from reading the configuration. Will update it to `get_scoped_project` if that sounds closer to reality

checks.d/openstack.py
talwai: Interesting. I wonder if there is a more reliable mechanism for inferring `hypervisor_hostname` from the physical host. Will implement the error handling now, and look for better implementations of `get_my_hostname`

mtougeron: Yeah, I've tried looking for a better way to do this too but I haven't come across anything in my initial searches. :/

checks.d/openstack.py
talwai: We actually catch `409` errors on [line 528](https://github.com/DataDog/dd-agent/blob/dd0917c1223855ddbbb5a61e9abb3405f2de34fc/checks.d/openstack.py#L527). Nova 2.1 API returns a 409 when the instance is powered off, and potentially other error codes for `paused` state. A bare `except Exception` wouldn't hurt though, as we generally want this portion of the check to be completely resilient to failures.

checks.d/openstack.py
talwai: Thanks, will fix

checks.d/openstack.py
talwai: Agreed, will fix

checks.d/openstack.py
talwai: Cool, did not know about that! will try out your method + filtering - hoping it will be more performant than the status quo.

checks.d/openstack.py
mtougeron: Ah, I see, that makes sense. I wonder though if it is appropriate to ignore the `check_all_projects` flag in this case. 

For example, in our environment, I plan on running this integration using the `local_only: true` flag on each of the hypervisors. But, because of the way the auth works, I would still want to have a way to gather the stats for each project. In that case, I would most likely want to disable the project check on the individual hypervisors. I would then enable *only* the project check on the controller nodes (with `local_only: true` and `check_all_projects: true` and the rest of the checks disabled).

Not sure if that'd be a common use case or even the most appropriate way to set it up.

checks.d/openstack.py
mtougeron: Yeah, Nova 2.1 is much better than 2.0. Unfortunately there are a lot of people, like us, who are still running Icehouse with Nova v2 api. :(

checks.d/openstack.py
talwai: Hm, interesting. `local_only` is a very coarse-grained option right now, a little bit by design - so that  existing users (such as yourself :)) can continue to get the old functionality just by flipping a single flag. I did not account for the possibility that `local_only` and a `check_all_*` option would be run in parallel.

Just thinking out loud about ways we can support your use case. In this version, even though a hypervisor can be involved in many projects, the agent is only scoped to a single project. This is kind of a glaring limitation and means that you would need to run one agent per unique `(hypervisor, project)` to get complete coverage of your project metrics - in the process duplicating the hypervisor-level metrics. I think the way to solve this is to have different project domains listed as `instances` in the check, rather than under `init_config`. This allows an agent running on a single host to perform `identity::get_project` on multiple projects, and allows you to deploy one agent per hypervisor in `local_only` mode without issue.

The above change should probably be made in a separate PR. For now, it looks like the best we can do is have one agent running `check_all_projects`, and a separate agent in `local_only` mode :(

Sorry for this, but you have pointed out a big improvement that should be made to the check

mtougeron: works for me

checks.d/openstack.py
talwai: Hm it appears that your method is [supported in Compute v2 extensions](http://developer.openstack.org/api-ref-compute-v2-ext.html#listHypervisorInstances) but not in [Compute v2.1](http://developer.openstack.org/api-ref-compute-v2.1.html#os-hypervisors-v2.1)

We will special case if the user specifies `nova_api_version: 'v2'` in his configuration

mtougeron: oh joy. that means that we are running a mix of v2 & v2.1 (or Mirantis did a special patch for this). :(

mtougeron: That's lame that they took that out. I wonder why. :/

checks.d/openstack.py
talwai: Actually, following off the above comments about the agent being scoped to the project, I think we should probably hold off on this. The behavior I'm seeing when `all_tenants=1` and `check_all_servers` is enabled, is that, on a multi-tenant host, the check tries to pull diagnostics for servers outside its authorization scope (i.e. in another project). This leads to `404`s. I suspect this problem will be helped when we move toward allowing multiple auth-scope `instances` per agent, but for now i think it's better to vendor this particular change if you need it.

mtougeron: Fair enough.

talwai:
Thanks for your review and feedback @mtougeron! I've made some updates, and hope I've addressed your comments - either through changes to the code or through more detailed docs. I'm going to go ahead and merge this to the mainline branch, and continue development from there

mtougeron:
Awesome! This is great. I'll update us internally to start using it in the next couple of days.

PR #2005, opened 2015-10-27
[core] improve detection of agent process from pid

degemer:
Sometimes the agent fails to restart after a "hard" reboot because the pidfile
contains the pid of a non-datadog running process, and the only check the agent
did before this commit was to check if the pid was associated with a running
process.

This commit implements a new function `is_my_process` which verifies on Linux if the
`psutil.Process.cmdline` ("command line this process has been called with") is
the good one (corresponding to the current executable). This function is used
instead of `pid_exists` when `psutil` is installed and checks the pid contained
in the pidfile, thus solving the above-mentioned problem.

-------

I tested it on Mac (both dmg install and source install), and Debian.

### DIFF
--- a/utils/process.py
+++ b/utils/process.py
@@ -1,5 +1,6 @@
 # stdlib
 import errno
+import inspect
 import os
 
 # 3p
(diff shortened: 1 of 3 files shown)


utils/process.py
```diff
@@ -9,7 +10,34 @@
...
+    if not psutil or not pid_existence:
+        return pid_existence
+
+    if Platform.is_windows():
+        # We can't check anything else on Windows
+        return True
+    else:
+        try:
+            command = psutil.Process(pid).cmdline() or []
+        except:
```
remh: You should avoid bare Exception btw

degemer: Yep, my bad. It only catches `psutil` errors now.

olivielpeau:
Looking great, thanks! I've also tested it on FreeBSD and SmartOS and it works like a charm.

Let's :shipit:

PR #2009, opened 2015-10-28
[win32_event_log] Allow optional tagging by event_id

JohnLZeller:
Add config option for allowing the user to have events tagged by event_id of the form `event_id:<num>`.

### DIFF
--- a/checks.d/win32_event_log.py
+++ b/checks.d/win32_event_log.py
@@ -61,7 +61,8 @@ class Win32EventLog(AgentCheck):
         # Save any events returned to the payload as Datadog events
         for ev in events:
             log_ev = LogEvent(ev, self.agentConfig.get('api_key', ''),
-                              self.hostname, tags, notify)
+                              self.hostname, tags, notify,
+                              self.init_config.get('tag_event_id', False))
 
             # Since WQL only compares on the date and NOT the time, we have to
             # do a secondary check to make sure events are after the last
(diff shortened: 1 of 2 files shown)


JohnLZeller:
Tested on Windows Server 2012 and it works beautifully.

JohnLZeller:
@degemer since you last touched this check, would you mind giving me a quick r? :)

conf.d/win32_event_log.yaml.example
```diff
@@ -1,4 +1,9 @@
 init_config:
+    # The (optional) tag_event_id setting will add an event id tag to each
+    # event sent from this check. Defaults to false.
+    # tag_event_id: true
+
+    tag_event_id: false
```
degemer: I think we kind of have an implicit convention that the default parameter should be commented, and nothing outside the comment, i.e. : 
```
 init_config:
    # The (optional) tag_event_id setting will add an event id tag to each
    # event sent from this check. Defaults to false.
    # tag_event_id: false
```

JohnLZeller:
@degemer fixed! All good now?

checks.d/win32_event_log.py
```diff
@@ -150,13 +151,14 @@ def _convert_event_types(self, types):
 
 
 class LogEvent(object):
-    def __init__(self, ev, api_key, hostname, tags, notify_list):
+    def __init__(self, ev, api_key, hostname, tags, notify_list, tag_event_id):
         self.event = ev
         self.api_key = api_key
         self.hostname = hostname
         self.tags = tags
```
degemer: I might be missing something, but why not use something like `self.tags = self._tags(tags, ev.EventCode) if tag_event_id else tags` ?
This way you don't need to have a `self.tag_event_id = tag_event_id`.

checks.d/win32_event_log.py
```diff
@@ -190,6 +192,15 @@ def _wmi_to_ts(self, wmi_ts):
                       second=second, microsecond=microsecond) + tz_delta
         return int(calendar.timegm(dt.timetuple()))
 
+    def _tags(self, event, tags):
```
degemer: Why not just pass the `EventCode` as argument ? (since you don't use/need `event`)

degemer:
Two last small comments, but LGTM otherwise !

JohnLZeller:
Tested, and small changes incorporated :)

Merging

PR #2010, opened 2015-10-28
Obfuscates passwords encoded in URI's

truthbk:
For lines including passwords in the URI of the form `fooproto://bar_user:somepass@somewhere....` this regex will strip somepass and substitute for asterisks.


remh:
:+1: will merge when CI is done!

Issue #1952, opened 2015-09-25
[wmi] Significant performance improvements on default Windows checks


TheCloudlessSky:
We're just starting out with Datadog and it's great. After a week of testing, I deployed the Datadog Agent to a couple of test Windows hosts (m3.medium EC2 instances). I noticed the CPU of the hosts jumping to 20-30% for 1-2 seconds every ~15 seconds (only the default checks were enabled). I checked and found that `WmiPrvSE.exe` and `ddagent.exe` were the culprits. This was reported to the support team. In the mean time, I wanted to investigate on my own. 

Because the `shell.exe` is shipped, it allowed me to easily debug the problems. I narrowed it down to the default Windows checks and the WMI calls. 

## Problems

1. There is a WMI query in [win32.Processes](https://github.com/DataDog/dd-agent/blob/76459c8e97aa2c31599b0d78b07b3ecca45c19b4/checks/system/win32.py#L43) that does not use the result.
2. The `wmi` Python package has lots of overhead to the queries.
3. The queries didn't explicitly specify the properties (too much data being returned).
4. Most of the queries use `Win32_PerfFormattedData_*` (which are **really** slow).

## Solutions

1. Remove this empty check.
2. Similar to #1917 (which I found and also support pointed me to), stop using the `wmi` package because of the large overhead.
3. Always specify the properties in the queries.
4. Use the `Win32_PerfRawData_*` classes and compute the values our self (this is the main performance gain). The MSDN documentation is a pain to work with, but once you [understand how to calculate the values](http://www.monitis.com/blog/2012/08/01/the-wmi-usage-peculiarities-at-monitoring/), it's easy and has *huge* performance improvements.

**DISCLAIMER:** I'm not primarily a Python programmer, so the code may not be idiomatic Python (sorry).

I originally implemented the changes *before* I learned about doing COM directly (from #1917). I found the `wmi` package still had a large amount of overhead. I re-implemented with direct COM and that's what I'm presenting to you today.

I created [this sample gist](https://gist.github.com/TheCloudlessSky/f5239cd0bf1991e42324). It shows the use of COM, explicit properties, and `Win32_PerfRawData_*` classes to improve performance. If this looks good, I'll create a PR integrating this with the real Windows checks. I created a `_WMISampler` class that holds the previous/current sample taken from the raw performance counter and does the calculation of the value.

## Results

On my dev laptop (Intel i7-3520M), the default Windows checks take:

- Originally: ~2200ms
[...]

PR #2011, opened 2015-10-28
[core][wmi] new WMI metric collection core

yannmh:
Original discussion thread: https://github.com/DataDog/dd-agent/issues/1952
Credits to @TheCloudlessSky (https://github.com/TheCloudlessSky) . Thanks a lot !!!

### WMI module + CounterType calculators
A lightweight Python WMI module wrapper built on top of `pywin32` and
`win32com` extensions.

**Specifications**
* Based on top of the `pywin32` and `win32com` third party extensions only
* Compatible with `Raw`* and `Formatted` Performance Data classes
    * Dynamically resolve properties' counter types
    * Hold the previous/current `Raw` samples to compute/format new values*
* Fast and lightweight
    * Avoid queries overhead
    * Cache connections and qualifiers
    * Use `wbemFlagForwardOnly` flag to improve enumeration/memory performance

*\* `Raw` data formatting relies on the avaibility of the corresponding calculator.
Please refer to `checks.lib.wmi.counter_type` for more information*

### Windows system check speedup
Speedup the Windows system check:
* Drop third party `wmi` Python package in favor of the local `checks.lib.wmi` module
* Use `PerfRawData` WMI Performance classes rather than `PerfFormattedData`

### WMI check speedup
Speedup the generic WMI check:
* Drop third party `wmi` Python package in favor of the local `checks.lib.wmi` module

### DIFF
--- a/checks.d/wmi_check.py
+++ b/checks.d/wmi_check.py
@@ -1,132 +1,285 @@
-'''
-Windows Only.
-
-Generic WMI check. This check allows you to specify particular metrics that you
-want from WMI in your configuration. Check wmi_check.yaml.example in your conf.d
-directory for more details on configuration.
-'''
-# 3rd party
-import wmi
+# stdlib
+from collections import namedtuple
 
 # project
 from checks import AgentCheck
+from checks.libs.wmi.sampler import WMISampler
 
-UP_METRIC = 'Up'
-SEARCH_WILDCARD = '*'
+WMIMetric = namedtuple('WMIMetric', ['name', 'value', 'tags'])
+
+
+class InvalidWMIQuery(Exception):
+    """
+    Invalid WMI Query.
+    """
+    pass
+
+
+class MissingTagBy(Exception):
+    """
+    WMI query returned multiple rows but no `tag_by` value was given.
+    """
+    pass
+
+
+class TagQueryUniquenessFailure(Exception):
+    """
+    'Tagging query' did not return or returned multiple results.
+    """
+    pass
 
 
 class WMICheck(AgentCheck):
+    """
+    WMI check.
+
+    Windows only.
+    """
     def __init__(self, name, init_config, agentConfig, instances):
         AgentCheck.__init__(self, name, init_config, agentConfig, instances)
-        self.wmi_conns = {}
-
-    def _get_wmi_conn(self, host, **kwargs):
-        key = "{0}:".format(host)
-        key += ":".join(str(v) for v in kwargs.values())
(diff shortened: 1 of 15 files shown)


remh:
Haven't reviewed it yet but really excited about this one!

TheCloudlessSky:
Awesome job @yannmh :star:!

1. Were you able to see similar performance improvements?
2. Given that implementors of custom checks will be using this (I already have several WMI classes I'd want to use it with), should it be possible to pass unimplemented counter type calculators to the `WMISampler`? Ideally, they'd be contributed back to the `dd-agent` repository. But, I don't think manually maintaining the previous/current sample *outside* of the `WMISampler` is a good idea (the value from `get_raw`). Perhaps we could expose it by inheritance or a dictionary parameter?

Thanks again for moving forward with this!

yannmh:
@TheCloudlessSky,

> Were you able to see similar performance improvements?

I built a custom Datadog Agent to benchmark those changes. It run both (old and new) system and WMI checks and add metrics to measure those checks' run times.

**Results** :trumpet: 
I made a Datadog screen board to visualize the results. 
screenshot 2015-11-03 11 58 33


To sum up: 
1. System check takes ~10 times less to run.
2. WMI check takes between ~5 and ~10 time less to run depending on the class data type (`RawData` vs `FormattedData`). On my machine that's ~10 seconds less to collect ~2k WMI metrics. 
2. Collecting `RawData` vs `FormattedData` shows ~2 'gain' factor.

Overall the performance gain is as significant as it was advertised before !:tada: :confetti_ball: :balloon: 


> Given that implementors of custom checks will be using this (I already have several WMI classes I'd want to use it with), should it be possible to pass unimplemented counter type calculators to the WMISampler? Ideally, they'd be contributed back to the dd-agent repository. But, I don't think manually maintaining the previous/current sample outside of the WMISampler is a good idea (the value from get_raw). Perhaps we could expose it by inheritance or a dictionary parameter?

Thanks for the suggestion. I agree, I think it would be a nice feature to have. 
We'll work on it on a separate PR and keep you informed about it !

olivielpeau:
Looking pretty awesome, let's :shipit: !

TheCloudlessSky:
Awesome @yannmh! I'm super excited about these changes landing in 5.6.0! :tada: !

irabinovitch:
@TheCloudlessSky Thank you for your excellent work on this refactor. Can you shoot me a note with your contact info via email? (ilan at datadoghq.com).  Thanks!

PR #2013, opened 2015-10-28
Enable running multiple datadog agents on the same machine.

mschulkind:
I needed to apply this patch so that I could configure a datadog agent install from source that ran alongside the system/ubuntu based install. Without this patch, if a set of configs is found in `/etc/...`, then they override the configs in the local directory tree even when installed from source.

### DIFF
--- a/config.py
+++ b/config.py
@@ -221,6 +221,14 @@ def get_config_path(cfg_path=None, os_name=None):
     if cfg_path is not None and os.path.exists(cfg_path):
         return cfg_path
 
+    # Check if there's a config stored in the current agent directory
+    try:
+        path = os.path.realpath(__file__)
+        path = os.path.dirname(path)
+        return _config_path(path)
+    except PathNotFound, e:
+        pass
+
     if os_name is None:
         os_name = get_os()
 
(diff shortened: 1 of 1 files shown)


remh:
Thanks a lot for the contribution @mschulkind !

Can you describe the use case where you need to run multiple agents on the same machine ?

mschulkind:
We've got a setup deployed on a customer's machine. They have their datadog
monitoring installed across their fleet, and we have our own monitoring for
our deployment. This means different sets of data go to different datadog
account.

remh:
@mschulkind It seems that having 2 agents running on the same machine is not the best solution for this use case. You'll have some port conflicts and use twice as many resources as usual. Can you reach out to our support team so we can discuss more in details about your use case ?

mschulkind:
It's been long enough now that I've since dug through the source and made
some small modifications to be able to run two parallel installations of
the agent, and now we don't even need it anymore anyway. I still don't
think there was a much better way to solve this.

For reference, it was really very hard to make this work. All I had to do
was change a few ports in the config, and then modify the get_confd and
config_path code to prioritize local configs over global configs, and then
anything worked. It's not clear to me why global configs ever get
prioritized. Seems like an oversight to me.

mschulkind:
I only just realized that this was my PR for the fix and not me reporting an issue. Is there a reason why this fix is not appropriate? I can't see why you would *NOT* want to apply this PR. Without this PR, there is a documented config option to use a local config, but then the code blatantly ignores it if a global config is around.

remh:
Thanks @mschulkind this does make sense. Merging it.
Thanks a lot!

PR #2015, opened 2015-10-29
Add the main_tags to the consul service_check tags

mtougeron:
Add the main_tags to the service_check tags. This will include the consul_datacenter tag in the health checks. This is needed for when you have the same service name in multiple datacenters

### DIFF
--- a/checks.d/consul.py
+++ b/checks.d/consul.py
@@ -188,7 +188,7 @@ class ConsulCheck(AgentCheck):
                 if check["ServiceID"]:
                     tags.append("service-id:{0}".format(check["ServiceID"]))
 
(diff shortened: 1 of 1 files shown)


remh:
@talwai can you take a quick pass at this one ?

mtougeron:
Is there anything I can do about that CI build failure? It looks like it is an issue with the test structure not with my code.

irabinovitch:
@mtougeron  I get the same sense here. I'm seeing this test failing on a few of the currently open PRs. We'll dig in on our end and see what we  can find.

yannmh:
:+1:

yannmh:
Thanks @mtougeron !!

PR #2021, opened 2015-10-31
[agent][dogstatsd] Add support for changing niceness on POSIX systems

armenb:
…ocesses on POSIX systems

let me know if you'd like a unit test for this, thanks :-)

### DIFF
--- a/agent.py
+++ b/agent.py
@@ -42,6 +42,7 @@ from util import (
 from utils.flare import configcheck, Flare
 from utils.jmx import jmx_command
 from utils.pidfile import PidFile
+from utils.process import renice_pid
 from utils.profile import AgentProfiler
 
 # Constants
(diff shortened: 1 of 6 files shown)


irabinovitch:
@armenb  @armenfirespotter  Thank you for the patch! Our team will review and be in touch soon with any feedback, most likely as part of the 5.7 release/review process. In the mean time, tests area always greatly appreciated!

armenb:
Hi DD team, do you have any initial comments/questions/concerns regarding this patch?  Thanks!

remh:
Thanks this is an interesting patch.

I'm not sure it makes sense to have the agent governs itself like that.

I think it would make more sense to patch your `/etc/dd-agent/supervisor.conf` file with nice/ionice commands.

What do you think ?

armenb:
remh - great idea, I think that's ultimately a better and simpler way to do this.  However, I think there is an issue with upgrading the datadog-agent via apt after doing this - it seems that the supervisor.conf file gets overwritten.

remh:
It does indeed. You could maybe manage this file through chef or something similar though. 
Or we could pass the niceness as an environment variable.

`DOGSTATSD_NICENESS=-5 /etc/init.d/datadog-agent restart`

What do you think?
Meanwhile i'm closing this PR.

olivielpeau:
Related: #3086

PR #2022, opened 2015-11-01
[go_expvar] add configuration option for custom metric namespace

### LINKED ISSUE #2016: [go_expvar] add ability to override default metric namespace
By default, the go_expvar check uses the `go_expvar` namespace. However, I'd much rather have these metrics available under the same namespace as my application metrics. To accomplish this, I'm thinking it would just be a `namespace` key within `init_config` that allows you to override the default value.

Looking at the `go_expvar` package it seems fairly simple, however I thought I'd open an issue before trying to submit a patch.

theckman:
This change adds the ability to configure a custom namespace for your `go_expvar` metrics. By default, the old value (`go_expvar`) is used. The mock tests were updated to confirm that the new behavior works as expected.

In addition to the changes around the check script, the Go utility the tests use was modified as well. A manual GC trigger was added to ensure garbage collection has run at least once.

fixes #2016

### DIFF
--- a/checks.d/go_expvar.py
+++ b/checks.d/go_expvar.py
@@ -24,7 +24,7 @@ SUPPORTED_TYPES = {
     RATE: AgentCheck.rate,
 }
 
-METRIC_NAMESPACE = "go_expvar"
+DEFAULT_METRIC_NAMESPACE = "go_expvar"
 
 
 # See http://golang.org/pkg/runtime/#MemStats
@@ -71,9 +71,10 @@ class GoExpvar(AgentCheck):
         data = self._get_data(url)
         metrics = DEFAULT_METRICS + instance.get("metrics", [])
         max_metrics = instance.get("max_returned_metrics", DEFAULT_MAX_METRICS)
-        return data, tags, metrics, max_metrics, url
+        namespace = instance.get('namespace', DEFAULT_METRIC_NAMESPACE)
+        return data, tags, metrics, max_metrics, url, namespace
 
-    def get_gc_collection_histogram(self, data, tags, url):
+    def get_gc_collection_histogram(self, data, tags, url, namespace):
         num_gc = data.get("memstats", {}).get("NumGC")
         pause_hist = data.get("memstats", {}).get("PauseNs")
         last_gc_count = self._last_gc_count[url]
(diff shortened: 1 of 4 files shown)


theckman:
I had an alternative implementation idea, which would to make it so that setting `namespace` in the configuration would prepend the namespace value while keeping `go_expvar`. Consider the following example config:

```YAML
init_config:
  namespace: testapp
```

The metric would become `testapp.go_expvar.memstats.alloc`. I almost personally prefer this prepend behavior, because I'd like to be able to put the expvar metrics under my application's namespace.

With that said, I wasn't sure which was more idiomatic with regards to agent checks so I figured I'd bring up both options.

checks.d/go_expvar.py
```diff
@@ -54,6 +54,7 @@ class GoExpvar(AgentCheck):
 
     def __init__(self, name, init_config, agentConfig, instances=None):
         AgentCheck.__init__(self, name, init_config, agentConfig, instances)
+        self._namespace = init_config.get('namespace', DEFAULT_METRIC_NAMESPACE)
```
remh: I think this should be an instance level parameter instead. 
This way if you have multiple go apps running on the same host you can assign them a different namespace.
What do you think ?

ci/resources/go_expvar/test_expvar.go
```diff
@@ -26,6 +27,44 @@ var numCalls = expvar.NewInt("num_calls")
...
+// forceAllocations is very rudimentary function to cause the Go runtime to
+// trigger garbage collection. In some situations, the CI tests for the
+// go_expvar package would fail due to the runtime not haivng run GC yet. The
+// reason this is needed is that get_gc_collection_histogram() function in
+// go_expvar.py short-circuits if there have been no GCs. This causes the
+// pause_ns metric to not be present thus causing tests to fail.
+//
+// The num parameter tells the function how many times to recurse, and size
+// is the number of random numbers to generate per iteration.
+func forceAllocations(num, size int) {
```
remh: :clap: thanks a lot !

remh:
Thanks a lot!

You should add this option to the example configuration file as well
https://github.com/DataDog/dd-agent/blob/master/conf.d/go_expvar.yaml.example

checks.d/go_expvar.py
theckman: That sounds much more forward thinking than my implementation. I'll happily adjust to make it an instance param.

theckman:
@remh Thank you for the Halloween CR. I've made the requested adjustments in an amended commit (to help keep history clean) including adding an example / blurb to the example config. Let me know if there are any further changes needed.

ci/resources/go_expvar/test_expvar.go
```diff
@@ -38,6 +77,11 @@ func HelloServer(w http.ResponseWriter, req *http.Request) {
 }
 
 func main() {
+	// test consumer requires that garbage collection has run
+	// see foceAllocations() comment for more information
+	// an a late 2013 MacBook Pro, with go1.5.1, this takes ~1 second to run
+	forceAllocations(1000, 10000)
+
```
LeoCavaille: Maybe forcing the garbage collection with the [runtime package](https://golang.org/pkg/runtime/#GC) would work too and be simpler? It's only one second so no biggie :), thank you for running the tests :100:

theckman: Something something, the path less traveled...

facepalm

A more idiomatic change has been pushed up. :smile:

irabinovitch:
@theckman Looks like tests are passing now.  I will check with the team, but I think we might be a bit late to the party to get this included in the upcoming 5.6 release.

theckman:
Bummer. Do you know when the next release is scheduled for? I'm not looking for any guarantees, just curious about the general timeline.

irabinovitch:
@theckman roughly every 6 weeks.

remh:
Thanks a lot ! This looks good !

PR #2023, opened 2015-11-02
[activemq_xml] add monitoring for standby hosts

joelvanvelden:
ActiveMQ standby hosts (for HA) are running but do not allow connections until they are failed over to. When a failover occurs we need to start capturing metrics, but until then we should silently wait.
This adds a new config parameter which if used will treat a offline host as 0 successful metrics rather than logging exceptions and presenting an "agent error" in the datadog console interface.

### DIFF
--- a/checks.d/activemq_xml.py
+++ b/checks.d/activemq_xml.py
@@ -42,27 +42,38 @@ class ActiveMQXML(AgentCheck):
         detailed_queues = instance.get("detailed_queues", [])
         detailed_topics = instance.get("detailed_topics", [])
         detailed_subscribers = instance.get("detailed_subscribers", [])
+        suppress_errors = bool(instance.get("suppress_errors", None))
 
         tags = custom_tags + ["url:{0}".format(url)]
 
         self.log.debug("Processing ActiveMQ data for %s" % url)
-        data = self._fetch_data(url, QUEUE_URL, username, password)
-        self._process_data(data, "queue", tags, max_queues, detailed_queues)
+        data = self._fetch_data(url, QUEUE_URL, username, password, suppress_errors)
+        if(data):
+            self._process_data(data, "queue", tags, max_queues, detailed_queues)
 
-        data = self._fetch_data(url, TOPIC_URL, username, password)
-        self._process_data(data, "topic", tags, max_topics, detailed_topics)
+        data = self._fetch_data(url, TOPIC_URL, username, password, suppress_errors)
+        if(data):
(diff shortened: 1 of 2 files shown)


irabinovitch:
@joelvanvelden It's been a few years since I've worked closely with ActiveMQ. is there any way via JMX or some other method to determine which host is active vs standby?

joelvanvelden:
@irabinovitch, it might be possible to use JMX to programmatically distinguish between a activemq master and its cold-standby, but i'm trying to avoid using datadog's relatively heavy JMX activemq probe and instead using the much lighter-weight XML probe (activemq_xml.py).
With activemq a cold-standby doesn't start listening on any tcp ports until it gets the lock, so can't probe it easily as it doesn't respond.

checks.d/activemq_xml.py
```diff
@@ -42,27 +42,38 @@ def check(self, instance):
...
         detailed_subscribers = instance.get("detailed_subscribers", [])
+        suppress_errors = bool(instance.get("suppress_errors", None))
 
         tags = custom_tags + ["url:{0}".format(url)]
 
         self.log.debug("Processing ActiveMQ data for %s" % url)
-        data = self._fetch_data(url, QUEUE_URL, username, password)
-        self._process_data(data, "queue", tags, max_queues, detailed_queues)
+        data = self._fetch_data(url, QUEUE_URL, username, password, suppress_errors)
+        if(data):
```
remh: why not just do:
```python
if not data and suppress_errors:
    # We don't have any data and we are suppressing errors. Let's just stop the check here
    return
```

This way you don't have to check for the data that many  times.
What do you think ?

joelvanvelden: @remh 
The three fetch_data's are independent and any of them may fail in any order.

checks.d/activemq_xml.py
```diff
@@ -42,27 +42,38 @@ def check(self, instance):
         detailed_queues = instance.get("detailed_queues", [])
         detailed_topics = instance.get("detailed_topics", [])
         detailed_subscribers = instance.get("detailed_subscribers", [])
+        suppress_errors = bool(instance.get("suppress_errors", None))
```
remh: You should use the `is_affirmative` method https://github.com/DataDog/dd-agent/blob/7a401319c6318eb63f16e68580cfacda9b373bb7/config.py#L212-L217

irabinovitch:
@joelvanvelden I was hoping to check back in and see if you had any further thoughts on @remh 's comments?

remh:
Closing in favor of #2430

PR #2024, opened 2015-11-02
[mysql] add mysql.replication.slave_running service check

c960657:
The mysql.replication.slave_running metric is boolean and is thus more suitable exposed as a service check (if I understand the concepts correctly).

This PR adds a new service check but keeps the old metric for backwards compatibility.


remh:
Thanks a lot @c960657 !

@truthbk can you add that in your refactoring please?

truthbk:
closing in favor of #2116 thanks @c960657

PR #2027, opened 2015-11-02
Mysql Handler Statistics for Index Usage

polynomial:
These statistics improve visibility into index performance
and usage. Documentation for the stats:
http://dev.mysql.com/doc/refman/5.7/en/server-status-variables.html#statvar_Handler_commit

### DIFF
--- a/checks.d/mysql.py
+++ b/checks.d/mysql.py
@@ -37,6 +37,7 @@ STATUS_VARS = {
     'Com_delete_multi': ('mysql.performance.com_delete_multi', RATE),
     'Com_replace_select': ('mysql.performance.com_replace_select', RATE),
     'Qcache_hits': ('mysql.performance.qcache_hits', RATE),
+    'Qcache_lowmem_prunes': ('mysql.performance.qcache_lowmem_prunes', RATE),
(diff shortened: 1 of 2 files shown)


remh:
Thanks a lot @polynomial we will review this for our 5.7.0 release.

irabinovitch:
@remh  @polynomial  I believe these maybe a subset of #2027 which is scheduled for 5.7.0 as well.  Thoughts on closing and following there?

polynomial:
1869 is missing Handler_prepare, otherwise that would be awesome.

irabinovitch:
@truthbk Can we get Handler_prepare added in #2116 ?

CC: @ovaistariq

ovaistariq:
@irabinovitch I think we should ad Handler_prepare to #2116 together with the other Handler_* metrics as "Additional (and optional) Status Metrics".

truthbk:
@polynomial @ovaistariq @irabinovitch these have now been added to #2116, including `Handler_prepare` which was missing. Thanks @polynomial!

irabinovitch:
@truthbk Thanks!

@polynomial  I'm going to close this in favor of #2116. Thanks for raising it!

PR #2029, opened 2015-11-02
[resources] Catch and log exceptions from the resources checks

olivielpeau:
They shouldn't make the collector exit when they raise an exception.

Also, raise an exception in the processes resource check instead of
returning False.


remh:
LGTM

PR #2030, opened 2015-11-02
Add more "core" Linux metrics

gphat:
# What's this PR do?

Adds some new system metrics:
* context switch count (for rates!)
* processes created count (for rates!)
* interrupts
* counts of processes by state, with tags
* counts of processes by priority, with tags

# Motivation

These metrics are used in our existing setup and, from what I can tell, are not available today in Datadog agent. We can get them from specific processes via `process.yaml`, but not for the whole system!

# Notes

I originally tried to add this stuff to `unix.py` but I'm unable to determine how those metrics end up as things like `sys.mem.free` and the like. So I settled on making a check that added the metrics I as after. I expect to add more.

Most of these metrics feel very "core" to Linux and therefore seem suitable to be included in something like `unix.py`. But for now I'll opt to take a conservative approach and assume this will just be a plugin. If you are not interested in having it in the agent, we'll just use it locally.

Lemme know and thanks for looking!


irabinovitch:
@gphat  Apologies for the delay here. We are mid prepping for the 5.6 release.  We will get back to you shortly with feedback on the change. In the mean time would you mind having a gander at the failing tests? It looks like at least some of them are just style issues that should be quick to address.

./checks/system/unix.py:580:55: F812 list comprehension redefines 'line' from line 572

gphat:
Done! Fyi, that change was in existing code, not in something I touched. But a fix nonetheless!

conf.d/linux_proc_extras.yaml.example
```diff
@@ -0,0 +1,4 @@
+init_config:
+
+instances:
+  - tags: []
```
remh: Can you add some comment here to say that the configuration it doesn't need any specific parameters ?

checks.d/linux_proc_extras.py
```diff
@@ -0,0 +1,87 @@
...
+    '<': 'high',
+    'N': 'low',
+    'L': 'locked'
+}
+
+class MoreUnixCheck(AgentCheck):
+    def check(self, instance):
+        tags = instance.get('tags', [])
+
+        state_counts = {
```
remh: you should just use a `collections.Counter` here

checks.d/linux_proc_extras.py
```diff
@@ -0,0 +1,87 @@
...
+            'uninterruptible': 0,
+            'runnable': 0,
+            'sleeping': 0,
+            'stopped': 0,
+            'paging': 0,
+            'dead': 0,
+            'zombie': 0
+        }
+
+        prio_counts = {
```
remh: same here

checks.d/linux_proc_extras.py
```diff
@@ -0,0 +1,87 @@
...
+                self.gauge('system.inodes.total', float(inode_stats[0]), tags=tags)
+                self.gauge('system.inodes.used', float(inode_stats[1]), tags=tags)
+
+            with open('/proc/stat', 'r') as stat_info:
+                lines = [line.strip() for line in stat_info.readlines()]
+
+                for line in lines:
+                    if line.startswith('ctxt'):
+                        ctxt_count = float(line.split(' ')[1])
+                        self.monotonic_count('linux.context_switches', ctxt_count, tags=tags)
```
remh: can you prefix your metrics with the `system` namespace please ?

checks.d/linux_proc_extras.py
```diff
@@ -0,0 +1,87 @@
...
+                state_tags = list(tags)
+                state_tags.append("state:" + state)
+                self.gauge('system.processes.states', float(state_counts[state]), state_tags)
+
+            for prio in prio_counts:
+                prio_tags = list(tags)
+                prio_tags.append("priority:" + prio)
+                self.gauge('system.processes.priorities', float(prio_counts[prio]), prio_tags)
+
+        except Exception:
```
remh: No need to catch that exception. You should just let it fail so that it would appear in the info page.

checks.d/linux_proc_extras.py
```diff
@@ -0,0 +1,87 @@
...
+                        self.monotonic_count('linux.processes_created', process_count, tags=tags)
+                    elif line.startswith('intr'):
+                        interrupts = int(line.split(' ')[1])
+                        self.monotonic_count('linux.interrupts', interrupts, tags=tags)
+
+            with open('/proc/sys/kernel/random/entropy_avail') as entropy_info:
+                entropy = entropy_info.readline()
+                self.gauge('system.entropy.available', float(entropy), tags=tags)
+
+            ps = [line.strip() for line in subprocess.Popen(['ps', '--no-header', '-eo', 'stat'], stdout=subprocess.PIPE, close_fds=True).communicate()[0]]
```
remh: You should use `utils.subprocess_output.get_subprocess_output` instead

checks.d/linux_proc_extras.py
gphat: I chose `linux` because I had no intention of dealing with other Unixes. Would you like me to do `system.linux.*` or just `system.*` here?

gphat:
Done! Waiting on the `linux` question above and ready to go.

checks.d/linux_proc_extras.py
```diff
@@ -0,0 +1,83 @@
...
+    'N': 'low',
+    'L': 'locked'
+}
+
+class MoreUnixCheck(AgentCheck):
+    def check(self, instance):
+        tags = instance.get('tags', [])
+
+        state_counts = collections.Counter(
+            uninterruptible=0,
```
remh: The good thing about the Counter is that you don't need to initialize your keys at 0 :)
```python
state_counts = Counter()
```
would give you the same result

checks.d/linux_proc_extras.py
```diff
@@ -0,0 +1,83 @@
+# project
+from checks import AgentCheck
+from utils.subprocess_output import get_subprocess_output
+import collections
```
remh: You should just import Counter:
```python
from collections import Counter
```

gphat:
Fixed! :)

remh:
Thanks Cory.
I think the metrics should be named `system.linux` 
Can you squash your commits as well please ?
Thanks!

gphat:
Squashed!  Omg I just learned how to squash merge when I've already done merges in to the branch. It's an early Christmas!

checks.d/linux_proc_extras.py
```diff
@@ -0,0 +1,71 @@
+# project
+from checks import AgentCheck
+from utils.subprocess_output import get_subprocess_output
+from collections import Counter
+import subprocess
```
remh: This is not used anymore btw :)

checks/system/unix.py
```diff
@@ -559,12 +559,13 @@ def check(self, agentConfig):
         processes = []
 
         for line in processLines:
-            line = line.split(None, 10)
-            processes.append(map(lambda s: s.strip(), line))
+            l = line.split(None, 10)
```
remh: Not sure why you are modifying this file though you should just let it as is.

gphat:
Fixed and rebased for brevity.

checks.d/linux_proc_extras.py
```diff
@@ -0,0 +1,70 @@
...
+    '<': 'high',
+    'N': 'low',
+    'L': 'locked'
+}
+
+class MoreUnixCheck(AgentCheck):
+    def check(self, instance):
+        tags = instance.get('tags', [])
+
+        state_counts = Counter()
```
remh: I know i'm the one who told you to use a Counter but i forgot that Counter were not available on python 2.6. Can you use a `defaultdict(int)` instead ?

remh:
Thanks @gphat ! Can you rebase against the latest version of master again please ? It should fix the zookeeper failing test.

Also could you write a test for this integration please ? It should be pretty straigthforward.

gphat:
Hey @remh, does all that look right? I'm still getting failures, but they seem to be elsewhere. I will clean up the commits once you think it's ok.

gphat:
Friendly ping, @remh!

remh:
Looks good to me @gphat ! Can you squash the commits and then i'll merge

gphat:
See #2202, I broke it when squash merging. Ugh.

PR #2045, opened 2015-11-04
[mongo] variable typo.

truthbk:
Not sure where this came from! 

Spotted this issue - dbstats weren't getting reported due to it.


olivielpeau:
Good catch, thanks for spotting and fixing this @truthbk! :+1:

PR #2048, opened 2015-11-04
[http_check] fixed check for weak ciphers

truthbk:
In the WeakCiphersAdapter we still need to check with getattr for the _tunnel_host attribute, we can't assume it's going to be populated automatically by httplib, requests lib, etc.


yannmh:
Thanks @truthbk, let's merge it.

PR #2049, opened 2015-11-04
[packaging][windows] force-include 'wmi' module

yannmh:
The module was previously dynamically imported from the [Windows System check](https://github.com/DataDog/dd-agent/blob/5.5.x/checks/system/win32.py#L10-L14) and needs to be manually included now.


olivielpeau:
:shipit:

PR #2050, opened 2015-11-04
[elasticsearch] Send cluster hostnames as elastic_hostname tag, not a host tag.

truthbk:
Sending the hostname as a `host:foobox` tag was creating duplicate hosts in the UI. By sending the hostname as a `elastic_hostname:foobox` we avoid that collateral.

### DIFF
--- a/checks.d/elastic.py
+++ b/checks.d/elastic.py
@@ -20,6 +20,7 @@ ESInstanceConfig = namedtuple(
     'ESInstanceConfig', [
         'pshard_stats',
         'cluster_stats',
+        'cluster_hosts_as_tags',
         'password',
         'service_check_tags',
         'tags',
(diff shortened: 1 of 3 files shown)


olivielpeau:
Looking good overall!

Now that I think of it what do you think about renaming it `elasticsearch_host`? (`elastic` is a bit vague IMO) .

The change has to be applied to the integration tests too.

Could you also update the `elastic.yaml.example` to document the new tag?

Also, we have to keep in mind that this change breaks BC for the users who have the `cluster_stats` parameter enabled and who use the `host` tag.

olivielpeau:
One last thing, unrelated: could you get rid of the `nodes_url` parameter on the `_process_stats_data` method [here](https://github.com/DataDog/dd-agent/blob/25dfab03f14e64f313bab71bae01ed586e7e37d7/checks.d/elastic.py#L419)?

truthbk:
@remh so this is the elasticsearch changes we made to avoid duplication of hosts - we hid the new behavior behind a feature flag to avoid breaking compatibility. Wondering if it wouldn't be better to tag with `server` as described here: https://github.com/DataDog/dogweb/pull/9089

Or maybe you prefer to fix this elsewhere. Let me know :)

remh:
Closing. This will break too much compatibility for an edge use case.

PR #2055, opened 2015-11-05
[docker_daemon] Remove spurious proc root container warnings

oeuftete:
With 5.6.0, this happens on every cycle:

```
2015-11-05 19:20:27 UTC | WARNING | dd.collector | checks.docker_daemon(__init__.py:649) | Couldn't find pid directory for container: . They'll be missing network metrics
```


yannmh:
Thanks for the quick fix @oeuftete !

PR #2064, opened 2015-11-06
[core] `supervisor.conf` to select Supervisor user

yannmh:
By default, Supervisor runs with the given `user` in the `supervisord`
section of the `supervisor.conf` configuration file.

> If supervisord is run as the root user, switch users to this UNIX user
account before doing any meaningful processing. This value has no effect
if supervisord is not run as root.
(c.f
http://supervisord.org/configuration.html#supervisord-section-settings)

Thus, forcing Supervisor to run as `dd-agent` in the init script file is
simply an extra precaution. However, it makes impossible to change UNIX
users running the agent subprocesses via the `supervisor.conf`
configuration file only.
Drop this constraint.

### DIFF
--- a/packaging/centos/datadog-agent.init
+++ b/packaging/centos/datadog-agent.init
@@ -104,7 +104,7 @@ start() {
     # no need to test for status before daemon,
     # the daemon function does the right thing
     echo -n "Starting Datadog Agent (using supervisord):"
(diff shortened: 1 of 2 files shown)


yannmh:
@degemer can you take a pass on it ?

remh:
Looks good to me, let's make sure to test that thoroughly. 

@olivielpeau possible to add more tests cases in dd-agent-testing with different values in the supervisord.conf ?

olivielpeau:
@remh I've added a few tests to dd-agent-testing

PR #2065, opened 2015-11-06
[windows][jmxfetch] fix logging initilization from jmxfetch python

olivielpeau:
Otherwise we don't actually log anything that's logged by jmxfetch.py


yannmh:
Looks good. Thanks.

PR #2071, opened 2015-11-12
[docker] Improve the way we extract image name and tag

hkaj:


remh:
Thanks.
Can you write a unit test for this function please ?

tests/checks/integration/test_docker_daemon.py
```diff
@@ -527,3 +526,24 @@ def test_container_size(self):
...
         for mname, tags in expected_metrics:
             self.assertMetric(mname, tags=tags, count=1, at_least=1)
+
+    def test_image_tags_extraction(self):
+        images = [
+            # ({'Image': image_name}, [expected_image_name, expected_image_tag])
+            ({'Image': 'nginx:latest'}, ['nginx', 'latest']),
+            ({'Image': 'localhost/nginx:latest'}, ['localhost/nginx', 'latest']),
+            ({'Image': 'localhost:5000/nginx:latest'}, ['localhost:5000/nginx', 'latest']),
+        ]
```
yannmh: Can we add tests with `RepoTags` ?

utils/dockerutil.py
```diff
@@ -95,3 +102,39 @@ def find_cgroup_filename_pattern(mountpoints, container_id):
...
+    if "Image" in co:
+        split = co["Image"].split(":")
+        if len(split) <= key:
+            return None
+        elif len(split) > 2:
+            # if the repo is in the image name and has the form 'docker.clearbit:5000'
+            # the split will be like [repo_url, repo_port/image_name, image_tag]. Let's avoid that
+            split = [':'.join(split[:-1]), split[-1]]
+        return [split[key]]
+    if "RepoTags" in co:
```
yannmh: Can `co` contain both `Image ` and `RepoTags ` keys ?

utils/dockerutil.py
```diff
@@ -95,3 +102,39 @@ def find_cgroup_filename_pattern(mountpoints, container_id):
...
+        splits = [el.split(":") for el in co["RepoTags"]]
+        tags = []
+        for split in splits:
+            if len(split) > 2:
+                split = [':'.join(split[:-1]), split[-1]]
+            if len(split) > key:
+                tags.append(split[key])
+        if len(tags) > 0:
+            return list(set(tags))
+    return None
```
yannmh: I find the method signature quite confusing.
Why don't we just drop the `key` argument and return a `(image_name, image_tags)` pair instead ? 
This would also avoid to go through the same code path multiple times when the `key` value changes.

utils/dockerutil.py
```diff
@@ -95,3 +102,39 @@ def find_cgroup_filename_pattern(mountpoints, container_id):
...
+    if "RepoTags" in co:
+        splits = [el.split(":") for el in co["RepoTags"]]
+        tags = []
+        for split in splits:
+            if len(split) > 2:
+                split = [':'.join(split[:-1]), split[-1]]
+            if len(split) > key:
+                tags.append(split[key])
+        if len(tags) > 0:
+            return list(set(tags))
```
yannmh:
```python
tags = set()
for split in splits:
  ...
if tags:
  return list(tags)
```
instead ?

utils/dockerutil.py
```diff
@@ -95,3 +102,39 @@ def find_cgroup_filename_pattern(mountpoints, container_id):
...
+    return None
+
+
+def container_name_extractor(co):
+    names = co.get('Names', [])
+    if names is not None:
```
yannmh: `names` will never be `None` as it's defaulted to `[]`.
Thus, this would be enough 
```python
if names:
```
but still **unnecessary** as your code below would still be fine when `names == []`

utils/dockerutil.py
```diff
@@ -95,3 +102,39 @@ def find_cgroup_filename_pattern(mountpoints, container_id):
...
+        # we sort the list to make sure that a docker API update introducing
+        # new names with a single "/" won't make us report dups.
+        names = sorted(names)
+        for name in names:
+            # the leading "/" is legit, if there's another one it means the name is actually an alias
+            if name.count('/') <= 1:
+                return [str(name).lstrip('/')]
```
yannmh: Would it be safer to cast it to `unicode` instead ?

utils/dockerutil.py
```diff
@@ -95,3 +102,39 @@ def find_cgroup_filename_pattern(mountpoints, container_id):
...
+    return co.get('Id')[:11]
```
yannmh: Why `[:11]` ? 
If it's an hardcoded limit, we should probably name this value at the module level. Also note that
```python
> ["Loremipsumdolorsitamet"][:11]
['Loremipsumdolorsitamet']
```

yannmh: Should `container_name_extractor ` return a single value rather than a list ?

yannmh:
Added my feedback.

utils/dockerutil.py
hkaj: no, `Image` is present for a container, `RepoTag` for an image. But you're right, that's not clear. I'll replace `co` with `entity` to avoid this confusion.

utils/dockerutil.py
hkaj: I believe it's to make the call simpler as it is used in a lambda: https://github.com/DataDog/dd-agent/blob/master/checks.d/docker_daemon.py#L120-L121
The method takes care of deciding whether it should return something or `None` which makes the call straight-forward.

tests/checks/integration/test_docker_daemon.py
hkaj: yessir!

utils/dockerutil.py
hkaj: actually `names` can be `None` if `co['Names'] == None` which happens because of a bug in docker < 1.9

utils/dockerutil.py
hkaj: I don't know why we chose `11`, I kept it for legacy reasons (for people migrating from the old docker check to the new one (it was already `11`). @LotharSee can you add your input about that?

`container_name_extractor`'s return value follows the same format as other `TAG_EXTRACTORS` (https://github.com/DataDog/dd-agent/blob/master/checks.d/docker_daemon.py#L118-L124) so that it's easier to process.

utils/dockerutil.py
hkaj: there shouldn't be any funny char in there: https://github.com/ajhager/docker/blob/master/runtime.go#L35

Do you think of something else?

hkaj:
Thanks @yannmh I fixed a bunch of things and replied to your other comments, let me know what you think.

utils/dockerutil.py
LotharSee: Because the short ID version of IDs consists in the first 12 digits of the ID. You can see it from most the Docker CLI commands, but it is also supported as ID from all the CLI commands.

So basically, that is the real identifier people see.  To see the full ID, you have to inspect or use the API.

hkaj: I figured it was something like that. The problem is `[:11]` returns characters 0 to 10, not 11. So we have 11 chars, not 12. Anyway, since it was in the

yannmh:
Feel free to merge :cactus:

PR #2077, opened 2015-11-13
[Docker] Return container_id when the container does not have a name

hkaj:
Work around the bug in docker engine where a container can have no name.


hkaj:
closing in favor of #2071

PR #2078, opened 2015-11-13
[dev] ignore tests/core/fixtures/flare/dd*

dylanvee:
These files are generated as a side effect of running tests and do not need to
be checked into the repository.

[skip ci]


remh:
Yup that's right.
Thanks @dylanvee !

PR #2079, opened 2015-11-13
[kafka] Improve example configuration

dougbarth:
Collecting MeanRate from Kafka's Meter JMX Mbeans isn't terribly helpful for graphing in DD because the rate is the [mean rate for the entire lifetime of the Kafka broker](http://metrics.dropwizard.io/3.1.0/manual/core/#meters).

Instead of collecting the MeanRate, we could collect the load average style rates or we can simply grab the monotonically increasing counts and calculate our own rate from that raw counters. Since DD captures metrics at smaller time intervals than a minute, calcuating our own rate gives the user the most granular view of cluster activity so that's what the example configuration file will contain.

In addition to fixing the collection of these rate metrics, the example configuration is also updated to collect nearly all the useful metrics [recommended by Confluent](http://docs.confluent.io/1.0.1/kafka/monitoring.html) for alerting and trending on a Kafka cluster.

### DIFF
--- a/conf.d/kafka.yaml.example
+++ b/conf.d/kafka.yaml.example
@@ -33,30 +33,30 @@ init_config:
         domain: 'kafka.server'
         bean: 'kafka.server:type=BrokerTopicMetrics,name=BytesOutPerSec'
         attribute:
-          MeanRate:
-            metric_type: gauge
-            alias: kafka.net.bytes_out
+          Count:
+            metric_type: rate
+            alias: kafka.net.bytes_out.rate
     - include:
         domain: 'kafka.server'
         bean: 'kafka.server:type=BrokerTopicMetrics,name=BytesInPerSec'
         attribute:
-          MeanRate:
-            metric_type: gauge
-            alias: kafka.net.bytes_in
+          Count:
+            metric_type: rate
+            alias: kafka.net.bytes_in.rate
     - include:
         domain: 'kafka.server'
         bean: 'kafka.server:type=BrokerTopicMetrics,name=MessagesInPerSec'
         attribute:
-          MeanRate:
-            metric_type: gauge
-            alias: kafka.messages_in
+          Count:
+            metric_type: rate
+            alias: kafka.messages_in.rate
     - include:
         domain: 'kafka.server'
         bean: 'kafka.server:type=BrokerTopicMetrics,name=BytesRejectedPerSec'
         attribute:
-          MeanRate:
-            metric_type: gauge
-            alias: kafka.net.bytes_rejected
+          Count:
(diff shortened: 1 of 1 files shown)


irabinovitch:
Tests are passing here. The one flake8 error will be addressed by PR #2080

remh:
Thanks a lot @dougbarth !

@DorianZaccaria can you review this one please ?

conf.d/kafka.yaml.example
```diff
@@ -129,41 +160,97 @@ init_config:
...
         attribute:
-          MeanRate:
+          Count:
+            metric_type: rate
+            alias: kafka.replication.unclean_leader_elections.rate
+    - include:
+        domain: 'kafka.controller'
+        bean: 'kafka.controller:type=KafkaController,name=OfflinePartitionsCount'
+        attribute:
+          Count:
```
DorianZaccaria: It looks like the only attribute available for this bean is `Value`.

conf.d/kafka.yaml.example
```diff
@@ -129,41 +160,97 @@ init_config:
...
+            metric_type: gauge
+            alias: kafka.replication.partition_count
+    - include:
+        domain: 'kafka.server'
+        bean: 'kafka.server:type=ReplicaManager,name=LeaderCount'
+        attribute:
+          Count:
             metric_type: gauge
-            alias: kafka.replication.unclean_leader_elections
+            alias: kafka.replication.leader_count
```
DorianZaccaria: Same comment for the three metrics above.

DorianZaccaria:
It makes sense to me to use the `Count` as a rate. The only issue I found was some metrics trying to fetch the `Count` attribute of some beans instead of the `Value`. Here the list of these metrics:
- kafka.replication.offline_partitions_count
- kafka.replication.active_controller_count
- kafka.replication.partition_count
- kafka.replication.leader_count

Everything was tested on Kafka 0.8.2.2.

olivielpeau:
Hi @dougbarth, have you been able to look at the comments above?

We'd very much like to include these changes in the next release of the agent (5.7.0), which will be out soon.

Thanks!

irabinovitch:
Hi @dougbarth  Have you had a chance to review the feedback here? Our merge window for 5.7 is closing soon and we'd love to include your improvements to the Kafka check.

dougbarth:
@irabinovitch @olivielpeau this PR is updated. So sorry for the delay!

@DorianZaccaria thank you very much for the detailed check against this PR. I appreciate it.

olivielpeau:
Thanks again @dougbarth! Everything looks good, merging.

PR #2080, opened 2015-11-14
fix indentation in docker_daemon

irabinovitch:
Just fixing a flake8 failure introduced in a recent commit.


irabinovitch:
The failures here seem to be issues with TravisCI and not specific tests failing or changes in my commit. Been seeing that a lot lately, we may want to reach out to them to see whats up.

remh:
Thanks @irabinovitch i fixed it in https://github.com/DataDog/dd-agent/pull/2082 as well

PR #2084, opened 2015-11-16
[flare][windows] Mention path to tar file in flare's UI

olivielpeau:
Especially useful because the path of the temp dir on Windows where the tar is saved can vary from machine to machine.

### DIFF
--- a/win32/gui.py
+++ b/win32/gui.py
@@ -775,18 +775,19 @@ def windows_flare():
     f.collect()
     email, ok = QInputDialog.getText(
         None, "Your email",
-        "Logs and configuration files have been collected"
+        "Logs and configuration files have been collected."
         " Please enter your email address:"
     )
     if not ok:
-        info_popup("Flare cancelled")
+        info_popup("Flare cancelled. You can still use {0}".format(f.tar_path))
         return
     try:
         case_id = f.upload(email=str(email))
         info_popup("Your logs were successfully uploaded. For future reference,"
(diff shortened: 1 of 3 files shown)


olivielpeau:
And btw, I've tested it on a windows build and it works well :)

win32/gui.py
```diff
@@ -775,18 +775,19 @@ def windows_flare():
...
+        info_popup("Flare cancelled. You can still use {0}".format(f.tar_path))
         return
     try:
         case_id = f.upload(email=str(email))
         info_popup("Your logs were successfully uploaded. For future reference,"
                    " your internal case id is {0}".format(case_id))
     except Exception, e:
-        warning_popup('The upload failed:\n{0}'.format(str(e)))
+        warning_popup('The upload failed. Please send the following file by email'
+                      ' to support: {0}\n\n{1}'.format(f.tar_path, str(e)))
```
truthbk: Could we maybe add the email here, for convenience?

truthbk:
Other than the minor comment about adding the email address for customer convenience - this looks :+1:

win32/gui.py
olivielpeau: We could, the only thing is that the customer may have a case already opened, in that case he/she should rather add the tar to the existing case and not send it directly to support's email...

truthbk:
:+1:

olivielpeau:
Thanks @truthbk! Will merge once the CI passes.

PR #2085, opened 2015-11-16
[jenkins] add support for reporting build parameters as event tags

dylanvee:
Jenkins checks can now be configured to report arbitrary build parameters as
event tags. For example, consider a config that looks like

```
init_config:

instances:
  - name: default
    jenkins_home: /var/lib/jenkins
    parameters:
      GIT_BRANCH: branch
```

and a Jenkins job where the `GIT_BRANCH` parameter is set to `master`. With
this patch the resulting DataDog event will have a `branch:master` tag
associated with it.

(This particular example illustrates why Stripe authored this patch: we don't
use the Jenkins Git plugin but still want branch information to be threaded
through to DataDog.)

While expanding `test_jenkins.py` to cover this new functionality, I replaced
the dict-to-XML helper with handwritten XML strings because it seemed simpler
than enhancing the helper to generate the more complex XML needed for the new
test coverage.

### DIFF
--- a/checks.d/jenkins.py
+++ b/checks.d/jenkins.py
@@ -89,6 +89,17 @@ class Jenkins(AgentCheck):
                     .text
             except Exception:
                 pass
+
+            try:
+                d['parameters'] = {
+                    e.find('name').text: e.find('value').text for e in
+                    tree.find('actions')
+                        .find('hudson.model.ParametersAction')
+                        .find('parameters')
+                }
+            except Exception:
+                d['parameters'] = {}
+
             return d
 
     def _get_build_results(self, instance_key, job_dir):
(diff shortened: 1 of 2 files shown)


gphat:
Woot! Thanks for sending this along @dylanvee. Tagging in @remh!

remh:
Thanks a lot @dylanvee @gphat 
We are going to deprecate the Jenkins check soon in favor of a proper jenkins-plugin
https://github.com/DataDog/jenkins-datadog-plugin

The new plugin will push events directly to Datadog instead of having the Agent watches the Jenkins data directory. It's way more scalable and easier to deploy.

So i'm not sure we'll add more features to the old check. But this feature would make sense in the new plugin.

What do you think ?

gphat:
Is that plugin ready for use now @remh? We'd almost certainly like to use that since it's probably better. ;)

remh:
Yup it is.

And it's already collecting the Git branch as a tag.
You can't pass a list of custom parameters to fetch though.
I'm closing this issue for now, feel free to open an issue or a PR in https://github.com/DataDog/jenkins-datadog-plugin

PR #2087, opened 2015-11-17
[http_check] Adding a days_critical to the SSL certificate expiration feature

truthbk:
The check will now raise a critical warning if we're within the specified CRITICAL range before a certificate expires.


MiguelMoll:
I was contemplating this exact feature! This would be really useful.

olivielpeau:
After some discussion, we'll keep the new `CRITICAL ` network check status.

This looks good to merge, thanks @truthbk!

PR #2088, opened 2015-11-17
[core] Fix host tags sending

remh:
If the create_dd_check_tags was enabled and if custom tags are passed
in datadog.conf, we have to send the dd_check and host_tags at the same
time.
Otherwise they’ll overwrite each other.

### DIFF
--- a/checks/collector.py
+++ b/checks/collector.py
@@ -173,10 +173,6 @@ class Collector(object):
                 'start': time.time(),
                 'interval': int(agentConfig.get('agent_checks_interval', 10 * 60))
             },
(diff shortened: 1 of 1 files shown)


olivielpeau:
I think we can safely remove these lines now: https://github.com/DataDog/dd-agent/blob/3042a1db079dc0771ac7d8db67ea8a70bd220014/checks/collector.py#L176-L179

Other than that it's looking good!

remh:
good point

remh:
Done

olivielpeau:
:+1:

PR #2091, opened 2015-11-18
[teamcity] Add ssl_validation flag to check

jslatts:
It would be helpful to have the ability to disable SSL validation in the teamcity check, same idea as issue #1782.

I went ahead and added a flag in with the same name as the nginx check. I did not try and write an integration test. Is a test needed for this?

### DIFF
--- a/checks.d/teamcity.py
+++ b/checks.d/teamcity.py
@@ -21,7 +21,7 @@ class TeamCityCheck(AgentCheck):
         # Keep track of last build IDs per instance
         self.last_build_ids = {}
 
-    def _initialize_if_required(self, instance_name, server, build_conf):
+    def _initialize_if_required(self, instance_name, server, build_conf, ssl_validation):
         # Already initialized
         if instance_name in self.last_build_ids:
             return
@@ -32,7 +32,7 @@ class TeamCityCheck(AgentCheck):
             build_conf=build_conf
         )
         try:
-            resp = requests.get(build_url, timeout=self.DEFAULT_TIMEOUT, headers=self.HEADERS)
+            resp = requests.get(build_url, timeout=self.DEFAULT_TIMEOUT, headers=self.HEADERS, verify=ssl_validation)
             resp.raise_for_status()
 
             last_build_id = resp.json().get('build')[0].get('id')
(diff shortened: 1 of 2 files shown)


remh:
Thanks @jslatts 
A test is not needed in that case as it would basically be testing the requests library.

We'll review more in details for our 5.7.0 release.
Thanks again!

checks.d/teamcity.py
```diff
@@ -21,7 +21,7 @@ def __init__(self, name, init_config, agentConfig, instances=None):
         # Keep track of last build IDs per instance
         self.last_build_ids = {}
 
-    def _initialize_if_required(self, instance_name, server, build_conf):
+    def _initialize_if_required(self, instance_name, server, build_conf, ssl_validation):
```
remh: Looks like you forgot to pass the ssl_validation flag when calling `_initialize_if_required`

irabinovitch: @jslatts The window for the 5.7  release will be closing soon. Do you have a moment to look at @remh 's feedback?

jslatts: @irabinovitch sorry, missed the original notification. I added it in. Do you want me to rebase the PR against current master?

conf.d/teamcity.yaml.example
```diff
@@ -22,6 +22,9 @@ instances:
     # rather than just that a successful build happened
     # is_deployment: true
 
+    # Optional, this turns of ssl ceritificate validation. Defaults to True.
```
olivielpeau: Small typos here: `turns of ssl ceritificate validation` -> `turns off ssl certificate validation`

checks.d/teamcity.py
```diff
@@ -91,6 +91,8 @@ def check(self, instance):
         if instance_name is None:
             raise Exception("Each instance must have a unique name")
 
+        ssl_validation = instance.get('ssl_validation', True)
```
olivielpeau: Could you use the `_is_affirmative` function here? i.e: 

```python
ssl_validation = _is_affirmative(instance.get('ssl_validation', True))
```

olivielpeau:
Hi @jslatts!

I've added two comments, could you address them? Also, could you only keep your commits in this PR? The other 3 commits seem to be unrelated.

Apart from that your changes look good to me. Thanks!

jslatts:
Fixed the typo and added the _is_affirmative call. Also rebased against current master. No idea how the other commits ended up in there. It looks clean now. Let me know if there is anything else.

olivielpeau:
Thanks again @jshirley!

Merging.

PR #2095, opened 2015-11-19
Add interpolation of environment variables.

mwolny:
Briefly the change allows to use all environment variables that begins with `DATADOG_` inside the configuration file.

### DIFF
--- a/config.py
+++ b/config.py
@@ -64,6 +64,8 @@ LEGACY_DATADOG_URLS = [
     "app.datad0g.com",
 ]
 
+# Environment variables that may be interpolated in the datadog configuration file.
+DATADOG_ENV_VARS = dict((k, v) for k, v in os.environ.iteritems() if k.startswith('DATADOG_'))
 
 class PathNotFound(Exception):
     pass
@@ -340,11 +342,7 @@ def get_config(parse_args=True, cfg_path=None, options=None):
         path = os.path.dirname(path)
 
         config_path = get_config_path(cfg_path, os_name=get_os())
-
-        # Take into account interpolation of enviroment variables that begins with "DATADOG_".
-        env_vars = dict((k, v) for k, v in os.environ.iteritems() if k.startswith('DATADOG_'))
-
-        config = ConfigParser.ConfigParser(env_vars)
+        config = ConfigParser.ConfigParser(DATADOG_ENV_VARS)
         config.readfp(skip_leading_wsp(open(config_path)))
 
         # bulk import
@@ -928,7 +926,7 @@ def get_logging_config(cfg_path=None):
         logging_config['log_to_syslog'] = True
 
     config_path = get_config_path(cfg_path, os_name=system_os)
-    config = ConfigParser.ConfigParser()
+    config = ConfigParser.ConfigParser(DATADOG_ENV_VARS)
     config.readfp(skip_leading_wsp(open(config_path)))
 
     if config.has_section('handlers') or config.has_section('loggers') or config.has_section('formatters'):


mwolny:
Looks like the first problem was the Python 2.6.x syntax issue. Anyway, it should be fixed now.

remh:
@mwolny 
Thanks for this.
There is a similar implementation in https://github.com/DataDog/dd-agent/pull/1343/files

We'll decide which direction we should go for 5.7.0

Thanks a lot!

mwolny:
It would be really awesome to have such functionality as soon as possible. Seriously, it simplifies a lot. Is there any deadline or planned date for 5.7.0?

irabinovitch:
@mwolny We're currently tracking 5.7 for January.

mwolny:
Awesome, really happy to hear it!

On Thu, Dec 17, 2015 at 11:05:06AM -0800, Ilan Rabinovitch wrote:
> @mwolny We're currently tracking 5.7 for January.
>
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/DataDog/dd-agent/pull/2095#issuecomment-165549990

--
Marcin Wolny
marcin@wutanic.com

remh:
@mwolny can you describe a bit more your use case ? 
What environment variables are you passing ?

mwolny:
Sure. Nowadays I work with docker containers mostly. As you probably
know enviroment variables as one of the most common method to pass
parameters (not necesarrly a good method). Anyway, in my case part of
containers use dogstatsd to send metrics to forwarders that resides
behind the AWS ELB. This means that I have to set `dogstatsd_target`.
Obviously environment variable is the easiest choice if you want to
treat your container stateless and keep it readonly. I would prefer to
avoid hardcoding values that may influence may infrastructure. Is this
answer your question?

I also use enviroment variables for checks. But here I can take
advantage of `pyaml` and `!!python`.

On Fri, Dec 18, 2015 at 10:50:11AM -0800, Remi Hakim wrote:
> @mwolny can you describe a bit more your use case ?
> What environment variables are you passing ?
>
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/DataDog/dd-agent/pull/2095#issuecomment-165868240

--
Marcin Wolny
marcin@wutanic.com

remh:
Thanks for the explanation. Can you describe where you'll be reading those settings from ?

From what i can see, all environment variables starting with `DATADOG_` will be passed to the config object but it's not clear where it will be read from.

Thanks!

mwolny:
This change allows you to use all environment variables starting with
`DATADOG_` in agent configuration, e.g. `%(DATADOG_API_KEY)s`.

In other way, all environment variables starting with `DATADOG_` you
will be able to interpolate in agent configuration file.

Is this answer your question?

On Tue, Dec 22, 2015 at 08:11:48AM -0800, Remi Hakim wrote:
> Thanks for the explanation. Can you describe where you'll be reading those settings from ?
>
> From what i can see, all environment variables starting with `DATADOG_` will be passed to the config object but it's not clear where it will be read from.
>
> Thanks!
>
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/DataDog/dd-agent/pull/2095#issuecomment-166662197

--
Marcin Wolny
marcin@wutanic.com

mwolny:
Looks like this change also introduces a problem with logging configuration. Basically the code creates ConfigParser object instance twice at two different places. This makes this path kind of unusable.

remh:
Yes it does answer my question, thanks!

You're right we initialize the config parser twice in the code. One for the logging config and one for the rest as we want to set up the logging as soon as possible. We should likely add the environment variable interpolation in the logging config set up code path as well. What do you think ?

mwolny:
Actually I am just in the middle of creating change where `env_vars` is changed into global variable and re-used at both places. I think this is the easiest possible change with current architecture.

andyshinn:
This change still seems to require a change to `datadog.conf` to add the variables, correct? It would be nice to be able to actually override the configuration with the environment variable without needing to edit the `datadog.conf` at all.

For example, I usually only ever set `api_key`. It would be great if I could just set the `DATADOG_API_KEY` environment variable to override whatever is in the config (then the stock installed config never needs to be touched).

PR #2097, opened 2015-11-19
[system-mac os] Fix memory metrics

remh:
We use the output of ps for the mem usage stats, but the values can
have different units (can be G or M).

We would need to parse the unit and convert the values to the same unit
(probably bytes), otherwise it doesn't make much sense.

The Mac Package is now bundling psutil so let’s use psutil to get these
metrics (and add more!)

### DIFF
--- a/checks/system/unix.py
+++ b/checks/system/unix.py
@@ -14,6 +14,12 @@ from util import get_hostname
 from utils.platform import Platform
 from utils.subprocess_output import get_subprocess_output
 
+# 3rd party
+try:
+    import psutil
+except ImportError:
+    psutil = None
+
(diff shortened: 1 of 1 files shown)


clutchski:
still compatible with the metadata?

remh:
Yeah we are just changing the values of the metrics (to actually send the proper ones)

olivielpeau:
Looks good to me!

Now that we need psutil in a core check, should we still allow installing the agent from source on OSX? (or at least should we output a warning of some kind?)

remh:
@olivielpeau done

olivielpeau:
:+1:

PR #2098, opened 2015-11-20
[windows][requests] Bundle default CA certs of `requests`

olivielpeau:
py2exe only includes the python files of the python packages in the
`library` zipped file. This causes the `requests` library to be shipped
without CA certs, so all the https requests made with it fail (unless
SSL cert validation is explicitely disabled).

This fixes the issue by:
1. shipping the default certificates of `requests` with the Windows
agent as an extra data file
2. overriding the path to the default CA bundle that `requests` uses
(which is done in quite an ugly way, but I haven't found any better
way)

Notes:
- we already ship CA certs for `tornado`, but to be as close as possible to the agent's behavior on other platforms I've used the default CA certs provided by `requests`/`certifi`
- this doesn't fix SSL cert validation on the `http_check` on Windows, because the check uses default paths that are only valid on unix platforms: see https://github.com/DataDog/dd-agent/blob/8fed4c926e8abbb2f16ebec47670f9f2a5ef8a61/checks.d/http_check.py#L127-L131 (unless the user sets a valid path in the check's config)

### DIFF
--- a/config.py
+++ b/config.py
@@ -620,6 +620,26 @@ def set_win32_cert_path():
     log.info("Windows certificate path: %s" % crt_path)
     tornado.simple_httpclient._DEFAULT_CA_CERTS = crt_path
 
+
+def set_win32_requests_ca_bundle_path():
+    """In order to allow `requests` to validate SSL requests with the packaged .exe on Windows,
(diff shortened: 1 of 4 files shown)


remh:
Thanks for doing this!

PR #2100, opened 2015-11-24
Add slab, page tables and cached swap.

gphat:
# What's this PR do?

Adds some new memory metrics:
* Slab:  The total amount of memory, in kilobytes, used by the kernel to cache data structures for its own use.
* PageTables: The total amount of memory, in kilobytes, dedicated to the lowest page table level.
* SwapCached: The amount of swap, in kilobytes, used as cache memory.

# Motivation

This is data we have in our existing munin + graphite setup. It's also stuff we've seen have a production impact during a nasty outage caused by a huge amount of VM pressure from slab allocation due to a weird problem we created by running a `find` across all of `/proc`. :P

# Notes

I still am not sure how this metric gets turned in to `system.memory.*`. Is this all I need to do? Will do some testing. :smile:

### DIFF
--- a/checks/collector.py
+++ b/checks/collector.py
@@ -318,7 +318,10 @@ class Collector(object):
                     'memSwapTotal': memory.get('swapTotal'),
                     'memCached': memory.get('physCached'),
                     'memBuffers': memory.get('physBuffers'),
-                    'memShared': memory.get('physShared')
+                    'memShared': memory.get('physShared'),
+                    'memSlab': memory.get('physSlab'),
+                    'memPageTables': memory.get('physPageTables'),
+                    'memSwapCached': memory.get('swapCached')
                 }
                 payload.update(memstats)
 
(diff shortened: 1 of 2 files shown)


gphat:
Friendly ping about this PR!

irabinovitch:
@gphat Thanks for your time on the phone and in IRC yesterday.  Looking to see this merged as well. Thanks for your patience while we streamline the integration release process.

remh:
Let's do it. The metric translation is done in our backend, we will add the needed rules for this to work.

gphat:
Yay! Lemme know if you need anything!

olivielpeau:
Hi @gphat! Adding memory metrics also requires updating this dict: https://github.com/DataDog/dd-agent/blob/8a0e3a06de10c83fd05bd4565cd826e991fa3596/checks/collector.py#L307-L320 with the new metrics, could you take care of it?

I'll add the metric translation to our backend.

Thanks!

gphat:
Added it, @olivielpeau!  Lemme know if I messed it up. :smile:

olivielpeau:
@gphat: I don't see any additional change in the commit you just pushed, could you check what you've committed? :)

gphat:
Oops! I used `--amend` and forgot to force, so the command failed and I left for the day!

Sorry @olivielpeau, should be fixed up now.

olivielpeau:
FYI the names of these metrics will probably be translated to:
```
system.mem.slab
system.mem.page_tables
system.swap.cached
```
in our backend.

If you have better names in mind we're open to suggestions :)
Merging, thanks again @gphat !

Issue #807, opened 2014-01-23
Dogstream doesn't support custom tags


remh:
It only supports device_name and aggregates the rest.

remh:
Closing as we are working on finding a better replacement for dogstream.

justiniso:
> Closing as we are working on finding a better replacement for dogstream.

@remh what's wrong with dogstream? Why is it being replaced?

ewdurbin:
this is not a well specified issue, we have determined that we are seeing funny issues with dogstreams and tags. we would like to work on a fix but it's unclear where to get started looking to resolve.

if more information was provided on what needs to be done to resolve dogstream tag handling, I'd be happy to take a closer look.

justiniso:
Hey @ewdurbin thanks for surfacing this again! Please don't close the issue as being unreproducible though. The problem is actually well-documented in some of the related issues/pr's linked above

https://github.com/DataDog/dd-agent/issues/1304
https://github.com/DataDog/dd-agent/pull/2103

Those were closed to consolidate with this one (which is admittedly not well specified) being the canonical issue. Please take a look at my PR https://github.com/DataDog/dd-agent/pull/2103. I've been running a version of the agent which implements that change on a service processing 8,000 log lines per second for several months now with the correct behavior. I'd love to remove that hack if this gets merged into the agent! 😄

justiniso:
@clutchski this is the dogstream issue I was talking about where you are limited to one unique series per second. Aggregation is rolling up custom tags to the first occurrence of that metric (https://github.com/DataDog/dd-agent/issues/1304 does a better job of describing the specific issue). When you have thousands of log lines per second, this is a show-stopper. When my PR was closed, it sounded like there was something coming in 6.0.0 that fixed the issue, but now it seems like that's not the case. I think this is a bug that can have a pretty significant impact on reporting for anyone using dogstream with tags. 

For my service ingesting 10,000 req/s, the math was very obviously wrong which led me to here. But for services logging with much lower volumes (e.g. 2-100 lines/s), this bug can go completely undetected and create wildly inaccurate reports.

alexef:
We are also affected by this issue - what is the plan. What kind of details do you need @ewdurbin ?

PR #2103, opened 2015-11-25
[dogstream] Do not aggregate dogstream metrics with different tags

justiniso:
This is a fix for https://github.com/DataDog/dd-agent/issues/807, which will no longer aggregate metrics with the same timestamp if they have different attributes.

### DIFF
--- a/checks/datadog.py
+++ b/checks/datadog.py
@@ -27,8 +27,10 @@ else:
 
 
 def point_sorter(p):
-    # Sort and group by timestamp, metric name, host_name, device_name
-    return (p[1], p[0], p[3].get('host_name', None), p[3].get('device_name', None))
+    # Sort and group by timestamp, metric name, and custom attributes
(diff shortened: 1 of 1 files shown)


irabinovitch:
@justiniso  Thanks for the pull request. Due to the way dogstream works at the moment this requires some changes to our internal pipeline in addition to the changes you have made here.  We are currently in the process of re-architecting this to make it more self contained with in the agent.

With that in mind, we're likely going to close out this PR but keep the issue open and track there.

justiniso:
Thanks @irabinovitch. We're currently running a forked version of the agent with the changes in this PR as a workaround. Do you know if the updates on your end likely to be backwards-compatible in a way that this PR will still work?

irabinovitch:
@justiniso We are still early in the process, so it is difficult to say what the impact will be. We do however strive for backwards compatibility.  As mentioned before, we're going to close this PR for now, but keep the issue open and will look to address this with the previously mentioned changes planned for 6.x.

PR #2104, opened 2015-11-26
Gauge delta support

### LINKED ISSUE #573: Add gauge deltas
Recently Etsy added Gauge deltas. See [here](http://statsd.readthedocs.org/en/latest/types.html) for usage and [here](https://github.com/etsy/statsd/commit/3eecd18) for the commit. I think we should add them to Datadog.

alq666:
To fix #573.

Note that we don't exactly follow the statsd implementation, which treats negative gauge values as gauge deltas (instead of using e.g. a new metric type). The canonical implementation for negative values then forces a reset of the value to 0 before the transmission of a negative delta.

Here we take the opposite view. There are as many negative numbers as there are positive numbers so we treat them as first class citizens and limit gauge deltas to increasing numbers.

### DIFF
--- a/aggregator.py
+++ b/aggregator.py
@@ -63,13 +63,14 @@ class Gauge(Metric):
         # This is a known oversight of the protocol
         try:
             if len(str(value)) > 0 and str(value)[0] in ('+'):
-                if self.value is None: self.value = 0
+                if self.value is None:
+                    self.value = 0
                 self.value += float(value)
             else:
                 self.value = float(value)
         except ValueError:
             log.warn("Cannot parse gauge value: %s" % value)
- 
+
         self.last_sample_time = time()
         self.timestamp = timestamp
 
--- a/aggregator.py
+++ b/aggregator.py
@@ -57,7 +57,16 @@ class Gauge(Metric):
         self.timestamp = time()
 
     def sample(self, value, sample_rate, timestamp=None):
-        self.value = value
+        # Parse optional leading + or - which means gauge delta
+        try:
+            if str(value)[0] in ('+', '-'):
+                if self.value is None: self.value = 0
+                self.value += int(value)
+            else:
+                self.value = value
+        except ValueError:
+            log.warn("Cannot parse gauge value: %s" % value)
+ 
         self.last_sample_time = time()
         self.timestamp = timestamp
 
@@ -392,7 +401,7 @@ class Aggregator(object):
     Abstract metric aggregator class.
     """
     # Types of metrics that allow strings
-    ALLOW_STRINGS = ['s', ]
+    ALLOW_STRINGS = ['s', 'g']
 
     def __init__(self, hostname, interval=1.0, expiry_seconds=300,
             formatter=None, recent_point_threshold=None,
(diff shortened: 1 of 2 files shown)


alq666:
And yes, I miss writing code sometimes :high_brightness: :five:

remh:
:thumbsup: I will give it a better review tomorrow but it looks good at first sight  
:tada:

alq666:
Here are some micro-benchmarks to compare with and without gauge delta support: https://gist.github.com/alq666/fee4d754305add822bf7

The interesting bit is that the stock code performs roughly 2x as slow as the new code, all this because of the `int()` casting operation that fails often and surprisingly:

```python
Python 2.7.10 (default, Oct 23 2015, 18:17:08) 
[GCC 4.2.1 Compatible Apple LLVM 7.0.0 (clang-700.0.59.5)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> int("+0.98284182484")
  File "<stdin>", line 1, in <module>
ValueError: invalid literal for int() with base 10: '+0.98284182484'
>>> float("+0.98284182484")
0.98284182484
```

remh:
Thanks a lot @alq666 

Could you squash your commits into one please ?

It seems weird to me to have a different behavior than etsy's statsd implementation regarding negative values but i understand the rationale. 

We could add support for it behind a feature flag to avoid breaking stuff. Not sure if we should do that now or not as it seems like negative deltas should be common when using gauge deltas.

What do you think ?

remh:
It's actually interesting to see how it's implemented in the clients.
https://github.com/jsocol/pystatsd/blob/master/statsd/client.py#L95-L106

Basically when not using deltas, the gauge is reset to 0 in case of negative values.

If we do the same assumption, then we could use negative deltas.

alq666:
@remh I don't think we can make that assumption. The other thing we can do is to not support gauge deltas, since the original implementation had such a glaring oversight.

remh:
Closing it.

The original implementation is pretty weird, it doesn't handle well negative deltas, also it would break compatibility with users / clients sending values with a "+" (edge case but can be happening).

chrisbenincasa:
Several tickets link to this and reference a “full explanation” as to why this feature isn’t supported. I don’t really see a detailed explanation here; would somebody mind elaborating? To that end, are there plans to suooort this feature? It generally makes gauges a TON more useful in distributed systems.

nickwilliams-eventbrite:
I'd like to chime in and request more details, per @chrisbenincasa's statement. As he said, several tickets say to "see #2104" for an explanation of why "gauge deltas" won't be supported, but there's no explanation here. Critically, we use gauge deltas extensively in our StatsD+Graphite installation. Instead of setting absolute gauge values, one worker may increment the gauge value (with `+1` as the metric value) and another worker may decrement the gauge value (with `-1` as the metric value) so that we arrive an a distributed value for the gauge. We have no recourse for migrating this code to use DataDog instead of Graphite. :-(

lby221:
I'm a little confused about the explanation as well. When thinking inside the distributed world, negative deltas are really important if some metrics are using that. There is no reason we cache locally and do an increase after reset to 0. And this is dangerous! Since we do have an issue OPEN for supporting this, just want to chime in and give a +1 on supporting this feature.

PR #2109, opened 2015-11-30
[redis] Optionally collect redis INFO COMMANDSTATS as metrics.

sethrosenblum:
The redis [INFO](http://redis.io/commands/INFO) COMMANDSTATS command returns:
* The number of times each command has been called.
* The amount of time spent on calling this command
* The average time per call of each command.

This parses that data into three gauges, tagged by command:
* `redis.command.calls`
* `redis.command.usec`
* `redis.command.usec_per_call`

These metrics are only collected if the `command_stats` flag is set to true for the instance, as there are 3 new metrics for every command run on a redis instance.

### DIFF
--- a/checks.d/redisdb.py
+++ b/checks.d/redisdb.py
@@ -244,6 +244,8 @@ class Redis(AgentCheck):
                         self.gauge('redis.key.length', 0, tags=key_tags)
 
         self._check_replication(info, tags)
+        if instance.get("command_stats", False):
+            self._check_command_stats(conn, tags)
 
     def _check_replication(self, info, tags):
 
@@ -324,6 +326,22 @@ class Redis(AgentCheck):
             self.histogram('redis.slowlog.micros', value, tags=tags + [command_tag])
 
         self.last_timestamp_seen[ts_key] = max_ts
+
+    def _check_command_stats(self, conn, tags):
+        """Get command-specific statistics from redis' INFO COMMANDSTATS command
+        """
+        try:
+            command_stats = conn.info("commandstats")
+        except Exception:
+            self.warning("Could not retrieve command stats from Redis."
+                         "INFO COMMANDSTATS only works with Redis >= 2.6.")
+            return
+
+        for key, stats in command_stats.iteritems():
+            command = key.split('_', 1)[1]
+            command_tags = tags + ['command:%s' % command]
+            self.gauge('redis.command.calls', stats['calls'], tags=command_tags)
+            self.gauge('redis.command.usec_per_call', stats['usec_per_call'], tags=command_tags)
 
     def check(self, instance):
         if ("host" not in instance or "port" not in instance) and "unix_socket_path" not in instance:
--- a/conf.d/redisdb.yaml.example
+++ b/conf.d/redisdb.yaml.example
@@ -39,3 +39,6 @@ instances:
     # set the value here.
     # Warning: It may impact the performance of your redis instance
     # slowlog-max-len: 128
+
+    # Collect INFO COMMANDSTATS output as metrics.
+    # command_stats: False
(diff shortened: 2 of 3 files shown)


checks.d/redisdb.py
```diff
@@ -325,6 +327,17 @@ def _check_slowlog(self, instance, custom_tags):
...
 
+    def _check_command_stats(self, conn, tags):
+        """Get command-specific statistics from redis' INFO COMMANDSTATS command
+        """
+        command_stats = conn.info("commandstats")
+        for key, stats in command_stats.iteritems():
+            command = key.split('_', 1)[1]
+            command_tags = tags + ['command:%s' % command]
+            self.gauge('redis.command.calls', stats['calls'], tags=command_tags)
+            self.gauge('redis.command.usec', stats['usec'], tags=command_tags)
```
remh: is this metric needed ? i would think that only the number of calls and the average time per call is enough.

checks.d/redisdb.py
```diff
@@ -244,6 +244,8 @@ def _check_db(self, instance, custom_tags=None):
                         self.gauge('redis.key.length', 0, tags=key_tags)
 
         self._check_replication(info, tags)
+        if instance.get("command_stats", False):
```
remh: can you add this to the redisdb.yaml.example as well please ?

checks.d/redisdb.py
sethrosenblum: I can't think of where I'd use it directly.  Maybe to hunt down where redis is spending it's time on a loaded system?

sethrosenblum:
Hmm, looks like redis 2.4 is not compatible with this info call.

checks.d/redisdb.py
remh: but that's basically redis.command.calls * redis.commands.user_per_call so i think we can remove it

remh:
Can you catch the error and issue a warning if redis <= 2.4 and command_stats is enabled ?

conf.d/redisdb.yaml.example
```diff
@@ -39,3 +39,6 @@ instances:
     # set the value here.
     # Warning: It may impact the performance of your redis instance
     # slowlog-max-len: 128
+
+    # Collect INFO COMMANDSTATS output as metrics.
+    # command_stats: True
```
remh: Should be False as it's the default value

checks.d/redisdb.py
```diff
@@ -325,6 +327,21 @@ def _check_slowlog(self, instance, custom_tags):
 
         self.last_timestamp_seen[ts_key] = max_ts
 
+    def _check_command_stats(self, conn, tags):
+        """Get command-specific statistics from redis' INFO COMMANDSTATS command
+        """
+        try:
+            command_stats = conn.info("commandstats")
+        except:
```
remh: Bare except are evil :p

checks.d/redisdb.py
```diff
@@ -325,6 +327,21 @@ def _check_slowlog(self, instance, custom_tags):
...
+            self.warning("Could not retrieve command stats from Redis."
```
remh: nitpick but you should explicitly return after this i think

sethrosenblum: vs. the else?  okay.

checks.d/redisdb.py
sethrosenblum: fair enough, can just catch Exception, but we don't want anything that fails this process to fail the full agent check, right?

remh: yup

but `except:` will also catch SystemInterrupt which is no good.

except Exception is fine

sethrosenblum:
@remh good to go?

remh:
Looks good. Tests are passing. Squashing your commits and merging.
Thanks!

PR #2110, opened 2015-11-30
Daemontools integration.

gphat:
# What's this PR do?

Adds an integration for the service manager [daemontools](http://cr.yp.to/daemontools.html).
# Motivation

Stripe uses daemontools to manage most of our services. Many of those services have first-class integrations within Datadog so verifying their liveness is handled elsewhere. Some of these services, however, are lightweight, internal daemons that will likely never have healthcheck endpoints or other metrics. This check would allow liveness checks and uptime tracking. The goal is to detect failed or flapping services.
# Notes
- Adds a gauge metric for `daemontools.service.uptime` (tagged with the service name) and a service check of `daemontools.is_running` tagged by service.
- `svstat` that emits the info we need usually requires root privs, so this integration needs some love from `sudoers`. This miiiight be a candidate for storage in [our repository](https://github.com/stripe/stripe-datadog-checks)?


irabinovitch:
Test failures look unrelated. Digging in.

irabinovitch:
Clearing caches at Travis fixed this.

checks.d/daemontools.py
```diff
@@ -0,0 +1,39 @@
+# project
+from checks import AgentCheck
+import re
+import subprocess
+
+dt_status = re.compile(".*: (down|up) \(pid (\d+)\) (\d+) seconds")
```
remh: As per PEP8, this should be upper case if it's a constant.

checks.d/daemontools.py
```diff
@@ -0,0 +1,39 @@
...
+
+class DaemonToolsCheck(AgentCheck):
+    def check(self, instance):
+        path = self.init_config.get('path', '/etc/service')
```
remh: We are considering deprecating init_config at a point. Can you pass that at the instance level instead ?

checks.d/daemontools.py
```diff
@@ -0,0 +1,39 @@
...
+        service = instance.get('service')
```
remh: We should likely raise an Exception if `service` is None

checks.d/daemontools.py
```diff
@@ -0,0 +1,39 @@
...
+        tags = instance.get('tags', [])
+        tags.append("service:" + service)
+
+        try:
+            status = subprocess.Popen(['svstat', path + "/" + service], stdout=subprocess.PIPE, close_fds=True).communicate()[0]
```
remh: You should use https://github.com/DataDog/dd-agent/blob/master/utils/subprocess_output.py#L15 instead. It's safer to use.

checks.d/daemontools.py
```diff
@@ -0,0 +1,39 @@
...
+
+            check_message = status
```
remh: I'm not sure why you need 2 variables for the same thing.

checks.d/daemontools.py
```diff
@@ -0,0 +1,39 @@
...
+                if dt_result.group(1) == "up":
+                    check_status = AgentCheck.OK
+                    self.gauge('daemontools.service.uptime', float(dt_result.group(3)), tags)
+                elif dt_result.group(1) == "down":
+                    check_status == AgentCheck.CRITICAL
+                else:
+                    check_status == AgentCheck.UNKNOWN
+
+            self.service_check(
+                "daemontools.{0}.is_running".format(service),
```
remh: You should pass the service name as a tag instead.

checks.d/daemontools.py
```diff
@@ -0,0 +1,39 @@
...
+                check_status,
+                message=check_message,
+                tags=tags
+            )
+
+        except Exception:
```
remh: No need to do that.

If it fails, the exception will show up in the logs and the info page.
So you should get rid of the try/except.

gphat:
Done! Working on my test box:

screenshot 2015-12-22 11 41 45

gphat:
Oops, just realized I didn't change how I was running the subprocess. Will do so!

gphat:
Ok, now this is done. Note that this integration requires some fiddling with `sudoers`, so it may be something with us keeping in our repo? Thoughts?

truthbk:
Thanks @gphat! As always your contributions are very much appreciated :)

We'll look into the privileges issues with `sudo`, I guess that could pose some sort of problem due to the vector for privilege escalation it introduces (albeit restricted to the svstat binary). We'll look into it and elaborate.

Thanks again, very much appreciated!

irabinovitch:
@truthbk did we come to a consensus on the sudoers issue?

truthbk:
This is a perfect candidate for `integrations-extras` I believe, more so if we bear in mind the `sudoers` issue. I believe this belongs there.

truthbk:
Hi @gphat,

Thanks again for your contribution, always very much appreciated. We're currently transitioning our checks to a new SDK-based model for the agent. An unfortunate side-effect is that we have to impose a freeze on the dd-agent repo. The SDK consists of two repositories https://github.com/DataDog/integrations-core and https://github.com/DataDog/integrations-extras, where all of our checks will reside. This PR would likely belong in the **[extras](https://github.com/DataDog/integrations-extras)** repository. Would you be so kind as to re-open the PR there? We have facilitated some tools to help you with the SDK development that should help you get up and running in no time: 

truthbk:
Hi @gphat to ease the pains of transitioning to the SDK I have taken the liberty of opening a PR there with your work: https://github.com/DataDog/integrations-extras/pull/29. Please don't hesitate to let us know if you have any issues. Hopefully we can merge that in its new home very soon! 🙇

gphat:
Sounds good to me!

irabinovitch:
@gphat would it be ok if we close this PR and continue to collaborate on DataDog/integrations-extras#29 instead?

gphat:
Yup!

PR #2111, opened 2015-11-30
[postgres] FIX: pg8000 wasn't doing a good job with multiple relations

truthbk:
A customer was having issues with the postgres check, after investigating it became apparent that metrics weren't reported for all specified relations. It looks like [pg8000](https://pypi.python.org/pypi/pg8000) wasn't doing a good job with the parameter substitution when making executing the SQL query. 

The fix consists in having us build the query manually instead.

### DIFF
--- a/checks.d/postgres.py
+++ b/checks.d/postgres.py
@@ -133,7 +133,7 @@ SELECT mode,
         'query': """
 SELECT relname,schemaname,%s
   FROM pg_stat_user_tables
- WHERE relname = ANY(%s)""",
+ WHERE relname = ANY(array[%s])""",
         'relation': True,
     }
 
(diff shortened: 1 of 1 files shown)


truthbk:
Customer confirmed fix is working for them :smile:

checks.d/postgres.py
```diff
@@ -437,10 +437,10 @@ def _collect_stats(self, key, db, instance_tags, relations, custom_metrics):
                 try:
                     # if this is a relation-specific query, we need to list all relations last
                     if scope['relation'] and len(relations) > 0:
-                        relnames = relations_config.keys()
+                        relnames = ', '.join("'{0}'".format(w) for w in relations_config.keys())
```
remh: nitpick but you can use iterkeys instead of keys

truthbk:
As per @remh comments, switching to iterkeys() for dict iteration.

olivielpeau:
LGTM!

PR #2112, opened 2015-11-30
[http_check] Requests must use proxy if enabled in datadog.conf.

truthbk:


hkaj:
:shipit:

checks.d/http_check.py
```diff
@@ -143,6 +143,12 @@ class HTTPCheck(NetworkCheck):
 
     def __init__(self, name, init_config, agentConfig, instances):
         self.ca_certs = init_config.get('ca_certs', get_ca_certs_path())
+        self.proxy_host = agentConfig.get('proxy_host','')
+        self.proxy_port = agentConfig.get('proxy_port','')
+        self.proxies = {}
```
remh: i think self.proxies should be None if proxy settings are not set, so that requests uses its default settings.

Also looks like you are not taking into account the username and password.

Also you shouldn't read agentConfig yourself but likely call  `util.proxy.get_proxy` https://github.com/DataDog/dd-agent/blob/master/utils/proxy.py#L26

remh:
@truthbk can you take a pass at my comment above ?

truthbk:
@remh fixed as per your comments - username and password now taken into account for basic `HTTPAuth`

checks.d/http_check.py
```diff
@@ -143,6 +144,23 @@ class HTTPCheck(NetworkCheck):
 
     def __init__(self, name, init_config, agentConfig, instances):
         self.ca_certs = init_config.get('ca_certs', get_ca_certs_path())
+        proxy_settings = get_proxy(agentConfig)
+        if not proxy_settings:
+            self.proxies = None
+        else:
+            uri = "{host}:{port}".format(
```
remh: this kind of string formatting is not compatible with python 2.6 unfortunately :(

olivielpeau: AFAIK it is compatible, the incompatible format is the one with the empty braces (`{}`).

remh: :+1:

olivielpeau:
Looking good, merging.

PR #2115, opened 2015-12-01
[vsphere] Check metric type first to determine how to report

JohnLZeller:
Currently, all metrics from vsphere are reported as gauges. We should attempt to read the metric type from the VMware metrics module and call self.gauge or self.rate, etc. accordingly.

First came up in conversation with @LeoCavaille here: https://github.com/DataDog/dd-agent/pull/1824

### DIFF
--- a/checks.d/vsphere.py
+++ b/checks.d/vsphere.py
@@ -15,6 +15,7 @@ from pyVmomi import vim
 from checks import AgentCheck
 from checks.libs.thread_pool import Pool
 from checks.libs.vmware.basic_metrics import BASIC_METRICS
+from checks.libs.vmware.all_metrics import ALL_METRICS
 from util import Timer
 
 SOURCE_TYPE = 'vsphere'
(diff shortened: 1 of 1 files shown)


JohnLZeller:
This has yet to be tested, but I'd love some feedback.

JohnLZeller:
After taking another look, I think this is correct. @truthbk could I get a r?

checks.d/vsphere.py
```diff
@@ -748,7 +748,13 @@ def _collect_metrics_atomic(self, instance, mor):
                     continue
                 instance_name = result.id.instance or "none"
                 value = self._transform_value(instance, result.id.counterId, result.value[0])
-                self.gauge(
+
+                # Most common metric types are absolute, delta, and rate
+                if self.metrics_metadata[i_key][result.id.counterId]['s_type'] == 'rate':
```
truthbk: I think this would throw a `KeyError` exception, from what I saw in the check the metadata object would only have `name`, `unit` and `instance_tag` keys.

JohnLZeller: Hmmm, this is an older PR, I may have made a mistake here. Somehow I came to the conclusion that `s_type` was present because of this https://github.com/DataDog/dd-agent/blob/zeller/vsphere-basic/checks/libs/vmware/basic_metrics.py

JohnLZeller: You're absolutely right, `name`, `unit` and `instance_tag` are the only keys there. I'll fix this pr :)

JohnLZeller:
@truthbk okay can you give this another look?

truthbk:
This looks good now! :+1:

PR #2117, opened 2015-12-02
[mysql] Flavored mysql test

truthbk:
# Description
Adding a flavored test that will allow us to finally have a disposable environment for our MySQL tests. The new test brings up two MySQL instances, one for the master, and a second one for the slave, and we setup replication. 

## Improvable
* tarballs should probably be put in our S3
* we should get tarballs for multiple versions.
* replication seems to be working as the logs indicate, but replication test is failing - not sure if the problem is with the actual check and the newer 5.7.x STATUS tables - 'Slave_running' is definitely not there. We will know once we merge with the revamped MySQL check.

## Caveats
* `libaio1` is a dependency for mysqld, unavailable in the Travis executors. We therefore have to download manually, extract and employ `LD_LIBRARY_PATH` for the dynamic linking to work. Currently we download the deb from the ubuntu repos, and extract the `deb` to a directory of our choice. Not sure if the Travis executors will allow all necessary commands (namely: `apt-cache` and `dpkg-deb`) - we can always put a tarball in our S3 with the lib.

* test has to resort to system() instead of sh to start-up the mysql instances because sh won't allow prepending env variables in the call. There may be some other way, but I'm not aware of it currently. If init fails, we won't notice until the actual mysql client tries to connect to the DB's but that should work fine and take us to the clean-up code.

### DIFF
--- a/ci/mysql.rb
+++ b/ci/mysql.rb
@@ -1,20 +1,68 @@
 require './ci/common'
 
-# FIXME: use our own brew of MySQL like other flavors
+def mysql_version
+  ENV['FLAVOR_VERSION'] || '5.7.9'
+end
+
+
+def mysql_rootdir
+  "#{ENV['INTEGRATIONS_DIR']}/mysql_#{mysql_version}"
+end
 
 namespace :ci do
   namespace :mysql do |flavor|
     task before_install: ['ci:common:before_install']
 
-    task install: ['ci:common:install']
+    task install: ['ci:common:install'] do
+      # Downloads
+      # https://github.com/postgres/postgres/archive/#{pg_version}.tar.gz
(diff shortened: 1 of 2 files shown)


truthbk:
Issue found after creating PR: we need to get libaio in Travis:
`error while loading shared libraries: libaio.so.1: cannot open shared object file: `

truthbk:
MySQL test now passes but several stylistic offenses in ruby file must be fixed ^_^

irabinovitch:
@truthbk Looks like this might require a rebase if we still plan to include it in 5.7.

olivielpeau:
Closing this PR: using containers would be way easier than the approach of this PR, but Travis doesn't support them well yet.

#2226 explores using CircleCI instead, but is blocked on the lack of support of test matrices.

PR #2119, opened 2015-12-03
[Process] Option to search processes by PID

handigarde:


checks.d/process.py
```diff
@@ -249,14 +251,18 @@ def check(self, instance):
...
-            raise KeyError('The "search_string" is mandatory')
-
-        pids = self.find_pids(
-            name,
-            search_string,
-            exact_match,
-            ignore_ad=ignore_ad
-        )
+            if pid is not None:
+                pids = {proc for proc in psutil.pids() if proc == pid}
```
remh: Syntax error.
This shoud be:
```python
pids = [proc for proc in psutil.pids() if proc == pid]
```

or even better:
```python
pids = [psutil.Process(pid)]
```

remh:
@handigarde can you fix according to the comment please ?

handigarde:
@remh Thanks - not sure why I thought it was a set. Changed to a list.

checks.d/process.py
```diff
@@ -308,3 +314,4 @@ def _process_service_check(self, name, nb_procs, bounds):
             tags=tag,
             message=message_str % (status_str[status], nb_procs, name)
         )
+
```
olivielpeau: This blank line is making Flake8 complain: https://travis-ci.org/DataDog/dd-agent/jobs/99521898#L270

@handigarde could you remove it? :)

checks.d/process.py
```diff
@@ -249,14 +251,18 @@ def check(self, instance):
...
-        pids = self.find_pids(
-            name,
-            search_string,
-            exact_match,
-            ignore_ad=ignore_ad
-        )
+            if pid is not None:
+                pids = [psutil.Process(pid)]
+            else:
+                raise KeyError('The "search_string" is mandatory')
```
remh: That shouldn't be a KeyError. That could be a ValueError or just a bare Exception

checks.d/process.py
```diff
@@ -235,13 +235,15 @@ def check(self, instance):
         exact_match = _is_affirmative(instance.get('exact_match', True))
         search_string = instance.get('search_string', None)
         ignore_ad = _is_affirmative(instance.get('ignore_denied_access', True))
+        pid = instance.get('pid', None)
```
remh: nitpick but no need for the `, None` part.
```python
pid = instance.get('pid')
```
would do the same thing.

remh:
Thanks @handigarde 2 more nitpicks to fix and then it will be good to go.

irabinovitch:
@handigarde Just a friendly ping since we're in the final stretch for 5.7.  We'd love to include this if you can address the final comments from Remi. Thanks again!

handigarde:
@remh and @irabinovitch  sorry forgot to add a comment, but Remi's last two comments should be addressed. Let me know if I missed anything and I'll push an update!

olivielpeau:
@handigarde Had a second look at this PR, and I think we could simplify its logic by having one `if else` structure like:

```python
if search_string is not None:
  # check validity of `search_string` and generate list of pids with `find_pids`
elif pid is not None:
  pids = [psutil.Process(pid)]
else:
  raise ValueError('The "search_string" or "pid" options are required for process identification')
```

What do you think?

handigarde:
@olivielpeau Definitely a better flow - I'll re-work it

olivielpeau:
@handigarde Pushing this to the `5.9.0` milestone as we'll close the merge window for `5.8.0` soon.

remh:
Thanks!

PR #2122, opened 2015-12-03
[packaging][rpm] Use HTTP for yum on RedHat <= 5

olivielpeau:
yum on RedHat <= 5 doesn't support TLS1.2 so it can't use
our https repo. Use http instead.

See https://github.com/DataDog/chef-datadog/pull/240 for a detailed description of the issue.


yannmh:
Looking good !

PR #2124, opened 2015-12-03
Relocatable procfs for process check

sethp-jive:
When running the datadog agent as a container, we need to point psutils at a different procfs to gather metrics on processes outside the current namespace.

Note that this PR requires a hefty version bump for psutils; I'm not quite sure how best to check everything's working but so far it seems fine.

### DIFF
--- a/checks.d/gunicorn.py
+++ b/checks.d/gunicorn.py
@@ -68,7 +68,7 @@ class GUnicornCheck(AgentCheck):
         for proc in worker_procs:
             # cpu time is the sum of user + system time.
             try:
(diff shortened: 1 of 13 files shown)


irabinovitch:
Thanks @sethp-jive. We are currently working on bumping psutils to 3.3.0 in #2134.
That should make this PR a bit easier to merge.

sethp-jive:
Good to know, I'll keep my eye on that patch. One question on this pull: since other parts of datadog may use psutil, and PROCFS_PATH is a package global, does it make more sense to pull it from process.py up to the agent configuration?

remh:
Closing in favor of #2163

PR #2126, opened 2015-12-04
 fix _disk_ typo in marathon.py

pidah:


remh:
Thanks ! Confirmed that with the API Docs: https://mesosphere.github.io/marathon/docs/rest-api.html#get-v2-apps

PR #2133, opened 2015-12-08
Add counter support for go_expvar.

gphat:
# What's this PR do?

Adds support for counters by way of `increment` to the go_expvar check.

# Motivation

I'm unsure why the existing integration does not support counters and only rate. I feel like maybe I'm missing something here. In our use a rate will often do what I want, but it seems odd to convert to a rate when the full counter value is available. Also someone might want to alert on a raw total instead of a rate?

### DIFF
--- a/checks.d/go_expvar.py
+++ b/checks.d/go_expvar.py
@@ -16,12 +16,14 @@ TAGS = "tags"
 
 GAUGE = "gauge"
 RATE = "rate"
+COUNTER = "counter"
 DEFAULT_TYPE = GAUGE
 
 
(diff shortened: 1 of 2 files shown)


remh:
Thanks a lot @gphat !
Could you also add something in the example yaml file to show that it's now supported? 
https://github.com/DataDog/dd-agent/blob/master/conf.d/go_expvar.yaml.example

gphat:
Done. Sorry for not doing that in the first place!

remh:
Thanks!

PR #2134, opened 2015-12-08
Updating psutil to 3.3.0

truthbk:


truthbk:
relevant doc: https://github.com/giampaolo/psutil/blob/master/HISTORY.rst

talwai:
@truthbk the default configuration for agent_metrics collection will also need an update: https://github.com/DataDog/dd-agent/blob/master/conf.d/agent_metrics.yaml.default

truthbk:
@talwai thanks for the comment, fixed accordingly. We've now tested both the windows and linux agents with the updated version.

olivielpeau:
:+1:

Let's remember to upgrade psutil on our windows build VMs.

PR #2135, opened 2015-12-08
[system] Add uptime metric on Windows

remh:


checks/system/common.py
```diff
@@ -0,0 +1,8 @@
+from checks import Check
+
+# 3rd party
+import uptime
+
+class System(Check):
+    def check(self, agentConfig):
+        return {"system.uptime": uptime.uptime()}
```
olivielpeau: Newline missing, don't know if we should care though.

olivielpeau:
:+1:

checks/system/common.py
remh: fixed

PR #2136, opened 2015-12-08
[WMI] Switching IIS and Win Service Checks and Win32EventLog check to win32com. WMI refactor.

truthbk:
# WMI via `win32com` for IIS, Service Checks and Win32EventLog
This PR aims to replace the older wmi module and use the WMI sampler based off `win32com` provided by @yann and @TheCloudlessSky in #2011 .
Because there is plenty of common code a new `WinWMICheck` class was introduced that is a parent to all our `wmi`related checks:
* System
* IIS
* Windows Service
* Windows Event Log

Tests have been re-written to support the new current code layout and behavior, and a totally new test has been introduced for the windows event log check.


checks/libs/wmi/sampler.py
```diff
@@ -250,33 +253,84 @@ def _get_connection(self):
...
+    inclusive to True.
+    """
     @staticmethod
-    def _format_filter(filters):
+    def _format_filter(filters, inclusive=False):
         """
         Transform filters to a comprehensive WQL `WHERE` clause.
         """
-        def build_where_clause(fltr):
+        def build_where_clause(fltr, inclusive):
```
yannmh: You don't change the value of `inclusive` across your recursive calls, so I guess you don't need to have this argument redefined in `build_where_clause ` but only in `_format_filter`.

checks.d/iis.py
```diff
@@ -1,114 +1,154 @@
...
-                # Submit the metric value with the correct type
-                value = float(getattr(iis_site, wmi_val))
-                metric_func = getattr(self, mtype)
-                metric_func(metric, value, tags=tags)
-
-        for remaining_site in expected_sites:
+            uptime = wmi_obj["ServiceUptime"]
+            status = AgentCheck.CRITICAL if uptime == 0 else AgentCheck.OK
+
+            self.service_check("iis.site_up", status, tags=['site:{0}'.format(self.normalize(sitename))])
```
yannmh: Can `self.normalize(sitename)` be unicode encoded ?

checks/libs/wmi/sampler.py
```diff
@@ -250,33 +253,84 @@ def _get_connection(self):
...
+
+            The filters also support lists/tuples as property 'values' to filter.
+            When such a list is provided, we OR filter for all such property values:
+            - [{'Property': [value1, value2, value3]},...] or
+            - [{'Property': [(cmp_op, value1), (cmp_op, value2)]},...] or
+
+    Normally the filter is expected to be exclusive, such that all filter conditions
+    should be met - therefore we AND all filters. If you'd like to OR, then set
+    inclusive to True.
+    """
```
yannmh: Docstring should be moved in the method body.

checks/libs/wmi/types.py
```diff
@@ -0,0 +1,24 @@
+# stdlib
```
yannmh: Why `types.py` ? Also I am not sure it makes sense to have it in `checks/libs/wmi` as it's related to `checks/wmi.py` not the sampler (which is quite stand alone).

checks/wmi.py
```diff
@@ -0,0 +1,226 @@
+
```
yannmh: Should we name it `wmi_checks.py` instead to be consistent with `network_checks.py`: I feel like both have a similar role.

yannmh:
Great changes @truthbk ! Thanks for taking care of it.
Can't wait to see it merged and drop `wmi` dependency once and for all.

It's a significant size PR, so i added a lot of nitpicks and questions :smile:

checks/libs/wmi/types.py
truthbk: I did this to avoid pulling in the sampler which in turn depends on pywintypes. 

Although we mock pywintypes, we do so later, so this it was causing import errors with the tests. I will try to do it more elegantly.

checks.d/iis.py
truthbk: I added c35d693 to deal with it the case sitename was unicode encoded, because normalize wasn't ready for that and I wasn't sure what ExecQuery returned. I felt it probably made sense to normalize to ascii, it's probably what one would expect when calling normalize on a unicode string.

yannmh:
For a reason that I haven't identified yet, the `run_check` command does not work on `WinWMICheck`
```
>>> from checks import run_check
>>> run_check('wmi_check')
  File "shell.py", line 14, in shell
  File "<string>", line 1, in <module>
  File "utils\debug.pyc", line 33, in run_check
  File "checks\__init__.pyc", line 790, in check
NotImplementedError
```

yannmh:
Looks great @truthbk. 

I've been running more tests to assess that we have no regression with the 5.6.x versions in metrics, events and service checks collections.
 
### Summary
```
+===================+=============================+============+============+================+
| Check             | 5.6.x / 5.7.0 check runtime | Metrics    | Event      | Service checks |
+===================+=============================+============+============+================+
| System            | ~ 1                         | OK         | N/A        | N/A            |
+-------------------+-----------------------------+------------+------------+----------------+
| WMI               | ~ 1                         | OK         | N/A        | N/A            |
+-------------------+-----------------------------+------------+------------+----------------+
| IIS               | 0.70                        | OK         | N/A        | OK             |
+-------------------+-----------------------------+------------+------------+----------------+
| Windows Service   | 80                          | N/A        | N/A        | OK             |
+-------------------+-----------------------------+------------+------------+----------------+
| Windows Log event | Not tested                  | Not tested | Not tested | Not tested     |
+===================+=============================+============+============+================+
```


Do you mind taking a pass on the 'Windows Log event' when you have a chance please ?

truthbk:
After the changes to the filtering in the last commit we might need to conduct a couple more tests to check regressions aren't present. Nothing has changed effectively, we just now place the parentheses a little differently in the WQL - we avoid adding more between AND clauses, and we isolate OR clauses with them.

truthbk:
Some things break after rebase due to issue with @yannmh timeout changes: https://github.com/DataDog/dd-agent/pull/2185

checks.d/iis.py
```diff
@@ -1,114 +1,152 @@
...
+        ('TotalAnonymousUsers', 'iis.users.anon', 'rate'),
+        ('TotalNonAnonymousUsers', 'iis.users.nonanon', 'rate'),
 
         # Requests
-        ('iis.requests.cgi', 'rate', 'TotalCGIRequests'),
-        ('iis.requests.isapi', 'rate', 'TotalISAPIExtensionRequests'),
+        ('TotalCGIRequests', 'iis.requests.cgi', 'rate'),
+        ('TotalISAPIExtensionRequests', 'iis.requests.isapi', 'rate'),
     ]
     SERVICE_CHECK = "iis.site_up"
```
yannmh: Looks like you don't use this variable.

yannmh:
@truthbk I think IIS and Windows Service checks need a quick refresh to comply well with timeouts (c.f https://github.com/DataDog/dd-agent/pull/2185).

checks.d/iis.py
truthbk: I've changed the service_check() calls to use it.

setup.py
```diff
@@ -52,7 +52,6 @@
     # Modules to force-include in the exe
     include_modules = [
         # 3p
-        'wmi',
```
yannmh: Let's keep it for now

truthbk:
## TODO

@yannmh: 
   - Raise `WMITimeoutException` when timeout
   - Cache connections in `list` instead of `set`
   - Gracefully handle `WMITimeoutException`

@truthbk:
   - Remove HashableConnection
   - Get rid of `done()`
   - Handle use-case where we're still sampling for service-checks (handle `WMITimeoutException`).

utils/containers.py
```diff
@@ -0,0 +1,17 @@
+
+
+def freeze(o):
+    """
+    Freezes any mutable object including dictionaries and lists for hashing.
+    Accepts nested dictionaries.
+    """
+    if isinstance(o, dict):
+        return frozenset({k:freeze(v) for k,v in o.items()}.items())
```
olivielpeau: python 2.6 doesn't support dict comprehension, you should change
```python
{k:freeze(v) for k,v in o.items()}
```
to
```python
dict([(k, freeze(v)) for k, v in o.items()])
```

truthbk: good feedback, thanks. Fixed.

tests/core/test_wmi.py
```diff
@@ -357,12 +379,131 @@ def test_wql_filtering(self):
...
+
+    def test_wql_filtering_op_adv(self):
+        """
+        Format the filters to a comprehensive WQL `WHERE` clause w/ mixed filter containing regular and operator modified properties.
+        """
+        from checks.libs.wmi import sampler
+        format_filter = sampler.WMISampler._format_filter
+
+        # Check `_format_filter` logic
+        filters = [{'Name': "Foo%"}, {'Name': "Bar*", 'Id': ('>=', "SomeId")}, {'Name': "Zulu"}]
```
olivielpeau: The wildcard character in WQL `WHERE` clauses is only `%` (not `*`), I've confirmed that by reading through https://technet.microsoft.com/en-us/library/ee176998.aspx and manually testing on a windows VM.

In terms of BC, the only check that was supporting the wildcard was the `wmi_check` (see https://github.com/DataDog/dd-agent/blob/5.5.x/checks.d/wmi_check.py#L51-L55), it was actually looking for a `*` in the filters and replacing it with a `%`.

Given that the wildcard feature was broken in 5.6.x I think we could simply only allow `%` as a wildcard in the `filters`. What do you think?

olivielpeau: Actually the wildcard was also implemented in the `event_log` check, but w/o the weird `*`-> `%` replacement, so I think we have another argument in favor of only allowing the `%` wildcard in filters :)

olivielpeau:
Went through all the filtering logic, only added one comment. It's great work! (huge :heart: for all the tests)

checks.d/iis.py
```diff
@@ -1,114 +1,165 @@
...
+            expected_sites.remove(sitename)
+
+        for site in expected_sites:
+            self.service_check(self.SERVICE_CHECK, AgentCheck.CRITICAL,
+                               tags=['site:{0}'.format(self.normalize(site))])
+
+
+    def _submit_metrics(self, wmi_metrics, metrics_by_property):
+        for m in wmi_metrics:
+            if m.name == "TotalBytesTransfered":
```
olivielpeau: could we keep the comment explaining why we're doing this? :)

olivielpeau:
Looking great! Had one tiny nitpick.

Could you squash your commits? Then feel free to merge once the CI passes.

PR #2138, opened 2015-12-09
[nginx] report nginx.server_zone.response.* as count

hkaj:
Asked by a user. After investigation I think it could also make sense for other metrics, let's discuss it.


irabinovitch:
ElasticSearch tests look to just be flakey, the passed on 2nd run. but the NGINX failures look relevant.

PR #2143, opened 2015-12-12
Lithium/elastic stats

bdharrington7:
Adds additional important statistics that can be monitored by the datadog agent. 

I wasn't sure how to verify the version of introduction of these statistics to elasticsearch, so I made the assumption that they've always been available. If this is incorrect or if there's a place to be able to find this information, please let me know

### DIFF
--- a/checks.d/elastic.py
+++ b/checks.d/elastic.py
@@ -98,6 +98,8 @@ class ESCheck(AgentCheck):
         "elasticsearch.search.fetch.total": ("gauge", "indices.search.fetch_total"),
         "elasticsearch.search.fetch.time": ("gauge", "indices.search.fetch_time_in_millis", lambda v: float(v)/1000),
         "elasticsearch.search.fetch.current": ("gauge", "indices.search.fetch_current"),
+        "elasticsearch.indices.segments.count": ("gauge", "indices.segments.count"),
+        "elasticsearch.indices.segments.memory_in_bytes": ("gauge", "indices.segments.memory_in_bytes"),
         "elasticsearch.merges.current": ("gauge", "indices.merges.current"),
         "elasticsearch.merges.current.docs": ("gauge", "indices.merges.current_docs"),
         "elasticsearch.merges.current.size": ("gauge", "indices.merges.current_size_in_bytes"),
@@ -158,6 +160,9 @@ class ESCheck(AgentCheck):
         "jvm.mem.non_heap_used": ("gauge", "jvm.mem.non_heap_used_in_bytes"),
         "jvm.threads.count": ("gauge", "jvm.threads.count"),
         "jvm.threads.peak_count": ("gauge", "jvm.threads.peak_count"),
+        "elasticsearch.fs.total.total_in_bytes": ("gauge", "fs.total.total_in_bytes"),
+        "elasticsearch.fs.total.free_in_bytes": ("gauge", "fs.total.free_in_bytes"),
+        "elasticsearch.fs.total.available_in_bytes": ("gauge", "fs.total.available_in_bytes"),
     }
 
     JVM_METRICS_POST_0_90_10 = {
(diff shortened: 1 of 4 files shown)


irabinovitch:
@bdharrington7  Thanks for the additions, have you had a chance to look into why tests are failing?

checks.d/elastic.py
```diff
@@ -158,6 +166,15 @@ class ESCheck(AgentCheck):
         "jvm.mem.non_heap_used": ("gauge", "jvm.mem.non_heap_used_in_bytes"),
         "jvm.threads.count": ("gauge", "jvm.threads.count"),
         "jvm.threads.peak_count": ("gauge", "jvm.threads.peak_count"),
+        "elasticsearch.fs.total.total_in_bytes": ("gauge", "fs.total.total_in_bytes"),
+        "elasticsearch.fs.total.free_in_bytes": ("gauge", "fs.total.free_in_bytes"),
+        "elasticsearch.fs.total.available_in_bytes": ("gauge", "fs.total.available_in_bytes"),
+        "elasticsearch.fs.total.disk_reads": ("gauge", "fs.total.disk_reads"),
```
remh: Should that metric be a rate instead ?

bdharrington7:
Hi @irabinovitch I haven't been able to dig into the failures, because I can't reproduce them locally. I've looked at the commands that are run in the build, but I get much different output. Here's my commands and output:

https://gist.github.com/bdharrington7/3d201d6651c8850c2c67

checks.d/elastic.py
bdharrington7: Should it? I've only seen examples using gauge, so I wasn't sure what "rate" is used for.

remh: If the value is continuously increasing (as i expect given the metric name) it makes more sense to send it as a rate. The agent would then compute a "rate per second" which would be way more actionable in a monitor or in a graph than a continuously increasing value. Feel free to ping me if you have more questions!

olivielpeau:
Hi @bdharrington7, could you change the type of the metrics that continuously increase to `rate` instead of gauge?

You can see in the output of the Travis jobs that some of the metrics are not available on all versions of Elasticsearch (see for instance https://travis-ci.org/DataDog/dd-agent/jobs/96399346#L304-L313 , on version `1.0.3` of ES). The `indices.segments.*` metrics in particular seem to exist only since ES 1.4, could you try to fetch these metrics only on ES 1.4+ ?

Also it seems that the following metrics:
```
fs.total.disk_read_size_in_bytes
fs.total.disk_writes
fs.total.disk_io_op
fs.total.disk_reads
fs.total.disk_io_size_in_bytes
fs.total.disk_write_size_in_bytes
```

are missing when Elasticsearch is run on Travis (or on any container-based environment). I think you can remove them from the list of the metrics that are tested in `test_elastic.py`.

FYI You can run the integration tests of the elasticsearch check locally by running:
```bash
$ FLAVOR_VERSION=1.0.3 rake ci:run[elasticsearch]
```
and replacing the value of `FLAVOR_VERSION` with the version of elasticsearch you want to test against.

Thanks!

bdharrington7:
@olivielpeau I'm trying to add new flavors / versions of elasticsearch (because it currently only goes to 1.6.0, and I know these metrics exist on 1.7.x), is there some step to add these package versions to the s3 repo for testing?

olivielpeau:
@bdharrington7 I've added the `1.7.4` flavor/version to our mirror, you'll just need to add the flavor here: https://github.com/DataDog/dd-agent/blob/4395b3661185cdc223c881e68bdb82e19afae87a/.travis.yml#L47 and it should run on Travis.

bdharrington7:
Thank you! Also, is the convention to use the latest patch version as it comes out? As in, can I change the flavor `1.6.0` to `1.6.2` (and do you also have to upload that as well)?

olivielpeau:
Yes, we try to use the latest patch version. I've uploaded `1.6.2` if you want to update it ;)

bdharrington7:
So I'm trying to run these tests locally, but I'm not seeing the download as usual, but I do see:
```
echo 70643 > $VOLATILE_DIR/elasticsearch.pid
sh: /Users/brian.harrington/github/dd-agent/embedded/es_1.6.2/bin/elasticsearch: No such file or directory
```

and it just times out. (because nothing is there)

I'm running this command, I get the same thing for 1.7.4...
```
FLAVOR_VERSION=1.6.2 rake ci:run[elasticsearch]
```

bdharrington7:
And in the ci tests on travis, I'm not really sure how to resolve this error: https://travis-ci.org/DataDog/dd-agent/jobs/100261151#L269

It looks like it might be a race condition where the test runs before the elasticsearch node is ready to respond...

tests/checks/integration/test_elastic.py
```diff
@@ -83,6 +85,8 @@
     "elasticsearch.merges.total.size": ("gauge", "indices.merges.total_size_in_bytes"),
     "elasticsearch.refresh.total": ("gauge", "indices.refresh.total"),
     "elasticsearch.refresh.total.time": ("gauge", "indices.refresh.total_time_in_millis", lambda v: float(v)/1000),
+    "elasticsearch.indices.translog.size_in_bytes": ("gauge", "indices.translog.size_in_bytes"),
+    "elasticsearch.indices.translog.operations": ("gauge", "indices.translog.operations"),
```
olivielpeau: I think the tests on `0.90.13` are failing because you're testing these 2 metrics on all the versions of elasticsearch. These metrics should only be listed in `ADDITIONAL_METRICS_POST_1_0_0 ` below.

bdharrington7:
Cool, so now all my checks pass, but the activemq one fails. Still ok to merge?

olivielpeau:
I've rerun the activemq test, it's passing now.

One last thing: could you squash your commits into one? Once that's done I'll merge, thanks!

bdharrington7:
Done! looks like the supervisord one needs to be kicked this time...

olivielpeau:
Looking good, thanks again @bdharrington7 !

Merging.

PR #2144, opened 2015-12-14
[core] do not log API keys

yannmh:
Create a `utils.logger` module:
* Move `log_exceptions` from `utils.debug`
* `log_no_api_key` patches a logger in a method scope to obfuscate API
 keys

Apply `log_no_api_key` on some forwarder methods.

### DIFF
--- a/checks/collector.py
+++ b/checks/collector.py
@@ -29,7 +29,7 @@ from util import (
     get_uuid,
     Timer,
 )
-from utils.debug import log_exceptions
+from utils.logger import log_exceptions
 from utils.jmx import JMXFiles
 from utils.platform import Platform
 from utils.subprocess_output import get_subprocess_output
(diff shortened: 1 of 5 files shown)


yannmh:
@talwai can you take a pass on this one please ?

talwai:
Looks nice. Can we get some benchmarks on the cost of your obfuscation method? Another implementation could rely on overriding the `LogRecord` class: http://stackoverflow.com/questions/29631171/custom-processing-on-python-logging-messages-in-a-generic-way

Would be interesting to see which is more performant

yannmh:
Thanks for sharing @talwai.
I like the custom `LogRecord` approach a lot more:
1. It's a more natural approach compared to a hacky decorator
2. It's easy to patch the entire `ddagent` module without having to repeat the `log_no_api_key` decoration

I made a new PR: https://github.com/DataDog/dd-agent/pull/2146. Would you mind taking a quick look at it please ? 

Still, I'll run some benchmarks to see what's the cost of obfuscation and compared both approaches. I'll let you know what I find.

PR #2145, opened 2015-12-15
Modify get_hostname to use the docker hostname when available

hkaj:


utils/dockerutil.py
```diff
@@ -138,3 +146,34 @@ def container_name_extractor(co):
...
+    if not os.path.exists(conf_path):
+        default_conf_path = os.path.join(confd_path, '%s.yaml.default' % CHECK_NAME)
+        if not os.path.exists(default_conf_path):
+            log.error("Couldn't find any configuration file for the docker check."
+                      " Not using the docker hostname.")
+            return None
+        else:
+            conf_path = default_conf_path
+
+    conf_path = os.path.join(confd_path, '%s.yaml' % CHECK_NAME)
```
remh: That's the same exact line as line 162

utils/dockerutil.py
```diff
@@ -138,3 +146,34 @@ def container_name_extractor(co):
...
+    check_config = check_yaml(conf_path)
+    init_config, instances = check_config.get('init_config', {}), check_config['instances']
+    init_config = {} if init_config is None else init_config
+    if len(instances) > 0:
+        set_docker_settings(init_config, instances[0])
+        name = get_client().info()['Name']
```
remh: you should return `get_client().info().get('Name')` instead.

PR #2146, opened 2015-12-15
[core] do not log API keys

yannmh:
Create a `utils.logger` module:
* Move `log_exceptions` from `utils.debug`
* `NoAPIKeyLogRecord` is a custom `LogRecord` that obfuscates API keys
  logging

### DIFF
--- a/checks/collector.py
+++ b/checks/collector.py
@@ -29,7 +29,7 @@ from util import (
     get_uuid,
     Timer,
 )
-from utils.debug import log_exceptions
+from utils.logger import log_exceptions
 from utils.jmx import JMXFiles
 from utils.platform import Platform
 from utils.subprocess_output import get_subprocess_output
--- a/ddagent.py
+++ b/ddagent.py
@@ -56,7 +56,10 @@ from util import (
     json,
     Watchdog,
 )
+from utils.logger import RedactedLogRecord
 
+
+logging.LogRecord = RedactedLogRecord
 log = logging.getLogger('forwarder')
 log.setLevel(get_logging_config()['log_level'] or logging.INFO)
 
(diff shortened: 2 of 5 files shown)


yannmh:
## Benchmarking

### Approaches
Available approaches are:

1. No processing, i.e API key is diplayed
2. Decorator hack to obfuscate (c.f. https://github.com/DataDog/dd-agent/pull/2144)
3. Custom `logging` LogRecord (exposed in this PR)
4. ~~Custom `logging` Filter.~~ Not a valid approach (used to exclude log lines or add extra logging information)

### Script
Available here: https://gist.github.com/yannmh/515f8560e0a3d1ca91c9

### Results

```
-> % python logging_benchmark.py
No processing: 0.028247
Decorator hack: 0.069303
Custom LogRecord: 0.044454
```

:trophy: Custom LogRecord :trophy: 

cc @talwai @remh

remh:
@yannmh i don't understand why is the filter not a good fit?

something like
```
class ApiKeyLoggingFilter(logging.Filter):

    def filter(self, record):
        record.msg = record.msg.replace(apikey, "REDACTED")
        return True
```

should work i think

yannmh:
@remh,
In this case `record.msg` returns the log message template, i.e `u"Sending %s to endpoint %s at %s"`. While `record.getMessage()` would return the formatted message,  there is no setter to change it. 

So to obfuscate the API key, there would be two options

1. Iterate over the `record` parameters to obfuscate API keys if found. It's slow and not elegant (same approach was done in the decorator approch)
2. Override `record.getMessage()` method, which is what is done with the custom `LogRecord ` approach.

talwai:
:+1: to using `LogRecord` - I like this approach. Couple of things:
1. Can we add a quick test for the overridden `getMessage` function?
2. I think `RedactedLogRecord` might be a nicer name, but that's just an opinion :)

Looks good otherwise

yannmh:
Thanks for you reviews @talwai, @remh 
 
1. Done :white_check_mark:
2. I agree. I switched the name to `RedactedLogRecord ` :lipstick:

remh:
:+1:

PR #2147, opened 2015-12-15
Minerkasch/yarn

wjsl:
PR contains a dd-agent check for YARN clusters.


checks.d/yarn.py
```diff
@@ -0,0 +1,264 @@
...
+YARN_NODES_PATH = '/ws/v1/cluster/nodes'
+
+# Metric types
+GAUGE = 'gauge'
+
+YARN_CLUSTER_METRICS_ELEMENT = 'clusterMetrics'
+
+# Cluster metrics for YARN
+YARN_CLUSTER_METRICS = {
+    'appsSubmitted': ('yarn.metrics.appsSubmitted', GAUGE),
```
remh: The usual datadog metrics name don't use camelCase usually but camel_case.

checks.d/yarn.py
```diff
@@ -0,0 +1,264 @@
...
+    def _get_cluster_id(self, rm_address):
+
+        # Get the cluster info
+        info_json = self._rest_request_to_json(rm_address, INFO_URI)
+
+        if info_json:
+            if info_json['clusterInfo']:
+                if info_json['clusterInfo']['id']:
+                    return info_json['clusterInfo']['id']
+
```
remh: Should we raise an exception here ?
```python
def _get_cluster_id(self, rm_address):
    info_json = self._rest_request_to_json(rm_address, INFO_URI)
    cluster_id = info_json.get("clusterInfo", {}).get("id")
    if cluster_id is not None:
        return cluster_id
    raise Exception("Unable to retrieve cluster id")
```

checks.d/yarn.py
```diff
@@ -0,0 +1,264 @@
...
+
+    def _yarn_cluster_metrics(self, hostname, cluster_id, rm_address):
+        metrics_json = self._rest_request_to_json(rm_address, YARN_CLUSTER_METRICS_PATH)
+
+        if metrics_json:
+            keys = ['cluster_id:' + str(cluster_id)]
+
+            if metrics_json[YARN_CLUSTER_METRICS_ELEMENT] is not None:
+
```
remh: Nitpick but those blank lines are not necessary

checks.d/yarn.py
```diff
@@ -0,0 +1,264 @@
...
+        return None
+
+
+    def _yarn_cluster_metrics(self, hostname, cluster_id, rm_address):
+        metrics_json = self._rest_request_to_json(rm_address, YARN_CLUSTER_METRICS_PATH)
+
+        if metrics_json:
+            keys = ['cluster_id:' + str(cluster_id)]
+
+            if metrics_json[YARN_CLUSTER_METRICS_ELEMENT] is not None:
```
remh: I think that you should assign
metrics_json.get(YARN_CLUSTER_METRICS_ELEMENT) to a variable and use it after.

checks.d/yarn.py
```diff
@@ -0,0 +1,264 @@
...
+            if metrics_json['apps'] is not None:
+                if metrics_json['apps']['app'] is not None:
+
+                    for app_json in metrics_json['apps']['app']:
+
+                        cluster_id = app_json['clusterId']
+                        app_id = app_json['id']
+
+                        keys = ['cluster_id:' + str(cluster_id)]
+                        keys.append('app_id:' + str(app_id))
```
remh: Why not adding this tag when instanciating the list ?
Also you should call the list `tags` instead of `keys`
```python
tags = ["cluster_id:%s" % cluster_id, "app_id:%s" % app_id]
```
Als

checks.d/yarn.py
```diff
@@ -0,0 +1,264 @@
...
+
+            if metrics_json[status] is not None:
+                self._set_metric(metric_name,
+                    metric_type,
+                    metrics_json[status],
+                    keys,
+                    hostname)
+
+
+    def _set_metric(self, metric_name, metric_type, value, tags=None, hostname=None, device_name=None):
```
remh: you shouldn't pass the hostname to your metrics as it will most likely be localhost.

checks.d/yarn.py
```diff
@@ -0,0 +1,264 @@
...
+
+    def _rest_request_to_json(self, address, object_path):
+        '''
+        Return None if error occurred
+        '''
+
+        response_json = None
+
+        try:
+            uri = HTTP + address + object_path
```
remh: You should use urljoin instead of concatenating the url yourself

checks.d/yarn.py
```diff
@@ -0,0 +1,264 @@
...
+            response = urllib2.urlopen(uri)
```
remh: Can you use the requests library instead? It's simpler to use

checks.d/yarn.py
```diff
@@ -0,0 +1,264 @@
...
+            response_json = json.load(response)
+
+        except (HTTPError, URLError, ValueError) as e:
```
remh: No need for an event here. But you should send service checks instead

remh:
Thanks that's a pretty nice first version! Nicely written and easy to read!

Can you ping me when the comments made are fixed please ?

Thanks again

irabinovitch:
Retrigered the ZK and ES. I dont think they failed a result of this PR.

checks.d/yarn.py
```diff
@@ -0,0 +1,287 @@
...
+
+    def __init__(self, name, init_config, agentConfig, instances=None):
+        AgentCheck.__init__(self, name, init_config, agentConfig, instances)
+        self._timeout = DEFAULT_TIMEOUT
+
+    def check(self, instance):
+
+        # Get properties from conf file
+        rm_address = instance.get('resourcemanager_uri', DEFAULT_RM_URI)
+        self._timeout = int(instance.get('timeout', DEFAULT_TIMEOUT))
```
remh: You shouldn't store instance level setting at the check level.

checks.d/yarn.py
zachradtka: I have updated the check to raise an exception if the cluster ID can't be found. I am just wondering if this is still an issue.

remh:
Thanks a lot @wjsl 
This looks almost ready ?
Can you make sure to respect the [contributing guidelines](https://github.com/DataDog/dd-agent/blob/master/CONTRIBUTING.md) please ?

Especially regarding the commit messages. Can you squash your commits into one and also write a more descriptive PR message ? 

Thanks again!

zachradtka:
Squashed and updated the commit message. Anything else?

wjsl:
I looked at the TravisCI failure and they seem to be related to ZK, which we shouldn't be touching. Anything on our end we can do to get this PR closed?

checks.d/yarn.py
remh: Thanks @zachradtka 
I think it would be better to have:
```python
cluster_id = info_json.get("clusterInfo", {}).get("id")
if cluster_id is not None:
   ...
```
instead of all the if statements. Also this would ensure that we wouldn't get a KeyError

remh:
Thanks @wjsl @zachradtka added one last comment.
Regarding the failing ZK test, can you rebase your branch with master ? That should fix the issue. We'll be able to merge after.

PR #2152, opened 2015-12-17
[snmp] convert into network-check to run instances in parallel

truthbk:
# Multi-threaded SNMP Check (as a Network Check)
This PR allows checking for multiple SNMP instances in parallel, each running in it's own thread (within the allocated thread-pool). The actual OID/walk is still done serially for each instance, but this PR should help organizations with multiple instances configured with a performance benefit as a factor of the thread pool size.

NOTE: tests involving rates now require three runs because network checks report the results of the nth run on the n+1 run.

### DIFF
--- a/checks.d/snmp.py
+++ b/checks.d/snmp.py
@@ -9,7 +9,7 @@ from pysnmp.smi import builder
 from pysnmp.smi.exval import noSuchInstance, noSuchObject
 
 # project
-from checks import AgentCheck
+from checks.network_checks import NetworkCheck, Status
 from config import _is_affirmative
 
 
@@ -41,29 +41,43 @@ def reply_invalid(oid):
         noSuchObject.isSameTypeWith(oid)
 
 
-class SnmpCheck(AgentCheck):
+class SnmpCheck(NetworkCheck):
 
+    SOURCE_TYPE_NAME = 'system'
     cmd_generator = None
     # pysnmp default values
     DEFAULT_RETRIES = 5
     DEFAULT_TIMEOUT = 1
+    SC_STATUS = 'snmp.can_check'
 
-    def __init__(self, name, init_config, agentConfig, instances=None):
-        AgentCheck.__init__(self, name, init_config, agentConfig, instances)
-
-        # Load Custom MIB directory
-        mibs_path = None
-        ignore_nonincreasing_oid = False
-        if init_config is not None:
-            mibs_path = init_config.get("mibs_folder")
-            ignore_nonincreasing_oid = _is_affirmative(
-                init_config.get("ignore_nonincreasing_oid", False))
-
-        # Create SNMP command generator and aliases
-        self.create_command_generator(mibs_path, ignore_nonincreasing_oid)
+    def __init__(self, name, init_config, agentConfig, instances):
+        for instance in instances:
+            if 'name' not in instance:
+                instance['name'] = instance['ip_address']
+            instance['skip_event'] = True
 
         # Set OID batch size
         self.oid_batch_size = int(init_config.get("oid_batch_size", DEFAULT_OID_BATCH_SIZE))
+
+        # Load Custom MIB directory
+        self.mibs_path = None
+        ignore_nonincreasing_oid = False
+        if init_config is not None:
+            self.mibs_path = init_config.get("mibs_folder")
+            self.ignore_nonincreasing_oid = _is_affirmative(
+                init_config.get("ignore_nonincreasing_oid", False))
+
+        NetworkCheck.__init__(self, name, init_config, agentConfig, instances)
+
+    def _load_conf(self, instance):
+        tags = instance.get("tags", [])
+        ip_address = instance["ip_address"]
+        metrics = instance.get('metrics', [])
+        timeout = int(instance.get('timeout', self.DEFAULT_TIMEOUT))
+        retries = int(instance.get('retries', self.DEFAULT_RETRIES))
+
+        return ip_address, tags, metrics, timeout, retries
+
 
     def snmp_logger(self, func):
         """
@@ -242,20 +256,24 @@ class SnmpCheck(AgentCheck):
         self.log.debug("Raw results: {0}".format(results))
         return results
 
-    def check(self, instance):
+    def _check(self, instance):
         '''
         Perform two series of SNMP requests, one for all that have MIB asociated
         and should be looked up and one for those specified by oids
         '''
-        tags = instance.get("tags", [])
-        ip_address = instance["ip_address"]
+
+        ip_address, tags, metrics, timeout, retries = self._load_conf(instance)
+
+        tags += ['snmp_device:{0}'.format(ip_address)]
+
         table_oids = []
         raw_oids = []
-        timeout = int(instance.get('timeout', self.DEFAULT_TIMEOUT))
-        retries = int(instance.get('retries', self.DEFAULT_RETRIES))
+
+        # Create SNMP command generator and aliases
+        self.create_command_generator(self.mibs_path, self.ignore_nonincreasing_oid)
 
         # Check the metrics completely defined
-        for metric in instance.get('metrics', []):
+        for metric in metrics:
             if 'MIB' in metric:
                 try:
                     assert "table" in metric or "symbol" in metric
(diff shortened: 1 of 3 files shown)


checks.d/snmp.py
```diff
@@ -242,20 +256,24 @@ def check_table(self, instance, oids, lookup_names, timeout, retries):
...
+
+        tags += ['snmp_device:{0}'.format(ip_address)]
+
         table_oids = []
         raw_oids = []
-        timeout = int(instance.get('timeout', self.DEFAULT_TIMEOUT))
-        retries = int(instance.get('retries', self.DEFAULT_RETRIES))
+
+        # Create SNMP command generator and aliases
+        self.create_command_generator(self.mibs_path, self.ignore_nonincreasing_oid)
```
yannmh: Any reason why this was moved from `__init__` to `_check` ?

checks.d/snmp.py
```diff
@@ -316,17 +342,15 @@ def report_raw_metrics(self, instance, results):
                 name = metric.get('name', 'unnamed_metric')
                 self.submit_metric(name, value, tags)
 
-    def report_table_metrics(self, instance, results):
+    def report_table_metrics(self, metrics, results, tags):
```
yannmh: :clap: for the cleanup :construction:

checks.d/snmp.py
```diff
@@ -300,14 +327,13 @@ def report_raw_metrics(self, instance, results):
...
-        tags = instance.get("tags", [])
-        tags = tags + ["snmp_device:" + instance.get('ip_address')]
-        for metric in instance.get('metrics', []):
+        for metric in metrics:
             if 'OID' in metric:
                 queried_oid = metric['OID']
                 for oid in results:
                     if oid.startswith(queried_oid):
                         value = results[oid]
+                        self.log.warning("Found matching oid: {0} - {1}".format(oid, value))
```
yannmh: I think you meant a `self.log.debug` here no ?

tests/checks/integration/test_snmp.py
```diff
@@ -107,19 +113,58 @@ class SNMPTestCase(AgentCheckTest):
...
+        Raise after
+        """
+        i = 0
+        while i < RESULTS_TIMEOUT:
+            self.check._process_results()
+            if len(getattr(self.check, attribute)) >= count:
+                return getattr(self.check, method)()
+            time.sleep(1)
+            i += 1
+        raise Exception("Didn't get the right count of service checks in time, {0}/{1} in {2}s: {3}"
```
yannmh: The method is generic, not the exception message mentioning "the right count of service checks"

tests/checks/integration/test_snmp.py
```diff
@@ -131,8 +176,11 @@ def test_command_generator(self):
 
         # Test command generator `ignoreNonIncreasingOid` parameter
         self.run_check(self.IGNORE_NONINCREASING_OID, force_reload=True)
+        self.wait_for_async_attrib('cmd_generator')
         self.assertTrue(self.check.cmd_generator.ignoreNonIncreasingOid)
 
+        self.coverage_report()
```
yannmh: Why a coverage report ? No service check or metric is tested.

tests/checks/integration/test_snmp.py
```diff
@@ -107,19 +113,58 @@ class SNMPTestCase(AgentCheckTest):
...
+            i += 1
+        raise Exception("Attribute not created in time.")
+
+
     def test_command_generator(self):
         """
         Command generator's parameters should match init_config
         """
-        self.run_check(self.MIBS_FOLDER)
+        self.run_check_twice(self.MIBS_FOLDER)
```
yannmh: I am not sure to understand why you need to run it twice here.

tests/checks/integration/test_snmp.py
```diff
@@ -169,17 +220,21 @@ def test_snpget(self):
...
         }
+        self.run_check_twice(config)
+        self.service_checks = self.wait_for_async('get_service_checks', 'service_checks', 1)
+
+        # Test service check
+        self.assertServiceCheck("snmp.can_check", status=AgentCheck.OK,
+                                tags=self.CHECK_TAGS, count=1)
+
         self.run_check(config)
+        time.sleep(1)
```
yannmh: Shall we `wait_for` `self.metrics` instead ?

tests/checks/integration/test_snmp.py
```diff
@@ -234,7 +295,10 @@ def test_invalid_metric(self):
         config = {
             'instances': [self.generate_instance_config(self.INVALID_METRICS)]
         }
-        self.assertRaises(Exception, self.run_check, config)
+        self.run_check(config)
+        self.service_checks = self.wait_for_async('get_service_checks', 'service_checks', 1)
+
+        # no exception raised as it's thrown in thread: self.assertRaises(Exception, self.run_check, config)
```
yannmh: Can we assert that something was `CRITICAL` or `EXCEPTION` logged instead ?

tests/checks/integration/test_snmp.py
```diff
@@ -253,8 +317,10 @@ def test_network_failure(self):
         config = {
             'instances': [instance]
         }
+        self.run_check(config)
+        self.service_checks = self.wait_for_async('get_service_checks', 'service_checks', 1)
 
-        self.assertRaises(Exception, self.run_check, config)
+        # no exception raised as it's thrown in thread: self.assertRaises(Exception, self.run_check, config)
```
yannmh: Same :)

checks.d/snmp.py
```diff
@@ -272,27 +290,36 @@ def check(self, instance):
...
 
             if raw_oids:
                 self.log.debug("Querying device %s for %s oids", ip_address, len(raw_oids))
                 raw_results = self.check_table(instance, raw_oids, False, timeout, retries)
-                self.report_raw_metrics(instance, raw_results)
+                self.report_raw_metrics(metrics, raw_results, tags)
         except Exception as e:
             if "service_check_error" not in instance:
                 instance["service_check_error"] = "Fail to collect metrics: {0}".format(e)
             raise
```
yannmh: This will never raise because of the `finally` returning below.

tests/checks/integration/test_snmp.py
```diff
@@ -142,6 +190,9 @@ def test_type_support(self):
                 self.SUPPORTED_METRIC_TYPES + self.UNSUPPORTED_METRICS)]
         }
         self.run_check_twice(config)
+        self.service_checks = self.wait_for_async('get_service_checks', 'service_checks', 1)
+        self.run_check(config)
+        self.service_checks = self.wait_for_async('get_service_checks', 'service_checks', 1)
```
yannmh: How about something like 
```python
self.run_check(config, repeat=3)
self.service_checks = self.wait_for_async('get_service_checks', 'service_checks', 3)[-1:]
```
???

yannmh:
Nice changes. Added a few comments here and there :cactus:

tests/checks/integration/test_snmp.py
truthbk: I don't need to :)

tests/checks/integration/test_snmp.py
truthbk: more copy+paste havoc

truthbk:
thanks a lot for your input @yannmh - if you could please take another pass sometime, that'd be great.

truthbk:
Tests pass locally but not on Travis - check out what's up!

yannmh:
Your changes look good to be merged.

Failing tests are due to some `flake8` missing dependencies on Travis's Python 2.6 environment. 
I am not sure what's going on here but it's certainly unrelated to your changes: let's fix this in a separate PR.

yannmh:
My bad, there are actually multiple Travis CI errors. While it looks unrelated to your changes and specific to Travis CI Python 2.6 , let's investigate more before merging any change.

truthbk:
After rebasing to master all `2.6` tests are failing with what seems a broken `virtualenv` on Travis. 

Also the `snmpd` test may sometimes flap because we collect some `rates` that are based on network stats (`udpDatagrams`, `ifInOctets`, etc) so if the stats don't change the rate is 0 and they're not flushed.

truthbk:
closing in favor of #2214 to fix TravisCI issues.

yannmh:
:shipit:

PR #2154, opened 2015-12-18
Update the mirror urls to use Apache archives for ZK tests

irabinovitch:
Update the mirror urls to use Apache.org archives, the previously used cogent mirror only includes the latest release. As such each time apache releases a new version the url we had here previously would begin returning 404 and fail tests.

Also start testing against 3.4.7, rather than 3.4.6.


remh:
:+1:

PR #2155, opened 2015-12-18
[packaging] Fix config reload on osx and source install

olivielpeau:
We don't package our own `kill` bin with the osx and source installs, so use the command provided by the OS for these install methods.

The cross-platform compatibility of the HUP signal seems to be good, so we can expect this to work well on most platforms.

Tested on OSX and FreeBSD.


hkaj:
LGTM

PR #2156, opened 2015-12-18
Rebased pr #1770: gather stats from zookeeper mntr command

talwai:
This is a rebased and slightly modified version of @jpittis #1770 
The addition puts the `zookeeper.instances.*` gauges behind an instance-level configuration flag (defaults to `True`). Users can flag this off if they prefer to rely solely on the `expected_mode` service check to monitor instance status


talwai:
cc @jpittis

PR #2159, opened 2015-12-18
[docker_daemon] add docker storage driver metrics

patrickbcullen:
Added the docker storage driver metrics for the devicemapper storage driver (used by CentOS/RHEL) using the docker info API.

The unit test shows the metrics names which are patterned after the system.disk.* metrics.

### DIFF
--- a/checks.d/docker_daemon.py
+++ b/checks.d/docker_daemon.py
@@ -5,7 +5,7 @@ import requests
 import time
 import socket
 import urllib2
-from collections import defaultdict, Counter, deque
+from collections import defaultdict, deque
 
 # project
 from checks import AgentCheck
@@ -264,8 +264,8 @@ class DockerDaemon(AgentCheck):
         must_query_size = self.collect_container_size and self._latest_size_query == 0
         self._latest_size_query = (self._latest_size_query + 1) % SIZE_REFRESH_RATE
 
-        running_containers_count = Counter()
-        all_containers_count = Counter()
+        running_containers_count = defaultdict(int)
+        all_containers_count = defaultdict(int)
 
         try:
             containers = self.client.containers(all=True, size=must_query_size)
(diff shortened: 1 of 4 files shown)


remh:
Thanks a lot @patrickbcullen 
We'll review this for 5.8.0
One concern i have is that in my experience i have noticed that in some cases the info command can take while to return. Have you experienced that kind of issue ?

patrickbcullen:
The info command takes 300-700 milliseconds on my system. Is that too long? Also how do I rerun the checks? It looks like my build failed on an unrelated test.

irabinovitch:
@patrickbcullen  I've re-triggered the tests.

patrickbcullen:
I found the cause of the test failure. It appears that collections.Counter is only supported in Python 2.7 and the Travis tests run in Python 2.6 (see http://stackoverflow.com/questions/10099013/importerror-cannot-import-name-counter).

I did not add the import code to dd-agent/checks.d/docker_daemon.py, but I was the first person to add a mocks test for it which is why the error shows up with this pull request.

Suggestions on how to handle this? Apparently docker_deamon only works on Python 2.7 and later now anyways so maybe the Travis tests should be upgraded to Python 2.7.

remh:
Thanks a lot @patrickbcullen you are right.
It was not an issue as no one is running the Docker Daemon check on python 2.6.
I think the easiest way to fix that would be to switch from `Counter` to a `defaultdict(int)` on these two lines.
Given how we use the Counter that would be the easiest way.
Can you take a pass at it while you're at it ?
Thanks a lot!

patrickbcullen:
It looks like you cannot write mock tests for docker_daemon.py since it was never designed to be tested in that way. I am not sure how to proceed with getting this pull request approved since I don't want to rewrite docker_daemon.py. I will just keep this change in my fork for now.

hkaj:
Closing as we already collect those. I think there was concurrent work going on when you opened this PR. Thank you anyway for the contribution!

PR #2165, opened 2015-12-23
Add support for forced SNMP data types to help w/ buggy devices

truthbk:
This is just a minor touch-up to #1331 by @chrissnell - adding tests and raising an exception in the event of a bad `forced_type`.

PR summary by @chrissnell:

# Add support for forced SNMP data types to help w/ buggy devices

Some SNMP-enabled devices return incorrect data types when certain OIDs are queried. The F5 1600LTM load balancer is an example of such a device. This PR allows for the forcing of data types (to 'gauge' or 'counter') in the event of an incorrect response from the remote agent.

It adds a parameter to the metric definition on the YAML, 'forced_type'.

NOTE: The changes in this PR were tested in a back-ported version of this improved snmp check that Datadog provided me. The check patched here is from a newer version of the agent that I don't have any easy way to test right now. The changes are the same between the versions and it should work but I have no way to test.

### DIFF
--- a/checks.d/snmp.py
+++ b/checks.d/snmp.py
@@ -328,7 +328,9 @@ class SnmpCheck(NetworkCheck):
 
         Submit the results to the aggregator.
         '''
+
         for metric in metrics:
+            forced_type = metric.get('forced_type')
             if 'OID' in metric:
                 queried_oid = metric['OID']
                 if queried_oid in results:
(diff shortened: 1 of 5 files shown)


olivielpeau:
Looks good! :+1: 

Could you squash your commits into one or two please?

PR #2070, opened 2015-11-12
Linux tcp stats

maatthc:
Added metrics from Linux stats:
/proc/net/sockstat
/proc/sys/net/ipv4/tcp_mem
/proc/sys/net/ipv4/tcp_max_orphans
/proc/sys/net/ipv4/netfilter/ip_conntrack_count
/proc/sys/net/ipv4/ip_conntrack_max Also from the command : /usr/bin/getconf PAGESIZE Those are critical metrics for high performance applications.

PR #2170, opened 2015-12-28
Rebased: Linux tcp stats

JohnLZeller:
Rebase of https://github.com/DataDog/dd-agent/pull/2070 fixing merge conflicts


checks.d/network.py
```diff
@@ -491,3 +496,117 @@ def _parse_solaris_netstat(self, netstat_output):
...
+            try:
+                lines = proc.readline()
+            finally:
+                proc.close()
+            return int(lines)
+        except IOError:
+            return 0
+
+    @staticmethod
+    def _linux_sockstat_get_max_orphans(self):
```
JohnLZeller: If you're using `self`, then `@staticmethod` won't work. Either remove the decorator, or use `@classmethod`

checks.d/network.py
```diff
@@ -491,3 +496,117 @@ def _parse_solaris_netstat(self, netstat_output):
...
+            return int(lines)
+        except IOError:
+            self.log.debug("Unable to read %s." % proc_name)
+
+    @staticmethod
+    def _linux_sockstat_get_size_of_page():
+        return long(check_output(["/usr/bin/getconf", "PAGESIZE"]))
+
+    @staticmethod
+    def _linux_sockstat_get_max_memory(self):
```
JohnLZeller: Ditto here. If you're using `self`, then `@staticmethod` won't work. Either remove the decorator, or use `@classmethod`

checks.d/network.py
```diff
@@ -491,3 +496,117 @@ def _parse_solaris_netstat(self, netstat_output):
...
+                proc.close()
+
+        except IOError:
+            self.log.debug("Unable to read %s." % tcpMemFilePath)
+        tcpmemline = tcpmem[0].split()
+        maxMemoryInBytes = sizeOfPageInBytes * tcpmemline[2]
+        return long(maxMemoryInBytes)
+
+    @staticmethod
+    def _linux_sockstat(self):
```
JohnLZeller: Ditto here. If you're using `self`, then `@staticmethod` won't work. Either remove the decorator, or use `@classmethod`

checks.d/network.py
JohnLZeller: If you're using `self`, then `@staticmethod` won't work. Either remove the decorator, or use `@classmethod`

JohnLZeller:
In several spots, you're using `self` in your functions., then `@staticmethod` won't work. Either remove the decorator, or use `@classmethod`

PR #2171, opened 2015-12-29
[freebsd][logging] Use correct file for syslog

### LINKED ISSUE #1956: Check for FreeBSD for syslog
https://github.com/DataDog/dd-agent/blob/master/config.py#L1027

* I check you want to use the Platform.is_darwin check here?
* For FreeBSD the file is /var/run/log

olivielpeau:
+ use the Platform helpers to determine the OS family

Fixes #1956


yannmh:
:+1:

PR #2174, opened 2016-01-04
[vsphere] Added option to disable strict SSL checks of vCenter server

austinlparker:
We ran into a problem using the vSphere check - if our vCenter server had a self-signed SSL certificate, dd-agent 5.6+ couldn't connect to it as there was no certification chain. This workaround allows for someone to disable the strict SSL check when connecting to a vCenter server if they are in a similar situation.

There's a better way to do this, but it would require updating pyvmomi to 6.0.0 so that SmartConnect will accept a ssl context parameter. The commit below was tested on Python 2.7.11 but should work on earlier versions as well.

### DIFF
--- a/checks.d/vsphere.py
+++ b/checks.d/vsphere.py
@@ -6,6 +6,7 @@ from Queue import Empty, Queue
 import re
 import time
 import traceback
+import ssl
 
 # 3p
 from pyVim import connect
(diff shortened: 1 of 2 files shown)


talwai:
Hi @austinlparker thanks for the contribution! I think the preferred way to do this now is to actually use the options merged in https://github.com/DataDog/dd-agent/pull/2180 to provide a custom SSL verification procedure rather than disabling strict SSL altogether. This in fact uses the better approach you mentioned (we have upgraded pyvmomi to `6.0.0`)

PR #2176, opened 2016-01-04
[dogstatsd] Recover gracefully from serialization errors

olivielpeau:
- try to skip errors so that we can serialize what we can
- add a metric on serialization status

### DIFF
--- a/dogstatsd.py
+++ b/dogstatsd.py
@@ -34,6 +34,7 @@ import simplejson as json
 # project
 from aggregator import get_formatter, MetricsBucketAggregator
 from checks.check_status import DogstatsdStatus
+from checks.metric_types import MetricTypes
 from config import get_config, get_version
 from daemon import AgentSupervisor, Daemon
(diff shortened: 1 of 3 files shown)


dogstatsd.py
```diff
@@ -61,9 +62,48 @@
...
+                metric[key] = tuple(value_list)
+        metrics[i] = metric
+    return metrics
+
+
+def serialize_metrics(metrics, hostname):
+    try:
+        metrics.append(add_serialization_status_metric("success", hostname))
+        serialized = json.dumps({"series": metrics})
+    except (UnicodeDecodeError, Exception) as e:
```
remh: stupid me :( will fix

dogstatsd.py
```diff
@@ -61,9 +62,48 @@
 EVENT_CHUNK_SIZE = 50
 COMPRESS_THRESHOLD = 1024
 
+def add_serialization_status_metric(status, hostname):
+    interval = 10.0
```
remh: Let's add some comments here

PR #2179, opened 2016-01-05
[docker_daemon] Ignore case when container are excluded

### LINKED ISSUE #1998: docker_daemon reports warnings when containers are excluded
Sample warnings are below. This occurs because _crawl_container_pids() tries to access container_dict[container_id] but since the container is excluded, _get_and_count_containers() does not create an entry for it in the dict. Presumably the fix is to skip excluded containers in _crawl_container_pids()

Datadog's docker_daemon integration is reporting:
Instance #0[WARNING]:Cannot parse /proc/4619/cgroup content: '14185c665d9131004725d9f7b8dc92fc797f5b58e8b7250608941751e2092f25'

remh:
Fix #1998

### DIFF
--- a/checks.d/docker_daemon.py
+++ b/checks.d/docker_daemon.py
@@ -718,6 +718,9 @@ class DockerDaemon(AgentCheck):
                 match = CONTAINER_ID_RE.search(cpuacct)
                 if match:
                     container_id = match.group(0)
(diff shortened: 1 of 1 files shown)


remh:
docker_daemon tests are passing locally

olivielpeau:
:+1:

PR #2180, opened 2016-01-05
[vsphere] Add SSL config options for certs

JohnLZeller:
As it stands, the vsphere check does not allow customization of SSL verification procedures in any way. This PR seeks to allow the ability to disable SSL verification, and to specify your own CA certificates.

### DIFF
--- a/checks.d/vsphere.py
+++ b/checks.d/vsphere.py
@@ -4,6 +4,7 @@ from datetime import datetime, timedelta
 from hashlib import md5
 from Queue import Empty, Queue
 import re
+import ssl
 import time
 import traceback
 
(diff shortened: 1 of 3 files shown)


checks.d/vsphere.py
```diff
@@ -429,12 +430,23 @@ def _get_server_instance(self, instance):
...
         if i_key not in self.server_instances:
             try:
                 server_instance = connect.SmartConnect(
-                    host=instance.get('host'),
-                    user=instance.get('username'),
-                    pwd=instance.get('password')
+                    host = instance.get('host'),
+                    user = instance.get('username'),
+                    pwd = instance.get('password'),
+                    sslContext = context if not ssl_verify or ssl_capath else None
```
truthbk: If the customer accidentally sets `ssl_verify` to False and also sets an `ssl_capath` in the config, the `not ssl_verify` context takes precedence (which I think makes sense) but we should probably log it somewhere to help identify that case come support-time.

truthbk:
Looks good :+1: I was just thinking we should maybe add a log line if there's a config "conflict"

checks.d/vsphere.py
```diff
@@ -429,12 +430,31 @@ def _get_server_instance(self, instance):
...
+        if not ssl_verify:
+            context = ssl.SSLContext(ssl.PROTOCOL_SSLv23)
+            context.verify_mode = ssl.CERT_NONE
+        elif ssl_capath:
+            context = ssl.SSLContext(ssl.PROTOCOL_SSLv23)
+            context.verify_mode = ssl.CERT_REQUIRED
+            context.load_verify_locations(capath=ssl_capath)
+
+        # If both configs are used, log a message explaining the default
+        if not ssl_verify and ssl_capath:
```
JohnLZeller: @truthbk how's this?

truthbk: looks good!

truthbk:
Ship it! :shipit:

PR #2183, opened 2016-01-05
Add timeout for requests to get go expvar metrics.

gphat:
# What's this PR do?
Adds a timeout for requests made to fetch go expvar metrics.

# Motivation
Locally we had a misbehaving app that never answered, causing the entire collector to be blocked and not getting metrics!

# Notes
Should this be configurable?

r? @antifuchs @remh

### DIFF
--- a/checks.d/go_expvar.py
+++ b/checks.d/go_expvar.py
@@ -59,7 +59,7 @@ class GoExpvar(AgentCheck):
         self._last_gc_count = defaultdict(int)
 
     def _get_data(self, url):
-        r = requests.get(url)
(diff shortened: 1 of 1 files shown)


remh:
Thanks @gphat 
The expvar endpoint should return pretty quickly usually. If it takes more than 10 secs, there is an issue. So it seems like a valid approach to me. Thanks!

gphat:
Excellent, thanks!

remh:
Merging, the failing tests are related to this issue https://github.com/DataDog/dd-agent/pull/2187

PR #2185, opened 2016-01-06
[core][wmi] gracefully timeout WMI queries

yannmh:
##### [utils] decorator to timeout a function
A decorator to timeout a function. Decorated method calls are executed in a separate new thread with a specified timeout.
Also check if a thread for the same function already exists before creating a new one.

##### [core][wmi] gracefully timeout WMI queries 
Timeout WMI queries after 10seconds with a warning message and no data.
Recover at next iteration retrieving the existing query.

##### [core][wmi] one active query / connection
Less thread safety error-prone `WMISampler`: have one active query per connection at a time.
* `_get_connection` → `get_connection`, a context manager that
  * returns an existing and available WMI connection, or create one if
    needed
  * release it, i.e. mark it as available when done

### DIFF
--- a/checks.d/wmi_check.py
+++ b/checks.d/wmi_check.py
@@ -151,12 +151,11 @@ class WMICheck(AgentCheck):
             self._format_tag_query(sampler, wmi_obj, tag_query)
 
         # Create a specific sampler
-        connection = sampler.get_connection()
         tag_query_sampler = WMISampler(
             self.log,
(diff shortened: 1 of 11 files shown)


olivielpeau:
Looks good to me, well done!

The failures on Travis look related to the python libs issues that have been solved in master, could you rebase the PR to make it work?

Once the tests pass I'm all for :ship: ing this!

PR #2189, opened 2016-01-07
[mesos_slave] Allow to configure mesos master port

### LINKED ISSUE #2188: Mesos master port in mesos_slave.py is hardcoded
https://github.com/DataDog/dd-agent/blob/master/checks.d/mesos_slave.py#L132

Effecting mesos_slave configuration doesn't work if the master port is other than 5050.
Probably should be added in configuration.

remh:
Fix #2188


olivielpeau:
:+1:

PR #2192, opened 2016-01-08
Mesos: Prevent requests from using chardet

GregBowyer:
Requests internally (and irritatingly imho) bundles `chardet`. It uses this library to figure out the encoding for a given response if there are no suitable HTTP headers detailing the character set of the response.

This library is pathologically slow (its loop heavy) on any python implementation _not_ running on top of a VM. There are better implementations of `chardet` for vanilla pythons but the requests library authors wish to avoid bundling C extensions. https://github.com/kennethreitz/requests/issues/2359

Taking a leaf from opentable, we now force the encoding of the mesos response to always be UTF8 which avoids requests pulling large mesos payloads through `chardet` and prevents the datadog agent taking large amounts of CPU time to run the check.

### DIFF
--- a/checks.d/mesos.py
+++ b/checks.d/mesos.py
@@ -106,6 +106,9 @@ class Mesos(AgentCheck):
                 self.warning(msg)
                 return None
 
+        if r.encoding is None:
+            r.encoding = 'UTF8'
+
         return r.json()
 
     def timeout_event(self, url, timeout, aggregation_key):
--- a/checks.d/mesos_master.py
+++ b/checks.d/mesos_master.py
@@ -148,6 +148,9 @@ class MesosMaster(AgentCheck):
                                    message=msg)
                 raise CheckException("Cannot connect to mesos, please check your configuration.")
 
+        if r.encoding is None:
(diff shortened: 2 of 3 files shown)


talwai:
Hi @GregBowyer , thanks for contributing! Looks like a worthwhile change, would you in fact be able to provide some benchmarks on performance improvement as a result of your changes to the check? [Agent Developer Mode](https://github.com/DataDog/dd-agent/wiki/Agent-Developer-Mode) might be helpful here.

GregBowyer:
> Hi @GregBowyer , thanks for contributing! Looks like a worthwhile change, would you in fact be able to provide some benchmarks on performance improvement as a result of your changes to the check? Agent Developer Mode might be helpful here.

I could but I would prefer not to, this gave us serious performance issues in our datacenter. Essentially datadog no-data alarms for all our monitors would start firing with the mesos check enabled, which comes through pagerduty which wakes our support staff up :P.

talwai:
No worries, and sincere apologies for any rude wake up calls this caused.
We'll profile the improvements on our end, and consider this for an
upcoming release

PR #2194, opened 2016-01-12
ECS Agent Check

ejholmes:
This adds an agent check that reports whether an ecs agent is connected or not. We've had cases where the ecs agent became "disconnected" and we had old docker images running without being aware that the agent was in a disconnected state.

### DIFF
--- a/tests/checks/mock/test_ecs.py
+++ b/tests/checks/mock/test_ecs.py
@@ -1,0 +1,65 @@
+# 3rd party
+import mock
+import json
+
+from tests.checks.common import AgentCheckTest, load_check
+
+MOCK_CONFIG = {
+    'init_config': {},
+    'instances' : [{
+        'region': 'us-east-1'
+    }]
+}
+
+def requests_get_mock(*args, **kwargs):
+    class MockResponse:
+        def __init__(self, data, status_code):
+            self.data = data
+            self.status_code = status_code
+
+        def json(self):
+            return self.data
+
+        def raise_for_status(self):
+            return True
+
+    return MockResponse({
+        'Cluster': 'default',
+        'ContainerInstanceArn': 'arn:aws:ecs:us-east-1:012345678910:container-instance/c9c9a6f2-8766-464b-8805-9c57b9368fb0',
+    }, 200)
+
+class MockECSClient:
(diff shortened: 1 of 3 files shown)


talwai:
Hey @ejholmes , thanks for contributing! There's a bit of noise in the CI logs, but one of the errors looks valid - perhaps related to using `assertRaises` as a context manager, which py2.6 does not allow. Would you mind taking a look at that?

remh:
Maybe it should actually be done in our crawler instead ? What do you think @talwai ?

ejholmes:
@talwai added a commit for python 2.6 support

@remh assuming the crawler is the AWS/ECS integration we see at https://app.datadoghq.com/account/settings? That would be pretty nice as well :)

ejholmes:
If something like this was added to the ECS integration, might I suggest scraping the event stream from all ECS services in a cluster. This is another area where we have little visibility into ECS scheduling problems and I'd love to not have to build it myself!

tests/checks/mock/test_ecs.py
```diff
@@ -0,0 +1,69 @@
...
+        container_instances.append({ 'agentConnected': agent_connected })
+    mock_client = mock.Mock()
+    mock_client.describe_container_instances.return_value = {
+            'DescribeContainerInstancesResponse': {
+                'DescribeContainerInstancesResult': {
+                    'containerInstances': container_instances }}}
+    return mock_client
+
+class TestCheckECS(AgentCheckTest):
+    CHECK_NAME = 'ecs'
```
ejholmes: Side note. Would you prefer this being called `ecs` or `ecs_agent`. I'm thinking that `ecs_agent` may be a little more descriptive.

tmichelet:
@ejholmes thanks for this contribution. I believe this could be done by our AWS/ECS integration.
Thanks also for suggesting to scrap the events. I'm adding that story to our backlog, we will keep you updated on it's status

remh:
@ejholmes we are going to close that PR. It's better to add that in our crawler instead. Feel free to keep using that check as a custom check until we've implemented it!

ejholmes:
Awesome. Thanks guys!

PR #2196, opened 2016-01-13
[fluentd] Allow the option to add custom tags

jgmchan:
This allows us to include an option in `fluentd.yaml` to add custom tags which appears in service checks and metrics:
```
- monitor_agent_url: http://example.com:24220/api/plugins.json
  tags:
    - tag1
    - tag2
```

### DIFF
--- a/checks.d/fluentd.py
+++ b/checks.d/fluentd.py
@@ -30,6 +30,8 @@ class Fluentd(AgentCheck):
             url = instance.get('monitor_agent_url')
             plugin_ids = instance.get('plugin_ids', [])
 
+            custom_tags = instance.get('tags', [])
+
             # Fallback  with `tag_by: plugin_id`
             tag_by = instance.get('tag_by')
             tag_by = tag_by if tag_by in self._AVAILABLE_TAGS else 'plugin_id'
@@ -37,7 +39,7 @@ class Fluentd(AgentCheck):
             parsed_url = urlparse.urlparse(url)
             monitor_agent_host = parsed_url.hostname
             monitor_agent_port = parsed_url.port or 24220
-            service_check_tags = ['fluentd_host:%s' % monitor_agent_host, 'fluentd_port:%s'
+            service_check_tags = custom_tags + ['fluentd_host:%s' % monitor_agent_host, 'fluentd_port:%s'
                                   % monitor_agent_port]
 
             r = requests.get(url, headers=headers(self.agentConfig))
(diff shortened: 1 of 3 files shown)


truthbk:
@jgmchan thank you very much for this. We'll review this as soon as possible. 

I have noticed that currently the `fluentd` checks are failing for your PR - if you could fix them that'd be fantastic.

We really appreciate your contibution! Thanks again.

checks.d/fluentd.py
```diff
@@ -45,13 +47,13 @@ def check(self, instance):
             status = r.json()
 
             for p in status['plugins']:
-                tag = "%s:%s" % (tag_by, p.get(tag_by))
+                custom_tags.append("%s:%s" % (tag_by, p.get(tag_by)))
```
truthbk: Here, you'd append to custom_tags for each `p in status['plugins']`, probably what you want is to just send the original custom_tags + ["%s:%s" % (tag_by, p.get(tag_by))] for that `p in status['plugins']`. I believe we should leave `custom_tags` untouched.

tests/checks/integration/test_fluentd.py
```diff
@@ -79,3 +79,23 @@ def test_fluentd_with_tag_by_plugin_id(self):
         self.assertServiceCheckOK(
             self.check.SERVICE_CHECK_NAME, tags=['fluentd_host:localhost', 'fluentd_port:24220'])
         self.coverage_report()
+
+    def test_fluentd_with_custom_tags(self):
```
truthbk: Thank you for writing the test! :clap:

tests/checks/integration/test_fluentd.py
```diff
@@ -79,3 +79,23 @@ def test_fluentd_with_tag_by_plugin_id(self):
...
+                    "monitor_agent_url": "http://localhost:24220/api/plugins.json",
+                    "tag_by": "plugin_id",
+                    "tags": ["tag1", "tag2"]
+                }
+            ]
+        }
+        self.run_check(config)
+        for m in self.CHECK_GAUGES:
+            self.assertMetric('{0}.{1}'.format(self.CHECK_NAME, m), tags=['tag1', 'tag2', 'plugin_id:plg1'])
+            self.assertMetric('{0}.{1}'.format(self.CHECK_NAME, m), tags=['tag1', 'tag2', 'plugin_id:plg2'])
```
truthbk: As per the comment I made above, I think these might fail due to how you're appending the plugin tags to `custom_tags`.

jgmchan:
Ok, no problem I'll take a look at this very soon.

truthbk:
@jgmchan thanks a lot for looking into it! :)

truthbk:
@jgmchan had some time so I fixed this myself. No need to do anything else from your side! Thanks again for your contribution.

Closing in favor of: https://github.com/DataDog/dd-agent/pull/2224

jgmchan:
Thanks @truthbk, sorry I've been swamped with work and haven't had any time to look at this properly.

PR #2197, opened 2016-01-13
[tests][mysql] use flavored mysql test

GHfan:
Add flavored mysql test to allow test different mysql version without
getting local environment involved.

Details:
* install and build from mysql source code
* init data directory and insecure root access
* run at port 6033
* change mysql integration test config port to 6033
* clean files no more needed
* local tests past over 5.5.47, 5.6.28, 5.7.10
* add three travis mysql versions

Flaws:
* not support replication, replication relative metrics are commented


truthbk:
@GHfan thank you very much for your contribution! Although I myself worked on a similar PR a while back, we have decided to put it on hold, as we might actually make an attempt to run flavored tests on CircleCI as opposed to TravisCI, due to the facilities they provide to use containers. Also, downloading the source and compiling on Travis implies a bunch of compute time on the nodes we'd rather use for testing... :disappointed: 

We will definitely consider this otherwise, I'll keep you posted. Thanks again!

GHfan:
@truthbk Well, this is my first pr though, i'm very happy for you choosing containers, container is perfect for this integration test.

PR #2200, opened 2016-01-13
[docker_daemon] catch exception when getting k8s labels fails

hkaj:
When a user misconfigures their kubernetes check, it make the docker check fails too. This PR modifies the behavior so that we just log a warning and omit the kubernetes tags.


yannmh:
:+1:

remh:
Let's add the exception message in the log at least.

PR #2201, opened 2016-01-13
[util] use proxy settings for EC2 tag collection

yannmh:
AWS EC2 custom tags collection should be using proxy settings.

### DIFF
--- a/util.py
+++ b/util.py
@@ -32,6 +32,7 @@ except ImportError:
 from utils.dockerutil import get_hostname as get_docker_hostname, is_dockerized
 from utils.pidfile import PidFile  # noqa, see ^^^
 from utils.platform import Platform
+from utils.proxy import get_proxy
 from utils.subprocess_output import get_subprocess_output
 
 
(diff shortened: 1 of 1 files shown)


util.py
```diff
@@ -360,7 +360,16 @@ def get_tags(agentConfig):
...
 
             import boto.ec2
-            connection = boto.ec2.connect_to_region(region, aws_access_key_id=iam_params['AccessKeyId'], aws_secret_access_key=iam_params['SecretAccessKey'], security_token=iam_params['Token'])
+            proxy_settings = get_proxy(agentConfig)
+            connection = boto.ec2.connect_to_region(
+                region,
+                aws_access_key_id=iam_params['AccessKeyId'],
+                aws_secret_access_key=iam_params['SecretAccessKey'],
+                security_token=iam_params['Token'],
+                proxy=proxy_settings['host'], proxy_port=proxy_settings['port'],
```
truthbk: If there's no proxy configured won't `get_proxy()` return None, and then you'd have a `TypeError` here?

yannmh: Oops :no_mouth: 

Thanks for catching that, I addressed your comment.

truthbk:
:+1:

PR #2163, opened 2015-12-22
[process] Relocatable procfs for process check

remh:
When running the datadog agent as a container, we need to point
psutils at a different procfs to gather metrics on processes
outside the current namespace.

-- 
Based on #2124

PR #2202, opened 2016-01-14
Add "more unix" plugin with some linux metrics.

gphat:
I somehow broke #2030 . Oops.

# What's this PR do?

Adds some new system metrics:
* context switch count (for rates!)
* processes created count (for rates!)
* interrupts
* counts of processes by state, with tags
* counts of processes by priority, with tags

# Motivation

These metrics are used in our existing setup and, from what I can tell, are not available today in Datadog agent. We can get them from specific processes via `process.yaml`, but not for the whole system!

# Notes

I originally tried to add this stuff to `unix.py` but I'm unable to determine how those metrics end up as things like `sys.mem.free` and the like. So I settled on making a check that added the metrics I as after. I expect to add more.

Most of these metrics feel very "core" to Linux and therefore seem suitable to be included in something like `unix.py`. But for now I'll opt to take a conservative approach and assume this will just be a plugin. If you are not interested in having it in the agent, we'll just use it locally.

Lemme know and thanks for looking!

### DIFF
--- a/checks.d/linux_proc_extras.py
+++ b/checks.d/linux_proc_extras.py
@@ -1,0 +1,70 @@
+# project
+from checks import AgentCheck
+from utils.subprocess_output import get_subprocess_output
+from collections import defaultdict
+
+PROCESS_STATES = {
+    'D': 'uninterruptible',
+    'R': 'runnable',
+    'S': 'sleeping',
+    'T': 'stopped',
+    'W': 'paging',
+    'X': 'dead',
+    'Z': 'zombie',
+}
+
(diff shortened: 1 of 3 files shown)


remh:
The failure in the tests is unrelated and was caused by a change in the version of rubocop. Merging. 
Thanks @gphat !

ashb:
This uses a hard-coded `/proc` which means it won't work when running in a docker container (which is what we do) -- This was fixed for the normal process check in #2163 one week before this was merged.

I guess we need to make it use a similar config mechanism here, or make this use psutil too?

PR #2203, opened 2016-01-14
[http check] Improve message when using default http code

### LINKED ISSUE #2190: checks.http_check(http_check.py:239) | Incorrect HTTP return code. Expected (1|2|3)\d\d, got 403
It seems that list INFO message is incomplete which makes it totally useless. We have more than 20 tests inside the http_check and if this like appears without telling which instance from the http_check caused this, it makes the entire message useless.
```
checks.http_check(http_check.py:239) | Incorrect HTTP return code. Expected (1|2|3)\d\d, got 403
```

remh:
Fix #2190

### DIFF
--- a/checks.d/http_check.py
+++ b/checks.d/http_check.py
@@ -28,6 +28,7 @@ from config import _is_affirmative
 from util import headers as agent_headers
 from utils.proxy import get_proxy
 
+DEFAULT_EXPECTED_CODE = "(1|2|3)\d\d"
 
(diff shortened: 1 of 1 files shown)


olivielpeau:
Looks good, but wasn't the original issue about the absence in the log message of the URL that returns the status code rather than the absence of the `http_response_status_code`?

remh:
What do you mean @olivielpeau , it's there! :p :trollface:

PR #2208, opened 2016-01-15
[go-metro] Adding supervisor entry and sample yaml config.

truthbk:
Alternatively, we could add a python launcher that will only launch the binary iff `conf.d/dd-tcp-rtt.yaml` exists.

### DIFF
--- a/conf.d/go-metro.yaml.example
+++ b/conf.d/go-metro.yaml.example
@@ -1,0 +1,40 @@
+# NOTE: for go metro to run unprivileged, you will have to set CAP_NET_RAW capabilities on the binary:
+#        (debian based) $ sudo apt-get update libcap
+#        (redhat based) $ sudo yum install libcap
+#        (redhat based - alternatively) $ sudo yum install compat-libcap1
+#        $ sudo setcap cap_net_raw+ep /opt/datadog-agent/bin/go-metro
+#
+#        Also, please note that go-metro logs to its own file - found in /var/log/datadog/go-metro.log.
+#        Additionally, go-mtro runs standalone so it will *NOT* currently appear on the agent's info page.
+
+init_config:
+    snaplen: 512            # should be >=104 (to accomodate for the largest possible TCP header)
+    idle_ttl: 300           # time after which an idle flow (no traffic received) is flushed.
+    exp_ttl: 60             # time after which a finished flow is flushed.
+    statsd_ip: 127.0.0.1
+    statsd_port: 8125
+    log_to_file: true
+    log_level: info         # available levels: trace, debug, info, warning, error, critical
+
+instances:
+- interface: eth0           # metrics will be also tagged by interface.
+  tags:
+    - foo:bar
+  ips:                        # Whitelist by IP - will perform name lookup and tag with hostname if available.
+    - 192.168.0.1             # <--- SAMPLE IP SET YOUR OWN LIST.
+  hosts:                      # Whitelist by hostname - will perform ip lookup and filter by hosts ip's
+    - somehost.somedomain.to  # <---SAMPLE HOST, SET YOUR OWN LIST.
+
+                              # NOTE: Whitelisting with ips and hosts is *highly* recommended to avoid having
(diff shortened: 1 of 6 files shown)


conf.d/go-metro.yaml.example
```diff
@@ -0,0 +1,31 @@
+# NOTE: for go metro to run unprivileged, you will have to set CAP_NET_RAW capabilities on the binary:
+#        (debian based) $ sudo apt-get update libcap
+#        (redhat based) $ sudo yum install compat-libcap1
+#        (redhat based - alternatively) $ sudo yum install compat-libcap1
```
olivielpeau: The alternative command is the same as the regular one ;)

olivielpeau: Unrelated: let's add a comment to mention that `go-metro` logs to a separate file

utils/flare.py
```diff
@@ -215,6 +216,7 @@ def _add_logs_tar(self):
         self._add_log_file_tar(self._forwarder_log)
         self._add_log_file_tar(self._dogstatsd_log)
         self._add_log_file_tar(self._jmxfetch_log)
+        self._add_log_file_tar(self._gometro_log)
```
olivielpeau: by default we're going to log a `warning` if the log file is not present. That could confuse users if they're running anything else than a packaged Linux agent...

conf.d/go-metro.yaml.example
truthbk: Are you sure it's the exact same `rpm`? I was expecting yum to know they were aliases, but it considered them distinct.

conf.d/go-metro.yaml.example
```diff
@@ -0,0 +1,39 @@
+# NOTE: for go metro to run unprivileged, you will have to set CAP_NET_RAW capabilities on the binary:
+#        (debian based) $ sudo apt-get update libcap
+#        (redhat based) $ sudo yum install libcap
+#        (redhat based - alternatively) $ sudo yum install compat-libcap1
+#        $ sudo setcap cap_net_raw+ep /opt/datadog-agent/bin/go-metro
+#
+#        Also, please note that go-metro logs to its own file - found in /var/log/datadog/go-metro.log.
```
olivielpeau: Last thing we could mention here: `go-metro` does _not_ appear as a check on the agent's `info` page (IMO this is counter-intuitive for users used to regular checks/JMXFetch checks)

utils/flare.py
```diff
@@ -215,13 +216,14 @@ def _add_logs_tar(self):
...
         self._add_log_file_tar(self._dogstatsd_log)
         self._add_log_file_tar(self._jmxfetch_log)
+        self._add_log_file_tar(self._gometro_log, warn=False)
         self._add_log_file_tar(
             "{0}/*supervisord.log".format(os.path.dirname(self._collector_log))
         )
 
-    def _add_log_file_tar(self, file_path):
+    def _add_log_file_tar(self, file_path, warn=True):
         for f in glob.glob('{0}*'.format(file_path)):
```
olivielpeau: actually the `glob.glob` step here prevents us from adding a log file that doesn't exist, so you can ignore my comment on the warnings and revert these changes, thanks!

packaging/supervisor.conf
```diff
@@ -57,5 +57,14 @@ priority=999
 startsecs=3
 user=dd-agent
 
+[program:go-metro]
+command=/opt/datadog-agent/bin/go-metro -cfg="/etc/dd-agent/conf.d/go-metro.yaml"
+stdout_logfile=NONE
+stderr_logfile=NONE
+redirect_stderr=true
+priority=999
+startsecs=3
```
olivielpeau: Let's lower this to 2 secs, and lower the number of seconds in go-metro too

utils/flare.py
truthbk: we're not really going to log a warning ;)

conf.d/go-metro.yaml.example
truthbk: added below.

olivielpeau:
:+1:

Issue #76, opened 2012-06-12
Fix perfdata parsing


alq666:
Related to having '[', ']', '*' in the perfdata templates.

PR #2210, opened 2016-01-15
[windows] 🔒 permissions on `datadog.conf`

yannmh:
Datadog Agent configuration file `datadog.conf` contains sensitive data
such as the API key.
Set the file permission, to allow the following access:
* Users. None
* Administrator & System. Full control

Similar to #76.

**Note**: Permissions are set at first install only. It is unchanged on upgrade.


truthbk:
:+1:

yannmh:
No regression on first install or upgrade. Merging.

PR #2211, opened 2016-01-15
[Ceph] New integration

remh:


checks.d/ceph.py
```diff
@@ -0,0 +1,137 @@
+"""ceph check
+Collects metrics from ceph clusters
+"""
+
+# stdlib
+import os
+import json
```
remh: Let's use simplejson instead.

checks.d/ceph.py
```diff
@@ -0,0 +1,137 @@
...
+
+class Ceph(AgentCheck):
+    """ Collect metrics and events from ceph """
+
+    DEFAULT_CEPH_CMD = '/usr/bin/ceph'
+
+    def __init__(self, name, init_config, agentConfig, instances=None):
+        AgentCheck.__init__(self, name, init_config, agentConfig, instances)
+
+    def _run_command(self, cmd):
```
remh: This is not used

checks.d/ceph.py
```diff
@@ -0,0 +1,137 @@
...
+        if 'status' in raw:
+            s_status = raw['status']['health']['overall_status']
+            if s_status.find('_OK') != -1:
+                status = AgentCheck.OK
+            else:
+                status = AgentCheck.CRITICAL
+            self.service_check(NAMESPACE + '.overall_status', status)
+
+    def check(self, instance):
+        ceph_cmd = instance.get('ceph_cmd', self.DEFAULT_CEPH_CMD)
```
remh: This should be:
```python
ceph_cmd = instance.get('ceph_cmd') or self.DEFAULT_CEPH_CMD
```
In case someone puts:
```yaml
instance:
    - ceph_cmd:
```

In the config

checks.d/ceph.py
```diff
@@ -0,0 +1,137 @@
...
+from checks import AgentCheck
+from utils.subprocess_output import get_subprocess_output
+from config import _is_affirmative
+
+NAMESPACE = "ceph"
+
+class Ceph(AgentCheck):
+    """ Collect metrics and events from ceph """
+
+    DEFAULT_CEPH_CMD = '/usr/bin/ceph'
```
remh: This should be at the module level to be more consistent.

checks.d/ceph.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def __init__(self, name, init_config, agentConfig, instances=None):
```
remh: Don't need that.

checks.d/ceph.py
```diff
@@ -0,0 +1,137 @@
...
+    def _run_command(self, cmd):
+        fd = os.popen(cmd)
+        raw = fd.read()
+        fd.close()
+        return json.loads(raw)
+
+    def _collect_raw(self, ceph_cmd, instance):
+        use_sudo = _is_affirmative(instance.get('use_sudo', False))
+        if use_sudo:
+            test_sudo = os.system('setsid sudo -l < /dev/null')
```
remh: We should ensure that the user has enough permissions to run the ceph command, and if not write a clear error message on how to fix that.

checks.d/ceph.py
```diff
@@ -0,0 +1,137 @@
...
+        if 'mon_status' in raw:
+            fsid = raw['mon_status']['monmap']['fsid']
+            tags.append(NAMESPACE + '_fsid:%s' % fsid)
+            tags.append(NAMESPACE + '_mon_state:%s' % raw['mon_status']['state'])
+
+        name = instance.get('name', None)
+        if name:
+            tags.append(NAMESPACE + '_name:%s' % name)
+
+        self.tags = tags
```
remh: You shouldn't store instance tags at the AgentCheck instance level.

checks.d/ceph.py
```diff
@@ -0,0 +1,137 @@
...
+
+            name = cmd.replace(' ', '_')
+            raw[name] = res
+
+        return raw
+
+    def _extract_tags(self, raw, instance):
+        tags = instance.get('tags', [])
+        if 'mon_status' in raw:
+            fsid = raw['mon_status']['monmap']['fsid']
```
remh: You should check that these keys exist

checks.d/ceph.py
```diff
@@ -0,0 +1,137 @@
...
+            tags.append(NAMESPACE + '_fsid:%s' % fsid)
```
remh: Nitpick but you shouldn't use two different types of string formatting.

checks.d/ceph.py
```diff
@@ -0,0 +1,137 @@
...
+            tags.append(NAMESPACE + '_mon_state:%s' % raw['mon_status']['state'])
+
+        name = instance.get('name', None)
```
remh: Let's remove that, as custom tags cover this use case already.

checks.d/ceph.py
```diff
@@ -0,0 +1,137 @@
...
+                status = AgentCheck.OK
+            else:
+                status = AgentCheck.CRITICAL
+            self.service_check(NAMESPACE + '.overall_status', status)
+
+    def check(self, instance):
+        ceph_cmd = instance.get('ceph_cmd', self.DEFAULT_CEPH_CMD)
+        raw = self._collect_raw(ceph_cmd, instance)
+        self._extract_tags(raw, instance)
+        if instance.get('publish_metrics', True):
```
remh: we don't need these feature flags.

checks.d/ceph.py
```diff
@@ -0,0 +1,137 @@
...
+        if name:
+            tags.append(NAMESPACE + '_name:%s' % name)
+
+        self.tags = tags
+
+    def _extract_metrics(self, raw):
+        if 'osd_perf' in raw:
+            for osdperf in raw['osd_perf']['osd_perf_infos']:
+                tags = self.tags + ['ceph_osd:osd%s' % osdperf['id']]
+                for k,v in osdperf['perf_stats'].iteritems():
```
remh: We should whitelist the list of metrics we want to collect just to avoid bad surprises.

checks.d/ceph.py
```diff
@@ -0,0 +1,137 @@
...
+            num_mons = len(raw['mon_status']['monmap']['mons'])
+            self.gauge(NAMESPACE + '.num_mons', num_mons, self.tags)
+
+        if 'df_detail' in raw:
+            stats = raw['df_detail']['stats']
+            self.gauge(NAMESPACE + '.total_objects', stats['total_objects'], self.tags)
+            used = float(stats['total_used_bytes'])
+            avail = float(stats['total_avail_bytes'])
+            if avail > 0:
+                self.gauge(NAMESPACE + '.aggregate_pct_used', 100.0*used/avail, self.tags)
```
remh: should it be `total` instead of `aggregate`?

remh: Also should it be `100.0 * used / (used + avail)` ?

checks.d/ceph.py
```diff
@@ -0,0 +1,137 @@
...
+            self.gauge(NAMESPACE + '.num_pools', len(l_pools), self.tags)
+            for pdata in l_pools:
+                tags = list(self.tags + [NAMESPACE + '_pool_name:%s' % pdata['name']])
+                stats = pdata['stats']
+                used = float(stats['bytes_used'])
+                avail = float(stats['max_avail'])
+                if avail > 0:
+                    self.gauge(NAMESPACE + '.pct_used', 100.0*used/avail, tags)
+                self.gauge(NAMESPACE + '.num_objects', stats['objects'], tags)
+                self.rate(NAMESPACE + '.rd_bytes', stats['rd_bytes'], tags)
```
remh: Let's call them `read_bytes` and `write_bytes`

PR #2116, opened 2015-12-01
[mysql] enhancing the catalog of metrics reported.

truthbk:
# MySQL Check Enhancement

This PR combines multiple efforts by @ovaistariq (huge effort!) @garnermccloud @c960657 @zdannar @polynomial @leucos to greatly improve the catalog of MySQL metrics reported. Metrics have been added to provide additional insight into:

* Performance Schema
* Replication
* Galera Cluster
* Per schema performance and size metrics

Additional work has gone into improving the test and attempt to test the entire catalog. 

This overhaul will now allow us to provide the following new metrics (hang tight):

## New Status Metrics
```
* MySQL name: 'Com_replace' - Datadog metric name: 'mysql.performance.com_replace' - Metric type: RATE
* MySQL name: 'Com_load' - Datadog metric name: 'mysql.performance.com_load' - Metric type: RATE
* MySQL name: 'Aborted_clients' - Datadog metric name: 'mysql.net.aborted_clients' - Metric type: RATE
* MySQL name: 'Aborted_connects' - Datadog metric name: 'mysql.net.aborted_connects' - Metric type: RATE
* MySQL name: 'Bytes_sent' - Datadog metric name: 'mysql.performance.bytes_sent' - Metric type: RATE
* MySQL name: 'Bytes_received' - Datadog metric name: 'mysql.performance.bytes_received' - Metric type: RATE
* MySQL name: 'Qcache_hits' - Datadog metric name: 'mysql.performance.qcache_hits' - Metric type: RATE
* MySQL name: 'Qcache_inserts' - Datadog metric name: 'mysql.performance.qcache_inserts' - Metric type: RATE
* MySQL name: 'Qcache_lowmem_prunes' - Datadog metric name: 'mysql.performance.qcache_lowmem_prunes' - Metric type: RATE
```

### MyISAM Metrics
```
* MySQL name: 'Key_buffer_bytes_unflushed' - Datadog metric name: 'mysql.myisam.key_buffer_bytes_unflushed' - Metric type: RATE
* MySQL name: 'Key_buffer_bytes_used' - Datadog metric name: 'mysql.myisam.key_buffer_bytes_used' - Metric type: GAUGE
* MySQL name: 'Key_read_requests' - Datadog metric name: 'mysql.myisam.key_read_requests' - Metric type: RATE
* MySQL name: 'Key_reads' - Datadog metric name: 'mysql.myisam.key_reads' - Metric type: RATE
[...]

PR #2213, opened 2016-01-18
Be sure MyISAM is enabled

nfo:
The MyISAM storage engine is not present in some MySQL compatible software, like MemSQL. These variables are thus missing when calling `SHOW STATUS`:

- `Key_blocks_unused`
- `key_cache_block_size`
- `key_buffer_size`

### DIFF
--- a/checks.d/mysql.py
+++ b/checks.d/mysql.py
@@ -148,11 +148,13 @@ class MySql(AgentCheck):
         del cursor
 
         # Compute key cache utilization metric
-        key_blocks_unused = self._collect_scalar('Key_blocks_unused', status_results)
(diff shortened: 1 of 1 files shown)


truthbk:
@nfo thank you very much for raising this issue! We're in the middle of a very big overhaul of the MySQL check (#2116) and this has raised a very valid point. We're going to follow a slightly different approach but will take this into account to make sure the check will report available metrics and be more resilient when others are not found.

:+1:

truthbk:
I'll be closing this as we've already incorporated the idea of this PR into the logic, but thanks again for your contribution @nfo!

PR #2214, opened 2016-01-19
SNMP check as network check (multi-threaded)

truthbk:
# Multi-threaded SNMP Check (as a Network Check)
This PR allows checking for multiple SNMP instances in parallel, each running in it's own thread (within the allocated thread-pool). The actual OID/walk is still done serially for each instance, but this PR should help organizations with multiple instances configured with a performance benefit as a factor of the thread pool size.

NOTE: tests involving rates now require three runs because network checks report the results of the nth run on the n+1 run.

### DIFF
--- a/checks.d/snmp.py
+++ b/checks.d/snmp.py
@@ -9,7 +9,7 @@ from pysnmp.smi import builder
 from pysnmp.smi.exval import noSuchInstance, noSuchObject
 
 # project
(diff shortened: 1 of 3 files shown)


truthbk:
This PR can cab ignored - the original PR #2152 is now fine on TravisCI after cleaning the cache.

PR #2216, opened 2016-01-20
[check] [mysql] Fix buggy tagging in service_checks

hkaj:
If the check is configured to use `mysql_sock` but the connection fails for whatever reason, the service check will report a `port:0` tag because the `service_check_tags` update happens after the connection.

This PR fixes that.

### DIFF
--- a/checks.d/mysql.py
+++ b/checks.d/mysql.py
@@ -353,8 +353,8 @@ class MySql(AgentCheck):
     @contextmanager
     def _connect(self, host, port, mysql_sock, user, password, defaults_file, ssl):
         self.service_check_tags = [
-            'server:{0}'.format(host),
-            'port:{0}'.format(port)
+            'server:%s' % (mysql_sock if mysql_sock != '' else host),
+            'port:%s' % ('unix_socket' if port == 0 else port)
         ]
 
         db = None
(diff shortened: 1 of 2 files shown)


checks.d/mysql.py
```diff
@@ -97,8 +97,8 @@ def _get_config(self, instance):
 
     def _connect(self, host, port, mysql_sock, user, password, defaults_file):
         service_check_tags = [
-            'host:%s' % host,
-            'port:%s' % port
+            'host:%s' % mysql_sock if mysql_sock != '' else host,
```
truthbk: Wondering if this should be `localhost` if `mysql_sock` is set, given that a sock connection would always be local... not sure.

truthbk:
Other than the metaphysical question this is :+1:

truthbk:
I realized when I saw this running on Travis, it'll probably break some tests!

checks.d/mysql.py
hkaj: that makes sense, I'm tempted to go with it but at the same time don't want to change the behavior too much if a lot of people are relying on this value. Thoughts @remh ?

remh: what was it before ? i guess it was None or the empty string, right ? in that case localhost makes more sense than that

hkaj: :+1:

remh:
Let's not change what it was then and keep the socket path

hkaj:
Fixed the tests, they pass for mysql.

PR #2218, opened 2016-01-20
[process map] Replace `processes` resource check with gohai

olivielpeau:
Drop in replacement: populates the same payload and sends data at the same rate. Should be unnoticeable for the backend.

Performance-wise, it takes roughly the same amount of time as the previous check (~0.5sec). Most of that time is spent on launching a subprocess (the `processes` check was executing `ps`, we now execute `gops`).

TODO: remove the `processes` resource check and the whole `resources/` directory altogether (it's a BC breaking change on paper, but the resources checks are most likely not used by anybody).


checks/collector.py
```diff
@@ -42,6 +44,16 @@
 FLUSH_LOGGING_INITIAL = 5
 DD_CHECK_TAG = 'dd_check:{0}'
 
+PROCESSES_FORMAT_DESCRIPTION = [
```
LeoCavaille: maybe add some bits of documentation to explain what this is used for and how?

checks/collector.py
```diff
@@ -378,6 +392,27 @@ def run(self, checksd=None, start_event=True, configs_reloaded=False):
...
+                        processes_payload['format_description'] = PROCESSES_FORMAT_DESCRIPTION
+                    payload['resources']['processes'] = processes_payload
+                    has_resource = True
+                except OSError as e:
+                    if e.errno == 2:  # file not found, expected when install from source
+                        log.info("gops file not found")
+                    else:
+                        raise e
+                except Exception as e:
+                    log.warning("gops command failed with error %s" % str(e))
```
LeoCavaille: nit: use logger interpolation rather than string interpolation

checks/collector.py
```diff
@@ -222,10 +240,8 @@ def __init__(self, agentConfig, emitters, systemStats, hostname):
             except Exception:
                 log.exception('Unable to load custom check module %s' % module_spec)
 
-        # Resource Checks
-        self._resources_checks = [
-            ResProcesses(log, self.agentConfig)
-        ]
+        # Resource Checks (deprecated)
+        self._resources_checks = []
```
truthbk: I can see you're keeping this in for legacy reasons, but I want to double check, as we now have a relatively large if branch that would never be executed [here](https://github.com/DataDog/dd-agent/blob/olivielpeau/use-go-process-map/checks/collector.py#L380-L395).

truthbk:
Looks good to me, just want to double check about `self._resources_checks` and then I would say we're good to merge.

olivielpeau:
Thanks @truthbk for the feedback, after our discussion I've removed the resources check logic entirely (along with all the dead code).

checks/collector.py
```diff
@@ -41,6 +43,18 @@
 FLUSH_LOGGING_INITIAL = 5
 DD_CHECK_TAG = 'dd_check:{0}'
 
+# Description of the format of the `processes` resource check, identical to the legacy check
+# Is sent on the first run of the collector
+PROCESSES_FORMAT_DESCRIPTION = [
```
truthbk: now that we're actually axing the legacy check maybe a description of what each field is about would be helpful, or a pointer to where to find it.

truthbk:
Ran a quick double-check in case we were using `resources` anywhere else. As expected, I found nothing).

Looks good to me! Just added that minor comment regarding the doc.

truthbk:
:shipit:

PR #2223, opened 2016-01-22
Low/No Disk space caused the Agent to crash and not recover.

truthbk:
It seems like a low/no disk space can cause the TemporaryFile calls made in `subprocess_output` to "fail" - and by fail we mean empty output that can lead to IndexErrors, etc, when parsing the output. In general we have to be ready to expect an empty string as output.

### DIFF
--- a/checks.d/network.py
+++ b/checks.d/network.py
@@ -176,6 +176,8 @@ class Network(AgentCheck):
                 metrics = self._parse_linux_cx_state(lines[2:], self.TCP_STATES['netstat'], 5)
                 for metric, value in metrics.iteritems():
                     self.gauge(metric, value)
+            except Exception:
+                self.log.exception("Error collecting connection stats.")
 
         proc = open('/proc/net/dev', 'r')
         try:
--- a/checks.d/network.py
+++ b/checks.d/network.py
@@ -7,7 +7,10 @@ import re
 # project
 from checks import AgentCheck
 from utils.platform import Platform
-from utils.subprocess_output import get_subprocess_output
+from utils.subprocess_output import (
+    get_subprocess_output,
+    SubprocessOutputEmptyError,
+)
 
 BSD_TCP_METRICS = [
     (re.compile("^\s*(\d+) data packets \(\d+ bytes\) retransmitted\s*$"), 'system.net.tcp.retrans_packs'),
@@ -176,7 +179,7 @@ class Network(AgentCheck):
                 metrics = self._parse_linux_cx_state(lines[2:], self.TCP_STATES['netstat'], 5)
                 for metric, value in metrics.iteritems():
                     self.gauge(metric, value)
-            except Exception:
+            except SubprocessOutputEmptyError:
                 self.log.exception("Error collecting connection stats.")
 
         proc = open('/proc/net/dev', 'r')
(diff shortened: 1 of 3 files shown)


JohnLZeller:
It seems like the default case when calling `get_subprocess_output()` is that we expect output. I think we should invert the option here.

truthbk:
@JohnLZeller I agree that we normally do use it to run commands where we want an output, but from a library point of view - and this being a library - I don't really want to enforce that kind of behavior were empty output is automatically akin to an error.

truthbk:
@JohnLZeller actually nevermind.... you're right, the function is `get_subprocess_output` after all!

utils/subprocess_output.py
```diff
@@ -37,6 +37,10 @@ def get_subprocess_output(command, log, shell=False, stdin=None):
 
         stdout_f.seek(0)
         output = stdout_f.read()
+
+    if output_expected and output == "":
```
JohnLZeller: nit: could change `output == ""` to `not output`. That way it'll catch `""` and `None`

checks.d/network.py
```diff
@@ -176,6 +176,8 @@ def _check_linux(self, instance):
                 metrics = self._parse_linux_cx_state(lines[2:], self.TCP_STATES['netstat'], 5)
                 for metric, value in metrics.iteritems():
                     self.gauge(metric, value)
+            except Exception:
+                self.log.exception("Error collecting connection stats.")
```
JohnLZeller: We should print the exception here as well, so we don't silence other important Exceptions

JohnLZeller:
Looking pretty good. I wonder if it'd be worth defining a small Exception class, and raising that. This way we can catch a specific exception rather than all of them.

Maybe something like this:
`class SubprocessOutputEmptyError(Exception): pass`

Then you can just:
`raise SubprocessOutputEmptyError('get_subprocess_output expected output but had none.')`

And later can catch `SubprocessOutputEmptyError`

utils/subprocess_output.py
truthbk: I think read() always returns something... but I agree None is more pythonic.

JohnLZeller: Yeah, you're right :) definitely a nit

checks.d/network.py
truthbk: Log or raise? Cause we're logging it with `log.exception()`.... If we raise, then we'd be unable to collect stats such as the network stats below which would be collectible due to procfs being a pseudo FS and not require calling any subcommand.

Not sure what you mean here.

JohnLZeller: Opps, sorry, I meant to do `except Exception as e:` and then log the Exception message `e`. If you end up doing the custom Exception class below, it'll look different obviously

config.py
```diff
@@ -570,17 +570,20 @@ def get_system_stats():
...
+            output, _, _ = get_subprocess_output(['grep', 'model name', '/proc/cpuinfo'], log)
+            systemStats['cpuCores'] = len(output.splitlines())
+
+        if Platform.is_darwin(platf):
+            output, _, _ = get_subprocess_output(['sysctl', 'hw.ncpu'], log)
+            systemStats['cpuCores'] = int(output.split(': ')[1])
+
+        if Platform.is_freebsd(platf):
+            output, _, _ = get_subprocess_output(['sysctl', 'hw.ncpu'], log)
+            systemStats['cpuCores'] = int(output.split(': ')[1])
```
LeoCavaille: not your code, but seems like this could be written as:
```python
if Platform.is_linux(platf):
...
elif Platform.is_darwin(platf) or Platform.is_freebsd(platf):
...
```

config.py
```diff
@@ -570,17 +570,20 @@ def get_system_stats():
...
+    except Exception as e:
+        log.warning("unable to retrieve number of cpuCores. Failed with error %s" % str(e))
```
LeoCavaille: nit here: but you should use the logger interpolation `log.warning("...%s", e)`

utils/subprocess_output.py
LeoCavaille: don't know if maybe we should `.strip()` before testing if the string is empty, some rude programs sometimes output newlines for instance.

checks.d/network.py
truthbk: log.exception() already adds the exception info - type, stacktrace to the logging statement.

I have added the custom exception - for future use, but I don't think it's necessary to go over all try...catch blocks that already handler `Exception` and add yet another block that basically will do nothing other than log the exception. Sounds like overkill. But that's just me, coming from a C/C++ background.

JohnLZeller: yeah, it's not necessary, I just want to make sure we aren't silencing errors we don't expect. Best practice is to attempt to keep from running `except Exception:` as much as possible, for that reason.

Could we do a `catch SubprocessOutputEmptyError:` here instead?

config.py
```diff
@@ -570,17 +570,16 @@ def get_system_stats():
...
-        systemStats['cpuCores'] = int(output.split(': ')[1])
+    try:
+        if Platform.is_linux(platf):
+            output, _, _ = get_subprocess_output(['grep', 'model name', '/proc/cpuinfo'], log)
+            systemStats['cpuCores'] = len(output.splitlines())
+
+        if Platform.is_darwin(platf) or Platform.is_freebsd(platf):
+            output, _, _ = get_subprocess_output(['sysctl', 'hw.ncpu'], log)
+            systemStats['cpuCores'] = int(output.split(': ')[1])
+    except Exception as e:
```
JohnLZeller: Ditto, Could we do a `catch SubprocessOutputEmptyError:` here instead?

JohnLZeller:
LGTM, other than the notes :+1:

PR #2226, opened 2016-01-26
[mysql] circleci flavored test

truthbk:


truthbk:
Not really as flexible as we would've wanted. Will give travis a shot with the containers approach. Closing.

PR #2228, opened 2016-01-27
[system][wmi_check] handle `TimeoutException`

yannmh:
#### [core][wmi] `list` to cache connections
Use a `list` to cache WMI connections instead of a `set`.

#### [system][wmi_check] handle `TimeoutException`
WMI queries can potentially raise a `TimeoutException` on query timeouts. Handle it.

### DIFF
--- a/checks.d/wmi_check.py
+++ b/checks.d/wmi_check.py
@@ -4,6 +4,7 @@ from collections import namedtuple
 # project
 from checks import AgentCheck
 from checks.libs.wmi.sampler import WMISampler
+from utils.timeout import TimeoutException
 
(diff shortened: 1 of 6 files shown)


truthbk:
Looks good, just as we had discussed... my only question here is where do we want to call `pythoncom.CoInitialize()` - one option could be `_query()`, the other would be in `get_connection()` - note that I believe that when you call CoInitialize() you must also call `Dispatch()` and `ConnectServer()` to get a new live connection. IMHO this should be in `get_connection()` and let the old connection get garbage collected... otherwise the sampler becomes kind of cumbersome to use.

PR #2229, opened 2016-01-27
Added tag parameter for custom  MySQL queries

gzysk8:
Allow tags for MySQL custom queries, separate from the global tags applied to the default metrics

### DIFF
--- a/checks.d/mysql.py
+++ b/checks.d/mysql.py
@@ -200,7 +200,8 @@ class MySql(AgentCheck):
         # Max of 20 queries allowed
         if isinstance(queries, list):
             for index, check in enumerate(queries[:self.MAX_CUSTOM_QUERIES]):
-                self._collect_dict(check['type'], {check['field']: check['metric']}, check['query'], db, tags=tags)
+                total_tags = tags + check.get('tags', [])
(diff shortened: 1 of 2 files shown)


checks.d/mysql.py
```diff
@@ -200,7 +200,12 @@ def _collect_metrics(self, host, db, tags, options, queries):
         # Max of 20 queries allowed
         if isinstance(queries, list):
             for index, check in enumerate(queries[:self.MAX_CUSTOM_QUERIES]):
-                self._collect_dict(check['type'], {check['field']: check['metric']}, check['query'], db, tags=tags)
+                if 'tags' in check:
+                    total_tags = tags + check['tags']
+                else:
+                    total_tags = tags
+
```
truthbk: you could probably just do it all in a single line:
`total_tags = tags + check.get('tags', [])`

check.get('tags', []) will return en empty list if there's no tags in check, or checks['tags'] if it is... So it shortens it a bit.

As you said, here it comes down to readability vs space efficiency.

truthbk: Just a comment - logic is :+1:

olivielpeau:
:shipit:

PR #2235, opened 2016-01-29
[hdfs_jmx] An HDFS agent that uses JMX

zachradtka:
An agent check for HDFS that uses the JMX interface to gather HDFS metrics from individual HDFS data nodes. 

The metrics collected are
```
hdfs.dfs_remaining                  The remaining disk space left in bytes
hdfs.storage_info                   Path to HDFS storage location
hdfs.dfs_capacity                   Disk capacity in bytes
hdfs.dfs_used                       Disk usage in bytes
hdfs.cache_capacity                 Cache capacity in bytes
hdfs.num_failed_volumes             Number of failed volumes
hdfs.last_volume_failure_date       Date the last volume failed
hdfs.estimated_capacity_lost_total  The estimated capacity lost in bytes
hdfs.num_blocks_cached              The number of blocks cached
hdfs.num_blocks_failed_to_cache     The number of blocks that failed to cache
hdfs.num_blocks_failed_to_uncache   The number of failed blocks to remove from cache
```

Authors:
@zachradtka
@wjsl


olivielpeau:
Thanks @zachradtka and @wjsl for your contributions!

We'll review your PRs in depth soon.

zachradtka:
I just pushed a quick change that added a few metrics for the HDFS namenode and split the agent-check for the datanodes and for the name node.

The metrics for the namenode are as follows.
```
hdfs.namenode.capacity_total                    Total disk capacity in bytes
hdfs.namenode.capacity_used                     Disk usage in bytes
hdfs.namenode.capacity_remaining                Remaining disk space left in bytes
hdfs.namenode.total_load                        Total load on the file system
hdfs.namenode.fs_lock_queue_length              Lock queue length
hdfs.namenode.blocks_total                      Total number of blocks
hdfs.namenode.max_objects                       Maximum number of files HDFS supports
hdfs.namenode.files_total                       Total number of files
hdfs.namenode.pending_replication_blocks        Number of blocks pending replication
hdfs.namenode.under_replicated_blocks           Number of under replicated blocks
hdfs.namenode.scheduled_replication_blocks      Number of blocks scheduled for replication
hdfs.namenode.pending_deletion_blocks           Number of pending deletion blocks
hdfs.namenode.num_live_data_nodes               Total number of live data nodes
hdfs.namenode.num_dead_data_nodes               Total number of dead data nodes
hdfs.namenode.num_decom_live_data_nodes         Number of decommissioning live data nodes
hdfs.namenode.num_decom_dead_data_nodes         Number of decommissioning dead data nodes
hdfs.namenode.volume_failures_total             Total volume failures
hdfs.namenode.estimated_capacity_lost_total     Estimated capacity lost in bytes
hdfs.namenode.num_decommissioning_data_nodes    Number of decommissioning data nodes
hdfs.namenode.num_stale_data_nodes              Number of stale data nodes
hdfs.namenode.num_stale_storages                Number of stale storages
```

Sorry for the late add, but I really felt these metrics would be helpful.

tests/checks/mock/test_hdfs_datanode.py
```diff
@@ -0,0 +1,56 @@
...
+        def raise_for_status(self):
+            return True
+
+
+    with open(Fixtures.file('hdfs_jmx'), 'r') as f:
+        body = f.read()
+        return MockResponse(body, 200)
+
+class HDFSDataNode(AgentCheckTest):
+    CHECK_NAME = 'hdfs_jmx'
```
olivielpeau: Don't forget to update `CHECK_NAME` to `hdfs_datanode` ;)

checks.d/hdfs_datanode.py
```diff
@@ -0,0 +1,180 @@
...
+    def _hdfs_datanode_metrics(self, jmx_uri, jmx_path):
+        '''
+        Get HDFS data node metrics from JMX
+        '''
+        response = self._rest_request_to_json(jmx_uri,
+            jmx_path,
+            JMX_SERVICE_CHECK,
+            qry='Hadoop:service=DataNode,name=FSDatasetState')
+
+        print response
```
olivielpeau: Should be removed from the final check :)

checks.d/hdfs_datanode.py
```diff
@@ -0,0 +1,180 @@
...
+        if args:
+            for directory in args:
+                url = self._join_url_dir(url, directory)
+
+        # Add kwargs as arguments
+        if kwargs:
+            query = '&'.join(['{}={}'.format(key, value) for key, value in kwargs.iteritems()])
+            url = urljoin(url, '?' + query)
+
+        print url
```
olivielpeau: Should be logged with a `log.debug` instead

checks.d/hdfs_datanode.py
```diff
@@ -0,0 +1,176 @@
...
+
+# stdlib
+from urlparse import urljoin
+
+# 3rd party
+import requests
+from requests.exceptions import Timeout
+from requests.exceptions import HTTPError
+from requests.exceptions import InvalidURL
+from requests.exceptions import ConnectionError
```
olivielpeau: Nitpick: you can put these imports on a single line:
```python
from requests.exceptions import ConnectionError, HTTPError, InvalidURL, Timeout
```

checks.d/hdfs_namenode.py
```diff
@@ -0,0 +1,193 @@
...
+    'EstimatedCapacityLostTotal' : ('hdfs.namenode.estimated_capacity_lost_total',  GAUGE),
+    'NumDecommissioningDataNodes' : ('hdfs.namenode.num_decommissioning_data_nodes',  GAUGE),
+    'NumStaleDataNodes' : ('hdfs.namenode.num_stale_data_nodes',  GAUGE),
+    'NumStaleStorages' : ('hdfs.namenode.num_stale_storages',  GAUGE),
+}
+
+class HDFSNameNode(AgentCheck):
+
+    def check(self, instance):
+        jmx_address = instance.get('hdfs_namenode_jmx_uri', JMX_URI)
```
olivielpeau: We usually don't make that parameter optional: we check that it exists and raise an error if not, so you can remove `JMX_URI ` and use something like:

```python
jmx_address  = instance.get('hdfs_namenode_jmx_uri')
if jmx_address is None:
    raise Exception("A url must be specified in the instance configuration")
```

It makes sense especially because it's the only parameter of the instance, so if it's not present in the configuration the instance is completely ignored by the collector.

checks.d/hdfs_namenode.py
```diff
@@ -0,0 +1,193 @@
...
+
+        # Get metrics from JMX
+        self._hdfs_namenode_metrics(jmx_address, JMX_PATH)
+
+    def _hdfs_namenode_metrics(self, jmx_uri, jmx_path):
```
olivielpeau: Don't see the need to pass `jmx_path` as an argument here, why not use `JMX_PATH` directly as an argument of `_rest_request_to_json`?

checks.d/hdfs_namenode.py
```diff
@@ -0,0 +1,193 @@
...
+            jmx_path,
+            JMX_SERVICE_CHECK,
+            qry='Hadoop:service=NameNode,name=FSNamesystemState')
+
+        beans = response.get('beans', [])
+
+        for bean in beans:
+            bean_name = bean.get('name')
+
+            if bean_name is not None \
```
olivielpeau: no need to check that `bean_name` is not `None` here as we already compare it with `HDFS_NAMENODE_BEAN_NAME` ;)

checks.d/hdfs_namenode.py
```diff
@@ -0,0 +1,193 @@
...
+        if args:
+            for directory in args:
+                url = self._join_url_dir(url, directory)
+
+        # Add kwargs as arguments
+        if kwargs:
+            query = '&'.join(['{}={}'.format(key, value) for key, value in kwargs.iteritems()])
+            url = urljoin(url, '?' + query)
+
+        try:
```
olivielpeau: Could you add some `debug`-level logging here, with a message that includes the URL that we're hitting?

checks.d/hdfs_datanode.py
```diff
@@ -0,0 +1,173 @@
...
+    def _set_metric(self, metric_name, metric_type, value, tags=None, device_name=None):
+        '''
+        Set a metric
+        '''
+        if metric_type == GAUGE:
+            self.gauge(metric_name, value, tags=tags, device_name=device_name)
+        else:
+            self.log.error('Metric type "%s" unknown' % (metric_type))
+
+    def _rest_request_to_json(self, address, object_path, service_name=JMX_SERVICE_CHECK, *args, **kwargs):
```
olivielpeau: Let's simplify the function signature a bit since we're only using the `address` and `object_path`. Keeping only these 2 parameters should be enough :)

checks.d/hdfs_namenode.py
```diff
@@ -0,0 +1,193 @@
...
+    def _set_metric(self, metric_name, metric_type, value, tags=None, device_name=None):
+        '''
+        Set a metric
+        '''
+        if metric_type == GAUGE:
+            self.gauge(metric_name, value, tags=tags, device_name=device_name)
+        else:
+            self.log.error('Metric type "%s" unknown' % (metric_type))
+
+    def _rest_request_to_json(self, address, object_path, service_name=JMX_SERVICE_CHECK, *args, **kwargs):
```
olivielpeau: Let's simplify the function signature a bit since we're only using the `address` and `object_path`. Keeping only these 2 parameters should be enough :)

checks.d/hdfs_namenode.py
```diff
@@ -0,0 +1,193 @@
...
+
+        # Add args to the url
+        if args:
+            for directory in args:
+                url = self._join_url_dir(url, directory)
+
+        # Add kwargs as arguments
+        if kwargs:
+            query = '&'.join(['{}={}'.format(key, value) for key, value in kwargs.iteritems()])
+            url = urljoin(url, '?' + query)
```
olivielpeau: Let's remove the `args` and `kwargs` since we're not using them.

checks.d/hdfs_namenode.py
```diff
@@ -0,0 +1,193 @@
...
+        else:
+            self.log.error('Metric type "%s" unknown' % (metric_type))
+
+    def _rest_request_to_json(self, address, object_path, service_name=JMX_SERVICE_CHECK, *args, **kwargs):
+        '''
+        Query the given URL and return the JSON response
+        '''
+        response_json = None
+
+        service_check_tags = ['instance:%s' % self.hostname]
```
olivielpeau: This tag is not very useful since the agent adds a `host:` tag which contains the agent's hostname to all service checks.

Having a `namenode_url:` tag with the instance's `hdfs_namenode_jmx_uri` would be better here

checks.d/hdfs_namenode.py
```diff
@@ -0,0 +1,193 @@
...
+                tags=service_check_tags,
+                message="Request failed: {0}, {1}".format(url, e))
+            raise
+
+        except JSONDecodeError as e:
+            self.service_check(service_name,
+                AgentCheck.CRITICAL,
+                tags=service_check_tags,
+                message='JSON Parse failed: {0}, {1}'.format(url, e))
+
```
olivielpeau: We should probably also re-raise the error in this case, or is there a reason not to do so?

olivielpeau:
Added a bunch of comments, most of them are nitpicks. The checks looks good overall, thanks!

I've only added comments on the namenode check, but since the datanode check is pretty similar could you also take them into account for the datanode check?

One last thing: we try to keep our metrics' and service checks' prefixes similar to the checks' names, so could you rename all the metrics and service checks to `hdfs_namenode.[...]` and `hdfs_datanode` respectively?

Thanks again!

checks.d/hdfs_namenode.py
zachradtka: I am using the kwargs in one of my calls:
```
        response = self._rest_request_to_json(jmx_uri,
            jmx_path,
            JMX_SERVICE_CHECK,
            qry='Hadoop:service=NameNode,name=FSNamesystemState')
```

I can remove the service name. Or did you want me to hardcode in the BeanName?

olivielpeau: Oh that's right, missed that, thanks.

Then instead could we have the `_rest_request_to_json` method take a `query_params` named param that would be a dict of the query parameters to use?

So having:
```python
def _rest_request_to_json(self, address, object_path, query_params=None):
```
and calling it with:
```python
self._rest_request_to_json(jmx_uri, JMX_PATH, query_params={'qry': HDFS_NAMENODE_BEAN_NAME})
```

What do you think?

checks.d/hdfs_namenode.py
zachradtka: I enabled the tag `namenode_url` for the namenode check and `datanode_url` for the datanode checks. I passed the `jmx_uri` into `_set_metric` so it wouldn't auto close this comment.

zachradtka:
Thanks for the comments, I completed all of them on both the datanode and namenode agent checks.

checks.d/hdfs_datanode.py
```diff
@@ -0,0 +1,169 @@
...
+        service_check_tags = ['instance:%s' % self.hostname]
+
+        url = address
+
+        if object_path:
+            url = self._join_url_dir(url, object_path)
+
+        # Add query_params as arguments
+        if query_params:
+            query = '&'.join(['{}={}'.format(key, value) for key, value in query_params.iteritems()])
```
olivielpeau: We support python 2.6 so you have to use `'{0}={1}'.format(key, value)` here (i.e. number the fields)

checks.d/hdfs_datanode.py
```diff
@@ -0,0 +1,169 @@
...
+        # Get metrics from JMX
+        self._hdfs_datanode_metrics(jmx_address)
+
+    def _hdfs_datanode_metrics(self, jmx_uri):
+        '''
+        Get HDFS data node metrics from JMX
+        '''
+        response = self._rest_request_to_json(jmx_uri,
+            JMX_PATH,
+            query_params={'qry':'Hadoop:service=DataNode,name=FSDatasetState'})
```
olivielpeau: Could you use the value of `HDFS_DATANODE_BEAN_NAME` here? (unless you have reasons not to)

checks.d/hdfs_namenode.py
olivielpeau: Thanks for changing that! Could you also use the `namenode_url:<jmx_uri>` tag instead of the `instance:<hostname>` tag for the service checks of the datanode check? (and same thing for the namenode check)

checks.d/hdfs_namenode.py
```diff
@@ -0,0 +1,189 @@
...
+            bean_name = bean.get('name')
+
+            if bean_name == HDFS_NAMENODE_BEAN_NAME:
+                for metric, (metric_name, metric_type) in HDFS_METRICS.iteritems():
+                    metric_value = bean.get(metric)
+
+                    if metric_value is not None:
+                        self._set_metric(metric_name, metric_type, metric_value, tags)
+
+    def _set_metric(self, metric_name, metric_type, value, tags=None, device_name=None):
```
olivielpeau: I _think_ the check never uses the `device_name` parameter, so we may as well remove it.

checks.d/hdfs_namenode.py
```diff
@@ -0,0 +1,189 @@
...
+            query_params={'qry':'Hadoop:service=NameNode,name=FSNamesystemState'})
+
+        beans = response.get('beans', [])
+
+        tags = ['namenode_url:' + jmx_uri]
+
+        for bean in beans:
+            bean_name = bean.get('name')
+
+            if bean_name == HDFS_NAMENODE_BEAN_NAME:
```
olivielpeau: Just have a question here: will the API only ever return one bean for our request?

Because otherwise we'd need to tag the metrics with something that would identify every bean that's returned.

olivielpeau:
Thanks for addressing my comments! I've added in a few comments and a question, once they're addressed the check should be good to go.

checks.d/hdfs_namenode.py
zachradtka: Only one bean should get returned. I used a list iterator because I thought it was cleaner than calling `if len(bean) == 1:`

zachradtka:
All comments are always welcome!

I have fixed all of the concerns for both the datanode and namenode agents. Let me know what to do next.

checks.d/hdfs_namenode.py
olivielpeau: Thanks for the clarification. The only thing that bothers me a bit here is that in cases when the API doesn't return what we expect (ie no beans or multiple beans with that name), the check would either silently send no metrics or silently overwrite metrics.

Could we maybe replace the `for` loop with the following:
```python
bean = next(iter(beans))  # get the bean
bean_name = bean.get('name')
if bean_name != HDFS_NAMENODE_BEAN_NAME:
  raise Exception("Unexpected bean name {0}".format(bean_name))
```

and then loop over the metrics we want to send.

This would make the check handle unexpected results from the API more explicitly, and also make it clear in the code that we're expecting one bean only.

What do you think?

olivielpeau:
Added one comment, once it's addressed could you squash your commits into one?

Thanks!

zachradtka:
OK, All commits squashed and rebased to the latest master.

olivielpeau:
Thanks again!

Looks good, I'll merge once the CI passes.

PR #2236, opened 2016-01-29
[mapreduce] An agent that reports MapReduce metrics

zachradtka:
An agent check that gathers metrics for MapReduce applications.

The metrics collected are:
```
MapReduce Job Metrics
---------------------
mapreduce.job.elapsed_ime                The elapsed time since the application started (in ms)
mapreduce.job.maps_total                 The total number of maps
mapreduce.job.maps_completed             The number of completed maps
mapreduce.job.reduces_total              The total number of reduces
mapreduce.job.reduces_completed          The number of completed reduces
mapreduce.job.maps_pending               The number of maps still to be run
mapreduce.job.maps_running               The number of running maps
mapreduce.job.reduces_pending            The number of reduces still to be run
mapreduce.job.reduces_running            The number of running reduces
mapreduce.job.new_reduce_attempts        The number of new reduce attempts
mapreduce.job.running_reduce_attempts    The number of running reduce attempts
mapreduce.job.failed_reduce_attempts     The number of failed reduce attempts
mapreduce.job.killed_reduce_attempts     The number of killed reduce attempts
mapreduce.job.successful_reduce_attempts The number of successful reduce attempts
mapreduce.job.new_map_attempts           The number of new map attempts
mapreduce.job.running_map_attempts       The number of running map attempts
mapreduce.job.failed_map_attempts        The number of failed map attempts
mapreduce.job.killed_map_attempts        The number of killed map attempts
mapreduce.job.successful_map_attempts    The number of successful map attempts
MapReduce Job Counter Metrics
-----------------------------
mapreduce.job.counter.reduce_counter_value   The counter value of reduce tasks
mapreduce.job.counter.map_counter_value      The counter value of map tasks
mapreduce.job.counter.total_counter_value    The counter value of all tasks
MapReduce Task Metrics
----------------------
mapreduce.job.task.progress         The progress of the task as a percent
mapreduce.job.task.elapsed_time     The elapsed time sine the application started (in ms)
maprduce.job.task.state.new         The total number of new tasks
maprduce.job.task.state.scheduled   The total number of tasks with a scheduled state
maprduce.job.task.state.running     The total number of tasks with a running state
maprduce.job.task.state.succeeded   The total number of tasks with a succeeded state
maprduce.job.task.state.failed      The total number of tasks with a failed state
maprduce.job.task.state.killWait    The total number of task with a kill_wait state
maprduce.job.task.state.killed      The total number of tasks with a killed state
maprduce.job.task.type.map          The total number of map tasks
maprduce.job.task.type.reduce       The total number of reduce tasks
MapReduce Task Counter Metrics
------------------------------
mapreduce.job.task.counter.map_input_records        The number of input records of a map task
mapreduce.job.task.counter.map_output_records       The number of output records of a map task
mapreduce.job.task.counter.combine_input_records    The number of input records of a combine task
mapreduce.job.task.counter.combine_output_records   The number of output records of a combine task
mapreduce.job.task.counter.reduce_input_records     The number of input records of a reduce task
mapreduce.job.task.counter.reduce_output_records    The number of output records of a reduce task
```

Authors:
@zachradtka
@wjsl


checks.d/mapreduce.py
```diff
@@ -0,0 +1,499 @@
...
+
+        return running_tasks
+
+    def _mapreduce_task_metric_states(self, tasks_json, job_id, cluster_id):
+        '''
+        Get metrics for task states and task types
+        '''
+        # Initialize the state and type counts
+        state_counts = {state: 0 for state in MAPREDUCE_TASK_STATE_METRICS}
+        type_counts = {task_type: 0 for task_type in MAPREDUCE_TASK_TYPE_METRICS}
```
olivielpeau: Just a quick comment: the `default` test fails here on python2.6 because of the dict comprehension that's only in python since 2.7.

You can probably initialize these vars to `defaultdict`s instead (from the [`collections`](https://docs.python.org/2/library/collections.html#collections.defaultdict) module).

checks.d/mapreduce.py
```diff
@@ -0,0 +1,487 @@
...
+# Project
+from checks import AgentCheck
+
+# Resource Manager URI
+DEFAULT_RM_URI = 'http://localhost:8088'
+
+# Service Check Names
+SERVICE_CHECK = 'cluster.can_connect'
+YARN_SERVICE_CHECK = 'resource_manager.can_connect'
+MAPREDUCE_SERVICE_CHECK = 'application_master.can_connect'
```
olivielpeau: The service checks, like metrics, should be prefixed with the check name, so `mapreduce.cluster.can_connect`, etc, instead

checks.d/mapreduce.py
```diff
@@ -0,0 +1,487 @@
...
+    'COMBINE_OUTPUT_RECORDS': ('mapreduce.job.task.counter.combine_output_records', GAUGE),
+    'REDUCE_INPUT_RECORDS': ('mapreduce.job.task.counter.reduce_input_records', GAUGE),
+    'REDUCE_OUTPUT_RECORDS': ('mapreduce.job.task.counter.reduce_output_records', GAUGE),
+}
+
+class MapReduceCheck(AgentCheck):
+
+    def check(self, instance):
+        # Get properties from conf file
+        rm_address = instance.get('resourcemanager_uri', DEFAULT_RM_URI)
```
olivielpeau: I'd say we should require `resourcemanager_uri` as it's the only parameter of the instance configuration, and we require instances to have at least one parameter in the yaml config. Also, the value in the example yaml file makes it clear what the default URI is.

checks.d/mapreduce.py
```diff
@@ -0,0 +1,481 @@
...
+    'FAILED': ('maprduce.job.task.state.failed', GAUGE),
+    'KILL_WAIT': ('maprduce.job.task.state.killWait', GAUGE),
+    'KILLED': ('maprduce.job.task.state.killed', GAUGE),
+}
+
+MAPREDUCE_TASK_TYPE_METRICS = {
+    'MAP': ('maprduce.job.task.type.map', GAUGE),
+    'REDUCE': ('maprduce.job.task.type.reduce', GAUGE),
+}
+
```
yannmh: Some typo here, `MAPREDUCE_TASK_METRICS ` & `MAPREDUCE_TASK_STATE_METRICS ` metrics are namespaced with `maprduce ` instead of `mapreduce` :pencil2: .

checks.d/mapreduce.py
```diff
@@ -0,0 +1,545 @@
...
+                        task_progress = task['progress']
+
+                        tags = ['cluster_id:' + str(cluster_id),
+                                'app_name:' + job_stats['app_name'],
+                                'user_name:' + job_stats['user_name'],
+                                'job_name:' + job_stats['job_name']
+                                ]
+
+                        if task_type == 'MAP':
+                            self._set_metric('mapreduce.job.task.map.progress', HISTOGRAM, task_progress, tags=tags)
```
olivielpeau: Just a quick comment: don't forget to add a tag with the `task_type` here if you're sending this metric for the two different task types ;)

zachradtka:
The mapreduce agent has the following metrics
```
MapReduce Job Metrics
---------------------
mapreduce.job.elapsed_ime                The elapsed time since the application started (in ms)
mapreduce.job.maps_total                 The total number of maps
mapreduce.job.maps_completed             The number of completed maps
mapreduce.job.reduces_total              The total number of reduces
mapreduce.job.reduces_completed          The number of completed reduces
mapreduce.job.maps_pending               The number of maps still to be run
mapreduce.job.maps_running               The number of running maps
mapreduce.job.reduces_pending            The number of reduces still to be run
mapreduce.job.reduces_running            The number of running reduces
mapreduce.job.new_reduce_attempts        The number of new reduce attempts
mapreduce.job.running_reduce_attempts    The number of running reduce attempts
mapreduce.job.failed_reduce_attempts     The number of failed reduce attempts
mapreduce.job.killed_reduce_attempts     The number of killed reduce attempts
mapreduce.job.successful_reduce_attempts The number of successful reduce attempts
mapreduce.job.new_map_attempts           The number of new map attempts
mapreduce.job.running_map_attempts       The number of running map attempts
mapreduce.job.failed_map_attempts        The number of failed map attempts
mapreduce.job.killed_map_attempts        The number of killed map attempts
mapreduce.job.successful_map_attempts    The number of successful map attempts

MapReduce Job Counter Metrics
-----------------------------
mapreduce.job.counter.reduce_counter_value   The counter value of reduce tasks
mapreduce.job.counter.map_counter_value      The counter value of map tasks
mapreduce.job.counter.total_counter_value    The counter value of all tasks

MapReduce Map Task Metrics
--------------------------
mapreduce.job.map.task.progress     The distribution of all map task progresses
mapreduce.job.map.task.elapsed_time The distribution of all map tasks elapsed time

MapReduce Reduce Task Metrics
--------------------------
mapreduce.job.reduce.task.progress      The distribution of all reduce task progresses
mapreduce.job.reduce.task.elapsed_time  The distribution of all reduce tasks elapsed time
```

The `task` metrics are histogram metrics and sample from each task running under a job.

The `job counter` metrics are custom metrics that the user must specify in the `init_config` section of `mapreduce.yaml`.

As an example, to get metrics for the counter with name `HDFS_BYTES_WRITTEN` for the counter group `org.apache.hadoop.mapreduce.FileSystemCounter` the `init_config` section would look like:
```
  counter_metrics:
     - job_name: 'Foo'
       metrics:
         - counter_group_name: 'org.apache.hadoop.mapreduce.FileSystemCounter'
           counters:
             - counter_name: 'HDFS_BYTES_WRITTEN'
```

checks.d/mapreduce.py
```diff
@@ -0,0 +1,449 @@
...
+                            running_jobs[str(job_name)] = {'app_name': str(app_name),
+                                                    'user_name': str(user_name),
+                                                    'tracking_url': self._join_url_dir(tracking_url, MAPREDUCE_JOBS_PATH, job_id)}
+
+                            tags = ['cluster_id:' + str(cluster_id),
+                                    'app_name:' + str(app_name),
+                                    'user_name:' + str(user_name),
+                                    'job_name:' + str(job_name)]
+
+                            self._set_metrics_from_json(tags, job_json, MAPREDUCE_JOB_METRICS)
```
olivielpeau: Actually there's still an issue here: we're sending the `MAPREDUCE_JOB_METRICS` (all gauges) for each job (ie for each `job_id`), but we only tag by `user_name`/`job_name`, which means that these metrics get overwritten when multiple jobs share the same `job_name`.

This means that once the check has run, for a given `job_name`, the values of these metrics will be the values from the last job in the list that has this `job_name`.

olivielpeau: Instead of this we should pre-aggregate the metrics for each `job_name`, probably using histograms.

checks.d/mapreduce.py
```diff
@@ -0,0 +1,449 @@
...
+                        job_name = job_json['name']
+                        user_name = job_json['user']
+
+                        if job_id is not None \
+                                and job_name is not None \
+                                and user_name is not None:
+
+                            running_jobs[str(job_name)] = {'app_name': str(app_name),
+                                                    'user_name': str(user_name),
+                                                    'tracking_url': self._join_url_dir(tracking_url, MAPREDUCE_JOBS_PATH, job_id)}
```
olivielpeau: I think that the `tracking_url` is specific to a `job_id`, not a `job_name`, so there should be multiple `tracking_url`s for a given `job_name`. Otherwise, when multiple jobs share the same `job_name`, we'd only retrieve the `MAPREDUCE_JOB_COUNTER_METRICS`, `MAPREDUCE_MAP_TASK_METRICS ` and `MAPREDUCE_REDUCE_TASK_METRICS` on one of the jobs that has that `job_name`, and the values on these metrics would only reflect that one job instead of all the jobs that share the same `job_name`.

For these metrics to be correct, we need to fetch the metrics for every job (even those that share the same `job_name`), and then compute the aggregates for each `job_name`.

Does that make sense?

checks.d/mapreduce.py
```diff
@@ -0,0 +1,449 @@
...
+
+        except ValueError as e:
+            self.service_check(service_name,
+                AgentCheck.CRITICAL,
+                tags=service_check_tags,
+                message=e)
+            raise
+
+        else:
+            self.service_check(service_name,
```
olivielpeau: Instead of sending an `OK` service check every time the request is successful, we should probably only send the `OK` status once at the end of a check run on an instance (so at the end the `check` method for instance).

checks.d/mapreduce.py
```diff
@@ -0,0 +1,449 @@
...
+        else:
+            self.log.error('Metric type "%s" unknown' % (metric_type))
+
+    def _rest_request_to_json(self, address, object_path, service_name=SERVICE_CHECK, *args, **kwargs):
+        '''
+        Query the given URL and return the JSON response
+        '''
+        response_json = None
+
+        service_check_tags = ['instance:%s' % self.hostname]
```
olivielpeau: Here, same comment as with the HDFS check, the service check tag should probably be something like `resourcemanager_url:<url>` instead of the hostname of the agent.

olivielpeau:
@zachradtka Just added a few comments on the check (see above), unfortunately there's still an tagging/aggregation issue on the metrics, feel free to reach out if you need more details.

checks.d/mapreduce.py
zachradtka: Is there a way with histograms to only produce `max`, `avg`, `median`, etc?

zachradtka: OK, I have updated all of the Job metrics to be Histograms.

zachradtka:
@olivielpeau I updated the Agent. It tracks applications and jobs by app ID and job ID respectively. I also updated the agent to utilize histograms for all of the checks. Please let me know if there is anything else I should change.

checks.d/mapreduce.py
olivielpeau: There's no built-in way to do that in the agent actually, so let's keep it simple and use normal histograms. Thanks!

olivielpeau:
Thanks @zachradtka! I'll make a few comments on some details, and we may have more comments once we've tested the check on our environment, but overall the check looks good.

checks.d/mapreduce.py
```diff
@@ -0,0 +1,451 @@
...
+        if rm_address is None:
+            raise Exception('The ResourceManager URL must be specified in the instance configuration')
+
+        # Get the cluster ID from Yarn
+        cluster_id = self._get_cluster_id(rm_address)
+
+        # Get the running MR applications from YARN
+        running_apps = self._get_running_app_ids(rm_address,
+            states=YARN_APPLICATION_STATES,
+            applicationTypes=YARN_APPLICATION_TYPES)
```
olivielpeau: Nitpick: coud you move the usage of these parameters further down, inside the `_get_running_app_ids` method, on the `_rest_request_to_json` function call?

olivielpeau:
Only had one comment for now, we'll get back to you once we've tested the check, thanks!

zachradtka:
OK, made that quick change.

conf.d/mapreduce.yaml.example
```diff
@@ -0,0 +1,23 @@
+instances:
+    -   resourcemanager_uri: http://localhost:8088
+
+init_config:
+  #
+  # Return metrics for specified counters
+  #
+
+  counter_metrics:
+  #   - job_name: 'Foo'
```
olivielpeau: Could you add a comment here in the yaml file with a link to the related hadoop doc page? (probably https://hadoop.apache.org/docs/current/hadoop-mapreduce-client/hadoop-mapreduce-client-core/MapredAppMasterRest.html#Job_Counters_API)

olivielpeau:
@zachradtka: I have some feedback from the team that's going to use the check:
- do you think it would make sense for some of the counter metrics to be retrieved by default? Specifically we were thinking about the ones that are useful for judging the work performed (bytes read/written, input/output records).
- could you add a config option that allows configuring a `counter_metrics` entry for all the job names instead of just for one name?

The team is going to test the check today, I'll get back to you with their feedback on that. Thanks!

zachradtka:
Ha, I originally had it written to be job agnostic and get all the specified counters from all jobs, but decided that having more granularity was better. I do like the option of having universal counters, that are collected for every job, and configurable in the `counter_metrics` section. And yes, I can do that.

Is this feature a blocker for testing, or can you test without it and I can add it later today?

olivielpeau:
Thanks for your answers. It's not a blocker, we're going to test the check as-is today.

zachradtka:
I updated the agent to include job-agnostic counter metrics configurable in the `mapreduce.yaml`. I also added some comments in the yaml file explaining how to configure the agent.

checks.d/mapreduce.py
```diff
@@ -0,0 +1,500 @@
...
+from requests.exceptions import Timeout, HTTPError, InvalidURL, ConnectionError
+from simplejson import JSONDecodeError
+
+# Project
+from checks import AgentCheck
+
+# Service Check Names
+SERVICE_CHECK = 'mapreduce.cluster.can_connect'
+YARN_SERVICE_CHECK = 'mapreduce.resource_manager.can_connect'
+MAPREDUCE_SERVICE_CHECK = 'mapreduce.application_master.can_connect'
```
olivielpeau: Actually hadn't noticed that there were different service checks when I suggested to only send an `OK` once.

The check should actually send an `OK` service check for each service kind when all the requests to that service have been made successfully.

So from what I see in the current state of the check, we should probably:
- remove the `SERVICE_CHECK` service check altogether (given that we only ever send `OK`s on it)
- send an `OK` `YARN_SERVICE_CHECK` service check right after `self._get_running_app_ids`, because, at this point of the check, all the requests made to the resource manager have been successful
- send an `OK` `MAPREDUCE_SERVICE_CHECK` at the end of the main `check` method, because it's only at that point of the check that we know that all the requests to the application master have been successful.

Does that sound sensible to you? If so could you make these changes and add a small comment on each `OK` service check explaining why we're sending the service check at that point of the check?

Thanks!

checks.d/mapreduce.py
```diff
@@ -0,0 +1,500 @@
...
+
+# URL Paths
+INFO_PATH = 'ws/v1/cluster/info'
+YARN_APPS_PATH = 'ws/v1/cluster/apps'
+MAPREDUCE_JOBS_PATH = 'ws/v1/mapreduce/jobs'
+
+
+TASK_COUNTER_GROUP_NAME = 'org.apache.hadoop.mapreduce.TaskCounter'
```
olivielpeau: This variable is not used anymore ;)

checks.d/mapreduce.py
```diff
@@ -0,0 +1,500 @@
...
+
+# Application type and states to collect
+YARN_APPLICATION_TYPES = 'MAPREDUCE'
+YARN_APPLICATION_STATES = 'RUNNING'
+
+# Metric types
+GAUGE = 'gauge'
```
olivielpeau: This one neither

checks.d/mapreduce.py
```diff
@@ -0,0 +1,500 @@
...
+                    metric_type,
+                    metric_status,
+                    tags)
+
+    def _set_metric(self, metric_name, metric_type, value, tags=None, device_name=None):
+        '''
+        Set a metric
+        '''
+        if metric_type == GAUGE:
+            self.gauge(metric_name, value, tags=tags, device_name=device_name)
```
olivielpeau: Let's remove these two lines as we're not using them anymore.

We'll add them back later if we need them.

checks.d/mapreduce.py
```diff
@@ -0,0 +1,500 @@
...
+        else:
+            self.log.error('Metric type "%s" unknown' % (metric_type))
+
+    def _rest_request_to_json(self, address, object_path, service_name=SERVICE_CHECK, *args, **kwargs):
+        '''
+        Query the given URL and return the JSON response
+        '''
+        response_json = None
+
+        service_check_tags = ['resourcemanager_url:%s' % address]
```
olivielpeau: Just realized that this `address` wouldn't always be the same for a given `service_name`. Could we only use the `scheme://hostname:port` part of the `address` as a tag here? The tag also needs to be consistent with the tag that we send with the `OK` service check on the same `service_name`.

Also, given that it's not always the resource manager's url, we should probably simply name the tag `url`.

olivielpeau:
@zachradtka Went through the check and made a few comments, tell me if anything needs to be clarified. Thanks!

checks.d/mapreduce.py
zachradtka: So regardless of service, you want to always send `scheme://hostname:port`? 

For example if we have two tests that are `http://localhost:8008/test1` and `http://localhost:8008/test2` you just want `http://localhost:8008`

zachradtka:
@olivielpeau I made the changes, squashed my changes, rebased with master and pushed. Let me know if there is anything else I can do.

Thanks for the helpful suggestions!

checks.d/mapreduce.py
```diff
@@ -0,0 +1,512 @@
...
+        # Get the cluster ID from Yarn
+        cluster_id = self._get_cluster_id(rm_address)
+
+        # Get the running MR applications from YARN
+        running_apps = self._get_running_app_ids(rm_address)
+
+        # Report success after gathering all metrics from ResourceManaager
+        self.service_check(YARN_SERVICE_CHECK,
+            AgentCheck.OK,
+            tags=['resourcemanager_url:%s' % rm_address],
```
olivielpeau: Given that we're using a `url` tag in the `_rest_request_to_json` method, we should also use `url` here to be consistent. (The rationale being that if we send a `CRITICAL` service check with a given service check name and tags on a check run, we want the `OK` service check that's sent when things go back to normal on the same service check name to have the same tags, so that scoping service checks with tags works as expected on the web UI).

checks.d/mapreduce.py
```diff
@@ -0,0 +1,512 @@
...
+        self._mapreduce_task_metrics(running_jobs, cluster_id)
+
+        # Report success after gathering all metrics from Application Master
+        if running_jobs:
+            job_id, metrics = running_jobs.items()[0]
+            am_address = self._get_url_base(metrics['tracking_url'])
+
+            self.service_check(MAPREDUCE_SERVICE_CHECK,
+                AgentCheck.OK,
+                tags=['applicationmanager_url:%s' % am_address],
```
olivielpeau: let's use a `url` tag here too instead of `applicationmanager_url`

tests/checks/mock/test_mapreduce.py
```diff
@@ -0,0 +1,231 @@
...
+                value=value,
+                tags=self.MAPREDUCE_REDUCE_TASK_METRIC_TAGS)
+
+        # Check the MapReduce job counter metrics
+        for metric, attributes in self.MAPREDUCE_JOB_COUNTER_METRIC_VALUES.iteritems():
+            tags = attributes['tags']
+            tags.extend(self.MAPREDUCE_JOB_COUNTER_METRIC_TAGS)
+            self.assertMetric(metric,
+                value=attributes['value'],
+                tags=tags)
```
olivielpeau: You can also add assertions on the service checks using `self.assertServiceCheckOK` for instance (see https://github.com/DataDog/dd-agent/blob/8e36e58ed13bdb735a6c75e2c7b21f8998457c6a/tests/checks/common.py#L484 for reference)

olivielpeau:
Added in 3 comments. Don't worry about the one on the service check assertions, it's a nice to have but not mandatory at all.

Thanks!

tests/checks/mock/test_mapreduce.py
zachradtka: I added tests for the status checks, but this comment was not removed.

zachradtka:
OK, all requirements have been addressed.

zachradtka:
Squashed and pushed!

olivielpeau:
Looks good, thanks @zachradtka! I'll merge once the CI passes.

olivielpeau:
Merging! :tada:

PR #2237, opened 2016-01-29
[mongo] oplog replicationInfo metrics

rhwlo:
# What’s this PR do?

This PR adds a few new metrics to dd-agent's mongo plugin, borrowed from [Mongo's db.getReplicationInfo()][0]:

* [`mongodb.oplog.totalsizemb`][1]
* [`mongodb.oplog.usedsizemb`][2]
* [`mongodb.oplog.timediffseconds`][3]

# Notes

The code for getting the metrics is a more-or-less direct translation from the Javascript `getReplicationInfo()` function from the mongo shell, with a few edits to exclude non-numerical fields. I also updated the metric names to include units where the mongo shell function didn’t.

[0]: https://docs.mongodb.org/v3.2/reference/method/db.getReplicationInfo/
[1]: https://docs.mongodb.org/v3.2/reference/method/db.getReplicationInfo/#db.getReplicationInfo.logSizeMB
[2]: https://docs.mongodb.org/v3.2/reference/method/db.getReplicationInfo/#db.getReplicationInfo.usedMB
[3]: https://docs.mongodb.org/v3.2/reference/method/db.getReplicationInfo/#db.getReplicationInfo.timeDiff

### DIFF
--- a/checks.d/mongo.py
+++ b/checks.d/mongo.py
@@ -141,6 +141,9 @@ class MongoDb(AgentCheck):
         "opcountersRepl.insert": RATE,
         "opcountersRepl.query": RATE,
         "opcountersRepl.update": RATE,
+        "oplog.totalSizeMB": GAUGE,
+        "oplog.usedSizeMB": GAUGE,
+        "oplog.timeDiffSeconds": GAUGE,
         "replSet.health": GAUGE,
         "replSet.replicationLag": GAUGE,
         "replSet.state": GAUGE,
@@ -482,6 +485,20 @@ class MongoDb(AgentCheck):
         Returns a dictionary that looks a lot like what's sent back by
         db.serverStatus()
         """
+
+        def total_seconds(td):
+            """
+            Returns total seconds of a timedelta in a way that's safe for
+            Python < 2.7
+            """
+            if hasattr(td, 'total_seconds'):
+                return td.total_seconds()
+            else:
+                return (
+                    lag.microseconds +
+                    (lag.seconds + lag.days * 24 * 3600) * 10**6
+                ) / 10.0**6
+
         if 'server' not in instance:
             raise Exception("Missing 'server' in mongo config")
 
@@ -623,14 +640,7 @@ class MongoDb(AgentCheck):
                 # If we have both we can compute a lag time
                 if current is not None and primary is not None:
                     lag = primary['optimeDate'] - current['optimeDate']
-                    # Python 2.7 has this built in, python < 2.7 don't...
-                    if hasattr(lag, 'total_seconds'):
-                        data['replicationLag'] = lag.total_seconds()
-                    else:
-                        data['replicationLag'] = (
-                            lag.microseconds +
-                            (lag.seconds + lag.days * 24 * 3600) * 10**6
-                        ) / 10.0**6
+                    data['replicationLag'] = total_seconds(lag)
 
                 if current is not None:
                     data['health'] = current['health']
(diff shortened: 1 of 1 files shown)


checks.d/mongo.py
```diff
@@ -730,3 +733,41 @@ def check(self, instance):
...
+
+                oplog_data['usedSizeMB'] = round(localdb.command("collstats", ol_collection_name)['size'] / 2.0 ** 20, 2)
+
+                op_asc_cursor = oplog.find().sort("_id", pymongo.ASCENDING).limit(1)
+                op_dsc_cursor = oplog.find().sort("_id", pymongo.DESCENDING).limit(1)
+
+                try:
+                    first_timestamp = op_asc_cursor[0]['ts'].as_datetime()
+                    last_timestamp = op_dsc_cursor[0]['ts'].as_datetime()
+                    oplog_data['timeDiffSeconds'] = (last_timestamp - first_timestamp).total_seconds()
```
olivielpeau: `total_seconds` only exists since python 2.7, and we want to support 2.6 too (see the Travis failure on 2.6 on the `mongodb` test).

Could you handle this case with something similar to what we do [here](https://github.com/DataDog/dd-agent/blob/9f022645960f48c6dfdd3f1430974a2cc3d5ee8d/checks.d/mongo.py#L610-L616)? (and handle it in a function if it makes sense)

olivielpeau:
Hi @rhwlo , thanks for your contribution!

I've had a quick look at your PR and only added one small comment for now, but we'll have a closer look soon. Thanks again!

checks.d/mongo.py
rhwlo: Would it make sense to you to break that out into a function of its own, e.g.

```
        def total_seconds(td):
            """
            Returns total seconds of a timedelta in a way that's safe for
            Python < 2.7
            """
            if hasattr(td, 'total_seconds'):
                return td.total_seconds()
            else:
                return (
                    lag.microseconds +
                    (lag.seconds + lag.days * 24 * 3600) * 10**6
                ) / 10.0**6
```
and then just call `data['replicationLag'] = total_seconds(lag)` and `oplog_data['timeDiffSeconds'] = total_seconds(last_timestamp - first_timestamp)`? Or would it be easier if I just reduplicated the `total_seconds` check from the `replicationLag` part?

olivielpeau: I'd go with breaking that out into a separate function, duplicating the code here wouldn't be justified IMO.

Thanks!

rhwlo:
@olivielpeau Thanks for the feedback — I’ve updated the PR to include breaking that out. It looks like there are still two failing tests in travis, but I’m not sure they’re related to the PR: [zookeeper flavor 3.3.6 looks like it’s having trouble with `psutil`](https://travis-ci.org/DataDog/dd-agent/jobs/106326708#L294), [even though it’s installed](https://travis-ci.org/DataDog/dd-agent/jobs/106326708#L405), and [network for Python2.6 is failing because of what look like a DNS issue](https://travis-ci.org/DataDog/dd-agent/jobs/106326626#L327).

rhwlo:
Friendly ping on this one. Is there anything more I need to do on my end, or is this pending for a later release?

yannmh:
Sorry for the late review @rhwlo.

We are planning a `5.7.4` release within the next 10 days. It will ship various improvements on the MongoDB check: thus, it'd be great to have your changes along them :).

I am rereading, checking/testing your changes. Overall it looks :+1: . 

Thanks again.

checks.d/mongo.py
```diff
@@ -748,3 +758,40 @@ def check(self, instance):
...
+
+        # Fetch information analogous to Mongo's db.getReplicationInfo()
+        localdb = cli['local']
+
+        oplog_data = {}
+
+        for ol_collection_name in ("oplog.rs", "oplog.$main", None):
+            ol_metadata = localdb.system.namespaces.find_one({"name": "local.%s" % ol_collection_name})
+            if ol_metadata is not None:
+                break
```
yannmh: Is there any reason why we can't simply do:
```python
for ol_collection_name in ("oplog.rs", "oplog.$main"):
  ol_metadata = ...
  if ol_metadata:
    break

if ol_metadata:
  ..
```
?

rhwlo:
Good catch — I’ve fixed the strangeness in that particular loop. A lot of this was as direct of a translation from `mongo`’s own JS as I could manage, so there was some odd un-pythonic-ness to weed out.

checks.d/mongo.py
```diff
@@ -141,6 +141,9 @@ class MongoDb(AgentCheck):
         "opcountersRepl.insert": RATE,
         "opcountersRepl.query": RATE,
         "opcountersRepl.update": RATE,
+        "oplog.totalSizeMB": GAUGE,
```
yannmh: Can we name it `oplog.logSizeMB` instead, to match  https://docs.mongodb.org/v3.2/reference/method/db.getReplicationInfo/ ?

checks.d/mongo.py
```diff
@@ -141,6 +141,9 @@ class MongoDb(AgentCheck):
...
+        "oplog.usedSizeMB": GAUGE,
+        "oplog.timeDiffSeconds": GAUGE,
```
yannmh: As for `oplog.logSizeMB`, can we name it `oplog.timeDiff` ? :bow:

checks.d/mongo.py
```diff
@@ -748,3 +758,41 @@ def check(self, instance):
...
+            if ol_metadata:
+                break
+
+        if ol_metadata:
+            try:
+                oplog_data['totalSizeMB'] = round(ol_metadata['options']['size'] / 2.0 ** 20, 2)
+
+                oplog = localdb[ol_collection_name]
+
+                oplog_data['usedSizeMB'] = round(localdb.command("collstats", ol_collection_name)['size'] / 2.0 ** 20, 2)
```
yannmh: Shall we catch potential `KeyError` here too ?

yannmh:
I just added a few nitpicks, other than that it's ready to be merged :ship:

yannmh:
Hey @rhwlo.

I rebased your PR, and added my two cents to address the couple nitpicks mentioned in this thread.
https://github.com/DataDog/dd-agent/pull/2429. Let's close this PR. Thanks again for the great additions.

PR #2241, opened 2016-02-02
integration(kong) Added Kong checks

shashiranjan84:
Kong is an open-source management layer for API. More details about Kong can be found at https://getkong.org/


JohnLZeller:
Hey @shashiranjan84, thanks for the PR!
We'll take a deeper look at this soon.

irabinovitch:
@shashiranjan84 Thank you for this check and the corresponding integration template.  Our team is reviewing and will be back in touch with next steps.

shashiranjan84:
Great, thanks.

shashiranjan84:
Any idea when it would be merged and released?

sinzone:
+1

irabinovitch:
@shashiranjan84 We merge PRs as we approach a release.  At the moment this has the 5.8 label on it.

shashiranjan84:
@JohnLZeller @irabinovitch  any update on the release date of version 5.8

checks.d/kong.py
```diff
@@ -0,0 +1,70 @@
...
+class Kong(AgentCheck):
+    """ collects metrics for Kong """
+    def check(self, instance):
+        metrics = self._fetch_data(instance)
+        for row in metrics:
+            try:
+                name, value, tags = row
+                self.gauge(name, value, tags)
+            except Exception:
+                self.log.error(u'Could not submit metric: %s' % repr(row))
```
degemer: Could you do something like this instead : ` self.log.error(u'Could not submit metric: %s', row)` ? Because of https://www.simonmweber.com/2014/11/24/python-logging-traps.html

checks.d/kong.py
```diff
@@ -0,0 +1,70 @@
...
+            response.raise_for_status()
+        except Exception:
+            self.service_check(service_check_name, AgentCheck.CRITICAL,
+                               tags=service_check_tags)
+            raise
+        else:
+            self.service_check(service_check_name, AgentCheck.OK,
+                               tags=service_check_tags)
+        if response.status_code != 200:
+            self.service_check(service_check_name, AgentCheck.CRITICAL,
```
degemer: If the status code is a 20x other than 200, two service checks are created, one OK and the other CRITICAL.

checks.d/kong.py
```diff
@@ -0,0 +1,70 @@
...
+        if 'kong_status_url' not in instance:
+            raise Exception('missing "kong_status_url" value')
+        tags = instance.get('tags', [])
+        url = instance.get('kong_status_url')
+
+        parsed_url = urlparse.urlparse(url)
+        host = parsed_url.hostname
+        port = parsed_url.port or 8001
+        service_check_name = 'kong.can_connect'
+        service_check_tags = ['host:%s' % host, 'port:%s' % port]
```
degemer: `host` is actually a reserved keyword : this tag is added by Datadog to specify the host (server running the agent) which sends the service checks. Could you use `kong_host` or something similar instead ?

tests/checks/integration/test_kong.py
```diff
@@ -0,0 +1,43 @@
...
+
+# 3p
+from nose.plugins.attrib import attr
+
+# project
+from tests.checks.common import load_check
+
+
+@attr(requires='kong')
+class TestKong(unittest.TestCase):
```
degemer: We actually created a class for the agent checks tests : https://github.com/DataDog/dd-agent/blob/master/tests/checks/common.py#L131

It should be easier to test and allows us to check the metrics coverage. (whether the agent receives all the metrics it should receive or not)

For a simple example : https://github.com/DataDog/dd-agent/blob/master/tests/checks/integration/test_activemq_xml.py
A more detailed one : https://github.com/DataDog/dd-agent/blob/master/tests/checks/integration/test_fluentd.py

degemer:
Thanks for your contribution @shashiranjan84 !

I added a few comments, it looks good otherwise (really nice setup for the integration tests).
Once they are addressed, we should be good to merge.

Let me know if you need more information !

shashiranjan84:
@degemer thanks, updated the code as you advised. Please review.

shashiranjan84:
too early, still need to update the test.

shashiranjan84:
@degemer I want to update Kong version in the Integration test, so checking if I have some time before next release. I am aiming to finish by tomorrow evening.

degemer:
@shashiranjan84 Yes, today or tomorrow is fine.

About the errors of the CI, the first one is a simple style issue on your Ruby code (https://travis-ci.org/DataDog/dd-agent/jobs/129063985#L319), and the other one is a metrics coverage issue. To fix it, it would be perfect if you could check the presence of all expected metrics and service checks. It's then way easier for us to test the new Kong versions.

The other changes look good !

shashiranjan84:
@degemer Please have a look, updated Kong version and test as you advised.

checks.d/kong.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def _fetch_data(self, instance):
+        if 'kong_status_url' not in instance:
+            raise Exception('missing "kong_status_url" value')
+        tags = instance.get('tags', [])
+        url = instance.get('kong_status_url')
+
+        parsed_url = urlparse.urlparse(url)
+        host = parsed_url.hostname
+        port = parsed_url.port or 8001
```
degemer: Shouldn't the default port be 80 in this case ?

checks.d/kong.py
```diff
@@ -0,0 +1,71 @@
...
+            tags = []
+        parsed = json.loads(raw)
+        metric_base = 'kong'
+        output = []
+        tagged_keys = ['database', 'server']
+        for key in tagged_keys:
+            metric_name = '%s.%s' % (metric_base, key)
+            metric_list = []
+            for tag_val, data in parsed.get(key).items():
+                tag = '%s.%s' % (metric_name, tag_val)
```
degemer: I didn't notice on my first review (sorry), but I think that what you call `metric_name` is actually more a `metric_prefix`, and the `tag` the actual `metric_name`.

checks.d/kong.py
```diff
@@ -0,0 +1,71 @@
...
+                metric_list.append((tag, data, tags + [tag]))
```
degemer: I don't really see why you're adding the metric name to the list of tags. Could you remove it ? (it is also in the tests)

degemer:
Thank you ! I tested one last time and noticed two small issues, could you fix them ?

checks.d/kong.py
shashiranjan84: Kong default admin port is 8001.

checks.d/kong.py
shashiranjan84: Your are right, wrong name choice. Updated.

checks.d/kong.py
shashiranjan84: updated, test file too.

shashiranjan84:
@degemer thanks, please review changes.

checks.d/kong.py
degemer: Indeed, but if a URL is provided without a specific port, wouldn't `requests` try the port 80 ? (Since this port is taken from the `kong_status_url` which is queried afterwards)

shashiranjan84: should be the parsed_url.port would be null in that case and port will be defaulted 8001?

degemer: I'm not sure I understand what you mean.
I think there are two cases : 
- `kong_status_url: http://something:port/status`, and `parsed_url.port` returns `port` as expected.
-  `kong_status_url: http://something/status`, and `parsed_url.port` returns `None`, and there is (I think) a small issue, because the current check tags with `kong_port: 8001` whereas `response = requests.get(url, headers=headers(self.agentConfig))` tries to connect on the port 80 (default HTTP port), contradicting the tag.

If you want to have a default for the `kong_port`, you could instead of asking for an URL in the config file (`kong.yaml.example`) ask for a `port`(set a default to 8001) and a `host` and then build the URL in the check `url = 'http://{}:{}/status'.format(host, port)`.
It would only be possible if the URL is always the same (the `/status` part).

It's really a nitpick (I guess every user of the check will specify the port), but please let me know what you think about it.

shashiranjan84: You are right it will fail while making the request. I have removed default port and expecting user to always provide the correct kong status url.

degemer:
Merged as 436d0fb9c10c22e7c1ed987639c86e652790d9d1.
Thanks again for your contribution and your reactivity !

shashiranjan84:
Thank you all your help @dspangen. I have also sent PR for IntegrationTemplate. Please review that too

https://github.com/DataDog/IntegrationTemplate/pull/11

degemer:
@shashiranjan84 Sure, I'll take have a look tomorrow.

irabinovitch:
@shashiranjan84 we're all set on the integration template. Once the 5.8 agent is released we'll be able to publish the tile and close out that PR.

sinzone:
@irabinovitch do you know when the 5.8 is going out?

irabinovitch:
This week.

PR #2242, opened 2016-02-02
[mysql] Improving replication status check

truthbk:
## Why
When we made the MySQL check revamp, this was probably the one point where I had noticed we were lacking a little bit - amongst other things because not all info is available everywhere (ie. the information on a slave node is different to the information available on `master`, and in turn, it'd be different in a system without replication enabled). This attempts to address all use cases.

## Extras
- Also squeezing in a fix to an issue regarding the hostkey that probably came up as a product of a badly resolved conflict.
- Fix to a bad service_check tag issue when the customers connect to their MySQL instance via a port and experience a connection failure. We were updating the tags _after_ the exception - which was naturally not working in the event of a failed check.

### DIFF
--- a/checks.d/mysql.py
+++ b/checks.d/mysql.py
@@ -260,6 +260,7 @@ SCHEMA_VARS = {
 
 REPLICA_VARS = {
     'Seconds_Behind_Master': ('mysql.replication.seconds_behind_master', GAUGE),
+    'Slaves_connected': ('mysql.replication.slaves_connected', COUNT),
 }
 
 SYNTHETIC_VARS = {
@@ -344,7 +345,6 @@ class MySql(AgentCheck):
         hostkey = self.host
         if self.mysql_sock:
             hostkey = "{0}:{1}".format(hostkey, self.mysql_sock)
-            return self.mysql_sock
         elif self.port:
             hostkey = "{0}:{1}".format(hostkey, self.port)
 
@@ -364,6 +364,10 @@ class MySql(AgentCheck):
             if defaults_file != '':
                 db = pymysql.connect(read_default_file=defaults_file, ssl=ssl)
             elif mysql_sock != '':
+                self.service_check_tags = [
+                    'server:{0}'.format(mysql_sock),
+                    'port:unix_socket'
+                ]
                 db = pymysql.connect(
                     unix_socket=mysql_sock,
                     user=user,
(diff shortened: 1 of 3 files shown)


checks.d/mysql.py
```diff
@@ -519,16 +520,25 @@ def _collect_metrics(self, host, db, tags, options, queries):
...
             # Get replica stats
             results.update(self._get_replica_stats(db))
+            results.update(self._get_slave_status(db, performance_schema_enabled))
             metrics.update(REPLICA_VARS)
 
             # get slave running form global status page
             slave_running_status = AgentCheck.UNKNOWN
             slave_running = self._collect_string('Slave_running', results)
+            binlog_running = results.get('Binlog_enabled', False)
+            # slaves will only be collected iff user has PROCESS privileges.
```
olivielpeau: Do we document this somewhere user-facing?

checks.d/mysql.py
```diff
@@ -810,9 +820,28 @@ def _get_replica_stats(self, db):
...
             return replica_results
 
+    def _get_slave_status(self, db, nonblocking=False):
+        with closing(db.cursor()) as cursor:
+            # querying threads instead of PROCESSLIST to avoid mutex impact on
+            # performance.
+            if nonblocking:
+                cursor.execute("SELECT THREAD_ID, NAME FROM performance_schema.threads WHERE NAME LIKE '%worker'")
+            else:
+                cursor.execute("SELECT * FROM INFORMATION_SCHEMA.PROCESSLIST WHERE COMMAND LIKE '%Binlog dump%'")
```
olivielpeau: Where should these be run from? The master or a slave?

Just want to make sure I understand exactly how we're choosing the status we send for the service check, both on the master and on a slave.

olivielpeau:
Just had 2 comments (more questions than comments actually), but the changes look good overall, thanks for improving this!

Also, could you rebase your PR on top of master to fix the rubocop issue that's making the `default` test fail?

checks.d/mysql.py
truthbk: I'm not sure, I can add this to the `YAML`.

checks.d/mysql.py
truthbk: The `slaves` should be handled successfully by the `get_replica_stats()`, because `SHOW SLAVE STATUS` should definitely come up non-empty in a working slave. In `get_replica_stats()` we also go over `SHOW MASTER STATUS`, which will let us know if binary logging is up and running (normally a sign of replication being OK). The final `_get_slave_status()`, is just one last failsafe check we can come up with to check for replication. I will have to put it in a try...except block so that if the user isn't configured with `PROCESS` credentials we don't crash and burn - your previous comment just made me realize this.

checks.d/mysql.py
olivielpeau: :+1: for the tile

checks.d/mysql.py
```diff
@@ -1155,21 +1200,26 @@ def _query_size_per_schema(self, db):
...
-                schema_name = str(row[0])
-                size = long(row[1])
+                    # set the tag as the dictionary key
+                    schema_size["schema:{0}".format(schema_name)] = size
 
-                # set the tag as the dictionary key
-                schema_size["schema:{0}".format(schema_name)] = size
+                return schema_size
+        except Exception as e:
+            self.log.warning("Avg exec time performance metrics unavailable at this time: %s", e)
```
olivielpeau: in the message, you probably meant `Size per schema metrics unavailable...` instead of `Avg exec time ...`

nitpick: same thing in the description of the `_query_size_per_schema` method

checks.d/mysql.py
truthbk: Tile PR: https://github.com/DataDog/dogweb/pull/11501

checks.d/mysql.py
```diff
@@ -803,21 +812,54 @@ def _is_innodb_engine_enabled(self, db):
...
+                cursor.execute("SHOW SLAVE STATUS;")
+                replica_results = dict(cursor.fetchall())
+                cursor.execute("SHOW MASTER STATUS;")
+                binlog_results = dict(cursor.fetchall())
+                if binlog_results:
+                    replica_results.update({'Binlog_enabled': True})
+
+                return replica_results
+        except (pymysql.err.InternalError, pymysql.err.OperationalError) as e:
+            self.log.warning("Privileges error getting replication status (must grant REPLICATION CLIENT): %s", e)
```
olivielpeau: Could you use `self.warning` instead? On top of logging at the `WARNING` level it will add the message to the `info` page and also show up as an "integration error" on the web UI.

olivielpeau: This also applies to all your other usages of `self.log.warning` IMO (haven't checked in details though)

checks.d/mysql.py
```diff
@@ -781,7 +790,7 @@ def _get_stats_from_variables(self, db):
 
     def _get_binary_log_stats(self, db):
         with closing(db.cursor()) as cursor:
-            cursor.execute("SHOW MASTER LOGS;")
+            cursor.execute("SHOW BINARY LOGS;")
```
olivielpeau: Shouldn't there be the same raise/except logic for this query, in cases when the user doesn't have `REPLICATION CLIENT` rights? (cf https://dev.mysql.com/doc/refman/5.6/en/show-binary-logs.html)

olivielpeau:
Added in 2 comments, tell me what you think ;)

checks.d/mysql.py
truthbk: Nice one, yeah definitely.

checks.d/mysql.py
truthbk: Totally right, missed this one.

olivielpeau:
:+1: good job! Feel free to squash your commits and :shipit:

olivielpeau:
(CI failures are unrelated)

PR #2243, opened 2016-02-02
[riak] added search metrics

gzysk8:
Adds metrics related to Riak Search that are exposed through Riak's HTTP API via the `/stats` endpoint


remh:
Thanks @gzysk8 !

PR #2244, opened 2016-02-02
[mongo] tag mongo instances by replset state

rhwlo:
# What’s this PR do?

This PR adds `mongo_replset_state` as a tag to represent [the various replica states](https://docs.mongodb.org/manual/reference/replica-states/) of mongodb instances.

# Motivation

In some cases, we’d like to look at stats only for replica primaries. This gives us that ability!

### DIFF
--- a/checks.d/mongo.py
+++ b/checks.d/mongo.py
@@ -18,6 +18,23 @@ class MongoDb(AgentCheck):
     SERVICE_CHECK_NAME = 'mongodb.can_connect'
     SOURCE_TYPE_NAME = 'mongodb'
 
+    """
+    MongoDB replica set states, as documented at
+    https://docs.mongodb.org/manual/reference/replica-states/
+    """
+    REPLSET_STATES = {
+        0: 'startup',
+        1: 'primary',
+        2: 'secondary',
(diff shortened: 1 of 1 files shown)


checks.d/mongo.py
```diff
@@ -619,6 +636,7 @@ def check(self, instance):
                     data['health'] = current['health']
 
                 data['state'] = replSet['myState']
+                tags.append('mongo_replset_state:%s' % self.REPLSET_STATES[data['state']])
```
JohnLZeller: Given that this is the only check using this tag, it's unnecessary to specify this as `mongo_*`. Could we change this instead to `replset_state`?

JohnLZeller:
Hey @rhwlo, thanks for the PR!

These changes look good, other than the nit that I added as a comment. Would you please make that change, make sure to squash and change your commit message to `[mongo] Tagging instances by replica set status`?

rhwlo:
@JohnLZeller Thanks for the quick response! I’ve made the changes you suggested.

JohnLZeller:
Looks good, thanks @rhwlo! :shipit:

PR #2246, opened 2016-02-03
[snmp] fix situation when ont requested oid is prefix for another requested oid

xkrt:
I was faced with the following situation, my simplified conf:
```
metrics:
  - OID: 1.2.3
    name: first
  - OID: 1.2.3.4.5
    name: second
```

As you see OID of `first` is prefix of `second` OID. Without my bug fix both metrics get the value of `first` (since `second` have prefix `1.2.3`).

### DIFF
--- a/checks.d/snmp.py
+++ b/checks.d/snmp.py
@@ -323,7 +323,7 @@ class SnmpCheck(NetworkCheck):
     def report_raw_metrics(self, metrics, results, tags):
         '''
         For all the metrics that are specified as oid,
-        the conf oid is going to be a prefix of the oid sent back by the device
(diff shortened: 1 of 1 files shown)


yannmh:
That's certainly an issue, thanks a lot for fixing it @xkrt !
We'll review and merge your fix for the upcoming 5.7.0 release.

yannmh:
Looks great @xkrt. Would you mind squashing your commits so we can merge your contribution ?

Thanks again.

xkrt:
@yannmh done

yannmh:
Thanks. Let's merge it :ship:

PR #2248, opened 2016-02-05
[ci][rubocop] Pin version of rubocop

olivielpeau:
- pin version to 0.37 (ie latest)
- fix offenses found by that version of rubocop


olivielpeau:
CI passes. Merging.

PR #2249, opened 2016-02-05
[DNS check] First pass at DNS check

### LINKED ISSUE #2052: DNS query response monitoring
It would be great to be able to monitor DNS responses. We had issues where we randomly got SERVFAIL status responses from our DNS and as a result this host could not be resolved, and it caused some trouble. 

Would be great to have DNS monitoring bundled to datadog that would do following:

- You can define which host to query for
- You can define which DNS server to use
- Define timeout of the DNS query
- Store response time in milliseconds as gauge metric

This script would create Datadog event if:

- We get timeout
- Hostname is not found
- We get any other status than "NOERROR"

There is DataDog agent check script that someone has created that does almost everything we need, but some features are missing:
- You cannot define DNS server you use
- Response time should be stored in milliseconds (using seconds is not practical because most DNS responses on LAN are on range of 1-30 milliseconds)
- Doesn't validate the status of the response (SERVFAIL or NOERROR etc.) so our errors went unnoticed.

Here is the script for reference:

- dns.py: https://gist.github.com/srkiNZ84/ef5a099bc4787b698fd7
- dns.yaml: https://gist.github.com/srkiNZ84/7bed4a5bd41ac602ad72

hkaj:
fixes #2052 

linked to https://github.com/DataDog/omnibus-software/pull/38 and https://github.com/DataDog/dd-agent-omnibus/pull/63

### DIFF
--- a/checks.d/dns_check.py
+++ b/checks.d/dns_check.py
@@ -1,0 +1,70 @@
+# stdlib
+import time
+
+# 3p
+import dns.resolver
+
+# project
+from checks import AgentCheck
+
+
+class DNSCheck(AgentCheck):
+    SERVICE_CHECK_NAME = 'dns.can_resolve'
+    DEFAULT_TIMEOUT = 5
+
+    def __init__(self, name, init_config, agentConfig, instances=None):
+        AgentCheck.__init__(self, name, init_config, agentConfig, instances)
+        self.default_timeout = init_config.get('default_timeout', self.DEFAULT_TIMEOUT)
+
+    def check(self, instance):
+        if 'hostname' not in instance:
+            self.log.info("Skipping instance, no hostname found.")
+            return
+
+        timeout = float(instance.get('timeout', self.default_timeout))
+        self.query_dns(instance, timeout)
+
+    def query_dns(self, instance, timeout):
+        """Perform the DNS query, and report its duration as a gauge"""
+        start_time = end_time = 0.0
+        resolver = dns.resolver.Resolver()
+        resolver.lifetime = timeout
+
+        hostname = instance.get('hostname')
+        tags = self._get_tags(instance)
+
+        # if a specific DNS server was defined use it, else use the system default
+        nameserver = instance.get('nameserver')
+        if nameserver is not None:
+            resolver.nameservers = [nameserver]
+
+        status = AgentCheck.CRITICAL
+        start_time = time.time()
+        try:
+            self.log.debug('Resolving hostname %s...' % hostname)
+            answer = resolver.query(hostname)
+            assert(answer.rrset.items[0].address)
+            end_time = time.time()
+        except dns.exception.Timeout:
+            self.log.error('DNS resolution of %s timed out' % hostname)
+            self.service_check(self.SERVICE_CHECK_NAME, status, tags=self._get_tags(instance))
+            raise
+        except Exception:
+            self.log.exception('DNS resolution of %s has failed.' % hostname)
+            self.service_check(self.SERVICE_CHECK_NAME, status, tags=self._get_tags(instance))
+            raise
+        else:
+            if end_time - start_time > 0:
+                self.gauge('dns.response_time', end_time - start_time, tags=tags)
+                self.service_check(self.SERVICE_CHECK_NAME, AgentCheck.OK, tags=self._get_tags(instance))
+
+    def _get_tags(self, instance):
+        tags = []
+        try:
+            nameserver = instance.get('nameserver') or dns.resolver.Resolver().nameservers[0]
+            tags.append('nameserver:%s' % nameserver)
+        except IndexError:
+            self.log.error('No DNS server was found on this host.')
+
+        tags.append('resolved_hostname:%s' % instance.get('hostname'))
+        return tags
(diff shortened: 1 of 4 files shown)


checks.d/dns_check.py
```diff
@@ -0,0 +1,73 @@
...
+            self.log.error('DNS resolution of %s failed.' % hostname)
+            self.log.exception(ex)
+            self.service_check(self.SERVICE_CHECK_NAME,
+                               AgentCheck.CRITICAL, tags=self._get_tags(instance))
+            return -1
+
+        return end_time - start_time
+
+    def _get_tags(self, instance):
+        tags = set()
```
yannmh: Any reason to use a `set` rather than a `list` ?

checks.d/dns_check.py
```diff
@@ -0,0 +1,73 @@
...
+            assert(answer.rrset.items[0].address)
+            end_time = time.time()
+        except dns.exception.Timeout:
+            self.log.error('DNS resolution of %s timed out' % hostname)
+            self.service_check(self.SERVICE_CHECK_NAME,
+                               AgentCheck.CRITICAL, tags=self._get_tags(instance))
+            return -1
+        except Exception as ex:
+            self.log.error('DNS resolution of %s failed.' % hostname)
+            self.log.exception(ex)
```
yannmh: The two lines above can be replaced by:
```
self.log.exception('DNS resolution of %s failed.' % hostname)
```

`log.exception` already logs the stack trace.

checks.d/dns_check.py
```diff
@@ -0,0 +1,73 @@
...
+            self.service_check(self.SERVICE_CHECK_NAME,
+                               AgentCheck.CRITICAL, tags=self._get_tags(instance))
+            return -1
```
yannmh: I'd add a fancy `try, except, else, finaly` here

```python
stauts = CRITICAL
try:
  do_something()
except Exception1:
  log(...)
except Exception2:
  log(..)
else:
  self.gauge(...)
  status = OK
finally:
  self.service_check(...)
```

tests/checks/mock/test_dns_check.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def test_timeout(self):
+        configs = [
+            # short default timeout
+            {'init_config': {'default_timeout': 0.1},
+             'instances': [{'hostname': 'www.example.org'}]},
+            # short timeout
+            {'instances': [{'hostname': 'www.example.org', 'timeout': 0.1}]},
+        ]
+        with mock.patch.object(Resolver, 'query', side_effect=Timeout()):
```
yannmh: To be consistent with above
```
with mock.patch.object(Resolver, 'query', side_effect=timeout_query_mock):
```
instead.

tests/checks/mock/test_dns_check.py
```diff
@@ -0,0 +1,82 @@
...
+            # invalid hostname
+            {'instances': [{'hostname': 'example'}]},
+            # invalid nameserver
+            {'instances': [{'hostname': 'www.example.org', 'nameserver': '0.0.0.0'}]}
+        ]
+        for config in configs:
+            self.run_check(config)
+            self.assertEquals(len(self.metrics), 0)
+            self.assertServiceCheck(SERVICE_CHECK_NAME, status=AgentCheck.CRITICAL)
+        self.coverage_report()
```
yannmh: `self.coverage_report` would only report for the last `run_check` here. I am wondering if it makes sense to keep it, since you are already making strong assertions on metrics and service checks.

checks.d/dns_check.py
yannmh: One additional thought: `AgentCheck`s usually raise on top of sending a `CRITICAL` service check.

yannmh:
It looks great overall. Besides the nitpicks, please see my point about `CRITICAL` service checks and exceptions.

checks.d/dns_check.py
```diff
@@ -0,0 +1,73 @@
...
+        hostname = instance.get('hostname')
+
+        # if a specific DNS server was defined use it, else use the system default
+        nameserver = instance.get('nameserver')
+        if nameserver is not None:
+            resolver.nameservers = [nameserver]
+
+        start_time = time.time()
+        try:
+            answer = resolver.query(hostname)
```
yannmh: Let's add a log DEBUG line here, so it's easy to follow the check progress while debugging.

hkaj:
thanks for the feedback @yannmh. Added a few changes, let me know what you think.

checks.d/dns_check.py
```diff
@@ -0,0 +1,70 @@
...
+            assert(answer.rrset.items[0].address)
+            end_time = time.time()
+        except dns.exception.Timeout:
+            self.log.error('DNS resolution of %s timed out' % hostname)
+            self.service_check(self.SERVICE_CHECK_NAME, status, tags=self._get_tags(instance))
+            raise
+        except Exception:
+            self.log.exception('DNS resolution of %s has failed.' % hostname)
+            self.service_check(self.SERVICE_CHECK_NAME, status, tags=self._get_tags(instance))
+            raise
```
yannmh: Exception messages are used to set instance statuses displayed in the `info` command (c.f. https://github.com/DataDog/dd-agent/blob/master/checks/__init__.py#L761-L766). 

We could set one for this exception.

yannmh:
CI failures seem legit. Feel free to merge once it's solved.

PR #2260, opened 2016-02-11
[hdfs] Bugfix updated bean name in datanode check

zachradtka:
This is just a quick bugfix that updates bean name being requested from the DataNode's JMX interface.

### DIFF
--- a/checks.d/hdfs_datanode.py
+++ b/checks.d/hdfs_datanode.py
@@ -35,7 +35,7 @@ JMX_PATH = 'jmx'
 GAUGE = 'gauge'
 
 # HDFS bean name
-HDFS_DATANODE_BEAN_NAME = 'Hadoop:service=DataNode,name=FSDatasetState'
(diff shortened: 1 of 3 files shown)


checks.d/hdfs_datanode.py
```diff
@@ -76,12 +76,10 @@ def _hdfs_datanode_metrics(self, jmx_uri):
 
         if beans:
 
+            # Only get the first bean
             bean = next(iter(beans))
             bean_name = bean.get('name')
```
olivielpeau: Could you `self.log.debug` the retrieved `bean_name` here, so that we can easily troubleshoot issues with the retrieved bean?

olivielpeau:
Only has one small comment, once it's addressed the PR should be good to go, thanks!

checks.d/hdfs_datanode.py
irabinovitch: That should also fix the flake8 tests.

zachradtka:
OK, I squashed commit's and pushed.

olivielpeau:
Looks good, thanks!

Merging.

PR #2268, opened 2016-02-16
[snmp] optional mib constraint enforcement + resilience

truthbk:


checks.d/snmp.py
```diff
@@ -200,53 +202,60 @@ def check_table(self, instance, oids, lookup_names, timeout, retries):
...
+                        message = "{0} for instance {1}".format(error_status.prettyPrint(),
+                                                                instance["ip_address"])
+                        instance["service_check_error"] = message
+                        self.warning(message)
+
+                    for table_row in var_binds_table:
+                        complete_results.extend(table_row)
+
+                all_binds.extend(complete_results)
+            except PySnmpError as e:
```
yannmh: Do you when this can happen ?

checks.d/snmp.py
```diff
@@ -200,53 +202,60 @@ def check_table(self, instance, oids, lookup_names, timeout, retries):
...
+                self.warning("Fail to collect metrics: {0}".format(e))
```
yannmh: Nitpick refactor
```python
except PySnmpError as e:
   if ...:
     ...
   if ...:
    ...
   self.warning(instance["service_check_error"])
```

yannmh:
Nice changes. Thanks @truthbk.

Let's get it merge.

checks.d/snmp.py
truthbk: This could happen if we *do* actually enforce constraints and a particular OID does not match it, like an agent returning 0 for a value that should be 1..65535 - then that batch would fail.... but we could probably still try to collect metrics from the other batches.

PR #2271, opened 2016-02-17
[cassandra] Exclude aggregated metrics not associated with a specific keyspace

johnaxel:
`ColumnFamily` metrics are available two ways: as global aggregates and per keyspace/column family. This regex from @olivielpeau limits metric collection to those metrics associated with a specific keyspace to omit the global aggregates. This eliminates confusingly similar metric names (e.g. `cassandra.total_disk_space_used` and `cassandra.total_disk_space_used.count`) and prevents the double-counting of certain metrics.


olivielpeau:
Thanks @johnaxel!

PR #2272, opened 2016-02-18
[hdfs] Deprecate previous hdfs check

olivielpeau:
- add a deprecation notice on the old check
- add to the new check a metric (`capacity_in_use`) that was sent by the old check

### DIFF
--- a/checks.d/hdfs.py
+++ b/checks.d/hdfs.py
@@ -21,7 +21,11 @@ DEFAULT_PORT = 8020
 
 
 class HDFSCheck(AgentCheck):
-    """Report on free space and space used in HDFS.
+    """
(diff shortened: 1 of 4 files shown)


remh:
You could also add a `self.warning` similar to https://github.com/DataDog/dd-agent/blob/master/checks.d/docker.py#L134-L135

olivielpeau:
Makes sense, thanks, I've added the warning.

olivielpeau:
CI failures are unrelated, merging.

irabinovitch:
Do we have a ticket to track the spurious CI failures?

olivielpeau:
@irabinovitch No, but I agree that it'd be good to have one, thanks. Could you create one?

PR #2273, opened 2016-02-18
[Config] Update config logic to look for checks in 3rd-party directory

tmichelet:
- Add tests on config's logic
- Refactor logic
- Update logic to look for checks in 3rd-party directory

Logic tested on staging

### DIFF
--- a/config.py
+++ b/config.py
@@ -30,7 +30,6 @@ from utils.subprocess_output import (
     SubprocessOutputEmptyError,
 )
 
-
 # CONSTANTS
 AGENT_VERSION = "5.8.0"
 DATADOG_CONF = "datadog.conf"
@@ -677,6 +676,18 @@ def get_checksd_path(osname=None):
     else:
         return _unix_checksd_path()
 
+def get_3rd_party_path(osname=None):
+    if not osname:
+        osname = get_os()
+    if osname in ['windows', 'mac']:
+        raise PathNotFound()
+
+    cur_path = os.path.dirname(os.path.realpath(__file__))
+    path = os.path.join(cur_path, '../3rd-party')
+    if os.path.exists(path):
+        return path
+    raise PathNotFound(path)
+
 
 def get_win32service_file(osname, filename):
     # This file is needed to log in the event viewer for windows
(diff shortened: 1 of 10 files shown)


config.py
```diff
@@ -742,157 +754,183 @@ def check_yaml(conf_path):
...
-            log.debug('Skipping check %s because it has already been loaded from another location', check)
-            continue
+    places = [lambda name: os.path.join(agentConfig['additional_checksd'], '%s.py' % name)]
 
-        # Let's see if there is a conf.d for this check
-        conf_path = os.path.join(confd_path, '%s.yaml' % check_name)
-        conf_exists = False
+    try:
+        thrird_party_path = get_3rd_party_path(osname)
+        places.append(lambda name: os.path.join(thrird_party_path, name, 'check.py'))
```
talwai: typo `third`

config.py
```diff
@@ -742,157 +754,183 @@ def check_yaml(conf_path):
...
+    except PathNotFound:
+        pass
```
talwai: Can we log something here for the missing path before continuing?

talwai:
For better coverage, how about a test for the case where the check class inherits from a subclass of `AgentCheck`? E.g. `class SnmpCheck(NetworkCheck)`

config.py
tmichelet: good question, maybe a debug log as windows and macos won't have this 3rd party folder and we don't want to spam them for that

config.py
```diff
@@ -742,157 +754,183 @@ def check_yaml(conf_path):
...
-                if AgentCheck in clsmember.__bases__:
-                    continue
-                else:
-                    break
+def _update_python_path(check_config):
+    # Add custom pythonpath(s) if available
+    if 'pythonpath' in check_config:
+        pythonpath = check_config['pythonpath']
+        if not isinstance(pythonpath, list):
+            pythonpath = [pythonpath]
```
talwai: Does it make sense to do some stricter checks here? For example `[os.path.exists(path) for path in pythonpath]`? You probably know better whether that case should fail the initialization, print a warning, or do nothing at all

config.py
```diff
@@ -742,157 +754,183 @@ def check_yaml(conf_path):
...
 
-        # Look for the per-check config, which *must* exist
-        if not check_config.get('instances'):
-            log.error("Config %s is missing 'instances'" % conf_path)
+    all_configs_paths = _all_configs_paths(osname, agentConfig)
+
+    checks_places = _checks_places(agentConfig, osname)
+
+    for config_path in itertools.chain(*all_configs_paths):
+        check_name = config_path.rsplit('.', 1)[0].rsplit('/', 1)[-1]
```
talwai: comment here with an example of the transformation would be nice

config.py
```diff
@@ -742,157 +754,183 @@ def check_yaml(conf_path):
...
-        # Add custom pythonpath(s) if available
-        if 'pythonpath' in check_config:
-            pythonpath = check_config['pythonpath']
-            if not isinstance(pythonpath, list):
-                pythonpath = [pythonpath]
-            sys.path.extend(pythonpath)
+            _update_python_path(check_config)
 
-        log.debug('Loaded check.d/%s.py' % check_name)
+            log.debug('Loaded check.d/%s.py' % check_name)
```
talwai: Should this be `'Loaded %s/%s.py' % (check_path, check_name)` ?

tests/core/test_config.py
```diff
@@ -110,3 +111,118 @@ def testDefaultChecks(self):
...
+            '%s/test_check.py' % self.TEMP_ETC_CHECKS_DIR)
+        checks = load_check_directory({"additional_checksd": self.TEMP_ETC_CHECKS_DIR}, "foo")
+        self.assertEquals(1, len(checks['initialized_checks']))
+        self.assertEquals('valid_check_1', checks['initialized_checks'][0].check(None))
+
+    def tearDown(self):
+        import config as Config
+        Config.get_checksd_path = self.patched_get_checksd_path
+        Config.get_confd_path = self.patched_get_confd_path
+        Config.get_3rd_party_path = self.patched_get_3rd_party_path
```
talwai: This pattern seems weird. Is it possible to use a `MockConfig` object with the dummy methods, and then decorate at the class-level: 
```
@patch.object(config, 'Config', MockConfig)
class TestConfigLoadCheckDirectory(unittest.TestCase):
```

talwai:
Nice cleanup, long overdue! Left a few comments

config.py
tmichelet: I'd rather keep it safe leave the current logic untouched if you don't mind

tmichelet:
@talwai thanks for the review, took your comments into account

talwai:
looks great :+1:

tmichelet:
Added a test on the deprecated nagios logic, and fix the logic for this case

irabinovitch:
@tmichelet is this intended for 5.7?

tmichelet:
no @irabinovitch we'll release it as part of 5.8

config.py
```diff
@@ -742,157 +754,184 @@ def check_yaml(conf_path):
...
+    try:
+        check_config = check_yaml(config_path)
+    except Exception, e:
+        log.exception("Unable to parse yaml config in %s" % config_path)
+        traceback_message = traceback.format_exc()
+        return False, None, {check_name: {'error': str(e), 'traceback': traceback_message}}
+    # Look for the per-check config, which *must* exist
+    if not check_config.get('instances'):
+        log.error("Config %s is missing 'instances'" % config_path)
+        return False, None, {}
```
olivielpeau: Isn't this check already taken care of by the `check_yaml` function?

tests/core/test_config.py
```diff
@@ -110,3 +112,126 @@ def testDefaultChecks(self):
 
         for c in DEFAULT_CHECKS:
             self.assertTrue(c in init_checks_names)
+
+
+TEMP_3RD_PARTY_CHECKS_DIR = '/tmp/dd-agent-tests/3rd-party'
+TEMP_ETC_CHECKS_DIR = '/tmp/dd-agent-tests/etc/checks.d'
+TEMP_ETC_CONF_DIR = '/tmp/dd-agent-tests/etc/conf.d'
+TEMP_AGENT_CHECK_DIR = '/tmp/dd-agent-tests'
+FIXTURE_PATH = 'tests/core/fixtures/checks'
```
olivielpeau: to be more flexible we could replace this with: `FIXTURE_PATH =  os.path.join(os.path.dirname(os.path.realpath(__file__)), 'fixtures', 'checks')`

tests/core/test_config.py
```diff
@@ -110,3 +112,126 @@ def testDefaultChecks(self):
 
         for c in DEFAULT_CHECKS:
             self.assertTrue(c in init_checks_names)
+
+
+TEMP_3RD_PARTY_CHECKS_DIR = '/tmp/dd-agent-tests/3rd-party'
```
olivielpeau: to be more flexible we could use `tempfile.gettempdir()` instead of `/tmp`

config.py
```diff
@@ -742,157 +754,184 @@ def check_yaml(conf_path):
...
+    deprecated_checks = {}
+    agentConfig['checksd_hostname'] = hostname
+    osname = get_os()
+
+    deprecated_checks.update(_deprecated_configs(agentConfig))
 
-        # Look for the per-check config, which *must* exist
-        if not check_config.get('instances'):
-            log.error("Config %s is missing 'instances'" % conf_path)
+    all_configs_paths = _all_configs_paths(osname, agentConfig)
```
olivielpeau: Not sure it makes sense for `all_config_paths` to be an list of lists instead of simply a list.

olivielpeau:
There's one thing that's not handled anymore if I read through the code correctly: the `.yaml.default` files should also be taken into account (but should be ignored when a `.yaml` file is present for the same check).

For instance we use default confs for all the checks that are enabled by default (`ntp`, `disk`, `network` and `agent_stats` if I remember correctly).

Tests that should pass:

```python

    def testConfigDefault(self, *args):
        copyfile('%s/valid_conf.yaml' % FIXTURE_PATH,
            '%s/test_check.yaml.default' % TEMP_ETC_CONF_DIR)
        copyfile('%s/valid_check_1.py' % FIXTURE_PATH,
            '%s/test_check.py' % TEMP_ETC_CHECKS_DIR)
        checks = load_check_directory({"additional_checksd": TEMP_ETC_CHECKS_DIR}, "foo")
        self.assertEquals(1, len(checks['initialized_checks']))

    def testConfigCustomOverDefault(self, *args):
        copyfile('%s/valid_conf.yaml' % FIXTURE_PATH,
            '%s/test_check.yaml' % TEMP_ETC_CONF_DIR)
        copyfile('%s/valid_conf_2.yaml' % FIXTURE_PATH,  # a 2nd valid conf file, slightly different so that we can test which one has been picked up (with 2 instances for instance)
            '%s/test_check.yaml.default' % TEMP_ETC_CONF_DIR)
        copyfile('%s/valid_check_1.py' % FIXTURE_PATH,
            '%s/test_check.py' % TEMP_ETC_CHECKS_DIR)
        checks = load_check_directory({"additional_checksd": TEMP_ETC_CHECKS_DIR}, "foo")
        self.assertEquals(1, len(checks['initialized_checks']))
        self.assertEquals(1, checks['initialized_checks'][0].instance_count())  # check that we picked the right conf
```

olivielpeau:
Oh and FYI flake8 is complaining about `tests/core/fixtures/checks/invalid_check_2.py:1:1: F401 'nothing' imported but unused`, a `# noqa` should fix it

tmichelet:
@olivielpeau rebased on master, and added last two commits, could you have a second look please?

tests/core/test_config.py
```diff
@@ -110,3 +113,149 @@ def testDefaultChecks(self):
 
         for c in DEFAULT_CHECKS:
             self.assertTrue(c in init_checks_names)
+
+
+TMP_DIR = tempfile.gettempdir()
+TEMP_3RD_PARTY_CHECKS_DIR = '%s/dd-agent-tests/3rd-party' % TMP_DIR
```
olivielpeau: `os.path.join(TMP_DIR, 'dd-agent-tests', '3rd-party')` would be slightly better here

(same thing for the 3 lines below)

olivielpeau:
One last nitpick, other than that feel free to squash and merge!

tmichelet:
@olivielpeau updated and squashed, can be merged I think

remh:
Needs a rebase

olivielpeau:
Closing in favor of #2467

PR #2275, opened 2016-02-19
DataDog agent check for apt package updates.

ejholmes:
This is a custom agent check that will generate metrics and service alerts for apt package updates. It uses the `/var/lib/update-notifier/updates-available` file to determine what package updates are available (see http://askubuntu.com/questions/49958/how-to-find-the-number-of-packages-needing-update-from-the-command-line).

If there are security related updates available, the `apt.updates` service check goes critical. If there are non-security related package updates, the `apt.updates` check warns.

### DIFF
--- a/checks.d/apt.py
+++ b/checks.d/apt.py
@@ -1,5 +1,4 @@
 # stdlib
-import requests
 import re
 
 # project
@@ -33,7 +32,7 @@ class APT(AgentCheck):
             self.service_check(SECURITY_CHECK, AgentCheck.OK)
 
     def updates(self, instance):
-        updates = { 'packages': 0, 'security': 0 }
+        updates = {'packages': 0, 'security': 0}
 
         with open(instance['updates_file'], 'r') as fd:
             content = fd.read()
--- a/checks.d/apt.py
+++ b/checks.d/apt.py
@@ -1,0 +1,44 @@
+# stdlib
+import requests
+import re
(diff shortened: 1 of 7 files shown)


olivielpeau:
Hi @ejholmes, thanks for your contribution! At first glance your PR looks good, but we'll be sure to spend more time reviewing it and testing it in the following weeks.

rabbitfang:
I believe the MOTD file (`/var/lib/update-notifier/updates-available`) is only updated when a user logs in (e.g. over SSH), so if you have a server that you tend to leave alone, the stats of available updates will never change. The `/usr/lib/update-notifier/update-motd-updates-available` script is what actually generates the text stored in `/var/lib/update-notifier/updates-available`, so calling (in an asynchronous fashion, every so often) that might be a better option.

directionless:
I think using this requires `update-notifier`, which claims to be a damon, and thus feels very heavy. Thoughts about things like `apt-get -s dist-upgrade | grep upgraded` or `apt-get -s dist-upgrade | grep ^Inst | wc -l`

truthbk:
Hi @ejholmes, since the transition to the SDK has begun I have taken the liberty to open a PR with your work here: DataDog/integrations-extras#34. You authorship should be preserved.

This should hopefully ease the pain of transitioning to the SDK. Let me know if you have any questions.

PR #2276, opened 2016-02-19
Adding detection of Arista EOS.

sysbot:
This is to add support for installing dd-agent package on Arista EOS. This have
been tested to work correctly at Fastly but with a caveat that user `dd-agent`
on Arista EOS does not persist between reboot so one must take care of adding
that back in by using event-handle system in `running-config` for the Arista.

In addition, Fastly package dd-agent into it's own virtualenv and run on Arista
which isolate the Python environment from the rest of EOS.

### DIFF
--- a/packaging/datadog-agent/source/install_agent.sh
+++ b/packaging/datadog-agent/source/install_agent.sh
@@ -57,8 +57,8 @@ fi
 
 # OS/Distro Detection
 # Try lsb_release, fallback with /etc/issue then uname command
-KNOWN_DISTRIBUTION="(Debian|Ubuntu|RedHat|CentOS|openSUSE|Amazon)"
-DISTRIBUTION=$(lsb_release -d 2>/dev/null | grep -Eo $KNOWN_DISTRIBUTION  || grep -Eo $KNOWN_DISTRIBUTION /etc/issue 2>/dev/null || uname -s)
+KNOWN_DISTRIBUTION="(Debian|Ubuntu|RedHat|CentOS|openSUSE|Amazon|Arista)"
+DISTRIBUTION=$(lsb_release -d 2>/dev/null | grep -Eo $KNOWN_DISTRIBUTION  || grep -Eo $KNOWN_DISTRIBUTION /etc/issue 2>/dev/null || grep -Eo $KNOWN_DISTRIBUTION /etc/Eos-release 2>/dev/null || uname -s)
 
 if [ $DISTRIBUTION = "Darwin" ]; then
     printf "\033[31mThis script does not support installing on the Mac.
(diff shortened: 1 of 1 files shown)


irabinovitch:
Thanks @sysbot.  Was this the only change required to get up and running on Arista?

@remh thoughts on this one? It looks like the test failures are unrelated, so I'm re-triggering them.

packaging/datadog-agent/source/install_agent.sh
```diff
@@ -57,8 +57,8 @@ fi
 
 # OS/Distro Detection
 # Try lsb_release, fallback with /etc/issue then uname command
-KNOWN_DISTRIBUTION="(Debian|Ubuntu|RedHat|CentOS|openSUSE|Amazon)"
-DISTRIBUTION=$(lsb_release -d 2>/dev/null | grep -Eo $KNOWN_DISTRIBUTION  || grep -Eo $KNOWN_DISTRIBUTION /etc/issue 2>/dev/null || uname -s)
+KNOWN_DISTRIBUTION="(Debian|Ubuntu|RedHat|CentOS|openSUSE|Amazon|Arista)"
+DISTRIBUTION=$(lsb_release -d 2>/dev/null | grep -Eo $KNOWN_DISTRIBUTION  || grep -Eo $KNOWN_DISTRIBUTION /etc/issue 2>/dev/null || uname -s || grep -Eo $KNOWN_DISTRIBUTION /etc/Eos-release)
```
remh: Shouldn't your addition be done before `uname -s` ? Or is `uname -s` failing on Arista EOS?

sysbot: This is what I got 

```
Last login: Tue Mar 22 21:25:51 2016 from 10.0.2.2

Arista Networks EOS shell

-bash-4.1# uname -s
Linux
```

sysbot: I can move it before `uname -s` if that's prefer.

remh: You should, otherwise `grep -Eo $KNOWN_DISTRIBUTION /etc/Eos-release` will never be executed.

packaging/datadog-agent/source/install_agent.sh
```diff
@@ -57,8 +57,8 @@ fi
 
 # OS/Distro Detection
 # Try lsb_release, fallback with /etc/issue then uname command
-KNOWN_DISTRIBUTION="(Debian|Ubuntu|RedHat|CentOS|openSUSE|Amazon)"
-DISTRIBUTION=$(lsb_release -d 2>/dev/null | grep -Eo $KNOWN_DISTRIBUTION  || grep -Eo $KNOWN_DISTRIBUTION /etc/issue 2>/dev/null || uname -s)
+KNOWN_DISTRIBUTION="(Debian|Ubuntu|RedHat|CentOS|openSUSE|Amazon|Arista)"
+DISTRIBUTION=$(lsb_release -d 2>/dev/null | grep -Eo $KNOWN_DISTRIBUTION  || grep -Eo $KNOWN_DISTRIBUTION /etc/issue 2>/dev/null || grep -Eo $KNOWN_DISTRIBUTION /etc/Eos-release || uname -s)
```
remh: Also you should redirect stderr to /dev/null to not pollute the console for non Arista users.

remh:
Thanks, added one another comment.
Could you squash your commits into 1 as well please ?
Thanks a lot!

remh:
Thanks !

PR #2277, opened 2016-02-22
[kubernetes] add kubernetes.pods.running metric

masci:
Using the kubelet api, query the number of pods and group them by the Replication Controller that spawned them.
Report the number of pods spawned by each RC, tagging with `kube_replication_controller` and `node_name` thus allowing sums.

### DIFF
--- a/checks.d/kubernetes.py
+++ b/checks.d/kubernetes.py
@@ -5,6 +5,8 @@ Collects metrics from cAdvisor instance
 import numbers
 from fnmatch import fnmatch
 import re
+import simplejson as json
+from collections import defaultdict
 
 # 3rd party
 import requests
@@ -12,7 +14,7 @@ import requests
 # project
 from checks import AgentCheck
 from config import _is_affirmative
-from utils.kubeutil import set_kube_settings, get_kube_settings, get_kube_labels
+from utils.kubeutil import set_kube_settings, get_kube_settings, extract_kube_labels
 from utils.http import retrieve_json
 
 NAMESPACE = "kubernetes"
@@ -39,6 +41,7 @@ FUNC_MAP = {
     GAUGE: {True: HISTO, False: GAUGE},
     RATE: {True: HISTORATE, False: RATE}
 }
+
 
 class Kubernetes(AgentCheck):
     """ Collect metrics and events from kubelet """
@@ -104,7 +107,6 @@ class Kubernetes(AgentCheck):
             self.service_check(service_check_name, AgentCheck.CRITICAL, message=str(e))
             self.log.warning('master checks url=%s exception=%s' % (url, str(e)))
             raise
-
 
     def check(self, instance):
         kube_settings = get_kube_settings()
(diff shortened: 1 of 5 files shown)


checks.d/kubernetes.py
```diff
@@ -5,6 +5,8 @@
 import numbers
 from fnmatch import fnmatch
 import re
+import json
```
hkaj: we usually work with simplejson instead:
`import simplejson as json`

checks.d/kubernetes.py
```diff
@@ -220,6 +222,8 @@ def _retrieve_metrics(self, url):
     def _retrieve_kube_labels(self):
         return get_kube_labels()
 
+    def _retrieve_pods_list(self, url):
```
hkaj: Nitpicking but the name `_retrieve_pods_list` sounds like it will do the work itself, without needing to pass it a url. The signature should be `_retrieve_pods_list(self)`. See `_retrieve_kube_labels` that follows this style.

checks.d/kubernetes.py
```diff
@@ -233,3 +237,27 @@ def _update_metrics(self, instance, kube_settings):
...
                 self.log.error("Unable to collect metrics for container: {0} ({1}".format(
                     subcontainer.get('name'), e))
+
+        pods_list = self._retrieve_pods_list(kube_settings["pods_list_url"])
+        self._update_pods_metrics(instance, pods_list)
+
+    def _update_pods_metrics(self, instance, pods):
+        controllers_map = defaultdict(list)
+        for pod in pods['items']:
+            node_name = pod['spec']['nodeName']
```
hkaj: not sure why you use the node name, shouldn't we use the pod name instead? If it's the case you can get it from `pod['metadata']['name']`

checks.d/kubernetes.py
```diff
@@ -233,3 +237,27 @@ def _update_metrics(self, instance, kube_settings):
...
+                    controllers_map[created_by['reference']['name']].append(node_name)
+            except KeyError:
+                continue
+
+        tags = instance.get('tags', [])
+        for ctrl, pods in controllers_map.iteritems():
+            _tags = tags[:]  # copy base tags
+            _tags.append('kube_replication_controller:{0}'.format(ctrl))
+            # at the moment kubelet api reports data only for the current node,
+            # we expect exactly on item in the set.
```
hkaj: typo: `on` --> `one`

hkaj:
Great job @masci :clap:
@remh you might want to have a quick look too.

remh:
Great first PR @masci 
Looks like you are querying the same endpoint as the one used to collect the labels https://github.com/DataDog/dd-agent/blob/massi/pods_running_metric/utils/kubeutil.py#L42

Could we refactor that so that we don't query it twice for the same data ?

checks.d/kubernetes.py
```diff
@@ -233,3 +238,27 @@ def _update_metrics(self, instance, kube_settings):
...
+                continue
+
+        tags = instance.get('tags', [])
+        for ctrl, pods in controllers_map.iteritems():
+            _tags = tags[:]  # copy base tags
+            _tags.append('kube_replication_controller:{0}'.format(ctrl))
+            # at the moment kubelet api reports data only for the current node,
+            # we expect exactly one item in the set.
+            for node_name in set(pods):
+                _tags.append('node_name:{0}'.format(node_name))
```
remh: There is no need to tag by node_name. As you mentioned, the kubelet will just return the data for the current node, and we already have the host name passed automatically so you don't have to add this tag.

hkaj:
nice optimization, you can squash your commits and :ship: it. It would also be cool to rebase it with master to make sure the tests are passing.

PR #2278, opened 2016-02-22
[wmi] if we timeout return an empty list to avoid exceptions.

truthbk:


checks/system/win32.py
```diff
@@ -54,12 +54,12 @@ def check(self, agentConfig):
                 u"Timeout while querying Win32_PerfRawData_PerfOS_System WMI class."
                 u" Processes metrics will be returned at next iteration."
             )
-            return
+            return []
 
         if not (len(self.wmi_sampler)):
             self.logger.info('Missing Win32_PerfRawData_PerfOS_System WMI class.'
```
yannmh: Can we take advantage of this PR to change these log levels to `warning` instead of `info` ? What do you think ?

yannmh:
Nice catch @truthbk, thanks for the fix. Feel free to merge.

checks/system/win32.py
truthbk: Sure, I agree that warning is probably  more adequate for the len()==0 cases! Will change to warning and merge.

PR #2279, opened 2016-02-23
[gunicorn] use psutil 3 compatible children func

talwai:
This slipped by [when upgrading to psutil 3.3.0](https://github.com/DataDog/dd-agent/pull/2134) . `get_children` is no longer valid.


truthbk:
:+1: thanks for catching this guy!

PR #2280, opened 2016-02-23
[pgbouncer] fixed UnboundLocalError for results var

hjkatz:
Fixed an error when there were no results for a query run against a pgbouncer instance. In that case there was a check against a variable `results` that was not yet initialized. I added an initialization before the checks for `results = None`, thus fixing the problem.


olivielpeau:
Good catch @hjkatz, thanks!

PR #2289, opened 2016-02-24
[dns] fixing check for windows.

truthbk:


checks.d/dns_check.py
```diff
@@ -7,6 +8,10 @@
 # project
 from checks import AgentCheck
 
+if sys.platform == 'win32':
+    from dns.rdtypes.ANY import *  # noqa
+    from dns.rdtypes.IN import *  # noqa
+
```
yannmh: :question: :question: :question:

checks.d/dns_check.py
```diff
@@ -7,6 +8,10 @@
 # project
 from checks import AgentCheck
 
+if sys.platform == 'win32':
```
yannmh: Let's use https://github.com/DataDog/dd-agent/blob/master/utils/platform.py#L54-L56 instead.

yannmh:
Did you had a chance to the check on Windows @truthbk ? In particular i am wondering if the nameserver resolution fallback works as expected, c.f. https://github.com/DataDog/dd-agent/blob/master/checks.d/dns_check.py#L64

checks.d/dns_check.py
truthbk: If we don't do this an exception is thrown because on windows they dynamic type resolution fails. Don't ask -  https://github.com/rthalley/dnspython/issues/39

truthbk:
@yannmh I tested it normally, but not the fallback. Not yet.

hkaj:
Good catch @truthbk! Thanks for fixing this

checks.d/dns_check.py
```diff
@@ -6,6 +6,11 @@
 
 # project
 from checks import AgentCheck
+from util import Platform
+
+if Platform.is_win32():
+    from dns.rdtypes.ANY import *  # noqa
+    from dns.rdtypes.IN import *  # noqa
```
olivielpeau: Could you add a comment explaining what why we have to do this?

olivielpeau:
:+1:

PR #2290, opened 2016-02-25
Use key:value in example tags

c960657:
The [guide to tagging](http://docs.datadoghq.com/guides/tagging/) recommends that tags are in the key:value syntax. This should be reflected in the example datadog.conf file.


olivielpeau:
Makes sense, thanks @c960657!

PR #2291, opened 2016-02-25
[wmi] no warnings on 'Name' property 🔇

yannmh:
Do not log a warning about non digit property values for the 'Name'
property.
Strengthen unit tests: less race condition error-prone.


olivielpeau:
:+1:

PR #2292, opened 2016-02-25
[windows][system] enable uptime collection

yannmh:
Enable uptime metric collection on Windows system check.


olivielpeau:
:+1: Thanks for the fix, totally missed the issue when I reviewed the original PR.

The way the collector handles the system checks would deserve some refactoring, but it's outside the scope of this fix obviously.

remh:
thanks for catching that!

PR #2293, opened 2016-02-25
[go-metro] fix init script to detect failed optional processes

truthbk:


olivielpeau:
Tested on debian and found 2 edge cases, one minor and one major:

- minor: when starting the agent, and with a `go-metro` that can't start with a non-zero exit code, supervisor tries to restart it for about ~15 seconds, so the process doesn't go into the `FATAL` state in the 10 second start period (https://github.com/DataDog/dd-agent/blob/b0012926397dd9228e4f03ac33171eceb8cbcba4/packaging/debian/datadog-agent.init#L98-L109) and we report the start of the agent as being successful when we wouldn't want to.
- major: when the agent is running and `go-metro` is in a `FATAL` state, if we execute `/etc/init.d/datadog-agent start`, `check_status` returns `1` [here](https://github.com/DataDog/dd-agent/blob/b0012926397dd9228e4f03ac33171eceb8cbcba4/packaging/debian/datadog-agent.init#L79-L83) so we regard the agent as not running and proceed to start the agent [here](https://github.com/DataDog/dd-agent/blob/b0012926397dd9228e4f03ac33171eceb8cbcba4/packaging/debian/datadog-agent.init#L93), which apparently causes supervisord's PID file (`datadog-supervisord.pid`) to be deleted, so we lose track of the current supervisord and cannot control the agent anymore.

Let's discuss how we could solve the second issue. I think using `check_status` to decide whether the agent is running or not before proceeding to start it is the core of the issue here.

conf.d/go-metro.yaml.example
```diff
@@ -25,7 +25,7 @@ instances:
   hosts:                      # Whitelist by hostname - will perform ip lookup and filter by hosts ip's
     - somehost.somedomain.to  # <---SAMPLE HOST, SET YOUR OWN LIST.
 
-                              # NOTE: Whitelisting with ips and hosts is *highly* recommended to avoid having
+                              # NOTE: Whitelisting with ips and hosts is *MANDATAORY*. We want to avoid having
```
olivielpeau: typo: `MANDATORY` ;)

packaging/centos/datadog-agent.init
```diff
@@ -51,11 +62,35 @@ check_status() {
         s=`$SUPERVISORCTL_PATH -c $SUPERVISOR_CONF status`
 
         # number of RUNNING supervisord programs (ignoring dogstatsd, jmxfetch and go-metro)
-	p=`echo "$s" | grep -Ev 'dogstatsd|jmxfetch|go-metro' | grep -c RUNNING`
+        p=`echo "$s" | grep -Ev $DD_OPT_PROC_REGEX | grep -c RUNNING`
+
+        # Number of non-failed OPTIONAL supervisord programs (dogstatsd, jmxfetch and go-metro)
+        p_o=`echo "$s" | grep -E $DD_OPT_PROC_REGEX | grep -c FATAL`
```
olivielpeau: The 2nd `grep` should be `grep -cv FATAL`

truthbk: Good catch! Must've inadvertedly removed it in vim, because this was a paste from the `debian` version!

olivielpeau:
LGTM! Feel free to merge

PR #2296, opened 2016-02-25
[wmi] add `tag_by` to the properties to collect

yannmh:
Append `tag_by` property to the list of properties to collect.


checks.d/wmi_check.py
```diff
@@ -42,7 +42,7 @@ def check(self, instance):
 
         wmi_sampler = self._get_wmi_sampler(
             instance_key,
-            wmi_class, properties,
+            wmi_class, properties, tag_by,
```
olivielpeau: Nitpick: I prefer naming the parameter explicitely when it's an optional parameter, so `tag_by=tag_by` here :)

olivielpeau:
:+1:

PR #2297, opened 2016-02-26
[packaging][osx] Use osx-specific supervisor.conf

olivielpeau:
Now that go-metro is bundled only on Linux, we use a different
`supervisor.conf` file for OSX.

We could go back to a shared supervisor conf once we have go-metro in the OSX pkg and instructions on how to run it unprivileged.


truthbk:
:+1:

PR #2299, opened 2016-02-26
[system] ✂ uptime collection on windows & unix

yannmh:
Split `system.uptime` metric collection check on Windows and Unix
systems to comply with their respective system metric formats.


remh:
:+1:
you need to remove this import though https://github.com/DataDog/dd-agent/blob/yann/split-uptime-unix-win/checks/collector.py#L24

yannmh:
Addressed and merged. Thanks!

PR #2301, opened 2016-02-29
[system] fixed import of missing module

masci:
In a [recent refactory](https://github.com/DataDog/dd-agent/commit/162c3fab358aa3f02c5beec4bd4189c9a70b40fc) we got rid of  `checks.system.common` module, still `test_sysstat.py` was importing it, causing failures on CI.


yannmh:
Thanks for the fix @masci !

PR #2302, opened 2016-02-29
Added Percentage Free Memory Stats

yapyj:
An additional stat variable for us to monitor the percentage of Free Memory.

### DIFF
--- a/checks/system/unix.py
+++ b/checks/system/unix.py
@@ -369,6 +369,7 @@ class Memory(Check):
 
                 if memData['physTotal'] > 0:
                     memData['physPctUsable'] = float(memData['physUsable']) / float(memData['physTotal'])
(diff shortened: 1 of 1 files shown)


olivielpeau:
Hi @yapyj and thanks for this addition! We'll review it shortly.

FYI for this addition to work you'll also need to update this dict: https://github.com/DataDog/dd-agent/blob/5.7.2/checks/collector.py#L309, and on our end we'll have to add the metric to our backend.

remh:
Thanks @yapyj !
We are already collection the % of usable memory as a metric. I'm not convinced that the % of free memory would provide more value.
What do you think ?

irabinovitch:
@yapyj  I was hoping to follow up and get your thoughts regarding Remi's comments.

thederby:
for what it's worth, `1 - sum:system.mem.pct_usable` gives percentage of free memory.

PR #2303, opened 2016-02-29
[openstack] new cloud admin check structure

talwai:
The existing `openstack.py` check is prohibitively hard to configure for cloud administrators wishing to monitor several projects spanning several domains. These changes introduce a new structure for OpenStack monitoring as a cloud admin, using an easier to configure domain-based auth scheme. **The new scheme, and its prerequisites, are described in detail [here](https://gist.github.com/talwai/7a30dede9ee23b7750a9). This doc should eventually be absorbed into the main OpenStack documentation**

This new monitoring scheme involves configuring 2 checks:
- `openstack_admin` : collects metrics on project limits / quotas for all projects accessible by the configured admin domain. Can run on just a single host which has access to the keystone server
- `openstack_hyp`: collects hypervisor-level metrics, and metrics about the servers managed by a single hypervisor. Should be deployed on each compute host.

Currently there is substantial code duplication between the two checks, which should probably be pulled out into a util module of some sort. But when used together, they output the exam same metrics and tags as the original `openstack.py` check with the advantage of being far easier to configure.


irabinovitch:
CC @mtougeron as he likely has interest in this one.

mtougeron:
Love it! This will make things a lot easier for us. :)

olivielpeau:
Moving back to Triage. If need be the checks can be used as custom checks before they're shipped in a future release.

truthbk:
Closing, reconsider on the SDK side.

truthbk:
Hi @talwai ,

Thanks again for your contribution, always very much appreciated. We're currently transitioning our checks to a new SDK-based model for the agent. An unfortunate side-effect is that we have to impose a freeze on the dd-agent repo. The SDK consists of two repositories https://github.com/DataDog/integrations-core and https://github.com/DataDog/integrations-extras, where all of our checks will reside. This PR would belong in the core/extras repository. Would you be so kind as to re-open the PR there? We have facilitated some tools to help you with the SDK development that should help you get up and running in no time: 
- [sdk guide](http://docs.datadoghq.com/guides/integration_sdk/) 
- [datadog sdk gem]( https://github.com/DataDog/datadog-sdk-testing). 

PR #2308, opened 2016-03-01
[docker] Avoid event collection failure if an event has no ID param

hkaj:
Events that don't have an `id` field also don't have a `from` field so `event['id']` bits in `_format_events` don't need to be modified.


olivielpeau:
:+1:

PR #2311, opened 2016-03-02
[sqlserver] Be consistent with the connection string generation.

truthbk:


olivielpeau:
Let's fix the logic of `open_db_connections` before merging this.

Martiflex:
@truthbk FYI Reports of error with `custom_metrics` activated for this modified check. More info https://datadog.zendesk.com/agent/tickets/47559

checks.d/sqlserver.py
```diff
@@ -307,34 +307,42 @@ def close_cursor(self, cursor):
...
-                connection['conn'].close()
-            except Exception as e:
-                self.log.warning("Could not close adodbapi db connection\n{0}".format(e))
+        conn_key = self._conn_key(instance)
+        if conn_key not in self.connections:
+            return
 
-    def open_db_connections(self):
+        try:
+            self.connections[conn_key]['conn'].close
```
olivielpeau: `close` -> `close()`  (probably)

olivielpeau:
Looks good overall!

As an aside, I think we could make the whole connection handling logic way clearer if we used context managers, i.e.:
- a `get_connection(instance)` context manager that would return the connection for the instance and take care of opening and closing it when used with `with get_connection(instance) as connection:`
- a `get_cursor(connection)` that would be passed the instance's connection (since the current `get_cursor` only ever uses an instance's connection anyway) and return a cursor. `adodbapi`'s `cursor` can already be used as a context manager that closes the cursor automagically.

Probably outside the scope of this PR, but the logic is so convoluted right now that it'd be worth spending some time on it at some point.

truthbk:
Totally agree, we should definitely rewrite this.

checks.d/sqlserver.py
truthbk: indeedy... typo!

olivielpeau:
:+1: Feel free to squash and merge! :tada:

checks.d/sqlserver.py
```diff
@@ -307,34 +307,42 @@ def close_cursor(self, cursor):
...
+        timeout = int(instance.get('command_timeout',
+                                   self.DEFAULT_COMMAND_TIMEOUT))
+        try:
+            rawconn = adodbapi.connect(self._conn_string(instance=instance),
+                                    timeout=timeout)
+            if conn_key not in self.connections:
+                self.connections[conn_key] = {'conn': rawconn, 'timeout': timeout}
+            else:
+                # Existing Connection object will be closed (if open) when GC'd
+                self.connections[conn_key]['conn'] = rawconn
```
olivielpeau: I don't really know how adodbapi's connections behave exactly, but given that we used to make sure to always close the connections explicitely (and not wait for GC to take care of it), we should probably explicitely close the old connection here before replacing it with the new one.

olivielpeau:
Added a comment, just want to make extra-sure that we won't have stale open connections as we seem to have had trouble with this in the past.

checks.d/sqlserver.py
truthbk: They call close() on __del__(), but I guess we can make it explicit.

olivielpeau:
:+1:

Can you squash your commits? Feel free to merge once that's done.

PR #2313, opened 2016-03-02
[packaging][go-metro] Remove go-metro from the i386 packages

olivielpeau:
We use 2 different supervisor confs on x64 and i386. The init script
should be able to handle both w/o problems.

### DIFF
--- a/conf.d/go-metro.yaml.example
+++ b/conf.d/go-metro.yaml.example
@@ -13,9 +13,9 @@
 #        Also, please note that go-metro logs to its own file - found in /var/log/datadog/go-metro.log.
 #        Additionally, go-mtro runs standalone so it will *NOT* currently appear on the agent's info page.
 #
-#        Finally, because the go-metro binary is only bundled with the RPM and DEB distributions of the
+#        Finally, because the go-metro binary is only bundled with the 64-bit RPM and DEB distributions of the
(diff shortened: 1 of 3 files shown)


conf.d/go-metro.yaml.example
```diff
@@ -13,9 +13,9 @@
 #        Also, please note that go-metro logs to its own file - found in /var/log/datadog/go-metro.log.
 #        Additionally, go-mtro runs standalone so it will *NOT* currently appear on the agent's info page.
 #
-#        Finally, because the go-metro binary is only bundled with the RPM and DEB distributions of the
+#        Finally, because the go-metro binary is only bundled with the 64-bit RPM and DEB distributions of the
 #        datadog-agent, it will only be available in those packaged versions (ie. go-metro is currently
-#        not available with the source install).
+#        available neither with the source install nor with the 32-bit packages).
```
truthbk: maybe better with `currently unavailable with the source install or the 32-bit packages`...

packaging/supervisor_wo_metro.conf
```diff
@@ -0,0 +1,61 @@
+[supervisorctl]
```
truthbk: Maybe name the file `packaging/supervisor_32.conf` in case this sticks?

truthbk:
Just two nits you are free to ignore, otherwise it looks good.

PR #2317, opened 2016-03-02
[zk] `mntr` command's `zk_packets*` as rate

yannmh:
Report `mntr` command's `zk_packets*` information as rate, not gauge.


olivielpeau:
:+1: Let's wait for Travis to pass before merging

olivielpeau:
Added a small commit to fix the tests, Travis passes now (it's been a while).

Merging.

PR #2320, opened 2016-03-04
[core] set watchdog duration in seconds

yannmh:
Watchdog constructor takes `duration` parameter in seconds, not
milliseconds. Initialize the forwarder accordingly.


olivielpeau:
:+1: Thanks for fixing this!

CI failures are unrelated, feel free to merge

PR #2322, opened 2016-03-04
[core] detect forwarder pathological activity

yannmh:
**[util] watchdog high activity detection :fire:**

New option to the Watchdog to detect suspiciously high activity:
restarts the process when too many resets were made for a given
timeframe.

**[core] detect forwarder pathological activity**
On some specific system date and scheduled transaction flush time, the
Datadog Agent forwarder's schedule breaks and enters in a frenetic loop.
It gets unreachable, flushes (no transaction) at a very high rate
(several ~100 flushes/second). At this point, only a restart fixes the
situation.

Under the hood, the issue is related to `tornado==3.2.2`. More
information is available here:
https://github.com/tornadoweb/tornado/issues/947

Before considering any upgrade, a simple fix is to use the Watchdog to
detect this pathological high activity.

### DIFF
--- a/ddagent.py
+++ b/ddagent.py
@@ -65,8 +65,14 @@ log.setLevel(get_logging_config()['log_level'] or logging.INFO)
 
 DD_ENDPOINT = "dd_url"
 
+# Transactions
 TRANSACTION_FLUSH_INTERVAL = 5000  # Every 5 seconds
+
+# Watchdog settings
 WATCHDOG_INTERVAL_MULTIPLIER = 10  # 10x flush interval
+WATCHDOG_HIGH_ACTIVITY_THRESHOLD = 1000  # Threshold to detect pathological activity
+
+# Misc
 HEADERS_TO_REMOVE = [
     'Host',
     'Content-Length',
@@ -79,7 +85,7 @@ MAX_WAIT_FOR_REPLAY = timedelta(seconds=90)
 # Maximum queue size in bytes (when this is reached, old messages are dropped)
 MAX_QUEUE_SIZE = 30 * 1024 * 1024  # 30MB
 
-THROTTLING_DELAY = timedelta(microseconds=1000000/2)  # 2 msg/second
+THROTTLING_DELAY = timedelta(microseconds=1000000 / 2)  # 2 msg/second
 
 
 class EmitterThread(threading.Thread):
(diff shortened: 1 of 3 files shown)


ddagent.py
```diff
@@ -65,8 +65,14 @@
 
 DD_ENDPOINT = "dd_url"
 
-TRANSACTION_FLUSH_INTERVAL = 5000  # Every 5 seconds
+# Transactions
+TRANSACTION_FLUSH_INTERVAL = 5000
```
olivielpeau: let's keep the comment `# Every 5 seconds` here :)

util.py
```diff
@@ -469,20 +470,27 @@ def get_instance_id(agentConfig):
 
 
 class Watchdog(object):
-    """Simple signal-based watchdog that will scuttle the current process
-    if it has not been reset every N seconds, or if the processes exceeds
-    a specified memory threshold.
+    """
+    Simple signal-based watchdog. Restart the process when:
```
olivielpeau: typo `Restart` -> `Restarts` (sounds better to me)

util.py
```diff
@@ -500,14 +515,38 @@ def self_destruct(signum, frame):
         finally:
             os.kill(os.getpid(), signal.SIGKILL)
 
+    def is_frenetic(self):
```
olivielpeau: nitpick: this method is "private" so should rather be named `_is_frenetic`

util.py
```diff
@@ -469,20 +470,27 @@ def get_instance_id(agentConfig):
...
+
+    **Warning**: Not thread-safe.
     Can only be invoked once per process, so don't use with multiple threads.
     If you instantiate more than one, you're also asking for trouble.
     """
-    def __init__(self, duration, max_mem_mb = None):
+    # Activity history timeframe
+    _RESTART_TIMEFRAME = 60
+
+    def __init__(self, duration, max_mem_mb=None, max_resets=None):
```
olivielpeau: Super nitpicky, but in terms of interface I'd prefer the `max_resets` parameter to be a value per second, otherwise we don't really know what the value means exactly w/o reading the code.

olivielpeau:
:+1: for the clean fix!

Added a few nitpicky comments

olivielpeau:
Moving to 5.8 as it's a change in the core of the agent and not a regression.

util.py
yannmh: I don't agree: `max_resets` as a value per second is not a sufficient information to understand how the watchdog behaves. You'd need to know how this rate is computed, i.e. a timeframe length.

Considering this, having  `max_reset` as  a plain threshold versus a rate over a timeframe, is, in my opinion, easier to comprehend.

PR #2323, opened 2016-03-07
[kafka] changed default Kafka broker port

vagelim:
changed the default Kafka broker port listed in the configuration to the _actual_ default port

[skip ci]


remh:
Thanks!

PR #2328, opened 2016-03-08
SqlProc integration

rlaveycal:
This integration calls a stored procedure to return a custom set of metrics from the database. It is based on the sqlserver integration and so uses SQLOLDDB for database access. It is thus tied to Windows but I'm sure could be made cross-platform if a suitable API were used (ODBC?).

As well as running a proc the integration can be configured to use an ``` only_if ``` guard condition. This allows deployment to backup servers but for metrics to be captured only when they become active.

### DIFF
--- a/checks.d/sqlproc.py
+++ b/checks.d/sqlproc.py
@@ -1,0 +1,185 @@
+'''
+Call a SQL stored procedure that returns metric values
+
+The proc must return a dataset with columns:
+ - metric : the name of the metric
+ - type   : gauge / rate / histogram
+ - value  : the value to be logged
+ - tags   : comma seprated list of tags
+
+It is based on the SQL Server check
+
+This could be made cross platform if we had access to a different DB library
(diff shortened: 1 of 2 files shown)


olivielpeau:
Hi @rlaveycal, and thanks for this addition!

Given that each SQL check is very specific to the DBMS it integrates with, and that we don't have plans to have a common base for all the SQL checks I think this feature should be added directly to the existing SQLServer check.

We could then work on porting the feature to other DBMSes if there's demand for it.

What do you think?

irabinovitch:
@rlaveycal Hey! Thanks again for this contribution to the SQL integration.  Do you have any thoughts regarding moving this to the core SQL integration?

rlaveycal:
@irabinovitch Hi, your suggestion makes sense. Will try and find some time to refactor.

rlaveycal:
Superseded by #2445

PR #2329, opened 2016-03-08
[mysql] Fix check failure when no tag is provided

### LINKED ISSUE #2326: MySQL checker throws "'NoneType' object is not iterable" iterable without tags
If a host that the MySQL checker is running on has no tags configured, the checker defaults to a `tags` value of `None` here:

https://github.com/DataDog/dd-agent/blob/master/checks.d/mysql.py#L317

Eventually, when `tags` is used in a context that requires an iterable value here:

https://github.com/DataDog/dd-agent/blob/master/checks.d/mysql.py#L598

a `'NoneType' object is not iterable` exception is thrown.

hkaj:
Fixes #2326

### DIFF
--- a/checks.d/mysql.py
+++ b/checks.d/mysql.py
@@ -250,7 +250,7 @@ GALERA_VARS = {
 }
 
 PERFORMANCE_VARS = {
-    'query_run_time_avg' : ('mysql.performance.query_run_time.avg', GAUGE),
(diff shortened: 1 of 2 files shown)


remh:
:+1: let's merge when the CI passes

hkaj:
failing on something unrelated, merging.

PR #2339, opened 2016-03-10
[cassandra] exclude new system_schema keyspace

johnaxel:
Cassandra added a new `system_schema` keyspace in version 3.0. If not excluded from the JMX check, this keyspace reports ~130 additional JMX metrics to Datadog. Tested on Cassandra 3.0.4.


remh:
:+1: :tada: Thanks!

PR #2341, opened 2016-03-10
[SDK] Minor changes to allow testing of the SDK check modules

truthbk:
Modules will normally live in `integrations-core/foo` or `integrations-extra/bar` and a file typically called (by convention) `check.py`...

These changes would allow us to use the AgentCheckTest facilities with CircleCI and the SDK. We might have to fine-grain a little bit how we load the checks if `is_sdk=True`.

Should be backward compatible.


remh:
Looks like this needs a rebase and a squash

olivielpeau:
👍

PR #2345, opened 2016-03-11
[utils] Add a default timeout parameter to retrieve_json

hkaj:
Add a timeout to avoid hanging requests that get the collector killed.


remh:
:+1:

PR #2347, opened 2016-03-11
Add metrics from Linux's /proc/vmstat

gphat:
### What's this PR do?
Adds metrics from Linux' vm subsystem.

* `system.linux.vm` from `/proc/vmstat`
  * `pages.in` for `pgpgin`
  * `pages.out` for `pgpgout`
  * `pages.swapped_in` for `pswpin`
  * `pages.swapped_out` for `pswpout`
  * `pages.faults` for `pgfault`
  * `pages.major_faults` for `pgmajfault`

### Motivation

We have swap totals but we don't have swap _activity_.

### Notes

There are lots of metrics in here and likely some others are useful. I defer to others as to what we might want.

I'd love it if @b0rk, @nelhage and @ebroder looked at this and let me know if there were additions / removals to make from `/proc/vmstat`.

cc @rhwlo

### DIFF
--- a/checks.d/linux_vm_extras.py
+++ b/checks.d/linux_vm_extras.py
@@ -19,4 +19,4 @@ class MoreLinuxVMCheck(AgentCheck):
 
             for line in content:
                 if line[0] in VM_COUNTS:
-                    self.count("system.linux.vm.{0}".format(VM_COUNTS[line[0]]), int(line[1]), tags=tags)
+                    self.monotonic_count("system.linux.vm.{0}".format(VM_COUNTS[line[0]]), int(line[1]), tags=tags)
--- a/checks.d/linux_vm_extras.py
+++ b/checks.d/linux_vm_extras.py
@@ -19,4 +19,4 @@ class MoreLinuxVMCheck(AgentCheck):
 
             for line in content:
                 if line[0] in VM_COUNTS:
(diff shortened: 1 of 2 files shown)


gphat:
Oh, for whomever from DD looks this over: I tried to use `monotonic_counter` for this, as I think it's appropriate. Unfortunately it just didn't export any metrics and I'm unsure why. :smile_cat: Can try again.

nelhage:
This seems credible to me. I don't see anything else at a glance that I'd want out of vmstat (that we don't already have via another metric).

olivielpeau:
Thanks @gphat, we'll test and review your PR soon.

You're right, using a `monotonic_count` here would make more sense here. With a `monotonic_count` the check needs to run at least twice before it starts sending data, that might be the reason why you didn't see any metrics?

Could you try again and let us know? Happy to answer your questions if you have any :)

Thanks again!

gphat:
Adjusted to use `monotonic_count`, thanks @olivielpeau!

truthbk:
Shouldn't this maybe be part of the linux system check @olivielpeau?

olivielpeau:
@truthbk: As a first step we could leave this in a separate check IMO, if the check is used a lot we could move it to the system check later. What do you think?

Also, assigning the PR review to me

olivielpeau:
Superseded by #2767, which adds a mock test. It'll be released with the Integration SDK.

Closing, thanks again @gphat!

PR #2351, opened 2016-03-15
[collector.py] ugly fix for the tzname encoding issue

deterralba:
tzname has an encoding issue with Japanese versions of windows.
See https://trello.com/c/cEhSEKpW/1231-windows-agent-is-not-reporting-metrics-and-strange-character-shows-in-time-zone-part

### DIFF
--- a/checks/collector.py
+++ b/checks/collector.py
@@ -753,7 +753,7 @@ class Collector(object):
             pass
 
         metadata["hostname"] = self.hostname
-        metadata["timezones"] = time.tzname
+        metadata["timezones"] = sanitize_tzname(time.tzname)
 
         # Add cloud provider aliases
         host_aliases = GCE.get_host_aliases(self.agentConfig)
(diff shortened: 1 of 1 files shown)


checks/collector.py
```diff
@@ -773,3 +773,12 @@ def _should_send_additional_data(self, data_name):
             return True
 
         return False
+
+def tzname_check(tzname):
+    """ Returns the tzname given, and deals with Japanses encoding issue
```
remh: typo

checks/collector.py
```diff
@@ -773,3 +773,12 @@ def _should_send_additional_data(self, data_name):
             return True
 
         return False
+
+def tzname_check(tzname):
```
olivielpeau: Nitpick: we could use a more descriptive name like `sanitize_tzname`

olivielpeau:
Added a comment, apart from that it looks good to me! Could you squash your commits once you've addressed my comment?

deterralba:
@olivielpeau

olivielpeau:
:+1: Merging

PR #2357, opened 2016-03-16
[snmp] fixing thread-safety issues with pysnmp cmd_generator

truthbk:
The cmd_generator was a per check instance, unfortunately it is not thread-safe. To address the issue we simply create a `cmd_generator` per instance (we also cache them to avoid too much overhead impact due to the change).

### DIFF
--- a/checks.d/snmp.py
+++ b/checks.d/snmp.py
@@ -45,7 +45,6 @@ def reply_invalid(oid):
 class SnmpCheck(NetworkCheck):
 
     SOURCE_TYPE_NAME = 'system'
-    cmd_generator = None
     # pysnmp default values
     DEFAULT_RETRIES = 5
     DEFAULT_TIMEOUT = 1
@@ -54,8 +53,10 @@ class SnmpCheck(NetworkCheck):
     def __init__(self, name, init_config, agentConfig, instances):
         for instance in instances:
             if 'name' not in instance:
-                instance['name'] = instance['ip_address']
+                instance['name'] = self._get_instance_key(instance)
             instance['skip_event'] = True
+
+        self.generators = {}
 
         # Set OID batch size
         self.oid_batch_size = int(init_config.get("oid_batch_size", DEFAULT_OID_BATCH_SIZE))
@@ -68,9 +69,6 @@ class SnmpCheck(NetworkCheck):
             self.ignore_nonincreasing_oid = _is_affirmative(
                 init_config.get("ignore_nonincreasing_oid", False))
 
-        # Create SNMP command generator and aliases
-        self.create_command_generator(self.mibs_path, self.ignore_nonincreasing_oid)
-
         NetworkCheck.__init__(self, name, init_config, agentConfig, instances)
 
     def _load_conf(self, instance):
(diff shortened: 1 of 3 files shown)


truthbk:
@olivielpeau So one thing I still have to do in `_get_instance_key()` is handle the case where `ip_address` or `host` or whatever is set, but port isn't. Because there could be cases where port doesn't apply - and we still shouldn't return `None`. Will add that soon.

checks/network_checks.py
```diff
@@ -183,33 +187,57 @@ def _process_results(self):
...
-
-            # if an exception happened, log it
-            if instance_name in self.jobs_results:
-                ret = self.jobs_results[instance_name].get()
-                if isinstance(ret, Exception):
-                    self.log.exception("Exception in worker thread: {0}".format(ret))
-                del self.jobs_results[instance_name]
+            self._clean_job(instance_key)
+
+    def _get_instance_key(self, instance):
```
olivielpeau: Shouldn't the instance key generation be in `snmp.py` instead? Because we enforce the existence and uniqueness of the `name` attribute in the `__init__` method of `NetworkCheck` (https://github.com/DataDog/dd-agent/blob/jaime/snmpv3fix/checks/network_checks.py#L72-L78), so we'll always return the `name` of the instance here.

Whereas what would probably make more sense would be to generate a unique `name` there: https://github.com/DataDog/dd-agent/blob/jaime/snmpv3fix/checks.d/snmp.py#L55-L57
and then use the `name` as an identifier of the instance in the check.

The method of generation of the unique name can also depend on the specifics of each kind of network check IMO.

olivielpeau:
@truthbk Added a comment, let me know what you think. Basically I think that generating the unique `name`s in `SnmpCheck.__init__` and then using the name as the instance key throughout the check would make more sense.

Other than that the changes look good to me!

checks.d/snmp.py
```diff
@@ -81,8 +79,32 @@ def _load_conf(self, instance):
         retries = int(instance.get('retries', self.DEFAULT_RETRIES))
         enforce_constraints = _is_affirmative(instance.get('enforce_mib_constraints', True))
 
-        return ip_address, tags, metrics, timeout, retries, enforce_constraints
-
+        instance_key = self._get_instance_key(instance)
```
olivielpeau: let's use `instance['name']` instead here

checks.d/snmp.py
```diff
@@ -298,18 +330,19 @@ def _check(self, instance):
...
             if raw_oids:
                 self.log.debug("Querying device %s for %s oids", ip_address, len(raw_oids))
-                raw_results = self.check_table(instance, raw_oids, False, timeout, retries,
+                raw_results = self.check_table(instance, cmd_generator, raw_oids, False, timeout, retries,
                                                enforce_constraints=False)
                 self.report_raw_metrics(metrics, raw_results, tags)
         except Exception as e:
             if "service_check_error" not in instance:
-                instance["service_check_error"] = "Fail to collect metrics: {0}".format(e)
+                key = "{0}:{1}".format(instance['ip_address'], instance.get("port", 161))
```
olivielpeau: same here, `instance['name']` would work instead

olivielpeau:
Good to go apart from my 2 nitpicks :+1:

olivielpeau:
Looks good, feel free to squash your commits and :shipit:

PR #2366, opened 2016-03-17
[wmi] disable timeout and re-use connections

### LINKED ISSUE #2364: [core] Memory/handle leak in Windows agent
It seems there's a memory and handle leak in v5.7.1 of the Windows agent. Please see the attached Process Explorer screenshot. Note that the amount of memory and handles used were built up in less than 24 hours. The exact same thing happened in three different servers.
The vast majority of the handles are file handles for `c:\windows\system32\wbem\wbemdisp.tlb` which is used in WMI yet I am not using `wmi_check`.

I'm running v5.6.1 in two other servers and it's not displaying this behaviour. Memory usage is stable around 20MB with only a couple hunded handles.

I made a full dump of the ddagent.exe process, let me know if that's useful, it's 3.5GB.

datadog-leak

yannmh:
Disable WMI timeouts, cache and re-use connections.

Fix #2364

### DIFF
--- a/checks/libs/wmi/sampler.py
+++ b/checks/libs/wmi/sampler.py
@@ -19,18 +19,18 @@ Credits to @TheCloudlessSky (https://github.com/TheCloudlessSky)
 """
 
 # stdlib
+from collections import defaultdict
 from contextlib import contextmanager
 from copy import deepcopy
 from itertools import izip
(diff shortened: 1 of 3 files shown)


olivielpeau:
LGTM :+1:

truthbk:
:+1:

PR #2379, opened 2016-03-23
[IIS] Deal with BytesTransfered vs BytesTransferred 2008sp2 typo

truthbk:
This PR intends to address the crazy typo made in windows 2008 IIS (not sure if it's IIS or windows version specific).

Also updated the tests to cover this scenario.


checks.d/iis.py
```diff
@@ -87,6 +95,10 @@ def check(self, instance):
                     filters=filters, instance_tags=instance_tags
                 )
             )
+        except pythoncom.com_error:
```
olivielpeau: We could be a little bit more specific here and log a warning only when the error string contains the error code that we expect, i.e.:

```python
except pythoncom.com_error as e:
    if '0x80041017' in str(e):
        self.warning("blah blah")
    raise e
```

Also, we could use `self.warning` instead of `self.log.warning`.

Thoughts?

checks/libs/wmi/sampler.py
```diff
@@ -83,6 +83,9 @@ def __init__(self, logger, class_name, property_names, filters="", host="localho
                 #   - Frequency_PerfTime
                 #   - Frequency_Object"
             ])
+        elif self.is_raw_perf_class:
+            self.logger.debug(u"Warning: Attempting to run check on raw_perf_class with no properties(==\"*\").")
```
olivielpeau: let's remove this :)

checks.d/iis.py
```diff
@@ -57,6 +60,7 @@ def check(self, instance):
         password = instance.get('password', "")
         instance_tags = instance.get('tags', [])
         sites = instance.get('sites', ['_Total'])
+        is_2008 = _is_affirmative(instance.get('is_2008', False))
```
olivielpeau: let's add the parameter in the example yaml file and explain its use

checks.d/iis.py
```diff
@@ -6,6 +6,9 @@
 from checks.wmi_check import WinWMICheck, WMIMetric
 from utils.containers import hash_mutable
 from utils.timeout import TimeoutException
+from config import _is_affirmative
+
+import pythoncom
```
olivielpeau: nitpick: we usually import 3rd party modules before project modules

checks.d/iis.py
```diff
@@ -6,6 +6,9 @@
 from checks.wmi_check import WinWMICheck, WMIMetric
 from utils.containers import hash_mutable
 from utils.timeout import TimeoutException
+from config import _is_affirmative
```
olivielpeau: another nitpick: the import order :p

olivielpeau:
@truthbk Added in a few comments. Looks good overall!

olivielpeau:
Let's :shipit: !

olivielpeau:
(could you squash your commits first though)

olivielpeau:
(the CI failures are unrelated)

PR #2383, opened 2016-03-24
[consul] Make max number of catalog services configurable

antifuchs:
Currently, the consul check cuts off the list of catalogued services that it interrogates at 50, even when using a whitelist to pick out the most important ones. (Which is slightly frustrating if you have more than 50 important services!)

This PR adds a new consul.yaml configuration key which allows setting that limit to be higher, lower, or entirely non-existent (by setting it to be <1). 

The configuration key is documented in consul.yaml.example (and I've taken the liberty of adjusting the config key docs for the whitelist to not imply that the whitelist can be used query for more than 50 services).
## Testing

I have added a test and a new configuration key; tests pass locally, hope that Travis is happy too. We tried this out on a consul cluster running ~160 services, and it picks up the configured number of services there.

### DIFF
--- a/checks.d/consul.py
+++ b/checks.d/consul.py
@@ -20,7 +20,7 @@ class ConsulCheck(AgentCheck):
     SOURCE_TYPE_NAME = 'consul'
 
     MAX_CONFIG_TTL = 300 # seconds
-    MAX_SERVICES = 50 # cap on distinct Consul ServiceIDs to interrogate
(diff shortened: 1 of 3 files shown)


hkaj:
Hi @antifuchs and thanks for opening this PR! This seems like a legit use case.
We'll review and test this PR shortly.

olivielpeau:
It appears that this has been addressed in https://github.com/DataDog/integrations-core/pull/140 (didn't realize that the present PR existed when I merged the other one), so I'm going to close this PR.
Thanks @antifuchs!

PR #2385, opened 2016-03-25
[mysql] resilience when InnoDB is not available or disabled.

truthbk:


checks.d/mysql.py
```diff
@@ -410,7 +410,7 @@ def _collect_metrics(self, host, db, tags, options, queries):
         results = self._get_stats_from_status(db)
         results.update(self._get_stats_from_variables(db))
 
-        if self._is_innodb_engine_enabled(db):
+        if (not _is_affirmative(options.get('disable_innodb_metrics', False)) and self._is_innodb_engine_enabled(db)):
```
olivielpeau: let's document the option in the yaml example: https://github.com/DataDog/dd-agent/blob/jaime/innodbfix/conf.d/mysql.yaml.example#L19

checks.d/mysql.py
```diff
@@ -806,13 +806,18 @@ def _is_innodb_engine_enabled(self, db):
...
         # from the output of SHOW ENGINES or from information_schema.ENGINES
         # table. Later is choosen because that involves no string parsing.
-        with closing(db.cursor()) as cursor:
-            cursor.execute(
-                "select engine from information_schema.ENGINES where engine='InnoDB'")
+        try:
+            with closing(db.cursor()) as cursor:
+                cursor.execute(
+                    "select engine from information_schema.ENGINES where engine='InnoDB' and \
+                    support not like 'no' and support not like 'disabled'"
```
olivielpeau: is there an advantage to using `not like` instead of `<>` here? :)

olivielpeau:
:+1: Apart from my comments!

Feel free to squash and merge once they're addressed

checks.d/mysql.py
truthbk: @olivielpeau I'm not totally sure about the possible "Disabled", "DISABLED", "disabled" combinations, that's what I did it this way....

conf.d/mysql.yaml.example
```diff
@@ -17,7 +17,11 @@ instances:
     #   extra_innodb_metrics: true
     #   extra_performance_metrics: true
     #   schema_size_metrics: false
+    #   disable_innodb_metrics: false
     #
+    #     NOTE: skip_innodb_metrics should only be used by users with older (unsupported) versions of
```
olivielpeau: typo: `skip_innodb_metrics` -> `skip_innodb_metrics` :)

PR #2389, opened 2016-03-28
riak security support

hamano:
I'ts nessessary when enable riak security.

### DIFF
--- a/conf.d/riak.yaml.example
+++ b/conf.d/riak.yaml.example
@@ -1,4 +1,6 @@
 init_config:
+  # cacert: /path/to/ca.pem
+  # disable_cert_verify: true
 
 instances:
   - url: http://127.0.0.1:8098/stats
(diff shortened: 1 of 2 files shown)


remh:
Thanks a lot for this ! (and sorry for the delay :( )
I think those settings would make more sense at the instance level instead of the init config level. Would it be possible to move them please ?

irabinovitch:
We should review if this is still relevant given the work being done in #2762

hamano:
@remh moved settings.

conf.d/riak.yaml.example
```diff
@@ -2,6 +2,8 @@ init_config:
 
 instances:
   - url: http://127.0.0.1:8098/stats
+    # cacert: /path/to/ca.pem
+    # disable_cert_verify: true
```
hush-hush: Since the default value in the code is `False`, could you use `false` in the configuration example ?

hush-hush:
Hello @Hamano,

I have rebase you PR to allow CI to succeed and change the example in the configuration to merge it quickly upstream.

Thanks again for your contribution. If you have any comments feel free to add them to #3168.

PR #2390, opened 2016-03-28
[tokumx] Handle int64 and replica connections

JohnLZeller:
Ran into a situation where some data points coming from mongo are of type int64 (ie `0L`, or `2L`). We should take these into account and then coerce them to floats.

Additionally, the integration panel for TokuMX mentions specifying a full MongoDB URI, but the yaml shows no examples.

### DIFF
--- a/checks.d/tokumx.py
+++ b/checks.d/tokumx.py
@@ -3,6 +3,7 @@ import time
 import types
 
 # 3p
+import bson
 from pymongo import (
     MongoClient,
     ReadPreference,
@@ -251,12 +252,7 @@ class TokuMX(AgentCheck):
             'host': hostname
         })
 
-    def _get_connection(self, instance):
-        if 'server' not in instance:
-            raise Exception("Missing 'server' in tokumx config")
-
-        server = instance['server']
-
+    def _get_ssl_params(self, instance):
         ssl_params = {
             'ssl': instance.get('ssl', None),
             'ssl_keyfile': instance.get('ssl_keyfile', None),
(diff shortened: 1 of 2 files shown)


JohnLZeller:
@olivielpeau could you take a quick pass? it's small

JohnLZeller:
I just added an additional commit to this PR which fixes an issue around reading replica sets.

As of pymongo version 3.0, the `read_preference` attribute of `MongoClient` is read-only.
We set `read_preference` to secondary when reading replicas, causing an AttributeError.
This commit creates a new database connection instead.

Also, database connections are created in a pool with MongoClient, and are automatically closed up for you. So there should be no need to close the connection ourselves before opening this new one, as pymongo should reuse connections.

checks.d/tokumx.py
```diff
@@ -356,7 +361,15 @@ def _get_replica_metrics(self, conn, tags, server, status):
...
-                    conn.read_preference = ReadPreference.SECONDARY
+                    self.log.debug("Current replSet member is secondary. "
+                                   "Creating new connection to set read_preference to secondary.")
+                    # need a new connection to deal with replica sets
+                    ssl_params = self._get_ssl_params(instance)
+                    conn = MongoClient(server,
+                                       socketTimeoutMS=DEFAULT_TIMEOUT*1000,
+                                       read_preference=ReadPreference.SECONDARY,
+                                       **ssl_params)
+                    db = conn[db.name]
```
olivielpeau: Looks like we don't handle the authentication on this connection. We should maybe extract the actual connection creation logic from the `_get_connection` method and use it here.

checks.d/tokumx.py
```diff
@@ -453,7 +469,10 @@ def collect_metrics(self, server, conn, db, tags):
                     continue
 
                 # value is now status[x][y][z]
-                assert type(value) in (types.IntType, types.LongType, types.FloatType)
+                if type(value) == bson.int64.Int64:
+                    value = float(value)
```
olivielpeau: let's cast to a `long` instead

olivielpeau:
Thanks for the fix! Added in 2 comments.

JohnLZeller:
@olivielpeau okay this should be good

olivielpeau:
Looks good, feel free to squash and merge!

PR #2391, opened 2016-03-29
[haproxy] add 'active' tag to metrics

gabetax:
The HAProxy CSV output contains two boolean columns, "act" and "bck" to denote whether a backend belongs to the "active" pool of connections normally used for servicing requests, or the "[backup](http://cbonte.github.io/haproxy-dconv/configuration-1.7.html#5.2-backup)" pool of connections used only when none of the "active" backends are available.

When a primary HAProxy backend fails over and recovers, there still may be connections to the backup backend. Having this tag allows for monitoring how much traffic is flowing to the primary vs backup.

"active" as a term seems ambiguous, but it does follow HAProxy's own language, so I think it's preferential over introducing our own terminology.

### DIFF
--- a/checks.d/haproxy.py
+++ b/checks.d/haproxy.py
@@ -373,6 +373,9 @@ class HAProxy(AgentCheck):
         tags = ["type:%s" % back_or_front, "instance_url:%s" % url]
         tags.append("service:%s" % service_name)
 
+        if 'act' in data:
+            tags.append("active:%s" % ('true' if data['act'] else 'false'))
+
         if self._is_service_excl_filtered(service_name, services_incl_filter,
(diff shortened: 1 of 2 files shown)


olivielpeau:
Hi @gabetax, sorry for getting back to you so late.

We'll review your PR soon.

One concern we might have is that the check can already potentially create many unique tag combinations, so having this additional tag by default might be an issue. We'll get back to you once we've discussed this internally. Thanks for your patience!

masci:
Let's do this but keep the feature behind a configuration flag.

ChristineTChen:
Moved this PR to the [integrations-core repo](https://github.com/DataDog/integrations-core).

PR #2392, opened 2016-03-29
[wmi] Updated metric_type lettercase to account for uppercase Gauge o…

avaughan89:
…r Counter being entered

This problem originated from ticket 46655.  Trying to grab attribute type of `Gauge` instead of `gauge` results in the following error:

```
  File "checks\__init__.pyc", line 746, in run
  File "C:\Program Files (x86)\Datadog\Datadog Agent\files\..\checks.d\wmi_check.py", line 65, in check
  File "checks\wmi_check.pyc", line 219, in _submit_metrics
Exception: Invalid metric type: Gauge
```

And derives from attributes being set:

https://github.com/DataDog/dd-agent/blob/master/conf.d/wmi_check.yaml.example#L20-L30

### DIFF
--- a/checks/wmi_check.py
+++ b/checks/wmi_check.py
@@ -214,7 +214,7 @@ class WinWMICheck(AgentCheck):
 
             metric_name, metric_type = metric_name_and_type_by_property[metric.name]
             try:
-                func = getattr(self, metric_type)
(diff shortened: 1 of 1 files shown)


olivielpeau:
Thanks @avaughan89 !

Merging

PR #2394, opened 2016-03-30
[elasticsearch] collect force_merge threadpool stats for ES instances version 2.1.0 or higher

samcday:
Force merge API was added in 2.1.0 (it's just optimize, but renamed because renaming things is fun!).

We do a lot of insanely long-running force merges in our cluster. How long? I've been waiting for this 140GB merge to finish for more than 2 days now :( Some metrics on force merge thread pool would be useful.


truthbk:
@samcday I know it's taken forever to get back to you! These metrics were actually already added to the agent in a superceding PR https://github.com/DataDog/dd-agent/pull/2758! Sorry the update got delayed so drasitcally.

Thanks for your patience and your contribution!

PR #2400, opened 2016-04-01
[http_check] Removes deprecated parameters 'threshold' and 'window' from the yaml

Martiflex:
These parameters are not used anymore in the http_check.py code.

Since the introduction of status_check based monitors, users can control from the app UI after how many CRITICAL statuses they want to be alerted etc.

The presence of these 2 parameters is confusing.

### DIFF
--- a/conf.d/http_check.yaml.example
+++ b/conf.d/http_check.yaml.example
@@ -33,13 +33,6 @@ instances:
     # This defaults to 1xx, 2xx and 3xx HTTP status code: (1|2|3)\d\d.
     # http_response_status_code: 401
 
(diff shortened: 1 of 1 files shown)


olivielpeau:
@Martiflex Just to be clear, these parameters are still used when `skip_event` is not specified or is set to `False` (which is deprecated since `5.3.0`, see https://github.com/DataDog/dd-agent/blob/5.7.3/checks/network_checks.py#L183-L189)

That said these parameters should definitely be removed from the example yaml since they've been deprecated for quite a while now.

Merging, thanks!

PR #2401, opened 2016-04-01
[wmi] enable timeout and remove connection caching

olivielpeau:
Re-enable the timeout feature (that was reverted in #2366), after some tests.

Sharing COM objects between threads leads to handle/memory
leaks. Because of this, take the safe approach of having one connection
COM object per thread, and never keep reference to the connection once
the thread is terminated.

This could probably be improved on by having a pool of long-lived
threads that would have their own COM objects.

Also, add a test to check that no handle leaks happen on the core check.

### DIFF
--- a/checks/libs/wmi/sampler.py
+++ b/checks/libs/wmi/sampler.py
@@ -19,18 +19,17 @@ Credits to @TheCloudlessSky (https://github.com/TheCloudlessSky)
 """
 
 # stdlib
-from collections import defaultdict
-from contextlib import contextmanager
 from copy import deepcopy
 from itertools import izip
-from utils.timeout import TimeoutException
 import pywintypes
 
 # 3p
+import pythoncom
 from win32com.client import Dispatch
 
 # project
 from checks.libs.wmi.counter_type import get_calculator, get_raw, UndefinedCalculator
+from utils.timeout import timeout, TimeoutException
 
 
 class CaseInsensitiveDict(dict):
@@ -51,8 +50,6 @@ class WMISampler(object):
     """
     WMI Sampler.
     """
-    # Shared resources
-    _wmi_connections = defaultdict(list)
 
     def __init__(self, logger, class_name, property_names, filters="", host="localhost",
                  namespace="root\\cimv2", username="", password="", and_props=[], timeout_duration=10):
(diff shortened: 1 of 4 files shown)


checks/libs/wmi/sampler.py
```diff
@@ -263,45 +259,32 @@ def _format_property_values(self, previous, current):
...
+        # that we run each query in its own thread, we don't cache connections
+        pythoncom.CoInitialize()
+        locator = Dispatch("WbemScripting.SWbemLocator")
+        connection = locator.ConnectServer(
+            self.host, self.namespace,
+            self.username, self.password
+        )
 
         # Yield the connection
         yield connection
```
truthbk: Blows my mind that holding a reference to the `connection` that would then get overwritten in subsequent runs wouldn't be garbage collected. Good catch.

tests/core/test_win32.py
```diff
@@ -51,3 +65,16 @@ def testNetwork(self):
...
+        proc = w32.Processes(log)
+        mem = w32.Memory(log)
+        net = w32.Network(log)
+        cpu = w32.Cpu(log)
+
+        self._checkHandleLeak(io.check, AGENT_CONFIG)
+        self._checkHandleLeak(proc.check, AGENT_CONFIG)
+        self._checkHandleLeak(mem.check, AGENT_CONFIG)
+        self._checkHandleLeak(net.check, AGENT_CONFIG)
+        self._checkHandleLeak(cpu.check, AGENT_CONFIG)
```
yannmh: :clap: for the test !

checks/libs/wmi/sampler.py
```diff
@@ -263,45 +259,32 @@ def _format_property_values(self, previous, current):
     @contextmanager
     def get_connection(self):
         """
-        Return an existing, available WMI connection or create a new one.
-        Release, i.e. mark as available at exit.
+        Create a new WMI connection
```
yannmh: I guess, it doesn't really make sense anymore to have this method as a `contextmanager`.

yannmh:
Looks great @olivielpeau.

Could you also remove this useless code line please https://github.com/DataDog/dd-agent/blob/master/tests/core/test_wmi.py#L267 ?

Feel free to merge

PR #2403, opened 2016-04-01
[process][win32] Add metric on opened handles

olivielpeau:
(would allow us to monitor that the agent doesn't have a handle leak)


yannmh:
Why is it Windows specific ?

olivielpeau:
@yannmh Very basically, `handles` on Windows (see https://pythonhosted.org/psutil/#psutil.Process.num_handles) are the equivalent of `file descriptors` on Unix (see https://pythonhosted.org/psutil/#psutil.Process.num_fds)

yannmh:
Looks good :+1:

PR #2405, opened 2016-04-04
[docker] add optional disk metrics

hkaj:
In cases where the docker volume is not mounted the disk check can't collect metrics about the volume usage. It happens in the ECS optimized AMI for example: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-optimized_AMI.html

Collecting these metrics using `docker info` makes sense in this case and is also useful in others.

### DIFF
--- a/checks.d/docker_daemon.py
+++ b/checks.d/docker_daemon.py
@@ -18,6 +18,7 @@ from utils.platform import Platform
 
 EVENT_TYPE = 'docker'
 SERVICE_CHECK_NAME = 'docker.service_up'
+DISK_METRICS_PREFIX = 'docker.disk.{0}'
 SIZE_REFRESH_RATE = 5  # Collect container sizes every 5 iterations of the check
 MAX_CGROUP_LISTING_RETRIES = 3
 CONTAINER_ID_RE = re.compile('[0-9a-f]{64}')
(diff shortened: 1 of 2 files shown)


checks.d/docker_daemon.py
```diff
@@ -32,6 +33,13 @@
     RATE: {True: HISTORATE, False: RATE}
 }
 
+UNIT_MAP = {
```
truthbk: should we have these as 2^10, 2^20, 2^30?

checks.d/docker_daemon.py
```diff
@@ -643,6 +656,39 @@ def _format_events(self, aggregated_events, containers_by_id):
...
+                self.gauge(DISK_METRICS_PREFIX.format(name), val, tags)
+
+    def _format_disk_metrics(self, metrics):
+        """Cast the disk stats to float and convert them to bytes"""
+        for name, raw_val in metrics.iteritems():
+            if raw_val:
+                val, unit = raw_val.split(' ')
+                # by default some are uppercased others lowercased. That's error prone.
+                unit = unit.lower()
+                val = int(float(val) * UNIT_MAP.get(unit, 1))
```
truthbk: I'm not convinced by this... maybe a log message saying `"unsupported unit %s" % unit`? I see your point of falling back to bytes, but maybe logging something would be helpful for support if we do end up reporting in the wrong unit - I can see `TB` coming into play. Or maybe adding `tb` to the `UNIT_MAP`? You've played with this, so I guess you know if this is a valid concern or not.

checks.d/docker_daemon.py
hkaj: actually no, the docker info command displays in 10^3, 10^6, 10^9 bytes form.

checks.d/docker_daemon.py
hkaj: yeah that makes sense. I'll add `tb` and error out with a log message when the unit's not recognized.

truthbk:
:+1: looks good!

PR #2407, opened 2016-04-06
Minerkasch/spark

zachradtka:
An AgentCheck that gathers metrics for Apache Spark

The metrics collected are:
```
Spark Job Metrics
-----------------
spark.job.num_tasks
spark.job.num_active_tasks
spark.job.num_completed_tasks
spark.job.num_skipped_tasks
spark.job.num_failed_tasks
spark.job.num_active_stages
spark.job.num_completed_stages
spark.job.num_skipped_stages
spark.job.num_failed_stages

Spark Stage Metrics
-------------------
spark.stage.num_active_tasks
spark.stage.num_complete_tasks
spark.stage.num_failed_tasks
spark.stage.executor_run_time
spark.stage.input_bytes
spark.stage.input_records
spark.stage.output_bytes
spark.stage.output_records
spark.stage.shuffle_read_bytes
spark.stage.shuffle_read_records
spark.stage.shuffle_write_bytes
spark.stage.shuffle_write_records
spark.stage.memory_bytes_spilled
spark.stage.disk_bytes_spilled

Spark Executor Metrics
----------------------
spark.executor.rdd_blocks
spark.executor.memory_used
spark.executor.disk_used
spark.executor.active_tasks
spark.executor.failed_tasks
spark.executor.completed_tasks
spark.executor.total_tasks
spark.executor.total_duration
spark.executor.total_input_bytes
spark.executor.total_shuffle_read
spark.executor.total_shuffle_write
spark.executor.max_memory

Spark RDD Metrics
-----------------
spark.rdd.num_partitions
spark.rdd.num_cached_partitions
spark.memory_used
spark.rdd.disk_used
```


checks.d/spark.py
```diff
@@ -0,0 +1,381 @@
...
+    'memoryUsed': ('spark.executor.memory_used', HISTOGRAM),
+    'diskUsed': ('spark.executor.disk_used', HISTOGRAM),
+    'activeTasks': ('spark.executor.active_tasks', HISTOGRAM),
+    'failedTasks': ('spark.executor.failed_tasks', HISTOGRAM),
+    'completedTasks': ('spark.executor.completed_tasks', HISTOGRAM),
+    'totalTasks': ('spark.executor.total_tasks', HISTOGRAM),
+    'totalDuration': ('spark.executor.total_duration', HISTOGRAM),
+    'totalInputBytes': ('spark.executor.total_input_bytes', HISTOGRAM),
+    'totalShuffleRead': ('spark.executor.total_shuffle_read', HISTOGRAM),
+    'totalShuffleWrite': ('spark.executor.HISTOGRAM', HISTOGRAM),
```
olivielpeau: typo in the metric name here

checks.d/spark.py
```diff
@@ -0,0 +1,381 @@
...
+        for app_id, (app_name, tracking_url) in running_apps.iteritems():
+
+            response = self._rest_request_to_json(tracking_url,
+                SPARK_JOBS_PATH,
+                SPARK_SERVICE_CHECK, app_name, 'jobs')
+
+            tags = ['cluster_id:' + str(cluster_id),
+                    'app_name:' + str(app_name)]
+
+            if len(response) > 0:
```
olivielpeau: This condition is unnecessary (even without it, no iteration of the `for` loop below will be run if `len(response) == 0`).

Applies to all the other `if len(response)` in the check too ;)

checks.d/spark.py
```diff
@@ -0,0 +1,381 @@
...
+        for status, (metric_name, metric_type) in metrics.iteritems():
+            metric_status = metrics_json.get(status)
+
+            if metric_status is not None:
+                self._set_metric(metric_name,
+                    metric_type,
+                    metric_status,
+                    tags)
+
+    def _set_metric(self, metric_name, metric_type, value, tags=None, device_name=None):
```
olivielpeau: let's remove `device_name` here as the check is never using it

checks.d/spark.py
```diff
@@ -0,0 +1,381 @@
...
+spark.rdd.num_partitions
+spark.rdd.num_cached_partitions
+spark.memory_used
+spark.rdd.disk_used
+'''
+
+# stdlib
+from urlparse import urljoin
+from urlparse import urlsplit
+from urlparse import urlunsplit
```
olivielpeau: Could you use a one-line import (`from urlparse import urljoin, urlsplit, urlunsplit`)?

olivielpeau:
Thanks @zachradtka! I've added a few comments here and there, but overall the check looks good!

Once you've addressed them we'll test the check on our own Spark setup, I may have more feedback for you then.

zachradtka:
@olivielpeau, Thanks for the change suggestions! All have been rectified.

buryat:
@zachradtka 
```
    spark
    -----
      - instance #0 [ERROR]: '404 Client Error: Not Found'
          File "/home/hadoop/dd-agent-spark/checks/__init__.py", line 746, in run
          File "/home/hadoop/dd-agent-spark/checks.d/spark.py", line 155, in check
            self._spark_job_metrics(running_apps, cluster_id)
          File "/home/hadoop/dd-agent-spark/checks.d/spark.py", line 220, in _spark_job_metrics
            SPARK_SERVICE_CHECK, app_name, 'jobs')
          File "/home/hadoop/dd-agent-spark/checks.d/spark.py", line 325, in _rest_request_to_json
            response.raise_for_status()
          File "/home/hadoop/dd-agent-spark/venv/local/lib/python2.7/site-packages/requests/models.py", line 834, in raise_for_status
            raise HTTPError(http_error_msg, response=self)
        HTTPError: 404 Client Error: Not Found
```

Instead of:
`http://hostname:20888/proxy/application_1460395170645_0001/api/v1/applications/Zeppelin/jobs`

this should be called:
`http://hostname:20888/proxy/application_1460395170645_0001/api/v1/application_1460395170645_0001/Zeppelin/jobs`

zachradtka:
@buryat: Thanks for the feedback. 

What version of Spark are you using and what version of YARN are you using?

The URL that you gave me is a bit interesting, because it uses `api/v1/[app-id]/jobs`. The spark [REST API](http://spark.apache.org/docs/latest/monitoring.html#rest-api) says that the endpoint should be `api/v1/applications/[app-id]/jobs`. It is a subtle difference, but there should be the word `applications` in the endpoint. Could there have been a typo in the URL you posted?

Also, could you possibly run zepplin again and give me the output of `http://hostname:20888/proxy/application_1460395170645_0001/api/v1/applications`? That should give you the list of running Spark applications for that YARN app-id. I am testing with HDP and the `id` of my Spark applications happen to also be the `name` of the YARN application. I am guessing that the `id`s of your Spark applications is the YARN `id`.

With that I should be able to fix the errors you are seeing.

Thanks!

buryat:
@zachradtka yeah, I'm sorry, I made a type when I was redacting the hostname. I've corrected my original comment.

I use AWS EMR-4.3.0, it runs Hadoop 2.7.1, Spark 1.6.0.

```
curl "hostname:20888/proxy/application_1460395170645_0002/api/v1/applications"
[ {
  "id" : "application_1460395170645_0002",
  "name" : "Zeppelin",
  "attempts" : [ {
    "startTime" : "2016-04-12T00:42:28.934GMT",
    "endTime" : "1969-12-31T23:59:59.999GMT",
    "sparkUser" : "",
    "completed" : false
  } ]
} ] 
```

zachradtka:
@buryat Thanks for the response and for the output.

I updated the integration to get the Spark application ID's from the Spark rest interface. That should fix the problems you were seeing before.

I also updated the unit tests to reflect my code changes.

Please let me know how it goes!

zachradtka:
@olivielpeau 
I added the following metrics for the driver:
```
spark.driver.rdd_blocks
spark.driver.memory_used
spark.driver.disk_used
spark.driver.active_tasks
spark.driver.failed_tasks
spark.driver.completed_tasks
spark.driver.total_tasks
spark.driver.total_duration
spark.driver.total_input_bytes
spark.driver.total_shuffle_read
spark.driver.total_shuffle_write
spark.driver.max_memory
```

I added the following metrics to easily get the job, stage, and executor count:
```
spark.job.count
spark.stage.count
spark.executor.count
```

I also added the ability to add custom tags in the yaml file and required users to enter a `cluster_id`.

Let me know if there are any more requests.

checks.d/spark.py
```diff
@@ -0,0 +1,444 @@
...
+
+        # Report success after gathering all metrics from the ApplicationMaster
+        if running_apps:
+            app_id, (app_name, tracking_url) = running_apps.items()[0]
+            am_address = self._get_url_base(tracking_url)
+
+            self.service_check(SPARK_SERVICE_CHECK,
+                AgentCheck.OK,
+                tags=['url:%s' % am_address],
+                message='Connection to ApplicationMaster "%s" was successful' % rm_address)
```
olivielpeau: You probably want `am_address` instead of `rm_address` here

checks.d/spark.py
```diff
@@ -0,0 +1,444 @@
...
+                SPARK_SERVICE_CHECK, app_id, 'jobs')
+
+            tags = ['app_name:' + str(app_name)]
+            tags.extend(addl_tags)
+
+            for job in response:
+                self._set_metrics_from_json(tags, job, SPARK_JOB_METRICS)
+
+            if len(response):
+                self._set_metric('spark.job.count', GAUGE, len(response), tags)
```
olivielpeau: We'll get into the issue of the gauges overwriting one another if multiple apps have the same `app_name`. I'd recommend using `increment` instead of `gauge` (see https://github.com/DataDog/dd-agent/blob/5.7.3/checks/__init__.py#L364) for the `spark.*.count` metrics that you've added.

Here, you can simply call `self.increment` for each app with the number of jobs of that app, and the metric that's ultimately sent would contain the sum of the numbers that have been passed to `increment`.

So here, if multiple apps share the same `app_name`, the metric sent for that `app_name` would be the total number of jobs over these apps, which sounds okay to me.

checks.d/spark.py
```diff
@@ -0,0 +1,444 @@
...
+    'diskUsed': ('spark.driver.disk_used', GAUGE),
+    'activeTasks': ('spark.driver.active_tasks', GAUGE),
+    'failedTasks': ('spark.driver.failed_tasks', GAUGE),
+    'completedTasks': ('spark.driver.completed_tasks', GAUGE),
+    'totalTasks': ('spark.driver.total_tasks', GAUGE),
+    'totalDuration': ('spark.driver.total_duration', GAUGE),
+    'totalInputBytes': ('spark.driver.total_input_bytes', GAUGE),
+    'totalShuffleRead': ('spark.driver.total_shuffle_read', GAUGE),
+    'totalShuffleWrite': ('spark.driver.total_shuffle_write', GAUGE),
+    'maxMemory': ('spark.driver.max_memory', GAUGE)
```
olivielpeau: Unfortunately the gauges from apps that share the same app_name will overwrite one another here too

zachradtka: I was just looking at that. What if I included the `application_id` as part of the tag. This would allow me to keep gauges and ensure that multiple jobs with the same name wouldn't write over each other. I don't really like the use of increment, because if there are multiple jobs that share the same name, you won't be able to get metrics on individual jobs, just the sum of the metrics for all the jobs. If I add the `application_id` as part of the tag, then the gauges will not overwrite each other.

One potential problem would be the number of unique tags over time would increase since the `application_id` is unique each time an application is launched. Would this be a problem with tagging, or is it acceptable?

conf.d/spark.yaml.example
```diff
@@ -0,0 +1,20 @@
...
+  # ResourceManager's hostname and port.
+  #
+  # The ResourceManager hostname can be found in the yarn-site.xml conf file
+  # under the property yarn.resourcemanager.address
+  #
+  # The ResourceManager port can be found in the yarn-site.xml conf file under
+  # the property yarn.resourcemanager.webapp.address
+  #
+  - resourcemanager_uri: http://localhost:8088
+    cluster_id: MySparkCluster
```
olivielpeau: Using `cluster_id` in the config could be misleading as users may think that they'd _have to_ set the `cluster_id` that's defined by Yarn. 

I realize we may have had a slight misunderstanding here, so to make things clearer, I think that we should:
- remove this `cluster_id` parameter in the yaml
- tag the metrics with the actual `cluster_id` as provided by Yarn if and only if the check is meant to be used along with another check that also tags by `cluster_id` (the Yarn check for instance). If it's not the case then we could: 1. have a parameter named `cluster_name` that would set a `cluster_name` tag and 2. add a comment in the yaml to explain that the parameter is mandatory but can be set to any user-defined value

What do you think?

zachradtka: I think it is best to tag metrics with the `cluster_id`, provided by YARN, so all applications running on YARN can be grouped. I also think that the users should have the ability to give their cluster a friendly name, by adding the `cluster_name` in the YAML file.

I am not sure if it should be mandatory though. When making a tag mandatory in the YAML I see two options: providing a default value, and not providing a default value. If you provide a default value, that is great, it will work out of the box, but I am sure most users will not change the default value so everyone will have the tag set as the default value, e.g. `MySparkCluster`. If you don't provide a default value, it won't work out of the box, and users will have to debug the error message and/or read the instructions.

What is the preferred setting: mandatory with no default value?

olivielpeau: Ok, makes sense.

So let's keep the `cluster_id` provided by Yarn. That'll also ensure that metrics from different clusters have different sets of tags (and don't overwrite one another), so the `cluster_name` can be optional and defined by the user as one of the `tags` in the instance config.

We could recommend the user to set a `cluster_name` tag in the yaml by setting the first optional tag to `cluster_name:MySparkCluster` for instance, but keep it commented out.

zachradtka:
I added the `cluster_id` tag pulled from YARN and made the `cluster_name` optional in the YAML file. I also fixed an incorrectly name tag `spark.memory_used` -> `spark.rdd.memory_used` and added a tag to get the number of rdd's: `spark.rdd.count`.

buryat:
When I specify `cluster_name` I get the following error:
```
2016-04-19 21:04:30,813 | ERROR | dd.collector | checks.spark(__init__.py:763) | Check 'spark' instance #1 failed
  File "/home/hadoop/dd-agent-spark/checks/__init__.py", line 746, in run
  File "/home/hadoop/dd-agent-spark/checks.d/spark.py", line 175, in check
    raise Exception('The ResourceManager URL must be specified in the instance configuration')
Exception: The ResourceManager URL must be specified in the instance configuration
Metrics:
[]
Events:
[]
Service Checks:
[{'check': 'spark.resource_manager.can_connect',
  'host_name': 'i-ab0d2330',
  'id': 1,
  'message': u'Connection to ResourceManager "http://ip-10-154-242-131.ec2.internal:8088" was successful',
  'status': 0,
  'tags': ['url:http://ip-10-154-242-131.ec2.internal:8088'],
  'timestamp': 1461099870.813359}]
Service Metadata:
[{}, {}]
    spark
    -----
      - instance #0 [OK]
      - instance #1 [ERROR]: 'The ResourceManager URL must be specified in the instance configuration'
          File "/home/hadoop/dd-agent-spark/checks/__init__.py", line 746, in run
          File "/home/hadoop/dd-agent-spark/checks.d/spark.py", line 175, in check
            raise Exception('The ResourceManager URL must be specified in the instance configuration')
        Exception: The ResourceManager URL must be specified in the instance configuration

      - Collected 0 metrics, 0 events & 1 service check
```

Config file:
```yaml
init_config:

instances:
  - resourcemanager_uri: http://ip-10-154-242-131.ec2.internal:8088
  - cluster_name: MySparkCluster
```

zachradtka:
@buryat 

I just checked my setup and I was able to replicate your error and fix it by removing the `-` before `cluster_name`.

Try the following config:
``` yaml
init_config:

instances:
  - resourcemanager_uri: http://ip-10-154-242-131.ec2.internal:8088
    cluster_name: MySparkCluster
```

I am not the best at YAML, but I believe the `-` indicates a nested series, so anything at the same level would have to have the same indicator. If there is a way to clear this up or some convention you guys use for other agents, please let me know.

buryat:
@zachradtka oh, I'm so sorry, that was mistake completely.
I confirm that it works

checks.d/spark.py
olivielpeau: @zachradtka I'm not sure that using the `application_id` would be acceptable in terms of number of sets of tags that get created. Based on the behavior of the YARN check, which creates a lot of sets of tags because of the `app_id` tag, I wouldn't want the Spark check to have the same behavior unless the number of applications that would typically run on a Spark cluster is way lower than on a typical non-spark YARN cluster.

So unfortunately for the driver metrics I don't really see an alternative to using one of:
- `increment`
- manual sum by metric name and `app_name` and then `gauge`
- `histogram`

olivielpeau:
@buryat has some additional feedback on the `stage` metrics: they're currently useful to know the overall progress of the jobs but they don't capture very well the current state of the _active_ stages since  they're aggregated over all the stages (including the non-active stages).

To address this the check could send `stage.active` metrics that would capture the values of the active stages only, or tag the `stage` metrics with a `state` tag (in that case the histograms would be computed separately for each value taken by the `state`).

@zachradtka What do you think?

checks.d/spark.py
zachradtka: @olivielpeau I think the best solution is to require users to enter a `cluster_name` in the YAML, similar to the `resource_manager_uri`. I think this is best because it is an easy change that can be made across all of the YARN based agents. This will unify the agent's in terms of configuration.

I also like the idea of using the `app_name` because it will keep the metric as a gauge. I will look into how difficult it will be to do this.

zachradtka:
@olivielpeau It would probably be best to tag the metric with the state. I am not sure what complexities that will add. I will look into it though.

checks.d/spark.py
olivielpeau: @zachradtka Requiring a `cluster_name` definitely makes sense if you plan to also require it in the other Yarn-based checks, thanks!

conf.d/spark.yaml.example
```diff
@@ -0,0 +1,24 @@
...
+  # The ResourceManager hostname can be found in the yarn-site.xml conf file
+  # under the property yarn.resourcemanager.address
+  #
+  # The ResourceManager port can be found in the yarn-site.xml conf file under
+  # the property yarn.resourcemanager.webapp.address
+  #
+  - resourcemanager_uri: http://localhost:8088
+
+  # An optional friendly name can be specified for the cluster.
+  #   cluster_name: MySparkCluster
```
olivielpeau: nitpick: could you add two spaces of indentation on these two lines (given that they're at the same level as the `tags`, i.e. at the instance level)

zachradtka:
@olivielpeau I made a few updates:
* Required `cluster_name` in YAML and tagged all metrics with `cluster_name`
* Removed `cluster_id` as a tag to eliminate problems with excessive tagging combos
* Used `increment` for all of my `spark.*.count` metrics
* Utilized `increment` for the `spark.driver.*` metrics. There should be only one driver per application.
* Added `status` tags for the jobs and stages allowing a greater fidelity for metrics within a particular status.

checks.d/spark.py
```diff
@@ -0,0 +1,457 @@
...
+                status = job.get('status')
+
+                tags = ['app_name:%s' % str(app_name)]
+                tags.extend(addl_tags)
+                tags.append('status:%s' % str(status).lower())
+
+                self._set_metrics_from_json(tags, job, SPARK_JOB_METRICS)
+
+            if len(response):
+                self._set_metric('spark.job.count', INCREMENT, len(response), tags)
```
olivielpeau: The `tags` that are passed here are the `tags` of the last job of the loop above, so the `status` tag wouldn't be accurate.

Instead we should either:
- increment this count by 1 for every job, and with the `status` tag (in order to have a `count` per status)
- increment this count by `len(response)` only once per app, and without the `status` tag (in order to have only a global `count` for all status)

I would go with option 1 since having a count per status sounds useful to me, but let me know what you think.

This applies to the `spark.stage.count` metric too.

Also, if possible it'd be good to have tests that reflect how different statuses on the `jobs` and `stages` should be counted.

olivielpeau:
@zachradtka Added in one comment on your latest changes

zachradtka:
@olivielpeau Thanks for the feedback, I should have caught the count with the incorrect tag.

That has been fixed and tests have been added.

olivielpeau:
@zachradtka Looks good to me, thanks! Could you squash your commits into one?

zachradtka:
@olivielpeau Commits squashed!

olivielpeau:
Merging! 🎉

PR #2409, opened 2016-04-07
[mongo] Improve messaging&tags on MongoDB replset member state events

antifuchs:
This change summarized a lot of changes that improve how leadership election events look in the mongodb service check:

* The event title nowrefers to the explicit hostname if checking localhost
* The title now calls the member state by its official name and gives an the expanded state description in the body.
* Events are only generated if there is a previous, known, state. This elimilates a lot of events being generated whenever the dd-agent is restarted.
* Events now come with a set of tags featuring the previous and current states in a form that can unambiguously be queried, as well as the replSet in question.

This change also eliminates the duplicate replica set state table that got introduced in a previous PR.

## Testing

We're running this updated check in production at the moment. It generates events that we can easily overlay on top of graphs using queries like `tags:member_status:primary,previous_member_status:secondary` (this finds re-elections where the primary is different from before).

### DIFF
--- a/checks.d/mongo.py
+++ b/checks.d/mongo.py
@@ -7,7 +7,7 @@ import pymongo
 
 # project
 from checks import AgentCheck
-from util import get_hostname
+from urlparse import urlsplit
 
 DEFAULT_TIMEOUT = 30
 GAUGE = AgentCheck.gauge
@@ -46,10 +46,6 @@ class MongoDb(AgentCheck):
     """
     SOURCE_TYPE_NAME = 'mongodb'
 
-    # Service checks
-    SERVICE_CHECK_NAME = 'mongodb.can_connect'
-
-    # Metrics
     """
     MongoDB replica set states, as documented at
     https://docs.mongodb.org/manual/reference/replica-states/
@@ -67,6 +63,14 @@ class MongoDb(AgentCheck):
         10: 'removed'
     }
 
+    # METRIC LIST DEFINITION
+    #
+    # Format
+    # ------
+    #   metric_name -> (metric_type, alias)
+    # or
+    #   metric_name -> metric_type *
+    # * by default MongoDB metrics are reported under their original metric names
     """
     Core metrics collected by default.
     """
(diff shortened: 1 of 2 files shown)


checks.d/mongo.py
```diff
@@ -354,49 +352,52 @@ def __init__(self, name, init_config, agentConfig, instances=None):
...
-                return 'Starting up (forking threads)'
-            elif state == 6:
-                return 'Unknown'
-            elif state == 7:
-                return 'Arbiter'
-            elif state == 8:
-                return 'Down'
-            elif state == 9:
-                return 'Rollback'
+            return self.REPLSET_MEMBER_STATES[state][1]
```
yannmh: As for the `get_state_name ` method, can we consider the case where `state` does not exist in `REPLSET_MEMBER_STATES ` ?
``` python
if state in 
  pass
else:
  return 'UNKNOWN'
```

Even though, this should not happen (as opposed to `get_state_name` when `last_state==-1` ), I think that would make the check less error prone to MongoDB updates.

checks.d/mongo.py
```diff
@@ -354,49 +352,52 @@ def __init__(self, name, init_config, agentConfig, instances=None):
...
+        self._last_state_by_server[clean_server_name] = state
+        if last_state != state and last_state != -1:
+            return self.create_event(last_state, state, clean_server_name, replSet['set'], agentConfig)
+
+    def hostname_for_event(self, clean_server_name, agentConfig):
+        """Return a reasonable hostname for a replset membership event to mention."""
+        uri = urlsplit(clean_server_name)
+        hostname = uri.netloc.split(':')[0]
+        if hostname == 'localhost':
+            hostname = get_hostname(agentConfig)
```
yannmh: Let's use the instance variable :)
```python
hostname = self.hostname
```

yannmh:
Thanks a lot @antifuchs. Your changes look great: I simply added a few nitpicks :pencil2:. 

We are planning a `5.7.4` release with various improvements on the MongoDB check. I added your changes to the milestone :ship:

checks.d/mongo.py
antifuchs: Agreed, that would be desirable. I'll fix this up.

checks.d/mongo.py
antifuchs: Happy to do that - can you tell me what the difference is between the instance variable & `get_hostname`? (I cargo-culted that usage from other places where it was used (-%)

yannmh: The result is the same but when the hostname is not specified in `datadog.conf` file, the `get_hostname` method has to re-resolve it, when the value is already accessible, stored in an instance variable, https://github.com/DataDog/dd-agent/blob/master/checks/__init__.py#L326

yannmh:
Would you mind rebasing your changes on the last `master` branch state please ? I merged various changes on the check, which seem to conflict with your PR 😕.
It should be pretty straightforward to fix, let me know if you are busy I'll do it :).

antifuchs:
@yannmh - the two fixes are in, and I rebased against master. Please let me know if the two changes are ok as-is or if you prefer them squashed into one.

yannmh:
Squashed changes would be perfect 😄

antifuchs:
Squashed! (:

yannmh:
@antifuchs, I added a small commit on top of yours to fix the tests and restore a variable which got removed by accident in the rebase. 
Let's move to https://github.com/DataDog/dd-agent/pull/2432 😄

PR #2411, opened 2016-04-08
[file] New check to test for a file's existence

antifuchs:
The file.py check allows the agent to test for whether files exist in the local file system, and if they do, emit a metric for how long they have existed. This is particularly useful for files like Debian's "package upgrades require a reboot" marker in /var/run/, or local puppet facts meant to be kept around only for a short time.

I have added almost no documentation to this check (mostly out of ignorance what format / locations are desired!), but would be interested in learning how you all prefer the check docs to be written.
### Testing

I added an integration test for the metric and service_check behavior. We're also running this check on our infra to test for exactly the presence of `/var/run/reboot-required`, and it emits events and service checks as intended.

### DIFF
--- a/checks.d/file.py
+++ b/checks.d/file.py
@@ -1,0 +1,83 @@
+import errno
+import os
+import time
+
+from checks import AgentCheck
+
+class FileCheck(AgentCheck):
+
+    STATUS_ABSENT = 'absent'
+    STATUS_PRESENT = 'present'
+
+    def __init__(self, name, init_config, agentConfig, instances=None):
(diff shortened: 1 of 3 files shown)


truthbk:
Hi @ejholmes, since the transition to the SDK has begun I have taken the liberty to open a PR with your work here: DataDog/integrations-extras#35. You authorship should be preserved.

This should hopefully ease the pain of transitioning to the SDK. Let me know if you have any questions.

irabinovitch:
Closing in favor of   DataDog/integrations-extras#35

PR #2412, opened 2016-04-09
[graylog2] add agent check for graylog2-server metrics

brettlangdon:
In this PR we are adding in a new agent check for [graylog2](https://www.graylog.org/) server. This check uses the servers REST API to fetch metrics about the system health and report them to Datadog.

As a note, this agent check by default will produce A LOT of metrics (over 4k). I wasn't entirely sure how to approach this issue, but what I have setup is the following:
- By default collect all metrics
- Add a metric prefix whitelist and blacklist to allow users to limit which metrics are collected

For example, I am running this agent check in production right now with the following config:

``` yaml
instances:
  -
    transport_uri: http://127.0.0.1:12900
    prefix_whitelist:
      - jvm.
```

Which is only collecting `jvm` metrics. I will slowly add more to the whitelist as I learn which ones are important to us.

I am very open to hearing suggestions on whether this was an appropriate approach to handling.


truthbk:
@brettlangdon thank you very much for your contribution, we will review this as soon as possible.

masci:
Hi @brettlangdon ,

Thanks again for your contribution, always very much appreciated. We're currently transitioning our checks to a new SDK-based model for the agent. An unfortunate side-effect is that we have to impose a freeze on the dd-agent repo. The SDK consists of two repositories https://github.com/DataDog/integrations-core and https://github.com/DataDog/integrations-extras, where all of our checks will reside. This PR would belong in the core/extras repository. Would you be so kind as to re-open the PR there? We have facilitated some tools to help you with the SDK development that should help you get up and running in no time: [sdk guide](http://docs.datadoghq.com/guides/integration_sdk/) [datadog sdk gem]( https://github.com/DataDog/datadog-sdk-testing). We are very sorry for the inconvenience caused. 

PR #2413, opened 2016-04-12
Adds Python Interpreter Arch to Info Cmd

gmmeyer:
The header on the Python interpreter will now include the Python Interpreter Architecture as well as the version. It will look like this:

```
Status date: 2016-04-12 15:49:45 (2s ago)
Pid: 31865
Platform: Linux-3.13.0-32-generic-x86_64-with-debian-wheezy-sid
Python Version: 2.7.11, 64bit
Logs: <stderr>, /tmp/collector.log
```

### DIFF
--- a/win32/gui.py
+++ b/win32/gui.py
@@ -401,6 +401,7 @@ class HTMLWindow(QTextEdit):
                 platform=platform.platform(),
                 agent_version=get_version(),
                 python_version=platform.python_version(),
+                python_architecture=Platform.python_architecture(),
                 logger_info=logger_info(),
                 dogstatsd=dogstatsd_status.to_dict(),
                 forwarder=forwarder_status.to_dict(),
--- a/checks/check_status.py
+++ b/checks/check_status.py
@@ -182,7 +182,9 @@ class AgentStatus(object):
         fields += [
             ("Pid", self.created_by_pid),
             ("Platform", platform.platform()),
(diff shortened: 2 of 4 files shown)


olivielpeau:
Looks good on Unix, thanks!

For this to work on the windows agent you'll have to update this too: https://github.com/DataDog/dd-agent/blob/5.7.3/win32/gui.py#L399-L408 (and probably the [template](https://github.com/DataDog/dd-agent/blob/master/win32/status.html) too).

win32/gui.py
```diff
@@ -396,11 +396,16 @@ def latest_status(self):
             dogstatsd_status = DogstatsdStatus.load_latest_status()
             forwarder_status = ForwarderStatus.load_latest_status()
             collector_status = CollectorStatus.load_latest_status()
+            if sys.maxsize > 2**32:
+                python_architecture = "64bit"
+            else:
+                python_architecture = "32bit"
```
olivielpeau: Could you extract this logic in a method in `utils/platform.py`, and just call that method here (same thing in check_status)

gmmeyer: Yes I can

olivielpeau:
Looks good, could you squash your commits into one please?

Once that's done let's :shipit:

olivielpeau:
Merging :tada:

PR #2414, opened 2016-04-12
[core] Make sure to flush service metadata

remh:
Otherwise they will just leak. No need to store them or send them in the agent payload though (they are empty).


remh:
@yannmh can you review please ?

yannmh:
Looks good :bow:

PR #2417, opened 2016-04-13
[windows] new service and packaging

degemer:
This PR completely changes the agent on Windows.

Try it here : [](https://drive.google.com/a/datadoghq.com/file/d/0B6bdbouD69UhS3djVzBlaDRxZDQ/view?usp=sharing)https://drive.google.com/a/datadoghq.com/file/d/0B6bdbouD69UhS3djVzBlaDRxZDQ/view?usp=sharing

@elafarge is the main author of this PR, I fixed some bugs and rebased.
## Previous state

The agent on Windows has two parts : 
-   the service (`win32/agent.py`) : starts and watch 4 processes (collector, forwarder, dogstatsd, jmxfetch), which reuse a lot of code from the Unix agent (except the collector, not using `agent.py`).
-   the gui : awful GUI we all know, displays the state/logs/conf files. Only easy way to edit the configuration files on Windows. Interacts with the service to start/stop/restart it.

Both of them are shipped with `py2exe`, so we have to specify all their dependencies in `setup.py`, and it's really difficult to add a Python package once installed.
## Why change this ?
1.  It's really difficult to add a check dependency.
2.  It's a pain to debug anything on Windows since the main interface is the GUI (or the embedded python, but it's frozen by py2exe).
3.  Having to maintain a part of the code for Windows specifically for Windows (and duplicate the feature added to `agent.py`) is annoying.

Omnibus helps us prepare the dependencies for the agent, but it won't solve our problems if the agent is still completely shipped frozen using py2exe.
## What does this PR do ?

First things first : it doesn't change the GUI. (it adds two displayable logs files, but that's it)

However it completely changes the service : it now lives in `win32/service.py`, and only launches a "Windows supervisor" (`windows_supervisor.py`).

This service and the GUI are STILL shipped using py2exe. (see updated `setup.py`), but the Windows supervisor and the agent are not. They are using a python shipped by Omnibus.

The Windows supervisor launches the agent processes (collector, dogstatsd, forwarder, jmxfetch) and restarts them if they die. It relies on `psutil` to do so.

The agent is slightly modified for Windows compatibility (watchdog and os.signal).

It is now way easier to play with the Windows agent, and it's even possible to interact with it using the command line. And the python shipped with it is a real/full one, with pip and the possibility to install other dependencies.

This PR also adds "custom checks" for Windows, i.e. we could specify a directory where customers could put custom checks (Unix agent-like).
## Does it affect other platforms ?

It shouldn't, but the watchdog and the agent were modified/moved for Windows compatibility. It was only tested on Windows.
## Main review points
-   New service (`win32/service.py`) : launch the Windows supervisor and stops it.
-   Windows supervisor (`windows_supervisor.py`), written from scratch, starts and stops the agent processes.
-   Watchdog for Windows and Linux (`utils/watchdog.py`).
## Related PR

https://github.com/DataDog/dd-agent-omnibus/pull/74
https://github.com/DataDog/omnibus-software/pull/48
https://github.com/DataDog/omnibus-ruby/pull/20
https://github.com/DataDog/osx-dd-agent-build-box/pull/1


olivielpeau:
@degemer Thanks for the PR (and the super clear description :) )! Let's get this reviewed and merged for the 5.9 release.

This supersedes #1847

setup.py
```diff
@@ -30,85 +28,39 @@
...
+        'servicemanager',
     ]
 
     class Target(object):
         def __init__(self, **kw):
             self.__dict__.update(kw)
             self.version = get_version()
             self.company_name = 'Datadog, Inc.'
-            self.copyright = 'Copyright 2013 Datadog, Inc.'
+            self.copyright = 'Copyright 2015 Datadog, Inc.'
```
derekwbrown: 2016, or 2017 if this isn't going to be released until after first of the year?

utils/watchdog.py
```diff
@@ -0,0 +1,155 @@
...
+            self._restarts.append(time.time())
+            if self._is_frenetic():
+                self.destruct()
+
+        # Re arm alarm signal
+        log.debug("Resetting watchdog for %d" % self._duration)
+        signal.alarm(self._duration)
+
+
+def new_watchdog(duration, max_mem_mb=None, max_resets=None):
```
remh: This should be a classmethod of the Watchdog class:

```python
class Watchdog(object):
    
    @classmethod
    def create(cls, duration, max_mem_mb=None, max_resets=None):
        if Platform.is_windows():
            return WatchdogWindows(duration)
        return WatchdogPosix(duration, max_mem_mb=max_mem_mb, max_resets=max_resets)

File: utils/watchdog.py [hunk 3]
```diff
@@ -0,0 +1,155 @@
...
+    # Activity history timeframe
+    _RESTART_TIMEFRAME = 60
+
+    def __init__(self, duration, max_mem_mb=None, max_resets=None):
+        # Set the duration
+        self._duration = int(duration)
+        signal.signal(signal.SIGALRM, WatchdogPosix.self_destruct)
+
+        # Set memory usage threshold
+        if max_mem_mb is not None:
```
remh: Let's take the opportunity to kill the memory limit code. It was added as an experiment and it never really took off and is not documented anywhere

config.py
```diff
@@ -618,9 +582,62 @@ def get_config(parse_args=True, cfg_path=None, options=None):
...
+
     return agentConfig
 
 
+def get_registry_conf(agentConfig):
+    registry_conf = {}
+    try:
+        import _winreg
+        with _winreg.OpenKey(_winreg.HKEY_LOCAL_MACHINE,
+                             "SOFTWARE\\Datadog\\Datadog Agent") as reg_key:
```
remh: Let's create a constant for the registry key

config.py
```diff
@@ -336,7 +302,7 @@ def remove_empty(string_array):
     return filter(lambda x: x, string_array)
 
 
-def get_config(parse_args=True, cfg_path=None, options=None):
+def get_config(parse_args=True, cfg_path=None, options=None, can_write_conf=False):
```
remh: do we need this new arg ?

remh:
Can we move the windows specific stuff out of `config.py` to its own util module ?

e.g. the helpers to get the common data directory, the functions to write to the registry etc. Otherwise we are making the config.py code even messier

config.py
```diff
@@ -618,9 +582,62 @@ def get_config(parse_args=True, cfg_path=None, options=None):
...
+                option = _winreg.QueryValueEx(reg_key, attribute)[0]
+                if option != '':
+                    registry_conf[attribute] = option
+    except (ImportError, WindowsError) as e:
+        log.error('Unable to get config keys from Registry: %s', e)
+
+    return registry_conf
+
+
+def update_conf_file(registry_conf, config_path):
```
remh: Something like
```python
def update_conf_file(registry_conf, config_path):
    lines = []
    with open(config_path, 'r') as f:
        while True:
            line = f.readline()
            if not line:
                break

            for k, v in registry_conf.iteritems():
                if k in line:
                    line = '{}: {}'.format(k, v)
            lines.append(line)
    with open(config_path, 'w') as f:
        f.write(os.linesep.join(lines))
```

is easier to read. 

That being said, we should move that somewhere else.

setup.py
```diff
@@ -8,12 +8,10 @@
 import sys
 
 # 3p
-from setuptools import find_packages, setup
```
remh: do we still need setup.py if we use omnibus ?

win32/gui.py
```diff
@@ -111,7 +115,7 @@
     }
 
 
-log = logging.getLogger(__name__)
+log = logging.getLogger('gui')
```
remh: why this change ?

windows_supervisor.py
```diff
@@ -0,0 +1,323 @@
+"""
```
remh: Let's find a better location for this file.

windows_supervisor.py
```diff
@@ -0,0 +1,323 @@
...
+        # Watch JMXFetch restarts
+        self._MAX_JMXFETCH_RESTARTS = 3
+        self._count_jmxfetch_restarts = 0
+
+        # C:\Program Files(x86)\Datadog\Datadog Agent\agent
+        file_dir = os.path.dirname(os.path.realpath(__file__))
+        embedded_python = os.path.normpath(
+            os.path.join(file_dir, '..', 'embedded', 'python.exe')
+        )
+        # This allows us to use the system's Python in case there is no embedded python
```
remh: Tis scenario is not meant to happen. Let's remove this.

yannmh:
> It is now way easier to play with the Windows agent, and it's even possible to interact with it using the command line. And the python shipped with it is a real/full one, with pip and the possibility to install other dependencies.


Can we list some examples please?

setup.py
degemer: Yes, still used to build the GUI unfortunately.

win32/gui.py
degemer: An experiment to try to actually get the logs somewhere, but it didn't succeed. I'll revert it.

config.py
degemer: Conf file uses `\n` as eol even on Windows.
I'll move it to its own `utils/windows_configuration`.

win32/service.py
```diff
@@ -0,0 +1,183 @@
...
+
+    path_buf = wintypes.create_unicode_buffer(wintypes.MAX_PATH)
+    _SHGetFolderPath(0, CSIDL_COMMON_APPDATA, 0, 0, path_buf)
+    return path_buf.value
+
+
+# Let's configure logging accordingly now (we need the above function for that)
+logging.basicConfig(
+    filename=os.path.join(_windows_commondata_path(), 'Datadog', 'logs', 'service.log'),
+    level=logging.DEBUG,
```
degemer: Yeah, that an issue. We could just not display it in the GUI but only send it with flares.

win32/service.py
```diff
@@ -0,0 +1,183 @@
...
+        win32serviceutil.ServiceFramework.__init__(self, args)
+
+        current_dir = os.path.dirname(os.path.realpath(__file__))
+        # Are we in a py2exed package or in a source install script or just a git pulled repo ?
+        if not os.path.isfile(os.path.join(current_dir, 'windows_supervisor.py')):
+            # py2exe package
+            # current_dir should be somthing like
+            # C:\Program Files(x86)\Datadog\Datadog Agent\dist\library.zip\win32s
+            self.agent_path = os.path.join(current_dir, '..', '..', '..', 'agent')
+        else:
```
degemer: Probably a feature which was abandoned. I'm going to remove this code.

win32/service.py
```diff
@@ -1,170 +1,271 @@
...
+            self._restarts.popleft()
+
+        return len(self._restarts) < self._max_restarts
+
+    def restart(self):
+        if not self._can_restart():
+            log.error(
+                "{0} reached the limit of restarts ({1} tries during the last {2}s"
+                " (max authorized: {3})). Not restarting."
+                .format(self._name, len(self._restarts),
```
yannmh: --> there are multiple occurrences of this.

utils/watchdog.py
```diff
@@ -0,0 +1,137 @@
...
+    def destruct(self):
+        try:
+            log.error("Self-destructing...")
+            log.error(traceback.format_exc())
+        finally:
+            # This will kill the current process including the Watchdog's thread
+            psutil.Process().kill()
+
+    def reset(self):
+        log.debug("Resetting watchdog for %d" % self._duration)
```
yannmh: I am worried this gets too spammy.

utils/windows_configuration.py
```diff
@@ -0,0 +1,61 @@
...
+                    line = '{}: {}\n'.format(k, v)
+            temp_config.write(line)
+    temp_config.close()
+    try:
+        os.remove(config_path)
+        os.rename(temp_config_path, config_path)
+    except OSError as e:
+        log.exception('Unable to save new datadog.conf')
+
+    log.debug('Sucessfully saved the new datadog.cong')
```
yannmh: `s/Sucessfully/Successfully` and `s/datadog.cong/datadog.conf`

win32/service.py
```diff
@@ -0,0 +1,288 @@
...
+
+            psutil.wait_procs([self._proc], timeout=3)
+
+            if self._proc.is_running():
+                log.debug("%s didn't exit. Killing it.", self._name)
+                self._proc.kill()
+
+            log.info("%s is stopped.", self._name)
+        else:
+            log.info('%s was not running.', self._name)
```
yannmh: Doesn't seem useful.

yannmh:
Issue: JMXFetch is not properly exited when the service stops.

yannmh:
Can we also document how to use the Datadog Agent from the command-line? We used to have a `shell.exe` Python entry point, but it's not the case anymore.

win32/service.py
degemer: Switched it to debug. I think it's still useful because it tells us the service didn't find the process. (more obvious than an absence of log)

utils/watchdog.py
degemer: We were already doing it in the Unix Watchdog, I think it's fine.

agent.py
```diff
@@ -14,18 +14,24 @@
...
 
 # For pickle & PID files, see issue 293
 os.umask(022)
 
+# 3p
+try:
+    import supervisor.xmlrpc
+except ImportError:
+    # Not used/shipped on Windows
+    pass
```
degemer: I respectfully disagree. I don't think it's needed for one dependency. Actually I would prefer extracting the function using `supervisor.xmlrpc` in its own `utils/` and have the logic there.

degemer:
JMXFetch now exits properly with the agent.

win32/service.py
```diff
@@ -0,0 +1,289 @@
...
+        AgentSvc.devnull = open(os.devnull, 'w')
+
+        config = get_config(parse_args=False, can_query_registry=False)
+        if config['api_key'] == 'APIKEYHERE':
+            self._update_config_file(config)
+
+        # Let's have an uptime counter
+        self.start_ts = None
+
+        # find the main agent dir, for instance C:\Program Files\Datadog\Datadog Agent\agent
```
olivielpeau: the comment is incorrect right? Think this should read `C:\Program Files\Datadog\Datadog Agent`

win32/service.py
```diff
@@ -0,0 +1,289 @@
...
+    def _prepare_agent_env(self, dd_dir):
+        # preparing a clean env for the agent processes
+        env = os.environ.copy()
+        if env.get('PYTHONPATH'):
+            del env['PYTHONPATH']
+        if env.get('PYTHONHOME'):
+            del env['PYTHONHOME']
+        if env['PATH'][-1] != ';':
+            env['PATH'] += ';'
+        env['PATH'] += "{};{};".format(os.path.join(dd_dir, 'bin'), os.path.join(dd_dir, 'embedded'))
```
olivielpeau: why do we also add `embedded` to the path?

degemer: Because `gohai` requires `python` to be in the path (to get the python version), and `python.exe` is in `embedded`.

setup.py
```diff
@@ -118,8 +70,8 @@ def __init__(self, **kw):
                 'compressed': True,
                 'bundle_files': 3,
                 'excludes': ['numpy'],
-                'dll_excludes': ['crypt32.dll',"IPHLPAPI.DLL", "NSI.dll",  "WINNSI.DLL",  "WTSAPI32.dll"],
-                'ascii':False,
+                'dll_excludes': ["IPHLPAPI.DLL", "NSI.dll",  "WINNSI.DLL",  "WTSAPI32.dll", "WTSAPI32.dll"],
```
degemer: Probably not, re-included it.

setup.py
```diff
@@ -30,85 +28,39 @@
...
-                install_requires.append(line)
 
     # windows-specific deps
     install_requires.append('pywin32==217')
 
     # Modules to force-include in the exe
     include_modules = [
+        # stdlib
+        'os',
+        'time',
```
degemer: Removed them, it still works.

win32/service.py
olivielpeau: let's remove it, we don't need this info in gohai anyway

olivielpeau: actually you can disregard my comment: we do the same path handling on Linux so let's do the same thing here

win32/service.py
```diff
@@ -0,0 +1,298 @@
...
+
+    def stop(self):
+        if self._proc is not None and self._proc.is_running():
+            log.info("Stopping %s...", self._name)
+            for child_proc in self._proc.children():
+                child_proc.kill()
+
+            self._proc.terminate()
+
+            self._proc.wait(timeout=self._STOP_TIMEOUT)
```
olivielpeau: let's also document that this `wait` step is not supposed to do anything right now (since `terminate` simply kills the process on Windows)

raboof:
Cool stuff!

PR #2418, opened 2016-04-13
better 413 response handling

gmmeyer:
When propjoe responds with a 413 error (payload too big), the agent used to retry the request. It shouldn't. The request shouldn't ever succeed.

This stops it from resubmitting the request. It logs a warning, saying what happened. It also adds a warning to the info page that a request was rejected.

### DIFF
--- a/checks/check_status.py
+++ b/checks/check_status.py
@@ -757,7 +757,7 @@ class ForwarderStatus(AgentStatus):
     NAME = 'Forwarder'
 
     def __init__(self, queue_length=0, queue_size=0, flush_count=0, transactions_received=0,
-                 transactions_flushed=0):
+                 transactions_flushed=0, too_big_count=0):
         AgentStatus.__init__(self)
         self.queue_length = queue_length
         self.queue_size = queue_size
@@ -767,6 +767,7 @@ class ForwarderStatus(AgentStatus):
         self.proxy_data = get_config(parse_args=False).get('proxy_settings')
         self.hidden_username = None
         self.hidden_password = None
+        self.too_big_count = too_big_count
         if self.proxy_data and self.proxy_data.get('user'):
             username = self.proxy_data.get('user')
             hidden = len(username) / 2 if len(username) <= 7 else len(username) - 4
@@ -780,6 +781,7 @@ class ForwarderStatus(AgentStatus):
             "Flush Count: %s" % self.flush_count,
             "Transactions received: %s" % self.transactions_received,
             "Transactions flushed: %s" % self.transactions_flushed,
+            "Transactions rejected: %s" % self.too_big_count,
             ""
         ]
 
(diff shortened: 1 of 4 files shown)


checks/check_status.py
```diff
@@ -797,6 +798,17 @@ def body_lines(self):
                     "  Password: %s" % self.hidden_password
                 ]
 
+        if self.too_big_count:
+            if self.too_big_count > 1:
+                lines += [
+                    "%d messages were rejected by the server as too big." % self.too_big_count
```
olivielpeau: To stay consistent with the rest of the info page, I'd prefer this message to be displayed with the other stats on the transactions, i.e. something like:

```
Transactions rejected (too large): 15
```

right below `Transactions flushed`. What do you think?

transaction.py
```diff
@@ -212,6 +215,23 @@ def tr_error(self,tr):
           (tr.get_id(), tr.get_error_count(), plural(tr.get_error_count()),
            tr.get_next_flush()))
 
+    def tr_error_too_big(self,tr):
+        tr.inc_error_count()
+        log.warn("Transaction %d is too large. It will not be replayed." % (tr.get_id()))
```
olivielpeau: I think we should also have the transaction size in the log message here.

olivielpeau:
@gmmeyer Cool stuff! Added in 2 comments

checks/check_status.py
gmmeyer: That sounds good to me. Should it appear there even if none have been rejected?

win32/status.html
```diff
@@ -242,6 +242,7 @@ <h1 id="status-header">Agent info</h1>
             <dt>Flush count</dt>        <dd>{{ forwarder['flush_count'] }}</dd>
             <dt>Queue size</dt>         <dd>{{ forwarder['queue_size'] }} bytes</dd>
             <dt>Queue length</dt>       <dd>{{ forwarder['queue_length'] }}</dd>
+            <dt>Transactions Rejected (too large)</dt>       <dd>{{forwarder['too_big_count']}}</dd>
```
olivielpeau: nitpick: could you add spaces between `{{`, `}}` and the var that's displayed (just for consistency)

Also, it's unrelated but while we're at it: could you add the `Transactions received` and `Transactions flushed` stats on this page too. Thanks!

gmmeyer: Yup, I can do all of that.

checks/check_status.py
```diff
@@ -757,7 +757,7 @@ class ForwarderStatus(AgentStatus):
     NAME = 'Forwarder'
 
     def __init__(self, queue_length=0, queue_size=0, flush_count=0, transactions_received=0,
-                 transactions_flushed=0):
+            transactions_flushed=0, too_big_count=0):
```
olivielpeau: nitpick: any reason for changing the indentation here? :)

olivielpeau:
One last nitpcik, otherwise looks gtg!

checks/check_status.py
gmmeyer: I don't think so. I'll change it back. I think it was just what my autoindent did.

olivielpeau:
👍  

Feel free to squash your commits and then merge 🎉

PR #2371, opened 2016-03-21
[kafka] updates to yaml file

vagelim:
updated Kafka configuration file to support collection of more metrics,
to be discussed in upcoming series on monitoring Kafka

PR #2420, opened 2016-04-15
Vagelim/kafka yaml

vagelim:
cc: @olivielpeau


conf.d/kafka.yaml.example
```diff
@@ -17,15 +21,161 @@ instances:
...
+            metric_type: gauge
+            alias: kafka.producer.io_wait
+
+
+    # v0.8.2.x Consumers
+    - include:
+        domain: 'kafka.consumer'
+        bean_regex: 'kafka\.consumer:type=FetcherLagMetrics,name=ConsumerLag,clientId=.*'
+        attribute:
+            alias: kafka.consumer.lag
```
olivielpeau: I think you forgot to fully specify the attribute here

irabinovitch: Do we need to make these changes in the tile and docs as well?

olivielpeau:
@evan: added a comment.

As @irabinovitch has said it would make sense to add these metrics in the metrics metadata (some of the metrics depend on the version of Kafka but I think it's ok to add them all)

vagelim:
@olivielpeau it seems that if you do not specify a metric_type, it defaults to gauge: http://docs.datadoghq.com/integrations/java/

@irabinovitch will update metadata accordingly

vagelim:
@olivielpeau that metric was a duplicate; removed

@irabinovitch I will update the metadata accordingly

olivielpeau:
Implemented the diff between this PR and #2371 (and fixes the change in #2425) in #2435

PR #2422, opened 2016-04-18
[core] adds swap (pagefile) to windows stats

gmmeyer:
Our Windows pagefile statistics are not very robust. We've had a few requests to add more, better. This PR adds such metrics.

Now we're capturing the total size of the pagefile, how much is used, how much is available and the percent that's still free.

I've tested it on Windows, it seems to work fine, and the metrics come out looking pretty nice.

### DIFF
--- a/checks/system/win32.py
+++ b/checks/system/win32.py
@@ -109,6 +109,11 @@ class Memory(Check):
         # usable = free + cached
         self.gauge('system.mem.usable')
         self.gauge('system.mem.pct_usable')
+        #  details about the usage of the pagefile.
+        self.gauge('system.mem.page_total')
+        self.gauge('system.mem.page_used')
+        self.gauge('system.mem.page_free')
+        self.gauge('system.mem.page_pct_free')
 
     def check(self, agentConfig):
         try:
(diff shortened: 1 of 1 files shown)


irabinovitch:
elasticsearch failures seem unrelated. checking.

gmmeyer:
The failure is gone now, so I think it was unrelated.

checks/system/win32.py
```diff
@@ -178,6 +183,13 @@ def check(self, agentConfig):
             pct_usable = float(usable) / total
             self.save_sample('system.mem.pct_usable', pct_usable)
 
+        page = psutil.virtual_memory()
+        if page.total is not None:
```
yannmh: Do you know if it's possible to fetch this information from the `Win32_PerfRawData_PerfOS_Memory` WMI class ? See https://github.com/DataDog/dd-agent/blob/master/checks/system/win32.py#L91

gmmeyer: I don't know. I'll investigate.

yannmh:
@gmmeyer Feel free to merge ;)

checks/system/win32.py
gmmeyer: The answer is, in short, we can't because they're different. I cannot, at this moment, articulate the exact difference. I'm pretty sure that `Win32_PerfRawData_PerfOS_Memory` does not give you the actual pagefile usage, which is what we're trying to get here.

PR #2423, opened 2016-04-18
[mongo] more default metrics ↗

yannmh:
**[mongo] more default metrics ↗**
The following metric topics are now collected by default:
* `durability` - journaling-related operations and performance report.
* `locks` - serverstatus locks report.
* `wiredtiger` - wiredtiger storage option.

Log a warning to inform the user when any of these metric topics is
listed in the configuration file 🔊 .

**[mongo] additional `wiredtiger` cache metrics**
Collect additional WiredTiger metrics, i.e.
* `wiredTiger.cache.modified_pages_evicted`,
* `wiredTiger.cache.unmodified_pages_evicted`
   Number of pages evicted from the cache
* `wiredTiger.cache.tracked_dirty_bytes_in_cache`
   Amount of space taken by dirty data in the cache (bytes).

### DIFF
--- a/checks.d/mongo.py
+++ b/checks.d/mongo.py
@@ -301,7 +301,6 @@ class MongoDb(AgentCheck):
 
     """
     WiredTiger storage engine.
-
     """
     WIREDTIGER_METRICS = {
         "wiredTiger.cache.bytes currently in the cache": (GAUGE, "wiredTiger.cache.bytes_currently_in_cache"),  # noqa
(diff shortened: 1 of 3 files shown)


checks.d/mongo.py
```diff
@@ -407,25 +448,28 @@ def _build_metric_list_to_collect(self, additional_metrics):
...
             additional_metrics = self.AVAILABLE_METRICS.get(option)
-
             if not additional_metrics:
-                self.log.warning(
-                    u"Failed to extend the list of metrics to collect:"
-                    " unrecognized {option} option".format(
-                        option=option
+                if option in self.DEFAULT_METRICS:
+                    self.log.warning(
+                        u"`%s` option is deprecated and will be removed in a future version."
```
remh: Not sure if `will be removed in a future version`, as it just does nothing anymore.
You can just say that it's not needed anymore.

checks.d/mongo.py
```diff
@@ -407,25 +448,28 @@ def _build_metric_list_to_collect(self, additional_metrics):
 
         # Defaut metrics
         metrics_to_collect.update(self.BASE_METRICS)
+        for default_metrics in self.DEFAULT_METRICS.itervalues():
```
remh: shouldn't `BASE_METRICS` be part of the `DEFAULT_METRICS` dictionary ?

remh:
all good besides my 2 nitpicks

PR #2427, opened 2016-04-19
[docker] Ignore mountpoints not accessible by current user

c960657:
On Ubuntu 16.04, both `/run/lxcfs/controllers`- and `/sys/fs`-style mount points are present. Only the latter are accessible by the **dd-agent** user, but `find_cgroup()` returns the former.

With this PR, `find_cgroup()` checks that the mount point is accessible by the current user.


remh:
Seems reasonable to me! Thanks a lot!

PR #2430, opened 2016-04-20
[activemq_xml] add monitoring for standby hosts

remh:
ActiveMQ standby hosts (for HA) are running but do not allow connections
until they are failed over to. When a failover occurs we need to start
capturing metrics, but until then we should silently wait.
This adds a new config parameter which if used will treat a offline host
as 0 successful metrics rather than logging exceptions and presenting an
"agent error" in the datadog console interface.

----
Rebase from #2023
Thanks to @joelvanvelden

### DIFF
--- a/checks.d/activemq_xml.py
+++ b/checks.d/activemq_xml.py
@@ -6,6 +6,7 @@ import requests
 
 # project
 from checks import AgentCheck
+from config import _is_affirmative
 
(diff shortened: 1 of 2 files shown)


olivielpeau:
Looks good, merging.

Thanks again @joelvanvelden!

PR #2433, opened 2016-04-21
[yarn] add extra tags to apps and filter per app state

alanbover:
The commit implements two changes for the datadog yarn plugin.

First, it adds more tags at the applications events. Those are useful to have so we can make aggregations:
* application_name (In experimentation you might have multiple executions of the same application)
* application_queue
* application_user
* application_type

Also it add an option for filter in which states the yarn applications must be to send metrics. This is extremelly useful for long-term running cluster, to allow not to send events for the big amount of FINISHED tasks.


truthbk:
@alanbover thanks a lot for this! 😄 

we'll review as soon as possible!

PR #2440, opened 2016-04-22
[core] Allow empty 'instances' section

c960657:
Having a config YAML file with an empty or missing `instances` section is not allowed.

I suggest lifting this restriction so that an empty section is allowed. Having an empty `instances` section should be equivalent to the YAML file not existing.

Under some circumstances it is convenient to always have the file present and dynamically add the individual entries, e.g when auto-generating the file by your configuration management software (Puppet/Salt/etc.).

This PR removes this restriction.

### DIFF
--- a/config.py
+++ b/config.py
@@ -917,9 +917,9 @@ def load_check_directory(agentConfig, hostname):
         else:
             pass
 
-        # Look for the per-check config, which *must* exist
+        # Look for the per-check config.
         if not check_config.get('instances'):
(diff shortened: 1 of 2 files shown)


truthbk:
@c960657 thank you for your contribution, we will review this shortly.

remh:
Thanks a lot @c960657 .
I get your use case, unfortunately we won't merge this pull request, as in most cases, not having an instance defined in a *.yaml file would be a symptom of a configuration error that would silently be ignored with your proposal.

PR #2441, opened 2016-04-25
[win32_event_log] `event_format` option 🎨

yannmh:
The (optional) `event_format` parameter will instruct the check to
generate
Datadog's event bodies with the specified list of event properties.
If unspecified, the EventLog's `Message` or `InsertionStrings` are used
by default.
Available properties are:
  - Logfile
  - SourceName
  - EventCode
  - Message
  - InsertionStrings
  - TimeGenerated
  - Type

### DIFF
--- a/checks.d/win32_event_log.py
+++ b/checks.d/win32_event_log.py
@@ -17,22 +17,32 @@ from utils.timeout import TimeoutException
 SOURCE_TYPE_NAME = 'event viewer'
 EVENT_TYPE = 'win32_log_event'
 
+
 class Win32EventLogWMI(WinWMICheck):
+    # WMI information
     EVENT_PROPERTIES = [
-        "Message",
+        "EventCode",
         "SourceName",
         "TimeGenerated",
         "Type",
-        "User",
+    ]
+    EXTRA_EVENT_PROPERTIES = [
         "InsertionStrings",
-        "EventCode"
+        "Message",
+        "Logfile",
     ]
     NAMESPACE = "root\\CIMV2"
-    CLASS = "Win32_NTLogEvent"
+    EVENT_CLASS = "Win32_NTLogEvent"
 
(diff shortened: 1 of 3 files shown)


checks.d/win32_event_log.py
```diff
@@ -155,16 +178,43 @@ def _msg_title(self):
...
+            # Override when verbosity
+            if 'Message' in self.event:
+                msg_text = "{message}\n".format(message=self.event['Message'])
+            elif 'InsertionStrings' in self.event:
+                msg_text = "\n".join([i_str for i_str in self.event['InsertionStrings']
+                                      if i_str.strip()])
+
+            if self.notify_list:
+                msg_text += "\n{notify_list}".format(
+                    notify_list=' '.join([" @" + n for n in self.notify_list]))
```
olivielpeau: The `notify_list` should probably be added to the message body even when a specific `_format` is specified, right? (even though that option is not documented and should probably be deprecated)

checks.d/win32_event_log.py
```diff
@@ -107,16 +127,18 @@ def check(self, instance):
                 u"[Win32EventLog] WMI query timed out."
                 u" class={wmi_class} - properties={wmi_properties} -"
                 u" filters={filters} - tags={tags}".format(
-                    wmi_class=self.CLASS, wmi_properties=self.EVENT_PROPERTIES,
+                    wmi_class=self.EVENT_CLASS, wmi_properties=self.EVENT_PROPERTIES,
```
olivielpeau: `self.EVENT_PROPERTIES` -> `event_properties`

olivielpeau:
All good apart from my 2 small comments!

checks.d/win32_event_log.py
yannmh: I agree.

PR #2442, opened 2016-04-25
[source] Pin setuptools version

olivielpeau:
Fixes the following error when installing the agent with the source install script:

```
* Setting up setuptools
It looks like you hit an issue when trying to install the Datadog agent.
###
  File "/usr/lib64/python2.7/zipfile.py", line 811, in _RealGetContents
    raise BadZipfile, "File is not a zip file"
zipfile.BadZipfile: File is not a zip file
###
```

Cause: the latest versions of setuptools (`20.10.0` and `20.10.1`) can't be downloaded by
`ez_setup.py`, so make ez_setup download a version that is actually
available.

See also https://github.com/pypa/setuptools/issues/557


remh:
👍

PR #2443, opened 2016-04-25
[marathon] Remove versions metric

### LINKED ISSUE #1861: [marathon] Marathon plugin slows down agent when marathon has many apps running
We are monitoring a marathon framework using datadog which has over 150 apps, and the marathon check seems to be slowing down the entire datadog process.

After investigating what the plugin actually does, the problem seems to be this loop: https://github.com/DataDog/dd-agent/blob/5.4.4/checks.d/marathon.py#L46.  It appears that the agent is sequentially hitting the API 150 times, which is enough to stop the agent from reporting metrics long enough to trigger some of our other alerts.

remh:
This metric is not really useful (nor used) but it’s really costly to collect.
Let’s remove it.

Fix #1861

### DIFF
--- a/checks.d/marathon.py
+++ b/checks.d/marathon.py
@@ -56,12 +56,6 @@ class Marathon(AgentCheck):
                     if attr in app:
                         self.gauge('marathon.' + attr, app[attr], tags=tags)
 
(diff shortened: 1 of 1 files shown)


olivielpeau:
Makes sense, merging

PR #2448, opened 2016-04-26
[http_check] add support for no_proxy environment variable

truthbk:
## Why this PR?
If a `no_proxy` environment variable is set we should honor it when making requests. This PR addresses the issue by adding the environment variable to the `proxies` dict used by `requests` to make the request.

### Questions
- Should we allow overriding of `no_proxy` in the yaml configuration file?
- Should we "inherit" `http` and `https` system-wide environment variables as well if they're not set in the configuration file?

### Answers
- I have allowed to override any proxy settings with a per instance `no_proxy` flag.
- If no `datadog.conf` proxy is enabled, but environment variables are found for https or http proxies, these will be enforced.

#### Notes
Unfortunately this is now a little more convoluted then I would've liked. Passing the `self.proxies` dict with `http` and/or `https` and a `no_proxy` key (or `no` key `requests` also takes) will not work, only after popping the relevant entries off the dictionary did I see the expected behavior. This has forced us to enforce the proxy settings explicitly per iteration.

### DIFF
--- a/checks.d/http_check.py
+++ b/checks.d/http_check.py
@@ -5,6 +5,7 @@
 # stdlib
 from datetime import datetime
 import os.path
+from os import environ
 import re
 import socket
 import ssl
@@ -150,9 +151,11 @@ class HTTPCheck(NetworkCheck):
     def __init__(self, name, init_config, agentConfig, instances):
         self.ca_certs = init_config.get('ca_certs', get_ca_certs_path())
         proxy_settings = get_proxy(agentConfig)
-        if not proxy_settings:
-            self.proxies = None
-        else:
+        self.proxies = {
+            "http": None,
+            "https": None,
+        }
+        if proxy_settings:
             uri = "{host}:{port}".format(
                 host=proxy_settings['host'],
                 port=proxy_settings['port'])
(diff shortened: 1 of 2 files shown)


checks.d/http_check.py
```diff
@@ -161,10 +164,15 @@ def __init__(self, name, init_config, agentConfig, instances):
...
                     uri=uri)
-            self.proxies = {
-                'http': "http://{uri}".format(uri=uri),
-                'https': "https://{uri}".format(uri=uri)
-            }
+            self.proxies['http'] = "http://{uri}".format(uri=uri)
+            self.proxies['https'] = "https://{uri}".format(uri=uri)
+        else:
+            self.proxies['http'] = environ.get('http', None)
+            self.proxies['https'] = environ.get('https', None)
```
gmmeyer: Why `http` and `https` as the environment variables? Those seem very prone to unintended overlap. 

(The requests library uses [HTTP_PROXY and HTTPS_PROXY](http://docs.python-requests.org/en/v2.6.2/user/advanced/#proxies).)

truthbk: Good catch! That was my bad, indeed the env variables on *nix systems have `_PROXY` appended. Fixing!

gmmeyer:
I think the apache failure is unrelated, if you're more concerned about it, rerun it. 

So, it looks good to me. 👍

n0ts:
Hi, Why no_proxy default setting is `False`?
I think that normally no use proxy.

PR #2449, opened 2016-04-26
[core] Status JSON and Flare options

JohnLZeller:
This PR has the agent drop json for use by the new gui. In addition, I've also added some command line options to the flare command, to make it usable without interaction at the command line.

### DIFF
--- a/checks/check_status.py
+++ b/checks/check_status.py
@@ -10,10 +10,10 @@ of checks.
 from collections import defaultdict
 import cPickle as pickle
 import datetime
-import json
 import logging
 import os
 import platform
+import simplejson as json
 import sys
 import tempfile
 import time
--- a/checks/check_status.py
+++ b/checks/check_status.py
@@ -10,6 +10,7 @@ of checks.
 from collections import defaultdict
 import cPickle as pickle
 import datetime
+import json
 import logging
 import os
 import platform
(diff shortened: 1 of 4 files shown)


checks/check_status.py
```diff
@@ -10,6 +10,7 @@
 from collections import defaultdict
 import cPickle as pickle
 import datetime
+import json
```
remh: simplejson

remh:
Can you describe a bit more how and why those features will be used ?

JohnLZeller:
Sure! So, I've made two changes here.

Firstly, currently we pickle python objects for that status of the collector, forwarder, and dogstatsd. The current GUI loads these pickled objects, and calls `.to_dict()` on each of them, before accessing its' attributes. With the changes here, the pickling still happens (can be removed later), but we are also simply dumping the dictionaries for each of these status objects, as json. There are now JSON dumps for the collector, forwarder, dogstatsd, and an additional one with basic agent stats like version, architecture, etc. That last one doesn't need to be dumped as often, but this gets the information accessible for now.

Secondly, the flare command currently only takes a single command line arg, for Zendesk ticket number. I've added option parameters for supplying an email, and for skipping any interactive command line stuff (such as confirming that you'd really, really like to send a flare). This makes it very easy to call the flare command from the node application (the gui).

JohnLZeller:
@hkaj would you mind taking a look?

gmmeyer:
😢

PR #2451, opened 2016-04-28
[util] Fix get_hostname for windows ec2 instances

hkaj:
Right now `get_hostname` never resolves to an EC2 instance id for windows host because the condition to look for it didn't include empty hostnames. This PR fixes that.

### DIFF
--- a/util.py
+++ b/util.py
@@ -223,8 +223,12 @@ def get_hostname(config=None):
             if unix_hostname and is_valid_hostname(unix_hostname):
                 hostname = unix_hostname
 
-    # if we have an ec2 default hostname, see if there's an instance-id available
-    if (Platform.is_ecs_instance()) or (hostname is not None and True in [hostname.lower().startswith(p) for p in [u'ip-', u'domu']]):
+    # if the host is an ECS worker, or has an EC2 hostname
+    # or it's a windows machine and the EC2 config service folder exists
+    # try and find an EC2 instance ID
+    if (Platform.is_ecs_instance()) or \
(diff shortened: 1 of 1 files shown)


util.py
```diff
@@ -223,8 +223,11 @@ def _get_hostname_unix():
...
                 hostname = unix_hostname
 
-    # if we have an ec2 default hostname, see if there's an instance-id available
-    if (Platform.is_ecs_instance()) or (hostname is not None and True in [hostname.lower().startswith(p) for p in [u'ip-', u'domu']]):
+    # if the host is an ECS worker, or has an EC2 hostname, or no hostname but it's a windows machine
+    # try and find a EC2 instance ID
+    if (Platform.is_ecs_instance()) or \
+       (hostname is not None and True in [hostname.lower().startswith(p) for p in [u'ip-', u'domu']]) or \
+       (hostname is None and os_name == 'windows'):
         instanceid = EC2.get_instance_id(config)
```
yannmh: Up to this bloc, the code never attempts to resolve the Windows hostname as `_get_hostname_unix` is called on UNIX system only. The `hostname == None`, and we have no clue whether the host belongs or not to EC2. 

It'd be better if we can get a sense of it and spare undesired call to the `EC2.get_instance_id` method. We could for instance eval the `socket. gethostname` value sooner on Windows systems.

hkaj:
@yannmh Thanks for the feedback, `socket.gethostname` is not relevant in this case as it will return something like `WIN-XXXXXX` which doesn't help figuring out wether this is an EC2 host or not, but I used something less costly to determine that (`os.path.exists`). Let me know what you think.

yannmh:
Looks good. Thanks @hkaj.

PR #2459, opened 2016-05-02
[integration][vsphere] Add optional vm include parameter

JohnLZeller:
This adds the ability to set a custom field from the vSphere PowerCLI which acts as a marker for which VMs you would like to have monitored by the agent. In order to add this custom field, a user can use this command: `Get-VM <MyVMName> | Set-CustomField -Name "DatadogMonitored" -Value "DatadogMonitored"`

The ideal would have been to use the new tagging feature brought in vSphere 6.0 (usable also through the vSphere Web Client), but the vSphere API (at least in vCenter 5.5) does not expose these tags. There is an attribute on the managed object for VirtualMachine, called tag. Unfortunately, this is actually not the same tag set as the ones set in the vSphere Web Client (at least in vCenter 5.5). The tags set in the web client are instead stored in the Inventory Service.

Additionally, Custom Attributes could have been an option, which can be set via PowerCLI or Web Client (I think), but Custom Attributes were deprecated in exchange for Tags in vSphere 6.0.

Alas, this fix gives the ability for all. When vSphere releases another update that makes it easy to access the tags set on VirtualMachine managed objects, then we can add a feature to use tags.

### DIFF
--- a/checks.d/vsphere.py
+++ b/checks.d/vsphere.py
@@ -17,6 +17,7 @@ from pyVim import connect
 from pyVmomi import vim # pylint: disable=E0611
 
 # project
+from config import _is_affirmative
 from checks import AgentCheck
 from checks.libs.thread_pool import Pool
 from checks.libs.vmware.basic_metrics import BASIC_METRICS
@@ -26,6 +27,8 @@ from util import Timer
 SOURCE_TYPE = 'vsphere'
 REAL_TIME_INTERVAL = 20  # Default vCenter sampling interval
 
+# Metrics are only collected on vSphere VMs marked by custom field value
+VM_MONITORING_FLAG = 'DatadogMonitored'
 # The size of the ThreadPool used to process the request queue
 DEFAULT_SIZE_POOL = 4
 # The interval in seconds between two refresh of the entities list
@@ -455,6 +458,8 @@ class VSphereCheck(AgentCheck):
 
         if i_key not in self.server_instances:
             try:
+                # Object returned by SmartConnect is a ServerInstance
+                #   https://www.vmware.com/support/developer/vc-sdk/visdk2xpubs/ReferenceGuide/vim.ServiceInstance.html
                 server_instance = connect.SmartConnect(
                     host = instance.get('host'),
                     user = instance.get('username'),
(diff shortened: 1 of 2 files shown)


JohnLZeller:
I've tested this. I had the agent running with `include_only_marked: false`, then I set it to `true`, and then I added the custom field to the `japan` vm. As you can see from this screenshot, the metrics coming in, reported correctly. I've included a host in this screenshot as proof that the host itself didn't also stop reporting.

Note: `Staging` and `japan` are the VMs.

screen shot 2016-05-02 at 6 18 53 pm

checks.d/vsphere.py
```diff
@@ -633,9 +649,10 @@ def _cache_morlist_raw(self, instance):
             'host_include': instance.get('host_include_only_regex'),
             'vm_include': instance.get('vm_include_only_regex')
         }
+        include_only_marked = instance.get('include_only_marked')
```
truthbk: Should probably default to `False`:
`include_only_marked = instance.get('include_only_marked', False)`

we should also probably also use `is_affirmative()` so the user can be a little more loose in the way they specify if the flag is enabled.

truthbk:
Looks good other than the nitpick.

JohnLZeller:
@truthbk thanks for the note! I took your suggestions, and made the changes. How's it look now?

checks.d/vsphere.py
```diff
@@ -602,6 +607,18 @@ def _cache_morlist_raw_atomic(self, i_key, obj_type, obj, tags, regexes=None):
                 if not match:
                     self.log.debug(u"Filtered out VM {0} because of vm_include_only_regex".format(obj.name))
                     return
+            # Also, if include_only_marked is true, then check if there exists a
+            # custom field with the value DatadogMonitored
+            if _is_affirmative(include_only_marked):
```
truthbk: nitpick, but feel free to ignore: I'd probably put `is_affirmative()` here:

`include_only_marked = _is_affrmative(instance.get('include_only_marked', False))` on line 653

truthbk:
👍  squash and merge as soon as CI passes! Thanks!

PR #2462, opened 2016-05-02
[core] fixes regression in windows build process

gmmeyer:
There was a regression in the windows build. Honestly, I think that I caused it. 

This PR will fix it.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -92,6 +92,7 @@ if sys.platform == 'win32':
         'checks.libs.vmware.*',
         'httplib2',
         'utils.containers',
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -92,6 +92,7 @@
         'checks.libs.vmware.*',
         'httplib2',
         'utils.containers',
+        'scandir',
```
truthbk: dependy brought in by 6ede57d6

truthbk:
Looks good to me. Travis error looks unrelated (postgres). 👍

PR #2463, opened 2016-05-03
[mysql] improving accuracy for query exec time. Fixes issue #2406.

### LINKED ISSUE #2406: [mysql] Inaccurate averaging of query run time
The averaging of MySQL query run time is prone to biasing, because it's an unweighted average of per statement averages. Imagine a common statement runs 100 times, with an average latency (avg_timer_wait) of 10 milliseconds, and a rare statement runs only once, but it takes a full second to execute, so its avg_timer_wait is 1000 milliseconds. Then the AVG(avg_timer_wait) as calculated currently would be 505 milliseconds, which is artificially inflated by the one slow query. See https://github.com/DataDog/dd-agent/blob/master/checks.d/mysql.py#L1186
```
sql_avg_query_run_time = """
    SELECT schema_name, 
    SUM(count_star) cnt,
    ROUND(AVG(avg_timer_wait)/1000000) AS avg_us
FROM performance_schema.events_statements_summary_by_digest
WHERE schema_name IS NOT NULL
GROUP BY schema_name"""
```

truthbk:
See #2406.

- `ROUND((SUM(sum_timer_wait) / SUM(count_star)) / 1000000)`: represents the real average `timer_wait` per query for a given schema. Here, `sum_timer_wait` represents the sum of all `timer_wait` times for a certain query, and `count_star` is the number of times a certain query was run. 
- formerly `ROUND(AVG(avg_timer_wait)/1000000)` actually represented the average of averages.

The discrepancy between the two computations may be considerable.

### DIFF
--- a/checks.d/mysql.py
+++ b/checks.d/mysql.py
@@ -1188,7 +1188,7 @@ class MySql(AgentCheck):
         # Fetches the avg query execution time per schema and returns the
         # value in microseconds
 
-        sql_avg_query_run_time = """SELECT schema_name, SUM(count_star) cnt, ROUND(AVG(avg_timer_wait)/1000000) AS avg_us
+        sql_avg_query_run_time = """SELECT schema_name, ROUND((SUM(sum_timer_wait) / SUM(count_star)) / 1000000) AS avg_us
             FROM performance_schema.events_statements_summary_by_digest
             WHERE schema_name IS NOT NULL
             GROUP BY schema_name"""
@@ -1204,7 +1204,7 @@ class MySql(AgentCheck):
                 schema_query_avg_run_time = {}
                 for row in cursor.fetchall():
                     schema_name = str(row[0])
-                    avg_us = long(row[2])
+                    avg_us = long(row[1])
 
                     # set the tag as the dictionary key
                     schema_query_avg_run_time["schema:{0}".format(schema_name)] = avg_us


truthbk:
- `ROUND((SUM(sum_timer_wait) / SUM(count_star)) / 1000000)`: represents the real average `timer_wait` per query for a given schema. Here, `sum_timer_wait` represents the sum of all `timer_wait` times for a certain query, and `count_star` is the number of times a certain query was run. 
- formerly `ROUND(AVG(avg_timer_wait)/1000000)` actually represented the average of averages.

Surprisingly the discrepancy between the two computations is considerable.

checks.d/mysql.py
```diff
@@ -1188,7 +1188,7 @@ def _query_exec_time_per_schema(self, db):
         # Fetches the avg query execution time per schema and returns the
         # value in microseconds
 
-        sql_avg_query_run_time = """SELECT schema_name, SUM(count_star) cnt, ROUND(AVG(avg_timer_wait)/1000000) AS avg_us
+        sql_avg_query_run_time = """SELECT schema_name, SUM(count_star) cnt, ROUND((SUM(sum_timer_wait) / SUM(count_star)) / 1000000) AS avg_us
```
olivielpeau: kind of unrelated, but we could remove `SUM(count_star) cnt` from the `SELECT` as we don't use it right?

olivielpeau:
Had only one comment, feel free to ignore it though. Looks good to me! 👍

truthbk:
Addressed @olivielpeau comment. Tested, CI tests passing, so merging. :)

PR #2465, opened 2016-05-04
Check if a container is excluded before counting as running

yenif:
I don't know if the current behavior of including excluded containers in docker.containers.running is expected behavior or if changing this will trigger compatibility issues, but this is how I would expect excluded containers to behave.

Thanks
Brian

### DIFF
--- a/checks.d/docker_daemon.py
+++ b/checks.d/docker_daemon.py
@@ -306,16 +306,16 @@ class DockerDaemon(AgentCheck):
         for container in containers:
             container_name = DockerUtil.container_name_extractor(container)[0]
 
+            # Check if the container is included/excluded via its tags
+            if self._is_container_excluded(container):
+                self.log.debug("Container {0} is excluded".format(container_name))
+                continue
+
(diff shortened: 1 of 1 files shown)


gmmeyer:
First, thank you very much for your contributions. We're always really happy for them!

This does seem more reasonable, and more like the behavior I'd expect, but I can't say more without being more thorough. 

We'll test and review it soon! Thank you very much! 😄

yenif:
In my experience this was required to ignore kubernetes pause containers `gcr.io/google_containers/pause` so that dd-agent correctly counts and reports the number of containers running. My understanding is that the number of containers running directly affects billing, thus without this patch containers are effectively double billed because k8s runs a pause container for every real container.

How are DataDog's other customers running k8s dealing with the above? Are they just eating the cost because they are unaware of the issue? Is there some other method of dealing with this I'm not aware of?

hkaj:
Billing is not impacted as we count the docker.cpu.user metric for the number of running containers.
You can use the same metric to get the filtered count, but it's still useful to get the full count, closing.

PR #2467, opened 2016-05-04
[Config] Update config logic to look for checks in 3rd-party directory (rebased)

olivielpeau:
Based on @tmichelet's work in #2273, rebased on top of master (i.e. it now handles configs from the service discovery).

When we load the checks, we take into account the configs from files (`.yaml` and `.default.yaml`) first, and then the configs from service discovery for the checks that haven't been loaded yet.

Also, this changes the logic of `get_check_class` in `utils.checkfiles` to take into account checks from 3rd-party dirs.

cc @hkaj 

Original PR message:
```
- Add tests on config's logic
- Refactor logic
- Update logic to look for checks in 3rd-party directory
```

### DIFF
--- a/config.py
+++ b/config.py
@@ -694,6 +694,19 @@ def get_checksd_path(osname=None):
         return _unix_checksd_path()
 
 
+def get_3rd_party_path(osname=None):
+    if not osname:
+        osname = get_os()
+    if osname in ['windows', 'mac']:
+        raise PathNotFound()
+
+    cur_path = os.path.dirname(os.path.realpath(__file__))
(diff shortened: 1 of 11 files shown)


olivielpeau:
@hkaj I've tested the changes on what's directly related to 3rd-party checks, but I haven't set up and tested the service discovery part. Could you give this a try?

hkaj:
@olivielpeau tested service discovery with these changes, it works as expected. Do you also need a review or that was already taken care of?

hkaj:
LGTM :shipit:

PR #2472, opened 2016-05-05
[dogstatsd][sc] Fix parsing of tag key ending with `m`

### LINKED ISSUE #2373: [dogstatsd] Service checks: message field parsing confused by `m:`
See https://github.com/DataDog/dd-agent/commit/1ca4a67163a4e563e558e4d7b2953453ee3ce152

The test fails because `m:` is regarded as the start of a message field when it shouldn't be (it's just part of the tag).

See also https://github.com/DataDog/dd-agent/blob/b09211d5410bc4ff874e09381908286914cf5271/aggregator.py#L557 for the line of code that's probably responsible

olivielpeau:
On service checks, tags which keys finished with `m:` were mistaken for
the beginning of a message metadata field.

Fix the behavior, and add a test.

Fixes #2373

### DIFF
--- a/aggregator.py
+++ b/aggregator.py
@@ -558,7 +558,7 @@ class Aggregator(object):
                 'status': int(status)
             }
 
-            message_delimiter = '|m:' if '|m:' in metadata else 'm:'
+            message_delimiter = 'm:' if metadata.startswith('m:') else '|m:'
             if message_delimiter in metadata:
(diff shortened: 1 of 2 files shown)


remh:
👍 if tests pass

olivielpeau:
CI passes. Merging.

gphat:
Thanks @olivielpeau and @remh!

PR #2473, opened 2016-05-06
[yarn] Consolidate Tags

zachradtka:
This is mainly a bug fix to reduce the number of tags created by the YARN agent check.

The major fixes are:
* Only get metrics from running applications
* Tag metrics with `cluster_name` instead of `cluster_id`
* Tag metrics with `app_name` instead of `app_id`


gmmeyer:
Thanks a lot for your contribution!

This looks pretty decent, it passes the checks. We will review and test this soon to make sure it all works! 😄

checks.d/yarn.py
```diff
@@ -146,61 +150,60 @@ def check(self, instance):
...
 
-                        cluster_id = app_json['clusterId']
-                        app_id = app_json['id']
+                        app_name = app_json['name']
 
-                        tags = ['cluster_id:' + str(cluster_id), 'app_id:' + str(app_id)]
+                        tags = ['app_name:%s' % str(app_name)]
+                        tags.extend(addl_tags)
 
                         self._set_yarn_metrics_from_json(tags, app_json, YARN_APP_METRICS)
```
olivielpeau: The check now tags these metrics by `app_name` instead of `app_id`, so there can be duplicate sets of tags for the same metric (when multiple apps have the same name). So we should use `increment` instead of `gauge` to send these metrics

checks.d/yarn.py
```diff
@@ -146,61 +150,60 @@ def check(self, instance):
...
 
-    def _get_cluster_id(self, rm_address):
-        '''
-        Return the cluster ID, otherwise raise an exception
-        '''
-        info_json = self._rest_request_to_json(rm_address, INFO_URI)
+        # Get the cluster name from the conf file
+        cluster_name = instance.get('cluster_name')
+        if cluster_name is None:
+            raise Exception('The cluster_name must be specified in the instance configuration')
```
olivielpeau: The YARN check has already been released so it'd better to be compatible with the previous configuration format. Instead of raising an exception here could you use `self.warning("The cluster_name must be specified in the instance configuration, defaulting to 'default_cluster'")` and set the `cluster_name` to `default_cluster` by default?

This would let the user know that a `cluster_name` needs to be set for the check to work best, without making the check suddenly stop working when the agent is upgraded.

olivielpeau:
@zachradtka Made 2 comments, once they're addressed we'll test the check on our clusters, and we may have more feedback then :)

Thanks!

checks.d/yarn.py
```diff
@@ -146,61 +150,60 @@ def check(self, instance):
...
+        if tags is None:
+            tags = []
+        else:
+            tags = list(set(tags))
 
-    def _get_cluster_id(self, rm_address):
-        '''
-        Return the cluster ID, otherwise raise an exception
-        '''
-        info_json = self._rest_request_to_json(rm_address, INFO_URI)
```
olivielpeau: the `INFO_URI` var should be removed from the check too

tests/checks/mock/test_yarn.py
```diff
@@ -7,19 +7,16 @@
...
-# Resource manager address
+# IDs
+CLUSTER_NAME = 'SparkCluster'
+
+# Resource manager URI
 RM_ADDRESS = 'http://localhost:8088'
 
-# Path to retrieve cluster info
+# Service URLs
 INFO_URL = urljoin(RM_ADDRESS, '/ws/v1/cluster/info')
```
olivielpeau: This is not used anymore afaik

zachradtka:
@olivielpeau I have updated the check and tests to remove the `INFO_URL` and to ensure that a warning is displayed if the user does not specify a `cluster_name` in the YAML

olivielpeau:
Looks good @zachradtka! Could you squash your commits please? I'll merge once that's done, thanks!

PR #2474, opened 2016-05-06
[mapreduce] Consolidate tags

zachradtka:
This is mainly a bugfix to reduce the number of tags collected with the MapReduce agent check.

The major fixes are:
* Commenting out the counter options in the YAML
* Tag metrics with `cluster_name` instead of `cluster_id`


gmmeyer:
Thank you very much for your contribution! 

We'll review and test this very soon, but it looks pretty good. The failing test looks unrelated, so I wouldn't worry about that.

checks.d/mapreduce.py
```diff
@@ -127,8 +127,19 @@ def check(self, instance):
...
+        tags = instance.get('tags', [])
+        if tags is None:
+            tags = []
+        else:
+            tags = list(set(tags))
+
+        # Get the cluster name from the conf file
+        cluster_name = instance.get('cluster_name')
+        if cluster_name is None:
+            raise Exception('The cluster_name must be specified in the instance configuration')
```
olivielpeau: same comment as the YARN check, it'd be better to add a warning and set a default value instead of raising an exception.

olivielpeau:
@zachradtka Looks pretty good, only had one comment similar to my suggestion on the yarn PR.

zachradtka:
@olivielpeau I have updated the check and tests to remove the `INFO_URL` and to ensure that a warning is displayed if the user does not specify a `cluster_name` in the YAML

olivielpeau:
Thanks @zachradtka, looks good, can you squash your commits please?

olivielpeau:
Thanks! Merging

PR #2475, opened 2016-05-06
[source_install] updating setuptools ez_setup.py to new recommended location.

truthbk:
### Why?

Setuptools has moved to github from bitbucket. As per the github readme the new preferred location is:
https://bootstrap.pypa.io/ez_setup.py


olivielpeau:
👍

PR #2359, opened 2016-03-17
[disk] add a regex option to exclude disks based on mountpoint

rhwlo:
#### summary

This changeset adds an `exclude_mountpoint_re` option the disk integration configuration and corrects a small typo in the disk integration configuration template.

#### motivation

Nagios’s `check_disk` plugin has a `-i` option to ignore paths or partitions based off of regular expressions, and this will add the same functionality to Datadog.

#### testing

I’ve added tests to `mock/test_disk.py` alongside the other disk tests to verify that the regexes exclude what we want and nothing else.

PR #2476, opened 2016-05-06
[disk] Fix disk check exclusion

masci:
#2359 introduced a regression under the following circumstances:
 * device name is `''`
 * `excluded_disk_re` is not specified in the config file.

In that case, the default pattern for `excluded_disk_re` was `^$` thus making `_exclude_disk` return before evaluating `excluded_mountpoint_re`.

Also added a specific test for this issue and fixed the logic with a more verbose but easier to read pattern.

### DIFF
--- a/checks.d/disk.py
+++ b/checks.d/disk.py
@@ -150,7 +150,7 @@ class Disk(AgentCheck):
         # device name matches `excluded_disk_re`
         elif not name_empty and self._excluded_disk_re.match(name):
             return True
-        # device name matches `excluded_mountpoint_re`
+        # device mountpoint matches `excluded_mountpoint_re`
         elif self._excluded_mountpoint_re.match(mountpoint):
             return True
         # fs is listed in `excluded_filesystems`
--- a/checks.d/disk.py
+++ b/checks.d/disk.py
@@ -135,13 +135,30 @@ class Disk(AgentCheck):
                                           part.fstype == '')) or
                 self._exclude_disk(part.device, part.fstype, part.mountpoint))
 
-    # We don't want all those incorrect devices
     def _exclude_disk(self, name, filesystem, mountpoint):
-        return (((not name or name == 'none') and not self._all_partitions) or
-                name in self._excluded_disks or
-                self._excluded_disk_re.match(name) or
-                self._excluded_mountpoint_re.match(mountpoint) or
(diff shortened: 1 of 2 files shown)


checks.d/disk.py
```diff
@@ -135,13 +135,30 @@ def _exclude_disk_psutil(self, part):
...
+        # allow empty names if `all_partitions` is `yes` so we can evaluate mountpoints
+        if name_empty and not self._all_partitions:
+            return True
+        # device is listed in `excluded_disks`
+        elif not name_empty and name in self._excluded_disks:
+            return True
+        # device name matches `excluded_disk_re`
+        elif not name_empty and self._excluded_disk_re.match(name):
+            return True
+        # device name matches `excluded_mountpoint_re`
```
masci: this should be "device mountpoint"

checks.d/disk.py
```diff
@@ -135,13 +135,30 @@ def _exclude_disk_psutil(self, part):
...
+        name_empty = not name or name == 'none'
+
+        # allow empty names if `all_partitions` is `yes` so we can evaluate mountpoints
+        if name_empty and not self._all_partitions:
+            return True
+        # device is listed in `excluded_disks`
+        elif not name_empty and name in self._excluded_disks:
+            return True
+        # device name matches `excluded_disk_re`
+        elif not name_empty and self._excluded_disk_re.match(name):
```
olivielpeau: small regression in the logic here: by adding the condition `not name_empty` before evaluating `self._excluded_disk_re.match(name)`, in the case where no exclusion is specified in the yaml, devices with actual empty names (`''`) would not get excluded anymore.

olivielpeau: IMO we should have a way to determine whether an exclude regex rule has been specified or not instead of applying the default of `^$`. Then we could simply apply the exclude regex _only if_ it has been specified in the yaml file. What do you think?

olivielpeau: Now that I think about it what I've described is a very minor regression, the behavior you implemented makes more sense.

The worst possible outcome of this "regression" would be that users who have enabled the collection on all partitions end up with slightly more metrics collected. So safe to ignore my comment :)

olivielpeau:
@masci: Had a little discussion with myself, but apart from your own comment the PR looks good to me!

PR #2477, opened 2016-05-06
[checks] [process] add pagefault stats

gmmeyer:
Mostly from PR #2363.

I fixed the failing test and rebased against master.

### DIFF
--- a/checks.d/process.py
+++ b/checks.d/process.py
@@ -34,10 +34,10 @@ ATTR_TO_METRIC = {
     'w_bytes':          'iowrite_bytes',  # FIXME: namespace me correctly (6.x) io.w_bytes
     'ctx_swtch_vol':    'voluntary_ctx_switches',  # FIXME: namespace me correctly (6.x), ctx_swt.voluntary
     'ctx_swtch_invol':  'involuntary_ctx_switches',  # FIXME: namespace me correctly (6.x), ctx_swt.involuntary
-    'minflt':           'mem.minflt',
-    'cminflt':          'mem.cminflt',
-    'majflt':           'mem.majflt',
-    'cmajflt':          'mem.cmajflt',
+    'minflt':           'mem.page_faults.minor_faults',
+    'cminflt':          'mem.page_faults.children_minor_faults',
+    'majflt':           'mem.page_faults.major_faults',
+    'cmajflt':          'mem.page_faults.children_major_faults',
 }
 
 
@@ -252,6 +252,11 @@ class ProcessCheck(AgentCheck):
                 st['cminflt'].append(cminflt)
                 st['majflt'].append(majflt)
                 st['cmajflt'].append(cmajflt)
+            else:
+                st['minflt'].append(None)
+                st['cminflt'].append(None)
+                st['majflt'].append(None)
+                st['cmajflt'].append(None)
 
         return st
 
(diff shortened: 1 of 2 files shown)


gmmeyer:
I cancelled the failed check, it's just two lines of comments and I didn't think it was necessary to run the whole test suite again for two comments. The checks got run in the previous commit and passed.

checks.d/process.py
```diff
@@ -241,8 +245,33 @@ def get_process_state(self, name, pids):
...
+
+        def file_to_string(path):
+            with open(path, 'r') as f:
+                res = f.read()
+            return res
+
+        # http://man7.org/linux/man-pages/man5/proc.5.html
+        try:
+            data = file_to_string('/proc/%s/stat' % pid)
+        except:
```
olivielpeau: let's avoid the bare exception here, and do `except Exception:` instead

checks.d/process.py
```diff
@@ -241,8 +245,33 @@ def get_process_state(self, name, pids):
...
+            return None
```
olivielpeau: we should log something when this happens. A `debug` level is probably enough though

olivielpeau:
@gmmeyer Added some comments. Could you also squash the commits you authored into one?

gmmeyer:
Yes I can squash them

checks.d/process.py
gmmeyer: Okay

checks.d/process.py
gmmeyer: Okay

gmmeyer:
The failure was activemq. I'm pretty sure it's unrelated. I can run the tests again if you want.

checks.d/process.py
```diff
@@ -34,6 +34,10 @@
     'w_bytes':          'iowrite_bytes',  # FIXME: namespace me correctly (6.x) io.w_bytes
     'ctx_swtch_vol':    'voluntary_ctx_switches',  # FIXME: namespace me correctly (6.x), ctx_swt.voluntary
     'ctx_swtch_invol':  'involuntary_ctx_switches',  # FIXME: namespace me correctly (6.x), ctx_swt.involuntary
+    'minflt':           'mem.minflt',
```
truthbk: Wondering if we should keep these names as (presently) defined on the `manpage` - or if we should make them more human readable. Let's check with the team....

checks.d/process.py
```diff
@@ -241,8 +245,34 @@ def get_process_state(self, name, pids):
             st['r_bytes'].append(ioinfo.get('read_bytes'))
             st['w_bytes'].append(ioinfo.get('write_bytes'))
 
+            pagefault_stats = self.get_pagefault_stats(pid)
+            if pagefault_stats is not None:
+                (minflt, cminflt, majflt, cmajflt) = pagefault_stats
+                st['minflt'].append(minflt)
+                st['cminflt'].append(cminflt)
+                st['majflt'].append(majflt)
+                st['cmajflt'].append(cmajflt)
```
truthbk: I think that `if pagefault_stats is None`, we should append `None`, otherwise I fear it might mess up the map for PID -> stats for subsequent `pids`. See `st['real']` above.

tests/checks/integration/test_process.py
```diff
@@ -266,6 +284,17 @@ def test_check(self, mock_process):
                     value=expected_value
                 )
 
+            # these are just here to ensure it passes the coverage report.
```
truthbk: Could we maybe define a list similar to `PROCESS_METRIC` (or even include the new page fault metrics in there - not sure if that could break other tests) so that we can do these asserts in a loop and repeat less code?

gmmeyer:
I addressed all of these issues in the most recent commit. Let me know if there's anything else. 😃

truthbk:
👍  looks good!

PR #2352, opened 2016-03-15
[network] Support netstat in Windows

mausch:
Implements #2350

PR #2478, opened 2016-05-06
[network] Support netstat in Windows

masci:
Original code from #2352 

Rebased on master + minor fixes

### DIFF
--- a/checks.d/network.py
+++ b/checks.d/network.py
@@ -7,6 +7,7 @@ Collects network metrics.
 """
 # stdlib
 import re
+import itertools
 
 # project
 from checks import AgentCheck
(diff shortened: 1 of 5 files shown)


olivielpeau:
Tested on windows 2012, looks good to me!

@masci Could you squash all the commits into one (and preserve the authorship)? Thanks!

olivielpeau:
Actually it'd be good to add a comment in `network.yaml.default` to specify that on Windows an instance needs to be there for the check to run but that its configuration is ignored.

olivielpeau:
Discussed with @remh about this addition and he was thinking about implementing this with `psutil` (`psutil.net_connections()`) instead of parsing `netstat`'s output.

masci:
Superseded by #2499

PR #2480, opened 2016-05-06
[core] changes coverage failure to opt out, instead of opt in

gmmeyer:
Currently, coverage reports only fail through an opt in system. You have to set `COVERAGE=true`. While it's a good idea not to always have the very verbose output or failures when you're testing something, it can be annoying, it should be opt out instead of opt in. That way, it's easier for people (both community and Datadog) to ensure they have proper coverage in their tests.

### DIFF
--- a/tests/checks/common.py
+++ b/tests/checks/common.py
@@ -300,7 +300,7 @@ WARNINGS
             untested_events=pformat(untested_events),
         ))
 
-        if os.getenv('COVERAGE'):
+        if not os.getenv('NO_COVERAGE'):
             self.assertEquals(coverage_metrics, 100.0)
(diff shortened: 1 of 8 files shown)


tests/checks/common.py
```diff
@@ -300,7 +300,7 @@ def coverage_report(self):
             untested_events=pformat(untested_events),
         ))
 
-        if os.getenv('COVERAGE'):
+        if not os.getenv('NO_COVERAGE'):
```
olivielpeau: nitpick: setting `NO_COVERAGE=false` would still disable the enforcement of the coverage :trollface:

olivielpeau:
👍

PR #2482, opened 2016-05-09
[linux] make /proc configurable.

truthbk:
Allows overriding of `/proc` by defining `procfs_path` in `datadog.conf` - should be enforcable by using `self.agentConfig.get("procfs_path")` from any of the checks. I believe we enforce that pretty much everywhere relevant with this PR.

Places where we haven't overriden it just yet: 
  - https://github.com/DataDog/dd-agent/blob/jaime/config_proc/daemon.py: because we haven't read the configuration yet.
  - https://github.com/DataDog/dd-agent/blob/jaime/config_proc/utils/process.py: because when those functions are called, we haven't read the configuration either.

### DIFF
--- a/checks.d/agent_metrics.py
+++ b/checks.d/agent_metrics.py
@@ -113,6 +113,7 @@ class AgentMetrics(AgentCheck):
         return self._collector_payload, self._metric_context
 
     def check(self, instance):
+
         if self.in_developer_mode:
             stats, names_to_metric_types = self._psutil_config_to_stats(instance)
             self._register_psutil_metrics(stats, names_to_metric_types)
--- a/config.py
+++ b/config.py
@@ -581,7 +581,7 @@ def get_config(parse_args=True, cfg_path=None, options=None):
     return agentConfig
 
 
-def get_system_stats():
+def get_system_stats(proc_path=None):
     systemStats = {
         'machine': platform.machine(),
         'platform': sys.platform,
@@ -593,7 +593,10 @@ def get_system_stats():
 
     try:
         if Platform.is_linux(platf):
-            output, _, _ = get_subprocess_output(['grep', 'model name', '/proc/cpuinfo'], log)
+            if not proc_path:
+                proc_path = "/proc"
+            proc_cpuinfo = os.path.join(proc_path,'cpuinfo')
+            output, _, _ = get_subprocess_output(['grep', 'model name', proc_cpuinfo], log)
             systemStats['cpuCores'] = len(output.splitlines())
 
         if Platform.is_darwin(platf) or Platform.is_freebsd(platf):
--- a/datadog.conf.example
+++ b/datadog.conf.example
@@ -111,6 +111,15 @@ use_mount: no
 # `/datadog/check_configs` key in the back-end. If you wish otherwise, uncomment this option
 # and modify its value.
 # sd_template_dir: /datadog/check_configs
+#
+# ========================================================================== #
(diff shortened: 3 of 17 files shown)


config.py
```diff
@@ -593,7 +593,10 @@ def get_system_stats():
 
     try:
         if Platform.is_linux(platf):
-            output, _, _ = get_subprocess_output(['grep', 'model name', '/proc/cpuinfo'], log)
+            if not proc_path:
+                proc_path = "/proc"
+            proc_cpuinfo = "{}/cpuinfo".format(proc_path.rstrip('/'))
```
hkaj: Using `os.path.join` is usually more reliable for this kind of thing.

datadog.conf.example
```diff
@@ -111,6 +111,15 @@ use_mount: no
...
 # and modify its value.
 # sd_template_dir: /datadog/check_configs
+#
+# ========================================================================== #
+# Other                                                                      #
+# ========================================================================== #
+#
+# In some environments we may have the procfs file system mounted in a 
+# miscellaneous location. The procfs_path configuration paramenter allows
+# us to override the standard default localtion '/proc'
```
hkaj: typo: `localtion` --> `location`

checks.d/agent_metrics.py
```diff
@@ -113,6 +114,11 @@ def get_metric_context(self):
         return self._collector_payload, self._metric_context
 
     def check(self, instance):
+
+        if Platform.is_linux():
+            procfs_path = self.agentConfig.get('procfs_path', '/proc').rstrip('/')
+            psutil.PROCFS_PATH = procfs_path
```
hkaj: Would it be possible to move this logic into the definition of the AgentCheck class? I'm pretty sure someone will forget to add this block in a new check at some point and we'll introduce bugs that could be easily avoided if this is called earlier and automatically.

The danger would remain for calls to `/proc` subfolders from `utils/`, `agent.py` and `config.py` but we don't modify these files as often and we are far more careful when doing so.

Let me know what you think.

hkaj:
A few nits, and hopefully we can move some of the duplicated logic in the mother class.

I'm also wondering about these 2:
- https://github.com/DataDog/dd-agent/blob/5.8.2/utils/kubeutil.py#L102-L108 I think it makes sense to support it here, what do you think?
- https://github.com/DataDog/dd-agent/blob/5.8.2/utils/dockerutil.py#L157 this one is trickier as it already starts with `self._docker_root`. It's probably not necessary, the user could just mount the host, custom `/proc` in `/host/proc` on the container so it should be fine.

And the tests are failing on snmpd, I triggered it again in case it's flaky but if it doesn't pass you should have a look.

checks.d/agent_metrics.py
olivielpeau: one other thing: in this check we can't assume that `psutil` has been imported successfully, so we have to check that `psutil is not None` before setting `psutil. PROCFS_PATH`.

Especially important in this check since it's run by default with the agent.

checks.d/agent_metrics.py
```diff
@@ -15,6 +15,7 @@
 from checks import AgentCheck
 from checks.metric_types import MetricTypes
 from config import _is_affirmative
+from utils.platform import Platform
```
hkaj: You need to remove this import, flake8 is failing on it as it's not used.

hkaj:
LGTM

PR #2483, opened 2016-05-09
[cassandra] Add read and write latency ColumnFamily metrics

sethrosenblum:
Read and Write latency are not only a way to figure out how long it takes to run the specified operation against the given columnfamily, the count metric is also the only way to figure out how often we're reading from and writing to a column family.

cc @johnaxel

### DIFF
--- a/conf.d/cassandra.yaml.example
+++ b/conf.d/cassandra.yaml.example
@@ -51,7 +51,7 @@ init_config:
         domain: org.apache.cassandra.metrics
         type: ColumnFamily
         bean_regex:
-          - .*keyspace=.*  
+          - .*keyspace=.*
         name:
           - TotalDiskSpaceUsed
           - BloomFilterDiskSpaceUsed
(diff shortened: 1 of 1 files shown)


johnaxel:
This is a very useful addition but I think it would balloon the number of metrics collected if we don't filter somewhat. For Cassandra 2.2 at least, `ReadLatency` and `WriteLatency` each contain 17 separate metric attributes, which would be collected for every CF. So each column family will add ~30 metrics to the JMX pile. I think we could filter this down to just grab a couple key metrics from this grouping for each CF, though.

olivielpeau:
I agree that it'd be better to scope the filter to the most useful attributes only. @sethrosenblum any insight on that? :)

(assigning to @yannmh)

sethrosenblum:
@johnaxel @olivielpeau I narrowed the attributes to Count (the most important), Mean, 99p and Max.  We could also swap 99p for 95p or drop them altogether if that makes more sense.

conf.d/cassandra.yaml.example
```diff
@@ -66,6 +66,14 @@ init_config:
           - MemtableLiveDataSize
           - MemtableSwitchCount
           - MinRowSize
+          - ReadLatency
+          - WriteLatency
+        attribute:
+          - Value
```
yannmh: `Value` did not return any metrics in my tests (Cassandra `2.1.13`). @sethrosenblum could you confirm that it is valid please?

yannmh:
Looks good to me ✅ 

On Cassandra `2.1`, the change creates 8 additional new metrics per (`keyspace`,`scope`).

PR #2485, opened 2016-05-10
Calculate aggregate replica set health.

bradobro:
The current replSet.health and replSet.state just return the state
of the current node, not the whole replica set. This patch adds
- mongodb.replSet.memberCount
- mongodb.replSet.healthyMemberCount
- mongodb.replSet.faultyMemberCount

healthyMemberCount + faultyMemberCount = memberCount

I also propose changing mongodb.replSet.health to something like mongodb.replSet.memberHealth or something like it, or set it to a healthyMemberCount / memberCount.

### DIFF
--- a/checks.d/mongo.py
+++ b/checks.d/mongo.py
@@ -151,7 +151,10 @@ class MongoDb(AgentCheck):
         "oplog.logSizeMB": GAUGE,
         "oplog.usedSizeMB": GAUGE,
         "oplog.timeDiff": GAUGE,
+        "replSet.faultyCount": GAUGE,
         "replSet.health": GAUGE,
+        "replSet.healthyCount": GAUGE,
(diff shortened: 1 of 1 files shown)


bradobro:
My first time contributing. We needed this right now in production, so I patched 5.7.x.

If you like it and want the patch against master or somewhere else, let me know!

irabinovitch:
@bradobro looks like there might be some syntax errors. Mind taking a look?

./tests/checks/integration/test_mongo.py:49:31: E901 SyntaxError: invalid syntax

bradobro:
Sure @irabinovitch, but I think close this PR and I'll port the check @master because of the recent API changes. It will probably be 1-2 weeks before I get space to do that. Thanks for the review.

PR #2487, opened 2016-05-10
[mapreduce] Switch most metrics to counters

olivielpeau:
With the following exceptions:
- `mapreduce.job.elapsed_time`
- `mapreduce.job.map.task.elapsed_time` (made non-default in config)
- `mapreduce.job.reduce.task.elapsed_time` (made non-default in config)

Also, removed `mapreduce.job.map.task.progress` and
`mapreduce.job.reduce.task.progress`.

We have to ping @kky once this PR is merged and part of a nightly/pre-release package so that he can update the default screenboard accordingly.


remh:
👍

PR #2489, opened 2016-05-11
[yarn] Add configurable application tags

alanbover:
Added application tags in the configuration file,
allowing set extra tags for app metrics like queue,
user, etc.

Also, added the tags functionality merged in
https://github.com/DataDog/dd-agent/pull/2473
on tests

### DIFF
--- a/checks.d/yarn.py
+++ b/checks.d/yarn.py
@@ -148,6 +148,7 @@ class YarnCheck(AgentCheck):
 
         # Get properties from conf file
         rm_address = instance.get('resourcemanager_uri', DEFAULT_RM_URI)
+        app_tags = instance.get('application_tags', [])
 
         # Get additional tags from the conf file
         tags = instance.get('tags', [])
@@ -166,7 +167,7 @@ class YarnCheck(AgentCheck):
 
         # Get metrics from the Resource Manager
         self._yarn_cluster_metrics(rm_address, tags)
-        self._yarn_app_metrics(rm_address, tags)
+        self._yarn_app_metrics(rm_address, app_tags, tags)
         self._yarn_node_metrics(rm_address, tags)
 
     def _yarn_cluster_metrics(self, rm_address, addl_tags):
@@ -182,7 +183,7 @@ class YarnCheck(AgentCheck):
             if yarn_metrics is not None:
                 self._set_yarn_metrics_from_json(addl_tags, yarn_metrics, YARN_CLUSTER_METRICS)
 
-    def _yarn_app_metrics(self, rm_address, addl_tags):
+    def _yarn_app_metrics(self, rm_address, app_tags, addl_tags):
         '''
         Get metrics for running applications
         '''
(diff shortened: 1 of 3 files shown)


alanbover:
@olivielpeau this should improve the recently merged PR #2473

checks.d/yarn.py
```diff
@@ -196,14 +197,19 @@ def _yarn_app_metrics(self, rm_address, addl_tags):
...
 
-                        app_name = app_json['name']
+                        tags = []
 
-                        tags = ['app_name:%s' % str(app_name)]
-                        tags.extend(addl_tags)
+                        for tag in app_tags:
+                            key, value = tag.split(':')
+                            tags.append("{tag}:{value}".format(
+                                tag=value, value=str(app_json[key])
```
degemer: If `app_json[key]` doesn't exist, this will raise a `KeyError`, failing the check.

I think it would be better to either test the existence of the value before appending it or try/catch it. (and emit a log.warning without failing if there is an issue)

conf.d/yarn.yaml.example
```diff
@@ -17,7 +17,12 @@ instances:
...
     # cluster_name: MyYarnCluster
 
+    # Tags from running applications that will be applied for application metrics, set as app_key:datadog_tag
+    #
+     application_tags:
+       - name:app_name
+
     # Optional tags to be applied to every emitted metric.
-    # tags:
+    # extra_tags:
```
degemer: I understand why you changed it to `extra_tags`, but for consistency reasons (all the checks use `tags`) it would be better to keep `tags`.

conf.d/yarn.yaml.example
```diff
@@ -17,7 +17,12 @@ instances:
     # A Required friendly name for the cluster.
     # cluster_name: MyYarnCluster
 
+    # Tags from running applications that will be applied for application metrics, set as app_key:datadog_tag
```
degemer: After seeing the code I understand what this means, but it could probably be clearer by adding an example (the `name:app_name` one), and a link indicating what can be used to tag applications.

degemer:
@alanbover Thanks for the PR, I added a few comments, looks good otherwise !

alanbover:
@degemer I've pushed again after some fix after your comments. Much appreciated!

checks.d/yarn.py
```diff
@@ -196,9 +197,22 @@ def _yarn_app_metrics(self, rm_address, addl_tags):
...
 
-                        app_name = app_json['name']
+                        tags = []
+
+                        for tag in app_tags:
+
+                            try:
+                                key, value = tag.split(':')
+                            except ValueError:
+                                self.log.error("Invalid value {0} for application_tag").format()
```
degemer: Typo ? Should probably be something like:
`self.log.error("Invalid value {0} for application_tag".format(tag)`

checks.d/yarn.py
```diff
@@ -196,9 +197,22 @@ def _yarn_app_metrics(self, rm_address, addl_tags):
...
+
+                            try:
+                                tags.append("{tag}:{value}".format(
```
degemer: This will fail if the previous block failed (if `ValueError` was raised).

I think it would be better to have something like:
```
try:
    key, value = tag.split(':')
    tags.append(...)
except ValueError:
    ...
except KeyError:
    ...
```

This way is the first fails, the second is not executed.

degemer:
Added a few comments, should be good after that.

alanbover:
Hi,

Thanks for your comments, I've missed that! I've pushed the changes

checks.d/yarn.py
```diff
@@ -196,9 +197,19 @@ def _yarn_app_metrics(self, rm_address, addl_tags):
...
+                        tags = []
+
+                        for tag in app_tags:
+                            try:
+                                key, value = tag.split(':')
+                                tags.append("{tag}:{value}".format(
+                                    tag=value, value=str(app_json[key])
+                                ))
+                            except ValueError:
+                                self.log.error("Invalid value {0} for application_tag").format(tag)
```
degemer: The parenthesis is still not correct 😉 .
Also could you use the logger interpolation instead of doing it with `format` ? https://www.simonmweber.com/2014/11/24/python-logging-traps.html
It should be `self.log.error("Invalid value %s for application_tag", tag)`.

degemer:
@alanbover There is still a small typo, 👍  after that.

irabinovitch:
@alanbover Were you by any chance able to review the typos @degemer mentioned?

degemer:
Closed in favor of https://github.com/DataDog/dd-agent/pull/3041

PR #2108, opened 2015-11-30
PowerDNS recursor check

AntoCard:
Check for PowerDNS recursor metrics. Needs pdns-recursor experimental web server active.

PR #2490, opened 2016-05-11
[powerdns_recursor] add integration

degemer:
This PR supersedes https://github.com/DataDog/dd-agent/pull/2108.

It adds a service check, restrict metrics being collected, and adds integration tests.


degemer:
@AntoCard and @janeczku I restricted the metrics collected to the ones you advised, and used rates to collect the counter metrics: I think it's more useful to know the questions rate than the number of questions, please let me know if I'm wrong.

@janeczku I think the metrics exposed by PowerDNS are too incomplete to be useful if converted to a histogram, so I also submitted them as rates.

@AntoCard Thanks again for your initial PR !

checks.d/powerdns_recursor.py
```diff
@@ -0,0 +1,79 @@
...
+        service_check_tags = ['recursor_host:{}'.format(config.host), 'recursor_port:{}'.format(config.port)]
+        headers = {"X-API-Key": config.api_key}
+        try:
+            request = requests.get(url, headers=headers)
+            request.raise_for_status()
+            result = json.loads(request.text)
+        except Exception:
+            self.service_check(self.SERVICE_CHECK_NAME, AgentCheck.CRITICAL,
+                               tags=service_check_tags)
+            raise
```
gmmeyer: Wouldn't we want to raise a custom error here rather than just the exception that was thrown by requests?

gmmeyer:
I added a single note, but otherwise I think this looks fine.

checks.d/powerdns_recursor.py
degemer: It could be nice, but since I have no idea what a non-2xx status means for PowerDNS, it wouldn't provide much more information than the default `requests` error.

So for now I'm going to keep it this way, if someone has an issue with the check, we'll update it to precise the error message.

degemer:
Thanks for the review @gmmeyer ! I'll wait for the 💚  CI before merging.

checks.d/powerdns_recursor.py
gmmeyer: Reasonable.

PR #2491, opened 2016-05-11
[memcached] Additional stats metrics

truthbk:
## Why
This PR adds support for memcached "items" statistics as reported by `stats items`. It also attempts to simplify the inclusion of additional optional metrics. To do so, it introduces a map of `stats` arguments (ie. items) to a list  of rate metrics, gauge metrics, and if needed a raw stat output handler.

In the particular case of the items statistics, we need an output handler because we need to parse the raw key value (with form "items:<slab_id>:<metric_name>" to extract the `<slab_id>` and actual metric name. The actual value remains untouched - no need to operate over it. We can then use the `slab_id` to tag the relevant metrics. All handlers are expected to return a tuple of metric_name, tags, and actual value. 

Enabling optional metrics requires configuring the `optional_metrics` flag in the memcached yaml. 

### Open Questions
  - should each of the optional metrics get its own feature flag?
  - currently namespacing optional metrics as follows: memcached.`<stats_arg>`.`<metric_name>` - yes? no?

### Reference
https://github.com/memcached/memcached/blob/master/doc/protocol.txt

### DIFF
--- a/checks.d/mcache.py
+++ b/checks.d/mcache.py
@@ -102,18 +102,71 @@ class Memcache(AgentCheck):
         "listen_disabled_num"
     ]
 
+    ITEMS_RATES = [
+        "evicted",
+        "evicted_nonzero",
+        "expired_unfetched",
+        "evicted_unfetched",
+        "outofmemory",
+        "tailrepairs",
+        "moves_to_cold",
+        "moves_to_warm",
+        "moves_within_lru",
+        "reclaimed",
+        "crawler_reclaimed",
+        "lrutail_reflocked",
+        "direct_reclaims",
+    ]
+
+    ITEMS_GAUGES = [
+        "number",
+        "number_hot",
+        "number_warm",
+        "number_cold",
+        "number_noexp",
+        "age",
+        "evicted_time",
+    ]
+
+    SLABS_RATES = [
+        "get_hits",
+        "cmd_set",
+        "delete_hits",
+        "incr_hits",
+        "decr_hits",
+        "cas_hits",
+        "cas_badval",
+        "touch_hits",
+        "used_chunks",
+    ]
+
+    SLABS_GAUGES = [
+        "chunk_size",
+        "chunks_per_page",
+        "total_pages",
+        "total_chunks",
+        "used_chunks",
+        "free_chunks",
+        "free_chunks_end",
+        "mem_requested",
+        "active_slabs",
+        "total_malloced",
+    ]
+
+    # format - "key": (rates, gauges, handler)
+    OPTIONAL_STATS = {
+        "items": [ITEMS_RATES, ITEMS_GAUGES, None],
+        "slabs": [SLABS_RATES, SLABS_GAUGES, None],
+    }
+
     SERVICE_CHECK = 'memcache.can_connect'
 
     def get_library_versions(self):
         return {"memcache": memcache.__version__}
 
-    def _get_metrics(self, server, port, tags):
-        mc = None  # client
-        service_check_tags = ["host:%s" % server, "port:%s" % port]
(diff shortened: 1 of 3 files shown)


degemer:
Memcached CI is failing on a coverage issue, it's probably legitimate. https://travis-ci.org/DataDog/dd-agent/jobs/129498510

degemer:
A few confirmations/nitpicks:
- The exceptions are not re-raised for the optional metrics, so no service check is sent. Is it because the optional metrics are really optional and could fail while the others would still be there ?

- Maybe you could add a bit more documentation in the check conf files, specifying what are the collected optional metrics.

Looks good otherwise.

For the feature flag question, I'd say that if it was possible to regroup the optional metrics by type, it would be then possible to hide each type behind one feature flag. But I don't know really know memcached, so it may not be the best idea.
Idem for the second question, you have to know memcached to be able to answer.

truthbk:
@degemer the reason I don't re-raise the exception for the optional metrics is precisely to avoid having the optional metrics trigger a failed service check if the default metrics succeed. It felt redundant. In principle, `items` stats wouldn't fail if the default metrics don't but I could foresee a few scenarios (like memcached not reporting all possible subsets of optional metrics due to some configuration directive, etc).

I think the right thing is probably to namespace the metrics, so I will make changes in that direction. We can then revisit and if we don't agree it's better, I'll revert.

truthbk:
@degemer I'm actually having a bit of a hard-time deciding if I want to make some of those gauges into rates as well.... So I'll take a look at that as well.

checks.d/mcache.py
```diff
@@ -170,22 +223,93 @@ def _get_metrics(self, server, port, tags):
...
+            raise
 
-        if mc is not None:
-            mc.disconnect_all()
-            self.log.debug("Disconnected from memcached")
-        del mc
+
+    def _get_optional_metrics(self, client, tags, options=None):
+        for arg, metrics_args in self.OPTIONAL_STATS.iteritems():
+            if not options or options.get(arg, False):
```
degemer: Shouldn't this be `if options and options.get(arg, False)` ?

checks.d/mcache.py
```diff
@@ -170,22 +223,93 @@ def _get_metrics(self, server, port, tags):
...
+        key: "items:<slab_id>:<metric_name>" format
+        value: return untouched
+
+        Like all optional metric handlers returns metric, tags, value
+        """
+        itemized_key = key.split(':')
+        slab_id = itemized_key[1]
+        metric = itemized_key[2]
+
+        tags = ["slab:{}".format(slab_id)]
```
degemer: Just to be sure: `items` and `slabs` are both tagged with `slab:{slab_id}` ?

degemer:
LGTM apart from the small remarks.

One question though: why didn't you enable `coverage_report` for the optional metrics tests ?

checks.d/mcache.py
truthbk: @degemer Not really, I supposed that if you called the method without setting the `options` dict, then you wanted all `options` (otherwise you wouldn't call the `_get_optional_metrics()` method). On the other hand if you do set the `options` dict, then only collect what is set to `True` there.

I enforce the logic you're thinking about in `check()` where we only call `_get_optional_metrics()` if `options` are set in the YAML.

Let me know if you disagree.

checks.d/mcache.py
truthbk: Correct :)

truthbk:
@degemer because I forgot... lol

checks.d/mcache.py
degemer: It makes sense, and the default of `options` is `{}` anyway, so 👍

degemer:
Feel free to merge once the CI is 💚

PR #2492, opened 2016-05-11
[windows services] Be more resilient to case

remh:
WMI queries are not sensitive to case when fetching services status.
So you could have a conf.d with a different case for the service name.
But as we were comparing names with the actual case here:
https://github.com/DataDog/dd-agent/blob/5.7.4/checks.d/windows_service.py#L71

It would result in a bad logic.

This fixes it by having a simpler logic.


degemer:
LGTM 👍

remh:
Failing on an unrelated ES set up.

PR #2493, opened 2016-05-12
[mysql] Fix slave_running for mysql 5.7 with slave binlog

nrw505:
When a MySQL 5.7 is a slave _and_ has a running binlog with no
downstream slaves attached, this check assumed that the instance was a
master with no running slaves.

Fix it so that it checks for Slave_IO_Running and Slave_SQL_Running
before checking for Binlog_enabled and Slaves_connected.

### DIFF
--- a/checks.d/mysql.py
+++ b/checks.d/mysql.py
@@ -524,40 +524,43 @@ class MySql(AgentCheck):
 
             # get slave running form global status page
             slave_running_status = AgentCheck.UNKNOWN
-            slave_running = self._collect_string('Slave_running', results)
+
+            if self._version_compatible(db, host, "5.7.0"):
+                # MySQL 5.7.x might not have 'Slave_running'. See: https://bugs.mysql.com/bug.php?id=78544
+                # look at replica vars collected at the top of if-block
+                slave_io_running = self._collect_string('Slave_IO_Running', results)
+                slave_sql_running = self._collect_string('Slave_SQL_Running', results)
+                if slave_io_running:
+                    slave_io_running = (slave_io_running.lower().strip() == "yes")
+                if slave_sql_running:
+                    slave_sql_running = (slave_sql_running.lower().strip() == "yes")
+
+                if not (slave_io_running is None and slave_sql_running is None):
+                    if slave_io_running and slave_sql_running:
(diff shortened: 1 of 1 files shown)


irabinovitch:
CC: @truthbk since he most recently did the MySQL clean up in dd-agent 5.7

jmackenziejr:
Hello there, this fix is great, and has allowed us to properly check if a slave is running on our databases that are set up with a master-master replication configuration. However, we do have a database that is configured with multiple slaves, and it seems this metric only verifies that the first instance of Slave_SQL_Running and Slave_IO_Running in the output of the "show slave status\G" command. It would be extremely helpful if this metric could track all instances of this output, so that the slaves running metric could report a value that is indicative of the amount of slaves running on that MySQL instance. This way if you have two slave processes configured that should be running, the metric will allow you to alert if any one of them shuts down. Right now, as it stands, I believe you will only be notified of a failure in the first slave that is displayed in the output of "show slave status\G". Hope that is clear, please let me know if there's anything I can do to clarify, or assist in providing a fix.

truthbk:
@nrw505 thanks for the feedback... we have actually rewritten part of the code, but this likely still applies because we hadn't accounted for this case in the logic. I will submit a similar updated PR, and would love if you could check it out as well :) Hopefully I can transplant your commit with no issues 👍

checks.d/mysql.py
```diff
@@ -524,40 +524,43 @@ def _collect_metrics(self, host, db, tags, options, queries):
...
+
+                if not (slave_io_running is None and slave_sql_running is None):
+                    if slave_io_running and slave_sql_running:
+                        slave_running_status = AgentCheck.OK
+                    elif not slave_io_running and not slave_sql_running:
+                        slave_running_status = AgentCheck.CRITICAL
+                    else:
+                        # not everything is running smoothly
+                        slave_running_status = AgentCheck.WARNING
+            else:
```
truthbk: Because of the bug mentioned in the comment above I'm not sure this can quite be mutually exclusive with the code block above. I'll revisit this in the PR I'll be submitting tomorrow.

nrw505:
Superseded by https://github.com/DataDog/dd-agent/pull/2610

PR #2495, opened 2016-05-12
[kong] tag metric by table/collection

degemer:
Instead of creating one metric per table.

### DIFF
--- a/checks.d/kong.py
+++ b/checks.d/kong.py
@@ -11,6 +11,9 @@ from util import headers
 
 
 class Kong(AgentCheck):
+
+    METRIC_PREFIX = 'kong.'
+
     """ collects metrics for Kong """
     def check(self, instance):
         metrics = self._fetch_data(instance)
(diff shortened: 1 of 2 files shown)


degemer:
@shashiranjan84 I was having a look at https://github.com/Mashape/IntegrationTemplate/blob/9e1cff13aa9fc4aac6e24dce34742f2f54a5d91d/Kong/metrics.csv, and I noticed that `kong.database.{table_name}` depends on the table name. What we usually do in this case is instead to tag the metric with the table name (but no in the metric name).

This is what this PR does, by adding two metrics : 
- `kong.database.count`, number of tables, (could be renamed to `kong.table.count`)
- `kong.database.items`, number of items, tagged by `table:{table_name}`. (could be renamed `kong.table.items`)

What do you think ?

shashiranjan84:
Looks great to me. Also `kong.table` makes more sense but I think its not big of a deal and Kong itself sends it as 'database'. Thank You

Seems like I have to update the Integrationtemplate  PR to match above changes?

degemer:
Thank you for the feedback, I'll update the PR.

No need to change the IntegrationTemplate PR, I'll take care of it.

PR #2496, opened 2016-05-12
[consul] Added maintenance status for metrics

brumschlag:
When a node is in 'maintenance' mode, it is a different measure than if a service is critical. 

This PR adds a 'maintenance' status -- it checks for a check named _node_maintenance. If it is critical, it marks the node and services as 'maintenance' status instead of critical.

This is helpful for cases when new nodes come up in a maintenance mode, ready to be turned on. 
Also useful for alerting only on 'critical' cases - not 'maintenance'.

### DIFF
--- a/checks.d/consul.py
+++ b/checks.d/consul.py
@@ -31,6 +31,7 @@ class ConsulCheck(AgentCheck):
         'passing': AgentCheck.OK,
         'warning': AgentCheck.WARNING,
         'critical': AgentCheck.CRITICAL,
+		'maintenance': "maintenance"
     }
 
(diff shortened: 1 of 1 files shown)


gmmeyer:
@brumschlag, thank you very much for your contribution. It's very much appreciated!

Unfortunately, we're not going to be able to add any more new features before the next release. It takes too long to test all of them, especially when we're in the middle of moving to a new model for how we package and distribute checks. This feature makes sense, however, and I think we'd happily add it into the check afterwards.

masci:
closing as superseded by https://github.com/DataDog/integrations-core/pull/1267

PR #2500, opened 2016-05-13
[mesos_master] Fix key error when a metric is missing.

gphat:
### What's this PR do?
Fixes a case where a missing metric can cause the mesos_master integration to report only a subset of metric values.

### Motivation
Our version of Mesos — 0.21.0 — was missing a key `master/tasks_error` and getting a `KeyError`. This caused it to only report 20 of the possible 60 metrics it should've been collecting!

### Notes
I didn't add any sort of logging for missing keys. Happy to do so if desired.

cc @rhwlo @antifuchs


remh:
Sounds reasonable. Thanks a lot for fixing it!

PR #2502, opened 2016-05-13
[mongo] Add additional tag on status change to make searching more precise.

gphat:
### What's this PR do?
Adds an additional tag `action:mongo_replset_member_status_change` to events generated for mongo replset status changes.

### Motivation
I don't like relying on full text search for events. I find tag filtering to be a more precise and ergonomic way of asking for a specific type of event. This tag lets us filter *just* for status change events easily.

cc @rhwlo @antifuchs


yannmh:
It makes sense. Thanks again @gphat  🙇  !

PR #2507, opened 2016-05-16
[docker daemon] Add the possibility to tag by container id

### LINKED ISSUE #2353: Include container id in docker_daemon metrics
Currently, the docker_daemon check [does not include the container id](https://github.com/DataDog/dd-agent/blob/master/checks.d/docker_daemon.py#L71-L82) as a tag in the metrics that are generated. Including this would be a great addition. Adding it would be pretty straightforward, but I'm not sure if there was any reasoning behind not including it in the first place.

remh:
Fix #2353


hkaj:
👍

PR #2509, opened 2016-05-17
Fix bad logging in case a check config file is missing

hkaj:
get_hostname tries calling DockerUtil in some cases which can lead to irrelevant error logs like
```
2016-05-11 20:23:09,243 | ERROR | dd.collector | utils.checkfiles(checkfiles.py:29) | Couldn't find any configuration file for the docker_daemon check.
2016-05-11 20:23:09,243 | ERROR | dd.collector | utils.dockerutil(dockerutil.py:79) | No instance was found in the docker check configuration. Docker related collection will not work.
```

This PR makes `get_conf_path` raise an IOError instead of logging, and manage this exception more precisely downstream.

### DIFF
--- a/utils/checkfiles.py
+++ b/utils/checkfiles.py
@@ -12,7 +12,7 @@ log = logging.getLogger(__name__)
 
 
 def get_conf_path(check_name):
-    """Return the yaml config file path for a given check name."""
+    """Return the yaml config file path for a given check name or raise an IOError."""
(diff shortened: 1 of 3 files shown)


degemer:
I'm not a huge fan of this solution (since it's only lowering the logs to `DEBUG`), but I guess that fixing the logic would probably require more time/changes, so :shipit:

(Build failing on elasticsearch, not related)

PR #2510, opened 2016-05-17
[checks][tokumx] fixes the div by zero error in tokumx

gmmeyer:
Tokumx has a bug where a key should be more differentiated than it is, as it is, it 4 different values share this key, so it's updated any time any of them update, sending false data back to the server. This PR fixes it.


remh:
CI is failing on ES.

PR #2511, opened 2016-05-17
[dogstatsd] improve logging of incorrect packet

degemer:
Copy/paste from the two commits:

## log metadata error instead of failing
Previously, if `dogstatsd` received a packet with an incorrect metadata (empty
for example), it would fail and raise an exception without explanation.

This commits changes this behaviour: if something is wrong with the metadata,
it is logged at warning level but does not fail.

## log useful parse errors
Previously dogstasd was only logging
```
            except Exception:
                log.exception('Error receiving datagram')
```
whereas the `parse_*` functions were trying to raise useful errors.

This shouldn't happen anymore, as the parse exception are now handled and
logged in the aggregator.

It has another benefit: if there are incorrect packets in the "datagram",
the correct ones will be processed correctly (it was not the case before,
`dogstatsd` was failing at the first incorrect packet).

### DIFF
--- a/aggregator.py
+++ b/aggregator.py
@@ -496,8 +496,7 @@ class Aggregator(object):
             except (IndexError, AssertionError):
                 log.warning(u'Incorrect metric metadata: metric_name:%s, metadata:%s',
                             name, u' '.join(value_and_metadata[2:]))
-                sample_rate = 1
-                tags = None
+                sample_rate = 1  # In case it's in a bad state
             parsed_packets.append((name, value, metric_type, tags, sample_rate))
 
         return parsed_packets
@@ -600,21 +599,20 @@ class Aggregator(object):
                 continue
 
             if packet.startswith('_e'):
-                self.event_count += 1
                 event = self.parse_event_packet(packet)
                 self.event(**event)
+                self.event_count += 1
             elif packet.startswith('_sc'):
-                self.service_check_count += 1
                 service_check = self.parse_sc_packet(packet)
                 self.service_check(**service_check)
+                self.service_check_count += 1
             else:
-                self.count += 1
                 parsed_packets = self.parse_metric_packet(packet)
+                self.count += 1
                 for name, value, mtype, tags, sample_rate in parsed_packets:
                     hostname, device_name, tags = self._extract_magic_tags(tags)
                     self.submit_metric(name, value, mtype, tags=tags, hostname=hostname,
-                        device_name=device_name, sample_rate=sample_rate)
-
+                                       device_name=device_name, sample_rate=sample_rate)
 
     def _extract_magic_tags(self, tags):
         """Magic tags (host, device) override metric hostname and device_name attributes"""
--- a/aggregator.py
+++ b/aggregator.py
@@ -441,7 +441,7 @@ class Aggregator(object):
         name_and_metadata = packet.split(':', 1)
 
         if len(name_and_metadata) != 2:
-            raise Exception('Unparseable metric packet: %s' % packet)
+            raise Exception(u'Unparseable metric packet: %s' % packet)
 
         name = name_and_metadata[0]
         broken_split = name_and_metadata[1].split(':')
@@ -462,7 +462,7 @@ class Aggregator(object):
             value_and_metadata = datum.split('|')
 
             if len(value_and_metadata) < 2:
-                raise Exception('Unparseable metric packet: %s' % packet)
+                raise Exception(u'Unparseable metric packet: %s' % packet)
 
             # Submit the metric
             raw_value = value_and_metadata[0]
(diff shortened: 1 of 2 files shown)


remh:
@degemer you need to carefully benchmark this.
This code path is ran in a really tightloop that can be called more than 20k per second. I think these try except will add a non negligible overhead so i'm not sure we want to do this.

degemer:
I benchmarked it, using https://github.com/DataDog/dd-agent/blob/master/tests/core/benchmark_aggregator.py
With the previous version, there was a 2% impact on perf, but it was not really useful (it was only useful if receiving multiple packet in the same datagram) so I updated it (removing the problematic `try.. except`), and there is no more perf impact (I didn't profile, just timed the tests).

https://github.com/DataDog/dd-agent/blob/master/tests/core/benchmark_aggregator.py#L14, https://github.com/DataDog/dd-agent/blob/master/tests/core/benchmark_aggregator.py#L41 and https://github.com/DataDog/dd-agent/blob/master/tests/core/benchmark_aggregator.py#L64 were tested.

**Master**:
```
python benchmark_aggregator.py  99.80s user 0.82s system 98% cpu 1:42.15 total
python benchmark_aggregator.py  6.08s user 0.11s system 99% cpu 6.230 total
python benchmark_aggregator.py  23.77s user 1.88s system 97% cpu 26.284 total
```
**Branch**:
```
python benchmark_aggregator.py  99.06s user 0.75s system 98% cpu 1:41.10 total
python benchmark_aggregator.py  5.87s user 0.14s system 96% cpu 6.231 total
python benchmark_aggregator.py  23.55s user 1.85s system 98% cpu 25.866 total
```

dogstatsd.py
```diff
@@ -362,8 +362,8 @@ def start(self):
                     raise
             except (KeyboardInterrupt, SystemExit):
                 break
-            except Exception:
-                log.exception('Error receiving datagram')
+            except Exception as e:
+                log.exception('Error receiving datagram %s', e)
```
yannmh: `log.exception` already logs the full trace.

aggregator.py
```diff
@@ -480,21 +480,25 @@ def parse_metric_packet(self, packet):
...
+                    if m[0] == '@':
+                        sample_rate = float(m[1:])
+                        assert 0 <= sample_rate <= 1
+                    elif m[0] == '#':
+                        tags = tuple(sorted(m[1:].split(',')))
+            except (IndexError, AssertionError):
+                log.warning(u'Incorrect metric metadata: metric_name:%s, metadata:%s',
+                            name, u' '.join(value_and_metadata[2:]))
+                sample_rate = 1
+                tags = None
```
yannmh: Do we need to reset the `sample_rate` & `tags` ?

aggregator.py
```diff
@@ -600,21 +600,20 @@ def submit_packets(self, packets):
                 continue
 
             if packet.startswith('_e'):
-                self.event_count += 1
                 event = self.parse_event_packet(packet)
                 self.event(**event)
+                self.event_count += 1
```
yannmh: 🍥

dogstatsd.py
```diff
@@ -363,7 +363,7 @@ def start(self):
             except (KeyboardInterrupt, SystemExit):
                 break
             except Exception:
-                log.exception('Error receiving datagram')
+                log.exception('Error receiving datagram %s', message)
```
yannmh: `message` is not necessarily defined.

aggregator.py
degemer: `sample_rate` yes, in case of an AssertionError, it's indeed not needed for `tags`.

dogstatsd.py
degemer: 🎣

degemer:
Thanks for the review @yannmh ! Updated the PR as per your comments.

gmmeyer:
Is there a way to do this without the try/except? 

I get that it might be a slightly more complex case statement, but it seems like it would be much more efficient. Even a small impact can have a large impact when you're calling it many times.

It also seems generally that, since we know that an exception here is possible in specific circumstances, we shouldn't be using an exception for control flow.

degemer:
I followed the "Easier to ask for forgiveness than permission." way 😃 : http://stackoverflow.com/a/1835844

This is really supposed to be an exception, doing real validation of the metadata may actually slow `dogstatsd`. I actually checked locally, there is no real difference (the uncertainty is greater than the difference) when the metric packet is correct, and there is a slight one when the packet is incorrect. (`if` is slightly faster) But the `if` only checked if the metadata was empty, which is probably not enough.

The main reason I chose this is that this is still supposed to be an exception (English sense, not the Python one), and we shouldn't adapt to all of the possibilities, just drop the metadata when something bad happens.

Happy to discuss more about it if you want!

gmmeyer:
I think that's reasonable. 

How often do you think it will trigger the except block?

degemer:
I only saw this happening in two support cases, and it was very likely to be an error from the program sending data. (I quickly checked our Ruby & Python libraries, and it can't happen) So nearly never with a standard setup, and the worst case had ~ 1 bad packet every 10-20 packets. (and it was irregular)

gmmeyer:
Okay, sounds good to me. 👍

degemer:
Thanks for the thorough review @gmmeyer ! 😃  
Merging.

PR #2513, opened 2016-05-17
[check][process] changes pagefaults stats from gauge to rate

gmmeyer:
Process pagefault stats aren't very good as a gauge. In most processes they're a roughly monotonically increasing count. But, they're pretty good as a rate. It gives you a baseline that you can then measure changes against.

### DIFF
--- a/tests/checks/integration/test_process.py
+++ b/tests/checks/integration/test_process.py
@@ -254,7 +254,6 @@ class ProcessCheckTest(AgentCheckTest):
     @patch('psutil.Process', return_value=MockProcess())
     def test_check(self, mock_process):
         (minflt, cminflt, majflt, cmajflt) = [1, 2, 3, 4]
-        pg_fault_stats = {'minor_faults': minflt,'children_minor_faults': cminflt,'major_faults': majflt,'children_major_faults': cmajflt}
 
         def mock_get_pagefault_stats(pid):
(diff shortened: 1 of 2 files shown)


tests/checks/integration/test_process.py
```diff
@@ -269,13 +269,13 @@ def mock_get_pagefault_stats(pid):
...
 
-        self.run_check(config, mocks=mocks)
+        self.run_check_twice(config, mocks=mocks)
 
         instance_config = config['instances'][-1]
         for stat_name in self.PAGEFAULT_STAT:
-            self.assertMetric('system.processes.mem.page_faults.' + stat_name, at_least=1,
+            self.assertMetric('system.processes.mem.page_faults.' + stat_name,
                 tags=self.generate_expected_tags(instance_config),
-                value=pg_fault_stats[stat_name])
```
truthbk: `pg_fault_stats` is now unused so the `default` test fails.

gmmeyer: Ah damn. I just ran the test with `nosetests tests/checks/integration/test_process.py:ProcessCheckTest.test_check`.

I'll fix it. Thank you!

truthbk:
Travis failures look unrelated. Please double check in case I missed anything and feel free to merge :shipit:

PR #2515, opened 2016-05-18
[sqlserver] send service checks at every run

degemer:
And not just at the beginning.
This also deletes the useless cache_failure feature.


checks.d/sqlserver.py
```diff
@@ -351,7 +327,19 @@ def open_db_connections(self, instance):
 
                 self.connections[conn_key]['conn'] = rawconn
```
truthbk: Wondering if we should explicity close the old connection if we replace something cached, wondering if this could lead to a leak. One of our wmi memory leaks was due to something like this - just replacing a cached connection hoping the old one would be garbage collected (they never were, but those checks were threaded so it was a little different case).

truthbk:
:shipit: looks good!

degemer:
Thanks for the review, merging it !

PR #2516, opened 2016-05-18
[jmxfetch] If statsd is bound to all interfaces, use localhost

remh:
Otherwise jmxfetch won't be able to send metrics


gmmeyer:
LGTM 👍

PR #2518, opened 2016-05-19
adds scandir to requirements-opt

gmmeyer:
Scandir requires both the python development headers and GCC in order to be built. So, it should be in requirements-opt rather than the main requirements file. If you want to install it from source, we should allow you to be able to do that without Python.h or gcc.


remh:
Looks good to me

PR #2521, opened 2016-05-19
[http_check] SNI support for cert_expiration

degemer:
Previously SNI wasn't supported for the cert_expiration part of the check,
because we were not passing the hostname to wrap_socket.

It can only be done with a SSLContext. The code is taken from the Python docs
and should act exactly the same as before.

Doc: https://docs.python.org/2.7/library/ssl.html#ssl.SSLContext.wrap_socket

============ Previous Description ============
In order to get the certificate, `ssl_` is now used.
This was done to fix a sneaky bug which seems to only happen with some hosts
and not others. (only thing in common these hosts had was that they were
running Apache and returning a 401 for the queried URL)

`server_hostname` argument is needed, as this fix doesn't work otherwise.
According to the doc, it is set when the shipped OpenSSL version is
SNI-compatible.

### DIFF
--- a/checks.d/http_check.py
+++ b/checks.d/http_check.py
@@ -74,11 +74,11 @@ class WeakCiphersHTTPSConnection(urllib3.connection.VerifiedHTTPSConnection):
 
         # Wrap socket using verification with the root certs in trusted_root_certs
         self.sock = ssl_.ssl_wrap_socket(conn, self.key_file, self.cert_file,
-                                        cert_reqs=resolved_cert_reqs,
-                                        ca_certs=self.ca_certs,
-                                        server_hostname=hostname,
-                                        ssl_version=resolved_ssl_version,
-                                        ciphers=self.ciphers)
+                                         cert_reqs=resolved_cert_reqs,
+                                         ca_certs=self.ca_certs,
+                                         server_hostname=hostname,
+                                         ssl_version=resolved_ssl_version,
+                                         ciphers=self.ciphers)
 
         if self.assert_fingerprint:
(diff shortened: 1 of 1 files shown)


MarijnKoesen:
+1 for this fix.

We also need this, as our server supports multiple hosts/domains on the same ip (using SNI), we need the hostname to be passed to the server so that it can serve the correct SSL certificate.

Without this fix we only get: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:590), because the wrong certificate is server.

With the fix we get the number of days remaining, as the correct one is served.

degemer:
Thanks for your feedback @MarijnKoesen !

I updated the PR, using `ssl` instead of the `requests` `ssl_`. (since we don't actually need it, I didn't correctly identified the bug before your comment)

checks.d/http_check.py
```diff
@@ -74,11 +74,11 @@ def connect(self):
...
-                                        cert_reqs=resolved_cert_reqs,
-                                        ca_certs=self.ca_certs,
-                                        server_hostname=hostname,
-                                        ssl_version=resolved_ssl_version,
-                                        ciphers=self.ciphers)
+                                         cert_reqs=resolved_cert_reqs,
+                                         ca_certs=self.ca_certs,
+                                         server_hostname=hostname,
+                                         ssl_version=resolved_ssl_version,
+                                         ciphers=self.ciphers)
```
hkaj: 👌

hkaj:
LGTM

PR #2532, opened 2016-05-24
Update ceph.py

gbodhi:
Add to ceph_args two parameters in order to take the name of the Ceph cluster, since If you set the cluster name at install time, then the ceph_args is not going to take the valid name and failed. It needs to change _collect_raw and check functions in order to pass the name parameter. Also it would be great to pass the name of the cluster on the yaml file.

### DIFF
--- a/checks.d/ceph.py
+++ b/checks.d/ceph.py
@@ -23,16 +23,16 @@ class Ceph(AgentCheck):
     DEFAULT_CEPH_CMD = '/usr/bin/ceph'
     NAMESPACE = "ceph"
 
-    def _collect_raw(self, ceph_cmd, instance):
+    def _collect_raw(self, ceph_cmd, instance, name):
         use_sudo = _is_affirmative(instance.get('use_sudo', False))
         ceph_args = []
         if use_sudo:
             test_sudo = os.system('setsid sudo -l < /dev/null')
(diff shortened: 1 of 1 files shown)


remh:
Thanks @gbodhi unfortunately this code doesn't work.
The check function doesn't have `name` in it's signature.. You should get the name from the instance dictionary instead.

Can you also add this option in the yaml file for documentation ?
Also, if no name is set up, it should keep the same behavior.

Thanks again for this!

gmmeyer:
Hi @gbodhi,

Thanks again for your contribution, always very much appreciated. 

We're currently transitioning our checks to a new SDK-based model for the agent. An unfortunate side-effect is that we have to impose a freeze on the dd-agent repo. 

The SDK consists of two repositories https://github.com/DataDog/integrations-core and https://github.com/DataDog/integrations-extras, where all of our checks will reside. This PR would belong in the core repository. Would you be so kind as to re-open the PR there? 

We have facilitated some tools to help you with the SDK development that should help you get up and running in no time: [sdk guide](http://docs.datadoghq.com/guides/integration_sdk/) [datadog sdk gem]( https://github.com/DataDog/datadog-sdk-testing). We are very sorry for the inconvenience caused. 

PR #2537, opened 2016-05-24
[docker_daemon] resolve hierarchy by looking at procfs <pid>/cgroup 

truthbk:
## Why
Containers may be run with custom cgroups, not necessarily falling under the default docker one. Typically I think we should expect these to be subgroups of the docker cgroup, although I'm not sure if that can be guaranteed (maybe we can make that a requirement - amongst other things because the docker group should probably have read privileges to the cgroup files - more testing in order). When customers decide to run the containers like this, we were unable to pick up on the cgroup stats and the check fails.

## How
If we are aware of the container <pids>, which we are (at least when running the agent on the host), we can quickly grab the cgroup file locations by parsing /proc/<pid>/cgroup. We can then iterate over the mountpoints, searching for these files in the particular subsystems of interest.

### Testing
Tested on Xenial - more testing pending across distros (systemd, upstart, etc). Also must test the dockerized agent.

When launching a container you can specify (to some degree) the parent cgroup under which that particular container should be executed. Specify using the docker `--cgroup-parent` option. Also, if you wish to create your own custom cgroups, you can do so with the linux `libcgroup-utils`: 
`cgcreate -a root:docker -g cpu,memory,blkio:docker/<somegroup> -t root:docker`

- [x] systemd - ubuntu (xenial)
- [x] upstart - ubuntu (trusty)
- [x] systemd - centos7
- [x] upstart - centos6(?)
- [x] dockerized
- [x] systemd-docker wrapper

### DIFF
--- a/checks.d/docker_daemon.py
+++ b/checks.d/docker_daemon.py
@@ -145,6 +145,7 @@ class DockerDaemon(AgentCheck):
         self.init()
         self._service_discovery = agentConfig.get('service_discovery') and \
             agentConfig.get('service_discovery_backend') == 'docker'
+        self._custom_cgroups = _is_affirmative(init_config.get('custom_cgroups', False))
 
     def is_k8s(self):
         return 'KUBERNETES_PORT' in os.environ
@@ -231,8 +232,11 @@ class DockerDaemon(AgentCheck):
                 self.log.warning('Could not retrieve kubernetes labels: %s' % str(e))
                 self.kube_labels = {}
 
+        # containers running with custom cgroups?
+        custom_cgroups = _is_affirmative(instance.get('custom_cgroups', self._custom_cgroups))
+
         # Get the list of containers and the index of their names
-        containers_by_id = self._get_and_count_containers()
+        containers_by_id = self._get_and_count_containers(custom_cgroups)
         containers_by_id = self._crawl_container_pids(containers_by_id)
 
         # Send events from Docker API
@@ -265,7 +269,7 @@ class DockerDaemon(AgentCheck):
             # It's not an important metric, keep going if it fails
             self.warning("Failed to count Docker images. Exception: {0}".format(e))
 
-    def _get_and_count_containers(self):
+    def _get_and_count_containers(self, custom_cgroups=False):
         """List all the containers from the API, filter and count them."""
 
         # Querying the size of containers is slow, we don't do it at each run
@@ -306,6 +310,16 @@ class DockerDaemon(AgentCheck):
                 continue
 
             containers_by_id[container['Id']] = container
+
+            # grab pid via API if custom cgroups - otherwise we won't find process when
+            # crawling for pids.
+            if custom_cgroups:
+                try:
+                    inspect_dict = self.docker_client.inspect_container(container_name)
+                    container['_pid'] = inspect_dict['State']['Pid']
+                except Exception as e:
+                    self.log.debug("Unable to inspect Docker container: %s", e)
+
 
         for tags, count in running_containers_count.iteritems():
             self.gauge("docker.containers.running", count, tags=list(tags))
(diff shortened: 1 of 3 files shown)


utils/dockerutil.py
```diff
@@ -177,6 +177,27 @@ def find_cgroup(self, hierarchy):
             raise Exception("Can't find mounted %s cgroups." % hierarchy)
 
     @classmethod
+    def find_cgroup_from_proc(cls, mountpoints, pid, subsys):
+        with open('/proc/{pid}/cgroup'.format(pid=pid), 'r') as fp:
```
remh: watchout you'll need to use `self._docker_root` here otherwise this won't work if the agent runs in a container where `/proc` is mounted on `/host/proc`

truthbk:
BTW, just as a note... I did notice that some containers (`nginx` to be precise) will have not one but two processes associated with the container (we already know in theory it should just be a single process per container, but in practice....). Although this doesn't affect the PR directly, it could mean some of the memory/cpu stats reported for a container are inaccurate. Just some food for thought....

checks.d/docker_daemon.py
```diff
@@ -806,6 +832,9 @@ def _crawl_container_pids(self, container_dict):
                     if line[1] in ('cpu,cpuacct', 'cpuacct,cpu', 'cpuacct') and 'docker' in line[2]:
```
hkaj: why not
```python
if line[1] in ('cpu,cpuacct', 'cpuacct,cpu', 'cpuacct') and ('docker' in line[2] or 'docker' in selinux_policy)
```
(parenthesis are not needed but add clarity)

checks.d/docker_daemon.py
```diff
@@ -509,7 +523,7 @@ def _report_performance_metrics(self, containers_by_id):
     def _report_cgroup_metrics(self, container, tags):
         try:
             for cgroup in CGROUP_METRICS:
-                stat_file = self._get_cgroup_file(cgroup["cgroup"], container['Id'], cgroup['file'])
+                stat_file = self._get_cgroup_from_proc(cgroup["cgroup"], container['_pid'], cgroup['file'])
```
hkaj: so is `_get_cgroup_file` used anywhere at this point?

checks.d/docker_daemon.py
```diff
@@ -307,6 +311,16 @@ def _get_and_count_containers(self):
 
             containers_by_id[container['Id']] = container
 
+            # grab pid via API if custom cgroups - otherwise we won't process find when
```
hkaj: `process find` --> `find process` I suppose

hkaj:
This new option should be introduced in the `docker_daemon.yaml`

checks.d/docker_daemon.py
```diff
@@ -307,6 +311,16 @@ def _get_and_count_containers(self):
...
+            # crawling for pids.
+            if custom_cgroups:
+                try:
+                    inspect_dict = self.docker_client.inspect_container(container_name)
```
hkaj: This call can be very slow sometimes.
We'll need to advise our users to use this feature as little as possible (preferably as a comment in the yaml file). Did you have a chance to try this on a lot of containers? I wonder if the collector run could get timeout killed because of this.

checks.d/docker_daemon.py
```diff
@@ -817,6 +846,12 @@ def _crawl_container_pids(self, container_dict):
                         continue
                     container_dict[container_id]['_pid'] = folder
                     container_dict[container_id]['_proc_root'] = os.path.join(proc_path, folder)
+                else: # if we match by pid that should be enough (?) - O(n) ugh!
```
hkaj: can we add a condition to only enter this loop if `custom_cgroup` is true?

remh:
@truthbk looks like it needs a rebase as well as a few fixes

checks.d/docker_daemon.py
truthbk: yup, totally right, should've cleaned this up myself.

checks.d/docker_daemon.py
truthbk: not really....

hkaj: alright let's kill it then

truthbk: burnt with 🔥

checks.d/docker_daemon.py
```diff
@@ -818,7 +835,13 @@ def _crawl_container_pids(self, container_dict):
                         continue
                     container_dict[container_id]['_pid'] = folder
                     container_dict[container_id]['_proc_root'] = os.path.join(proc_path, folder)
-            except Exception as e:
+                else: # if we match by pid that should be enough (?) - O(n) ugh!
```
hkaj: should we use something like `elif :` here? I'm just trying to see if we can avoid entering this loop if it's not absolutely necessary.

hkaj:
LGTM now. network check is failing but it's unrelated.
Good one @truthbk !

PR #2541, opened 2016-05-25
[core] expiration should adjust to longer collection intervals (>300s).

truthbk:
## Why
If the `min_collection_interval` > 300 (the default expiration time) then metrics would be expired before running the next interval. This would affect rates, post expiration messages on the log file, and remove metric contexts when it probably shouldn't.

### Proposed fix
Make the expiration time be `X + min_collection_time` where X would typically be the default expiration time.

### DIFF
--- a/checks/__init__.py
+++ b/checks/__init__.py
@@ -299,6 +299,8 @@ class AgentCheck(object):
 
     SOURCE_TYPE_NAME = None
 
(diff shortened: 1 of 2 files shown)


degemer:
This is really an edge case, your change shouldn't affect a lot of users, 👍 for your implementation.

truthbk:
CI error is unrelated. Merging! Thanks @degemer!

PR #2542, opened 2016-05-25
[kubernetes] disable use_histogram by default

hkaj:


conf.d/kubernetes.yaml.example
```diff
@@ -7,20 +7,20 @@ instances:
   # The kubernetes check retrieves metrics from cadvisor running under kubelet.
   # By default we will assume we're running under docker and will use the address
   # of the default router to reach the cadvisor api.
-  #
```
degemer: 🍰

degemer:
Indeed default is already `False` in the code (https://github.com/DataDog/dd-agent/blob/685da61a60b55a960277b23ae19a7d992e5ecbc1/checks.d/kubernetes.py#L105)
👍

PR #2545, opened 2016-05-26
When dd_url is empty, continue but log to debug level

wleese:
When testing our custom emitter, ddagent/forwarder tries to and complains about connecting to dd_url. 

This change allows dd_url to be empty (nothing new) and if so, logs to loglevel debug and does not attempt to connect to dd_url. Old behaviour was to simply try to connect, even if the url was invalid to begin with.

### DIFF
--- a/ddagent.py
+++ b/ddagent.py
@@ -212,62 +212,70 @@ class AgentTransaction(Transaction):
     def flush(self):
         for endpoint in self._endpoints:
             url = self.get_url(endpoint)
-            log.debug(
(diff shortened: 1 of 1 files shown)


degemer:
Hi @wleese! 
I completely understand why this was useful for your case, but it seems to be really specific and very few users could benefit from it.
It could actually be very confusing for some users (making a mistake in their config and not seeing anything in Datadog and in the logs), so we're not going to merge it.
Thanks for the PR, I hope the custom emitter works fine now! 😃

PR #2547, opened 2016-05-26
Added a note on how to get bundler installed

jeffallen:
If you're a doofus and don't know that bundler is a Ruby gem, then you won't know how to install it unless someone tells you.


remh:
Thanks ;)

PR #2550, opened 2016-05-27
[directory] Add subdir features.

gphat:
### What's this PR do?
Adds support for various "subdirectory" options and generally makes the directory check have some fancy new features.

### Motivation
We wanted to evaluate the shape of our Kafka on-disk stuffs. That means we want to look at the Kafka log directory and each subdirectory therein — as that's how topics are stored — and generate metrics for same.

This felt like a job for the directory check, but it turned out we needed a few options to influence it's behavior:

* `subdirtagname` the tag name for each subdir we descend in to
* `subdirgauges` is the boolean for controlling the emission of a per-subdir total-size gauge `system.disk.directory.bytes` with a `subdirtagname` tag. :)
* `histograms` is the boolean for controlling the emission of directory histograms (we don't need em!)

### Notes
* This adds the subdirtagname to each file gauge and histogram, if enabled.
* A updated the listings of options to match in all the places!
* Adds a `is_subdir` tag to things with that tag. The root dir gets a `false` and subdirs get `true`.

### DIFF
--- a/checks.d/directory.py
+++ b/checks.d/directory.py
@@ -27,10 +27,14 @@ class DirectoryCheck(AgentCheck):
         "directory" - string, the directory to gather stats for. required
         "name" - string, the name to use when tagging the metrics. defaults to the "directory"
         "dirtagname" - string, the name of the tag used for the directory. defaults to "name"
+        "subdirtagname" - string, the name of the tag used for the subdirectory. defaults to "subdir"
         "filetagname" - string, the name of the tag used for each file. defaults to "filename"
         "filegauges" - boolean, when true stats will be an individual gauge per file (max. 20 files!) and not a histogram of the whole directory. default False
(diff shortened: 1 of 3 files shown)


gphat:
I still need to add tests for this!

gphat:
Tests work, but the travisci failures seem to be limited to a specific instance that is having other problems. This is ready for review!

gphat:
We've decided we need a few more changes that make this integration really complex and are instead going to make a new integration and add it to our OSS repo.

jeffbyrnes:
@gphat any chance of a link to that integration? I’d love to see if it helps us out with some directory checking needs we have at Dark Sky.

gphat:
Hey @jeffbyrnes!

You can find [our subdir check here](https://github.com/stripe/datadog-checks#subdir-sizes).

jeffbyrnes:
Thanks @gphat!

PR #2551, opened 2016-05-27
[kubernetes] Ingest k8s events + limits and requests metrics

masci:
## What
 * Fetch events from the Kubernetes API and feed them to Datadog.
 * Add new `limits` and `requests` metrics.

Logic recap:
 * ~~Each agent fetches the whole list of available events and discards those not regarding pods running on the same node.~~
 * If the `collect_events` config param is set to `True`, the agent fetches the whole list of available events
 * ~~The strategy above could be replaced with delegating~~ Users should delegate one and only one agent pod to collect and send all the events - that agent possibly being the one running on k8s master. This unfortunately doesn't work on GCE where the master is not part of the cluster but it's provided by the platform as an external service instead. How to achieve this part is left to the users.
 * To avoid duplicates, the timestamp of the oldest event seen is stored in the `Kubeutil` singleton and used to filter out older events. 

Open issues:
 * There are events not directly tied to Pods, like ReplicaSet events. Those are discarded for the moment.
 * There are serious concerns about flooding since there's no way at the moment to perform selection queries through the event API. We rely on the fact that k8s events currently have a TTL limited to 1 hour.

### DIFF
--- a/checks.d/kubernetes.py
+++ b/checks.d/kubernetes.py
@@ -30,6 +30,7 @@ DEFAULT_ENABLED_RATES = [
     'diskio.io_service_bytes.stats.total',
     'network.??_bytes',
     'cpu.*.total']
+DEFAULT_COLLECT_EVENTS = False
 
 NET_ERRORS = ['rx_errors', 'tx_errors', 'rx_dropped', 'tx_dropped']
 
@@ -118,7 +119,8 @@ class Kubernetes(AgentCheck):
         self._update_metrics(instance, pods_list)
 
         # kubelet events
-        self._process_events(instance, pods_list)
+        if _is_affirmative(instance.get('collect_events', DEFAULT_COLLECT_EVENTS)):
+            self._process_events(instance, pods_list)
 
     def _publish_raw_metrics(self, metric, dat, tags, depth=0):
         if depth >= self.max_depth:
--- a/checks.d/kubernetes.py
+++ b/checks.d/kubernetes.py
@@ -306,7 +306,7 @@ class Kubernetes(AgentCheck):
                             continue
                         self.publish_gauge(self, '{}.{}.requests'.format(NAMESPACE, request), values[0], _tags)
                 except (KeyError, AttributeError) as e:
-                    self.log.error("Unable to retrieve container requests for %s: %s", c_id, e)
+                    self.log.error("Unable to retrieve container requests for %s: %s", c_name, e)
                     self.log.debug(container)
 
         self._update_pods_metrics(instance, pods_list)
--- a/checks.d/kubernetes.py
+++ b/checks.d/kubernetes.py
@@ -288,8 +288,7 @@ class Kubernetes(AgentCheck):
 
     def _process_events(self, instance, pods_list):
         """
-        Retrieve a list of events from kubelet that is relevant for the host the
-        agent is running on.
+        Retrieve a list of events from the kubernetes API.
 
         At the moment there is no support to select events based on a timestamp query, so we
         go through the whole list every time. This should be fine for now as events
@@ -302,8 +301,6 @@ class Kubernetes(AgentCheck):
         """
         node_ip, node_name = self.kubeutil.get_node_info()
         self.log.debug('Processing events on {} [{}]'.format(node_name, node_ip))
-        pods = self.kubeutil.filter_pods_list(pods_list, node_ip)
-        pod_uids = self.kubeutil.extract_meta(pods, 'uid')
 
         k8s_namespace = instance.get('namespace', 'default')
         events_endpoint = '{}/namespaces/{}/events'.format(self.kubeutil.kubernetes_api_url, k8s_namespace)
@@ -317,11 +314,7 @@ class Kubernetes(AgentCheck):
         self.log.debug('Found {} events, filtering out using timestamp: {}'.format(len(event_items), last_read))
 
         for event in event_items:
-            # skip if the event is from another node
             involved_obj = event.get('involvedObject', {})
-            if involved_obj.get('uid') not in pod_uids:
-                self.log.debug('Pod {} is not running on this node, skip...'.format(involved_obj.get('uid')))
-                continue
 
             # skip if the event is too old
             event_ts = int(time.mktime(time.strptime(event.get('lastTimestamp'), '%Y-%m-%dT%H:%M:%SZ')))
--- a/checks.d/kubernetes.py
+++ b/checks.d/kubernetes.py
@@ -10,15 +10,15 @@ from collections import defaultdict
 from fnmatch import fnmatch
 import numbers
 import re
-import simplejson as json
+import time
 
 # 3rd party
 import requests
+import simplejson as json
 
 # project
 from checks import AgentCheck
 from config import _is_affirmative
-from utils.http import retrieve_json
 from utils.kubeutil import KubeUtil
 
 NAMESPACE = "kubernetes"
@@ -45,6 +45,8 @@ FUNC_MAP = {
     GAUGE: {True: HISTO, False: GAUGE},
     RATE: {True: HISTORATE, False: RATE}
 }
+
+EVENT_TYPE = 'kubernetes'
 
 
 class Kubernetes(AgentCheck):
@@ -107,11 +109,16 @@ class Kubernetes(AgentCheck):
         self.publish_rate = FUNC_MAP[RATE][self.use_histogram]
         self.publish_gauge = FUNC_MAP[GAUGE][self.use_histogram]
 
+        pods_list = self.kubeutil.retrieve_pods_list()
+
         # kubelet health checks
         self._perform_kubelet_checks(self.kubeutil.kube_health_url)
 
         # kubelet metrics
-        self._update_metrics(instance)
+        self._update_metrics(instance, pods_list)
+
+        # kubelet events
+        self._process_events(instance, pods_list)
 
     def _publish_raw_metrics(self, metric, dat, tags, depth=0):
         if depth >= self.max_depth:
@@ -187,7 +194,6 @@ class Kubernetes(AgentCheck):
 
         return tags
 
-
     def _update_container_metrics(self, instance, subcontainer, kube_labels):
         tags = list(instance.get('tags', []))  # add support for custom tags
 
@@ -235,12 +241,8 @@ class Kubernetes(AgentCheck):
                               sum(float(net[x]) for x in NET_ERRORS),
                               tags)
 
-    def _retrieve_metrics(self, url):
-        return retrieve_json(url)
-
-    def _update_metrics(self, instance):
-        pods_list = self.kubeutil.retrieve_pods_list()
-        metrics = self._retrieve_metrics(self.kubeutil.metrics_url)
+    def _update_metrics(self, instance, pods_list):
+        metrics = self.kubeutil.retrieve_metrics()
 
         excluded_labels = instance.get('excluded_labels')
         kube_labels = self.kubeutil.extract_kube_labels(pods_list, excluded_keys=excluded_labels)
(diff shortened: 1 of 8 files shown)


tests/checks/fixtures/kubernetes/events.json
```diff
@@ -0,0 +1,532 @@
...
+      "source": {
+        "component": "replicaset-controller"
+      },
+      "firstTimestamp": "2016-05-27T16:37:13Z",
+      "lastTimestamp": "2016-05-27T16:37:13Z",
+      "count": 1,
+      "type": "Normal"
+    }
+  ]
+}
```
masci: This fixture is a little bit large but contains a different kind of events, could be useful later.

irabinovitch:
@masci @remh Given the discussion about above GKE vs K8S should we reach out to our friends at Google for feedback on approach?

irabinovitch:
Do we have examples of the "extraneous" events that aren't tied to pods or replica sets?

masci:
@irabinovitch generally speaking, every node having the `Kind` field set to something different from `Pod`, for example `Created pod: xxx` having `ReplicaSet` kind, or `Scaled up xxx` of `Deployment` kind.

therc:
For people with multiple masters, saying "make the pod running on the master be the one to get all events" doesn't work. Perhaps a PetSet might help? Or perhaps just run a DaemonSet on every node, then create an additional ReplicaSet/Deployment with a single agent pod that only reports events (the other metrics are left to the regular DaemonSet pod). So, on one machine, you'd have two agents, with no overlap between the data they report back.

onbjerg:
Hi @masci! I was notified about this PR by @irabinovitch. I just opened a new PR (#2728) about k8s deployment metrics. We should see if we could somehow combine the two.

A quick recap of what I did:

- Queried the deployments endpoint (``/apis/extensions/v1beta1/deployments``)
- Looped over each deployment and pulled data for the 3 metrics in #2728 (``kubernetes.pods.desired``, ``kubernetes.pods.available``, ``kubernetes.pods.unavailable``), tagged with ``kube_deployment:{name}``
- Resolved the deployment name for a pod based on the ReplicaSet name in ``kubernetes.io/created-by`` and the pod template hash (see [L108 of kubeutil.py in the PR](https://github.com/DataDog/dd-agent/pull/2728/files#diff-d82f35d6c4e640ac2e6e751eef6cc01fR108)). This is used to tag the subcontainers with ``kube_deployment:{name}``.

Other than that there's the obvious new fixture and some added cases in the tests.

onbjerg:
Also: Are there any events for deployments? Currently I do not have deploy markers on my graphs for my team, because we are deploying using kubectl. It would be cool to have some sort of way to detect a k8s deploy.

irabinovitch:
Re: https://github.com/DataDog/dd-agent/pull/2551#issuecomment-234003322

@therc One of the issues we're encountering here is that masters are generally not schedule-able by default.   So while DaemonSets are perfect for monitoring the containers or workloads on individual nodes, they generally don't help us with events or monitoring the master.  Any thoughts on how you would want to see this approached?

irabinovitch:
@onbjerg Just a quick note that @masci is AFK for a few days. He'll get back to you once hes back in the office.

onbjerg:
@irabinovitch Sure, thanks for the heads up.

dadux:
@irabinovitch A daemonsets will still deploy to your masters even if they are not schedulable. the The unschedulable field of a node is not respected by the daemon set controller.

checks.d/kubernetes.py
```diff
@@ -235,25 +245,72 @@ def _update_container_metrics(self, instance, subcontainer, kube_labels):
...
+                continue
+
+            for container in containers:
+                c_name = container.get('name')
+                _tags = container_tags.get(name2id.get(c_name), [])
+
+                # limits
+                try:
+                    for limit, value_str in container['resources']['limits'].iteritems():
+                        values = [float(s) for s in re.findall(r'[-+]?\d+[\.]?\d*', value_str)]
```
hkaj: maybe you should compile this regex, it's used in a loop here, and also for requests.

checks.d/kubernetes.py
```diff
@@ -235,25 +245,72 @@ def _update_container_metrics(self, instance, subcontainer, kube_labels):
...
+                try:
+                    for request, value_str in container['resources']['requests'].iteritems():
+                        values = [float(s) for s in re.findall(r'[-+]?\d+[\.]?\d*', value_str)]
+                        if len(values) != 1:
+                            self.log.warning("Error parsing requests value string: %s", value_str)
+                            continue
+                        self.publish_gauge(self, '{}.{}.requests'.format(NAMESPACE, request), values[0], _tags)
+                except (KeyError, AttributeError) as e:
+                    self.log.error("Unable to retrieve container requests for %s: %s", c_name, e)
+                    self.log.debug(container)
```
hkaj: Debug logs can be very sneaky, showing up later or earlier than expected in the log file and all. I'd recommend adding a few words of description before dropping the variable like that to avoid confusion.

checks.d/kubernetes.py
```diff
@@ -273,11 +330,73 @@ def _update_pods_metrics(self, instance, pods):
...
+    def _process_events(self, instance, pods_list):
+        """
+        Retrieve a list of events from the kubernetes API.
+
+        At the moment there is no support to select events based on a timestamp query, so we
+        go through the whole list every time. This should be fine for now as events
+        have a TTL of one hour[1] but logic needs to improve as soon as they provide
+        query capabilities or at least pagination, see [2][3].
+
+        [1] https://github.com/kubernetes/kubernetes/blob/master/cmd/kube-apiserver/app/options/options.go#L50
```
hkaj: Nitpick but you should link to a commit, not master. When master changes the link becomes obsolete. As an example, master changed since you opened this PR and now you want to link to L51.

checks.d/kubernetes.py
```diff
@@ -273,11 +332,73 @@ def _update_pods_metrics(self, instance, pods):
...
+
+        events = self.kubeutil.retrieve_json_auth(events_endpoint, self.kubeutil.get_auth_token())
+        event_items = events.get('items') or []
+        last_read = self.kubeutil.last_event_collection_ts[k8s_namespace]
+        most_recent_read = 0
+
+        self.log.debug('Found {} events, filtering out using timestamp: {}'.format(len(event_items), last_read))
+
+        for event in event_items:
+            involved_obj = event.get('involvedObject', {})
```
hkaj: This should probably be done after the timestamp comparison

checks.d/kubernetes.py
```diff
@@ -273,11 +332,73 @@ def _update_pods_metrics(self, instance, pods):
...
+            dd_event = {
+                'timestamp': event_ts,
+                'host': node_ip,
+                'event_type': EVENT_TYPE,
+                'msg_title': title,
+                'msg_text': msg_body,
+                'source_type_name': EVENT_TYPE,
+                'event_object': 'kubernetes:{}'.format(involved_obj.get('name')),
+            }
+            self.log.debug('Posting event: {}'.format(dd_event))
```
hkaj: i don't know how spammy events can get but I have a feeling that I'm going to hate this in the collector logs 😅

masci: Yeah it made sense with the first version implementing the event filtering but this way the event collector will be hammered. Better with or without it?

hkaj: I'm in favor of dropping it. It's a linux check anyway so we can ask customers to add it back on a per-user basis if we really need it for debugging.

utils/kubeutil.py
```diff
@@ -93,5 +106,101 @@ def extract_kube_labels(self, pods_list, excluded_keys=None):
 
         return kube_labels
 
+    def extract_meta(self, pods_list, field_name):
```
hkaj: do we need this? It's nice and it's tested ( 🍪 ) but I don't see any usage of it.

utils/kubeutil.py
```diff
@@ -93,5 +106,101 @@ def extract_kube_labels(self, pods_list, excluded_keys=None):
...
+        """
         return retrieve_json(self.pods_list_url)
+
+    def retrieve_metrics(self):
+        """
+        Retrieve metrics from Cadvisor.
+        """
+        return retrieve_json(self.metrics_url)
+
+    def filter_pods_list(self, pods_list, host_ip):
```
hkaj: ditto

hkaj: I suspect this is a remain of the other approach we tried. It could be useful for later in this case. If you think we should keep them around, feel free to.

utils/kubeutil.py
```diff
@@ -93,5 +106,101 @@ def extract_kube_labels(self, pods_list, excluded_keys=None):
...
+    def retrieve_json_auth(self, url, auth_token, timeout=10):
+        """
+        Kubernetes API requires authentication using a token available in
+        every pod.
+
+        We try to verify ssl certificate if available.
+        """
+        verify = self.CA_CRT_PATH if os.path.exists(self.CA_CRT_PATH) else False
+        log.debug('ssl validation: {}'.format(verify))
+        headers = {'Authorization': 'Bearer {}'.format(auth_token)}
```
hkaj: Isn't that a bit sensitive to log? Maybe we can hide bits of it

utils/kubeutil.py
masci: `extract_meta` might be generic enough to stay in a "util" class, `filter_pods_list` is very specific to the previous approach...
We could keep both of them for the sake of trying again the previous approach with 1.3 but I'd add a note on the docstrings mentioning the situation so we can get rid of them later

utils/kubeutil.py
masci: Good catch! My bad, it was meant for internal usage

utils/kubeutil.py
hkaj: Agreed. If we keep them, let's have a reminder in the docstring that they're not in use for now.

hkaj:
Let's make sure that it works nicely with 1.2 *and* 1.3, then it's good to go.

hkaj:
all good on 1.2 too, you can squash and 👍

therc:
Is this available for testing in any of the Docker images?

masci:
You can try the `nightly` tag from https://hub.docker.com/r/datadog/docker-dd-agent/tags/ but please consider the agent as unstable. Events collection is disabled by default, you can activate it in the check configuration file if you want. Later it'll be possible to instrument that feature through env variables, see https://github.com/DataDog/docker-dd-agent/pull/117

therc:
I tried nightly, but I'm not sure it has the latest:

-rw-r--r--. 4 root root 10843 Aug 19 04:04 kubernetes.py
-rw-r--r--. 1 root root  9823 Aug 31 15:41 kubernetes.pyc

It doesn't e.g. import the time module as done in this PR.

therc:
From `docker inspect`:

    "Created": "2016-08-29T16:41:53.174520676Z",

From https://hub.docker.com/r/datadog/docker-dd-agent/builds/bc5brbm22yzmrwsiruph9ua/

2016-08-29T16:40:33.009Z

Nothing else lines up exactly, either. E.g. I can't find a revision of kubernetes.py that is exactly 10843 bytes.

therc:
Nevermind, I was looking at the commit history in the PR, not in the repo. It turns out that the image has kubernetes.py from my Aug 7 change. :-)

https://github.com/DataDog/dd-agent/blob/ef3aa479ea2aae19cb74a5f14e70a45566197550/checks.d/kubernetes.py

So, despite the nightly name, it looks like the source tree has .py files from Aug 19.

rosskukulinski:
Important note:  This requires that the `kubernetes` service exists within the namespace dd-agent is running.  This is _not_ always the case (especially in namespaces that are not `default`).

The preferred mechanism to get the API Server name is by using the environmental variables injected into every pod: `KUBERNETES_SERVICE_HOST` and `KUBERNETES_SERVICE_PORT`.

rosskukulinski:
This also only gets the events for the `default` namespace.  I believe it should pull events from _all_ namespaces.

rosskukulinski:
cc: @remh who I was chatting with on IRC

masci:
Hi @rosskukulinski and thanks for your feedback, very precious.

You can specify any namespace other than `default` in the configuration file under the `instance` section. Anyway, we could (and probably should) collect all the events with no regards for the namespace, but I'd like this to be behind some sort of configuration parameter with reasonable defaults.

Since this code is already in `master` and is going to be shipped with 5.9, do you mind to open two  new, different issues to track and discuss:
a) the way we get the api URL
b) how we can improve events gathering when someone is interested in any namespace available

A PR would work well too 😜  !

Thanks!

rosskukulinski:
done, thanks @masci.  Up to my clients as to whether I invest the development time in submitting PRs.

checks.d/kubernetes.py
```diff
@@ -235,25 +245,74 @@ def _update_container_metrics(self, instance, subcontainer, kube_labels):
...
+                            continue
+                        self.publish_gauge(self, '{}.{}.limits'.format(NAMESPACE, limit), values[0], _tags)
+                except (KeyError, AttributeError) as e:
+                    self.log.debug("Unable to retrieve container limits for %s: %s", c_name, e)
+                    self.log.debug("Container object for {}: {}".format(c_name, container))
+
+                # requests
+                try:
+                    for request, value_str in container['resources']['requests'].iteritems():
+                        values = [float(s) for s in prog.findall(value_str)]
```
therc: Isn't this going to parse eg. 500m as 500 instead of 0.5?

therc: FYI: the Kubernetes code dealing with quantities lives in 

https://github.com/kubernetes/kubernetes/blob/master/pkg/api/resource/quantity.go
https://github.com/kubernetes/kubernetes/blob/master/pkg/api/resource/suffix.go

jonmoter:
@masci, you said:

> You can specify any namespace other than default in the configuration file under the instance section.

I'm looking at the [kubernetes.yaml.example](/DataDog/dd-agent/blob/master/conf.d/kubernetes.yaml.example) file, and I don't see anything documented on how to change the namespace where it looks for the kubernetes service. Could you clarify?

masci:
Hi @jonmoter 
that's because documentation and the example yaml file are not up to date unfortunately but basically all you have to do is
```
instance:
  namespace: my-other-namespace
```
for more details on the issue and provide feedback let's use #2838

PR #2557, opened 2016-05-31
[core] remove events & resources api_key

degemer:
Legacy, they are not used anymore.


olivielpeau:
👍 thanks for the cleanup!

PR #2559, opened 2016-05-31
[mongo] Only try to connect ReplicationInfo metrics when we can

### LINKED ISSUE #2520: [mongo] "database error: can't use 'local' database through mongos"
System Info:

> Collector (v 5.7.4)
> Platform: Linux-3.10.0-327.18.2.el7.x86_64-x86_64-with-centos-7.2.1511-Core
> Python Version: 2.7.11, 64bit
> MongoDB shell version: 3.2.3
> db.version(): 3.2.3

I'm seeing the following error for certain mongo checks:

    mongo
    -----
      - instance #0 [ERROR]: "database error: can't use 'local' database through mongos"
      - Collected 33 metrics, 0 events & 2 service checks

I see there was a pull request merged 29 days ago (#2429) that added code with the comment:

> Fetch information analogous to Mongo's db.getReplicationInfo()`.

It seems that pull request is the culprit here, since it explicitly uses the local database to fetch certain stats:

Line 763:

    localdb = cli['local']
    ...
    localdb.command("collstats", ol_collection_name)['size'] / 2.0 ** 20, 2

Up until now this check has been working fine when used through `mongos`. Is it correct to assume that this check is only meant to be used to directly hit a mongo server, instead of going through `mongos`? Or can this check be updated so it doesn't rely on the `local` database?

remh:
Fix #2520
If you connect to mongo through mongos, there won’t be a `local`
database. But we shouldn’t fail the check if this happens as other
metrics would still be available.

### DIFF
--- a/checks.d/mongo.py
+++ b/checks.d/mongo.py
@@ -828,44 +828,49 @@ class MongoDb(AgentCheck):
             except Exception, e:
                 self.log.warning('Failed to record `top` metrics %s' % str(e))
 
-        # Fetch information analogous to Mongo's db.getReplicationInfo()
-        localdb = cli['local']
 
-        oplog_data = {}
+        if 'local' in dbnames: # it might not be if we are connectiing through mongos
+            # Fetch information analogous to Mongo's db.getReplicationInfo()
+            localdb = cli['local']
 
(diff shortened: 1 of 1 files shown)


yannmh:
:shipit:

PR #2560, opened 2016-05-31
[vsphere] enhance topology support ✨

yannmh:
**[vsphere] enhance topology support ✨**
* Rewrite `_̶c̶a̶c̶h̶e̶_̶m̶o̶r̶l̶i̶s̶t̶_̶r̶a̶w̶_̶a̶t̶o̶m̶i̶c̶`→`_discover_mor` method to be more
  permissive on
  vCenter topologies.
* Move the exclusion logic to a `_is_excluded` method.
* Add tests coverage

**[vsphere] skip unknown metrics**

vSphere metrics are hardcoded and stored at in
`checks.libs.vmware.all_metrics`.

New versions of vSphere bring new metrics: thus, we should not consider
this list exhaustive, neither fail (raise a `KeyError` exceptions) on an
"unknown" metric.
```
2016-06-03 20:14:04 UTC | CRITICAL | dd.collector |
checks.vsphere(vsphere.py:932) | A worker thread crashed:
  File "/opt/datadog-agent/agent/checks.d/vsphere.py", line 299, in
wrapper
    method(*args, **kwargs)
  File "/opt/datadog-agent/agent/checks.d/vsphere.py", line 866, in
_collect_metrics_atomic
    if
ALL_METRICS[self.metrics_metadata[i_key][result.id.counterId]['name']]['s_type']
== 'rate':
KeyError: 'vflashModule.numActiveVMDKs'
```
Instead, skip these metrics.


checks.d/vsphere.py
```diff
@@ -560,19 +560,26 @@ def _cache_morlist_raw_atomic(self, i_key, obj_type, obj, tags, regexes=None, in
...
+                    self.pool.apply_async(
+                        self._cache_morlist_raw_atomic,
+                        args=(i_key, 'folder', resource, tags_copy, regexes)
+                    )
 
         elif obj_type == 'compute_resource':
-            if obj.__class__ == vim.ClusterComputeResource:
-                cluster_tag = "vsphere_cluster:%s" % obj.name
-                tags_copy.append(cluster_tag)
+            cluster_tag = "vsphere_cluster:%s" % obj.name
```
truthbk: Are we sure that if `obj_type == 'compute_resource'` that implies `obj.__class__ == vim.ClusterComputeResource`.... not sure if pyvmomi might let some of the other objects fall into a compute_resource category (ie. datastores or whatever). We should make sure of this.

checks.d/vsphere.py
```diff
@@ -582,6 +589,18 @@ def _cache_morlist_raw_atomic(self, i_key, obj_type, obj, tags, regexes=None, in
                     args=(i_key, 'host', host, tags_copy, regexes, include_only_marked)
                 )
 
+        elif obj_type == 'folder':
```
truthbk: Can we have nested folders? If those are possible, then we wouldn't recurse into them with the current logic. Also, can we only have folders in the `datacenter` level? Or can other resources be arranged into folders as well.

truthbk:
Looks good, just have that comment we should probably check on. Or maybe you already know. Once we double check that and CI passes, everything else looks good to merge!

checks.d/vsphere.py
```diff
@@ -523,112 +523,181 @@ def get_external_host_tags(self):
...
+                tags.append(u"vsphere_host:{}".format(obj.name))
+                for vm in obj.vm:
+                    if vm.runtime.powerState != 'poweredOn':
+                        continue
+                    self.pool.apply_async(
+                        browse_mor,
+                        args=(vm, tags, depth + 1)
+                    )
+
+            # Virtual Machine
```
truthbk: I don't think we need to make it part of this PR, but we should probably make a separate PR at some point to take a look at other types like `vim.Datastore` (so customers can keep track of the SAN's, NAS, NFS, etc).

checks.d/vsphere.py
```diff
@@ -523,112 +523,181 @@ def get_external_host_tags(self):
...
+            elif isinstance(obj, vim.VirtualMachine):
+                if regexes and regexes.get('vm_include') is not None:
```
truthbk: shouldn't we use `_is_excluded()` here? I see we're only using it for the tests.

checks.d/vsphere.py
```diff
@@ -523,112 +523,181 @@ def get_external_host_tags(self):
...
+                        continue
+
+                    self.pool.apply_async(
+                        browse_mor,
+                        args=(host, tags, depth + 1)
+                    )
+
+            # Host
+            elif isinstance(obj, vim.HostSystem):
+                if regexes and regexes.get('host_include') is not None:
```
truthbk: shouldn't we use _is_excluded() here? I see we're only using it for the tests.

checks.d/vsphere.py
```diff
@@ -523,112 +523,181 @@ def get_external_host_tags(self):
...
+                        - host5
+                            - vm1
+                            - vm2
+            ```
+
+        If it's a node we want to query metric for, queue it in `self.morlist_raw` that
+        will be processed by another job.
+        """
+        @atomic_method
+        def browse_mor(obj, prev_tags, depth):
```
truthbk: way more flexible, the less assumptions about the resource hierarchy the better! 💯

tests/checks/mock/test_vsphere.py
```diff
@@ -0,0 +1,248 @@
...
+        is_labeled = kwargs.get('label', False)
+
+        self.name = name
+        self.customValue = []
+
+        if is_labeled:
+            self.customValue.append(Mock(value="DatadogMonitored"))
+
+
+def create_topology(topology_json):
```
truthbk: 😍

truthbk:
Looks good overall! Just a couple of points regarding the `_is_excluded()` method, and a note we should bear in mind for the future regarding Datastores (and maybe others).

checks.d/vsphere.py
yannmh: 👍

checks.d/vsphere.py
yannmh: Oops my mistake

PR #2562, opened 2016-06-01
[mongo][doc] Add more info about ssl_cert_reqs config

spk:
Just adding more info about the `ssl_cert_reqs` config
Cheers

### DIFF
--- a/conf.d/mongo.yaml.example
+++ b/conf.d/mongo.yaml.example
@@ -18,6 +18,10 @@ instances:
     # ssl_keyfile: # Path to the private keyfile used to identify the local
     # ssl_certfile: # Path to the certificate file used to identify the local connection against mongod.
     # ssl_cert_reqs: # Specifies whether a certificate is required from the other side of the connection, and whether it will be validated if provided.
+    # Possible values:
+    #   * 0 for ssl.CERT_NONE (certificates ignored)
+    #   * 1 for ssl.CERT_OPTIONAL (not required, but validated if provided)
+    #   * 2 for ssl.CERT_REQUIRED (required and validated)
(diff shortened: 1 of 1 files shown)


conf.d/mongo.yaml.example
```diff
@@ -13,6 +13,10 @@ instances:
 
     # Optional SSL parameters, see https://github.com/mongodb/mongo-python-driver/blob/2.6.3/pymongo/mongo_client.py#L193-L212
     # for more details
+    # For `ssl_ca_certs` should be:
```
degemer: Small error probably: shouldn't it be `ssl_cert_reqs` ? https://github.com/mongodb/mongo-python-driver/blob/2.6.3/pymongo/mongo_client.py#L201-L208

Could you also put this under the `ssl_cert_reqs` explanation ?
I.e
```
# ssl_cert_reqs: # Specifies whether a certificate is required from the other side of the connection, 
# and whether it will be validated if provided.
# Possible values:
#   * 0 for ssl.CERT_NONE (certificates ignored)
#   * 1 for ssl.CERT_OPTIONAL (not required, but validated if provided)
#   * 2 for ssl.CERT_REQUIRED (required and validated)

degemer:
Thanks for taking the time to add documentation @spk!

I added a comment, let me know what you think about it.

conf.d/mongo.yaml.example
spk: yes exactly, thanks for the review will fix that

spk:
@degemer updated with you suggestions and rebased !

degemer:
LGTM, merging it.

Thanks again!

PR #2565, opened 2016-06-01
[wmi] set a `provider` to request data

yannmh:
### Changes
**[wmi] set a `provider` to request data**

`WMISampler` class has a new `provider` attribute to specify a WMI
provider to request data from.
Available options are:
* (Default) `ProviderArchitecture.DEFAULT`
   Request from the provider corresponding to the version of the Datadog
   Agent version installed, i.e. 32-bit on Datadog Agent 32-bit or
   64-bit otherwise.
* `ProviderArchitecture._32BIT`
  Request from the 32-bit provider.
* `ProviderArchitecture._64BIT`
  Request from the 64-bit provider.

**[wmi_check] support `provider` option**

Add a `provider` option to the WMI check's YAML configuration file.
Instructs the check to request WMI data from the specified provider.
Available options are: `32` or `64`.

**[iis] support `provider` option**

Add a `provider` option to the WMI check's YAML configuration file.
Instructs the check to request WMI data from the specified provider.
Available options are: `32` or `64`.

### Next ?

WMI fallbacks to the default provider when no data is available in the provided one. Thus, I reckon it would make sense to default to the 64-bit provider directly. Any thought ?

### DIFF
--- a/checks/libs/wmi/sampler.py
+++ b/checks/libs/wmi/sampler.py
@@ -47,18 +47,57 @@ class CaseInsensitiveDict(dict):
         return super(CaseInsensitiveDict, self).get(key.lower())
 
 
+class ProviderArchitectureMeta(type):
+    """
+    Metaclass for ProviderArchitecture.
+    """
+    def __contains__(cls, provider):
+        """
+        Support `Enum` style `contains`.
+        """
+        return provider in cls._AVAILABLE_PROVIDER_ARCHITECTURES
+
+
+class ProviderArchitecture(object):
+    """
+    Enumerate WMI Provider Architectures.
+    """
+    __metaclass__ = ProviderArchitectureMeta
+
+    # Available Provider Architecture(s)
+    DEFAULT = 0
+    _32BIT = 32
+    _64BIT = 64
+    _AVAILABLE_PROVIDER_ARCHITECTURES = frozenset([DEFAULT, _32BIT, _64BIT])
+
+
 class WMISampler(object):
     """
     WMI Sampler.
     """
+    # Properties
+    _provider = None
+    _formatted_filters = None
+
+    # Type resolution state
+    _property_counter_types = None
+
+    # Samples
+    _current_sample = None
+    _previous_sample = None
+
+    # Sampling state
+    _sampling = False
 
     def __init__(self, logger, class_name, property_names, filters="", host="localhost",
-                 namespace="root\\cimv2", username="", password="", and_props=[], timeout_duration=10):
+                 namespace="root\\cimv2", provider=None,
+                 username="", password="", and_props=[], timeout_duration=10):
         self.logger = logger
 
         # Connection information
         self.host = host
         self.namespace = namespace
+        self.provider = provider
         self.username = username
         self.password = password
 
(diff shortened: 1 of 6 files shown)


yannmh:
@olivielpeau can you take a pass on it please ? 🙇

checks/libs/wmi/sampler.py
```diff
@@ -47,18 +47,57 @@ def get(self, key):
...
+
+
+class ProviderArchitecture(object):
+    """
+    Enumerate WMI Provider Architectures.
+    """
+    __metaclass__ = ProviderArchitectureMeta
+
+    # Available Provider Architecture(s)
+    DEFAULT = 0
```
olivielpeau: nitpick: for consistency, why not use `_DEFAULT` instead of `DEFAULT` ? :)

checks/libs/wmi/sampler.py
```diff
@@ -47,18 +47,57 @@ def get(self, key):
...
+
+    # Type resolution state
+    _property_counter_types = None
+
+    # Samples
+    _current_sample = None
+    _previous_sample = None
+
+    # Sampling state
+    _sampling = False
```
olivielpeau: I don't understand why all these attributes have been changed to _class_ attributes now. Aren't they meant to be _object_ attributes? (If we have multiple instances of `WMISampler` aren't they going to overwrite the values and lead to weird behaviors?)

Tell me if I'm missing something :)

checks/libs/wmi/sampler.py
olivielpeau: hmm disregard actually, I think I get it (you're using `DEFAULT` in the `provider` setter)

conf.d/wmi_check.yaml.example
```diff
@@ -17,6 +17,10 @@ instances:
   #
   # `namespace` is the optionnal WMI namespace to connect to (default to `cimv2`).
   #
+  # `provider` is the optional WMI provider (default to `32` on Datadog Agent 32-bit or `64`).
+  # Usefull to request WMI data from the non-default provider. Available options are: `32` or `64`.
```
olivielpeau: typo: `usefull` -> `useful` ;)

conf.d/wmi_check.yaml.example
```diff
@@ -17,6 +17,10 @@ instances:
...
+  # For more information: https://msdn.microsoft.com/en-us/library/aa393067(v=vs.85).aspx
```
olivielpeau: IMO we should also document the option in `iis.yaml.example`, what do you think?

olivielpeau:
Thanks @yannmh! Made a few comments, let me know what you think :)

checks/libs/wmi/sampler.py
yannmh: It's still instances attributes, i.e.
```python
class Foo(object):
  bar = None
  
  def __init__(self, value):
    self.bar = value

m_foo1 =  Foo("hello")
m_foo2 = Foo("world")

assert m_foo1.bar == "hello"
assert m_foo2.bar == "world"
```

The only reason I moved this block out of the `__init__` constructor level is because these attributes are always initialized with the same values whichever parameters are passed to the constructor.

conf.d/wmi_check.yaml.example
yannmh: I am wondering if we should not just make the provider default to `64` bit directly. By default, it looks like the WMI query would fallback to the 32-bit provider when no data is available.

If so, the `provider` option would not provide any more value and we could keep it undocumented for debugging purpose. 

What do you think ?

conf.d/wmi_check.yaml.example
yannmh: 🙇

checks/libs/wmi/sampler.py
olivielpeau: Ok thanks, I got confused here.

Might just be me but I'd rather have all the instance attributes initialized in `__init__` though, because potentially having two different values for `Foo.bar` and `m_foo1.bar` confuses me a lot :/

conf.d/wmi_check.yaml.example
olivielpeau: So after discussing this: by default let's keep the default provider that WMI chooses. This way we're sure not to break anything for the existing users of the WMI checks.

Consequence of this: let's keep the documentation on the `provider` option

checks/libs/wmi/sampler.py
yannmh: The only thing that I don't like with this (and part of the reason why I did it), is that it doesn't look great when using a Python property, i.e.

```python
class WMISampler(object):
    """
    An OK looking WMISampler 🌧
    """
    def __init__(self, provider):
        self._provider = None
        self.provider = provider
```
```python
class WMISampler(object):
    """
    A good looking WMISampler ☀😎
    """
    _provider = None

    def __init__(self, provider):
        self.provider = provider
```

olivielpeau: I agree that it looks better your way :)

The thing we have to be careful about in the future is not to use a mutable value when initializing these properties in the class, i.e. if one day we do something like:

```python
class WMISampler(object):
    _property_counter_types = CaseInsentitiveDict()
    
    def _parse_results(self):
        # ...
        self._property_counter_types[wmi_property.Name] = counter_type
```

then the dict is going to be shared among all the instances.

So I think you can go ahead with your changes but let's be very careful about this behavior in the future :)

olivielpeau:
Responded to your comment, but looks good to me, feel free to merge! 🎆

PR #2567, opened 2016-06-02
[dev] locating the system python2 to create the virtualenv

palazzem:
## What's this PR do?

Enforces the use of ``python2`` while creating the ``venv/`` virtualenv, during the execution of ``rake setup_env``.

## Motivation

Without the ``-p`` option, virtualenv uses the ``python`` system executable. Unfortunately I'm running a system (ArchLinux) that uses Python 3 as a default and because of that, the virtualenv is not properly created.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -60,7 +60,7 @@ desc 'Setup a development environment for the Agent'
 task 'setup_env' do
   `mkdir -p venv`
   `wget -O venv/virtualenv.py https://raw.github.com/pypa/virtualenv/1.11.6/virtualenv.py`
-  `python venv/virtualenv.py  --no-site-packages --no-pip --no-setuptools venv/`
+  `python venv/virtualenv.py -p python2 --no-site-packages --no-pip --no-setuptools venv/`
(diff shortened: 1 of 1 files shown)


degemer:
Thanks @palazzem !
It seems that ArchLinux is the only distribution doing this (and caused PEP 394 https://www.python.org/dev/peps/pep-0394/). 😄 

I don't think the `which` is needed according to `virtualenv.py`:
```
    parser.add_option(
        '-p', '--python',
        dest='python',
        metavar='PYTHON_EXE',
        help='The Python interpreter to use, e.g., --python=python2.5 will use the python2.5 '
        'interpreter to create the new environment.  The default is the interpreter that '
        'virtualenv was installed with (%s)' % sys.executable)
```
Could you remove it and just use `-p python2` instead ? I tried on OS X (and it worked), but it would be perfect if you could double check on ArchLinux.

palazzem:
Hi @degemer , you're welcome!
I've updated and rebased the PR with your upstream. As you asked, I've double checked the ``Rakefile`` and it still works in ArchLinux.

Let me know if it is ready to merge,
Cheers! :D

degemer:
That's perfect, thanks again!

PR #2568, opened 2016-06-02
Changed the except to be more modern

ParthKolekar:


degemer:
Thanks @ParthKolekar!

PR #2571, opened 2016-06-04
Copying of API Key Into datadog.conf Fixed

sha1sum:
An apparently improper use of the `sed` command seems to break the copying of the API key into the `datadog.conf` file. This should be fixed with this revision.


olivielpeau:
@sha1sum The `sed` command on OS X (BSD) is slightly different from the GNU `sed`. Here the `-i` option requires a value. From the `man` page on OS X:

```
-i extension
             Edit files in-place, saving backups with the specified extension.  If a zero-length extension is given, no backup will be saved.
```

So the use of `sed` in the OS X install script looks correct. I've given it a quick test to make sure of that and it worked fine.

What issue are you experiencing with the copying of the API key?

sha1sum:
This makes a lot of sense. I have installed many GNU utilities, one of them being sed. #2580 replaces this PR with checks for the sed version before attempting the command.

PR #2572, opened 2016-06-06
[openstack] openstack should pick-up on proxy settings set on agent

truthbk:
## Why
Customers often have boxes that must route http/https traffic through a proxy. Their keystone server, and openstack API calls might have to use the proxy. The openstack check should be proxy aware, this check aims to provide that functionality.

Also, in case the keystone/openstack servers happen to be in the LAN and not require the proxy (even though the agent might), we also introduce a `use_agent_proxy` flag to the openstack config to override the behavior at will.

Also improving support for openstack Mitaka. On mitaka (despite being v2.1) the nova endpoints don't match the `novav21` pattern, instead they match directly to nova. The small changes to how we deal with nova endpoint versions should allow us to get the endpoints if the right `nova_api_version` is overridden in the YAML. Same goes for the hypervisors stats.  

Note: all checks should really potentially be able to make calls via the proxy defined for the agent, so I've added a property to the `AgentCheck` class to provide easy access to the agent's proxy settings.

### DIFF
--- a/checks.d/openstack.py
+++ b/checks.d/openstack.py
@@ -137,10 +137,32 @@ class OpenStackProjectScope(object):
         auth_scope = cls.get_auth_scope(instance_config)
         identity = cls.get_user_identity(instance_config)
 
+        exception_msg = None
         try:
             auth_resp = cls.request_auth_token(auth_scope, identity, keystone_server_url, ssl_verify)
         except (requests.exceptions.HTTPError, requests.exceptions.Timeout, requests.exceptions.ConnectionError):
-            raise KeystoneUnreachable()
+            exception_msg = "Failed kaystone auth with identity:{id} scope:{scope} @{url}".format(
+                id=identity,
+                scope=auth_scope,
+                url=keystone_server_url)
+
+        if exception_msg:
+            try:
+                identity['password']['user']['domain']['name'] = identity['password']['user']['domain'].pop('id')
+
+                if 'domain' in auth_scope['project']:
+                    auth_scope['project']['domain']['name'] = auth_scope['project']['domain'].pop('id')
+                else:
+                    auth_scope['project']['name'] = auth_scope['project'].pop('id')
+                auth_resp = cls.request_auth_token(auth_scope, identity, keystone_server_url, ssl_verify)
+            except (requests.exceptions.HTTPError, requests.exceptions.Timeout, requests.exceptions.ConnectionError) as e:
+                exception_msg = "{msg} and also failed kaystone auth with identity:{id} scope:{scope} @{url}: {ex}".format(
+                    msg=exception_msg,
+                    id=identity,
+                    scope=auth_scope,
+                    url=keystone_server_url,
+                    ex=e)
+                raise KeystoneUnreachable(exception_msg)
 
         auth_token = auth_resp.headers.get('X-Subject-Token')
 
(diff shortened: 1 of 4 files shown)


checks.d/openstack.py
```diff
@@ -333,6 +357,7 @@ def __init__(self, name, init_config, agentConfig, instances=None):
         AgentCheck.__init__(self, name, init_config, agentConfig, instances)
 
         self._ssl_verify = init_config.get("ssl_verify", True)
+        self._use_proxy = init_config.get("use_agent_proxy", True)
```
truthbk: should this be `True` or `False` by default?

masci: I think both are legit so it's a matter of choosing the default that is going to fit the most likely scenario. I'd leave it as it is but open to change it if we receive strong feedback by users to do this differently.

checks.d/openstack.py
```diff
@@ -346,13 +371,31 @@ def __init__(self, name, init_config, agentConfig, instances=None):
         # Mapping of Nova-managed servers to tags
         self.external_host_tags = {}
 
+        # Set proxy settings
+        self.proxies = {
```
masci: Maybe not for this PR, but we should move this boilerplate code upper in the hierarchy or in a free function to avoid duplication and ease maintenance

conf.d/openstack.yaml.example
```diff
@@ -27,6 +27,9 @@ init_config:
       # need to set to false when using self-signed certs
       # ssl_verify: true
 
+     # Whether the dd-agent proxy should be used for openstack API requests (if set)
```
masci: Speaking as a newbie, what about adding `also`? Like `... should be also used for openstack ...`

checks.d/openstack.py
```diff
@@ -137,10 +139,32 @@ def from_config(cls, init_config, instance_config):
         auth_scope = cls.get_auth_scope(instance_config)
         identity = cls.get_user_identity(instance_config)
 
+        exception_msg = None
         try:
-            auth_resp = cls.request_auth_token(auth_scope, identity, keystone_server_url, ssl_verify)
+            auth_resp = cls.request_auth_token(auth_scope, identity, keystone_server_url, ssl_verify, proxy_config)
         except (requests.exceptions.HTTPError, requests.exceptions.Timeout, requests.exceptions.ConnectionError):
-            raise KeystoneUnreachable()
+            exception_msg = "Failed kaystone auth with identity:{id} scope:{scope} @{url}".format(
```
masci: typo: `keystone`

checks.d/openstack.py
```diff
@@ -137,10 +139,32 @@ def from_config(cls, init_config, instance_config):
...
+                id=identity,
+                scope=auth_scope,
+                url=keystone_server_url)
+
+        if exception_msg:
+            try:
+                identity['password']['user']['domain']['name'] = identity['password']['user']['domain'].pop('id')
```
masci: I see that `identity` is validated by `cls.get_user_identity`, still we access multiple keys in YOLO mode :)
Maybe we could add unit tests to `cls.get_user_identity` and `cls.get_auth_scope ` to keep this section more resilient to future changes.

masci:
Apart the typo this LGTM and we could :shipit: immediately, other comments are open to discussion.

checks.d/openstack.py
truthbk: Ah yes! Definitely, I had forgotten to do this! I realized too when I fixed the `http_check`.

checks.d/openstack.py
truthbk: There are already some tests for that here: https://github.com/DataDog/dd-agent/blob/master/tests/checks/mock/test_openstack.py#L155-L205

From what I saw, the YOLO mode shouldn't be too dangerous due to the previous validation.

masci: Sorry my bad, completely missed the test cases, that's exactly what I was talking about, as long as we have them, accessing those dict is perfectly fine as it is now.

PR #2573, opened 2016-06-06
[service discovery] Improve logging, test coverage and template variable interpolation

hkaj:
# What
- improve test coverage for service discovery and logging
- support ip address collection from networks when available (for docker > 1.9)
- make template variable interpolation more resilient
- add the ability to specify keys on top of indexes for template variables (eg: `host_bridge` will return the IP address of the host on its bridge network)
- kill some dead code

### DIFF
--- a/config.py
+++ b/config.py
@@ -829,18 +829,24 @@ def _service_disco_configs(agentConfig):
     """ Retrieve all the service disco configs and return their conf dicts
     """
     if agentConfig.get('service_discovery') and agentConfig.get('service_discovery_backend') in SD_BACKENDS:
-        sd_backend = get_sd_backend(agentConfig=agentConfig)
-        service_disco_configs = sd_backend.get_configs()
+        try:
+            log.info("Fetching service discovery check configurations.")
+            sd_backend = get_sd_backend(agentConfig=agentConfig)
+            service_disco_configs = sd_backend.get_configs()
+        except Exception:
+            log.exception("Loading service discovery configurations failed.")
     else:
         service_disco_configs = {}
 
     return service_disco_configs
+
 
 def _conf_path_to_check_name(conf_path):
     f = os.path.splitext(os.path.split(conf_path)[1])
     if f[1] and f[1] == ".default":
         f = os.path.splitext(f[0])
     return f[0]
+
 
 def get_checks_places(osname, agentConfig):
     """ Return a list of methods which, when called with a check name, will each return a check path to inspect
--- a/utils/service_discovery/abstract_sd_backend.py
+++ b/utils/service_discovery/abstract_sd_backend.py
@@ -41,7 +41,7 @@ class AbstractSDBackend(object):
             for key in tpl:
                 # iterate over template variables found in the templates
                 for var in self.PLACEHOLDER_REGEX.findall(str(tpl[key])):
-                    var_value = variables[var.strip('%')]
+                    var_value = variables.get(var.strip('%'))
                     if var_value is not None:
                         # if the variable is found in a list (for example {'tags': ['%%tags%%', 'env:prod']})
                         # we need to iterate over its elements
--- a/utils/service_discovery/abstract_sd_backend.py
+++ b/utils/service_discovery/abstract_sd_backend.py
@@ -41,14 +41,25 @@ class AbstractSDBackend(object):
             for key in tpl:
                 # iterate over template variables found in the templates
                 for var in self.PLACEHOLDER_REGEX.findall(str(tpl[key])):
-                    if var.strip('%') in variables and variables[var.strip('%')]:
+                    var_value = variables[var.strip('%')]
+                    if var_value is not None:
                         # if the variable is found in a list (for example {'tags': ['%%tags%%', 'env:prod']})
                         # we need to iterate over its elements
                         if isinstance(tpl[key], list):
-                            for idx, val in enumerate(tpl[key]):
-                                tpl[key][idx] = val.replace(var, variables[var.strip('%')])
+                            # if the variable is also a list we can just combine both lists
+                            if isinstance(var_value, list):
+                                tpl[key].remove(var)
+                                tpl[key] += var_value
+                                # remove dups
+                                tpl[key] = list(set(tpl[key]))
+                            else:
+                                for idx, val in enumerate(tpl[key]):
+                                    tpl[key][idx] = val.replace(var, var_value)
                         else:
-                            tpl[key] = tpl[key].replace(var, variables[var.strip('%')])
+                            if isinstance(var_value, list):
+                                tpl[key] = var_value
+                            else:
+                                tpl[key] = tpl[key].replace(var, var_value)
                     else:
                         log.warning('Failed to interpolate variable {0} for the {1} parameter.'
                                     ' Dropping this configuration.'.format(var, key))
--- a/utils/service_discovery/sd_docker_backend.py
+++ b/utils/service_discovery/sd_docker_backend.py
@@ -32,49 +32,100 @@ class SDDockerBackend(AbstractSDBackend):
             self.config_store = get_config_store(agentConfig=agentConfig)
 
         self.VAR_MAPPING = {
-            'host': self._get_host,
-            'port': self._get_ports,
+            'host': self._get_host_address,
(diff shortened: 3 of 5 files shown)


config.py
```diff
@@ -829,19 +829,25 @@ def _service_disco_configs(agentConfig):
...
     """
     if agentConfig.get('service_discovery') and agentConfig.get('service_discovery_backend') in SD_BACKENDS:
-        sd_backend = get_sd_backend(agentConfig=agentConfig)
-        service_disco_configs = sd_backend.get_configs()
+        try:
+            log.info("Fetching service discovery check configurations.")
+            sd_backend = get_sd_backend(agentConfig=agentConfig)
+            service_disco_configs = sd_backend.get_configs()
+        except Exception as ex:
+            log.error("Loading service discovery configurations failed: %s" % ex)
```
remh: log.exception to keep the traceback ?

utils/service_discovery/abstract_sd_backend.py
```diff
@@ -41,14 +41,25 @@ def _render_template(self, init_config_tpl, instance_tpl, variables):
             for key in tpl:
                 # iterate over template variables found in the templates
                 for var in self.PLACEHOLDER_REGEX.findall(str(tpl[key])):
-                    if var.strip('%') in variables and variables[var.strip('%')]:
+                    if var.strip('%') in variables and variables[var.strip('%')] is not None:
```
remh: You are stripping your `var` twice here. (And even a third time below).

```python
var_value = variables.get(var.strip('%'))
if var_value is not None:
   ...
````

is simpler to read and more efficient.

File: utils/service_discovery/sd_docker_backend.py [hunk 3]
```diff
@@ -32,49 +32,100 @@ def __init__(self, agentConfig):
...
+    def _extract_ip_from_networks(self, ip_dict, tpl_var):
+        """Extract a single IP from a dictionary made of network names and IPs."""
+        if not ip_dict:
+            return None
+        tpl_parts = tpl_var.split('_')
+
+        # no specifier, try to pick the bridge key, falls back to the value of the last key
+        if len(tpl_parts) < 2:
+            log.debug("No key was passed for template variable %s." % tpl_var)
+            if 'bridge' in ip_dict.keys():
```
remh:
```python
if 'bridge' in ip_dict:
```
Otherwise you'll create a new copy of your keys

utils/service_discovery/sd_docker_backend.py
```diff
@@ -32,49 +32,100 @@ def __init__(self, agentConfig):
...
+                log.debug("Using the bridge network instead.")
+                return ip_dict['bridge']
+            else:
+                last_key = sorted(ip_dict.keys())[-1]
```
remh: use `iterkeys` instead of `keys` here

utils/service_discovery/sd_docker_backend.py
```diff
@@ -32,49 +32,100 @@ def __init__(self, agentConfig):
...
+                log.debug("Trying with the last key: '%s' instead." % last_key)
+                return ip_dict[last_key]
+        else:
+            res = ip_dict.get(tpl_parts[-1])
+            if res is None:
+                log.warning("The key passed for template variable %s was not found." % tpl_var)
+                if 'bridge' in ip_dict.keys():
```
remh: same as in line 94

utils/service_discovery/sd_docker_backend.py
```diff
@@ -32,49 +32,100 @@ def __init__(self, agentConfig):
...
+                    log.warning("Using the bridge network instead.")
+                    return ip_dict['bridge']
+                else:
+                    last_key = sorted(ip_dict.keys())[-1]
```
remh: same as in line 98

utils/service_discovery/sd_docker_backend.py
```diff
@@ -86,7 +137,33 @@ def _get_ports(self, container_inspect):
...
+            return None
+
+        tpl_parts = tpl_var.split('_')
+
+        if len(tpl_parts) == 1:
+            log.debug("No index was passed for template variable %s. "
+                      "Trying with the last element." % tpl_var)
+            return ports[-1]
+
+        idx = self._is_int(tpl_parts[-1])
```
remh: you use `_is_int` just once.
this could be rewritten in:
```python
try:
    return ports[int(index)]
except ValueError:
    log.error("Index is not an integer. Using the last element instead")
except IndexError:
    log.error("Index is out of range. Using the last element instead")
return ports[-1]
```

utils/service_discovery/abstract_sd_backend.py
```diff
@@ -41,14 +41,25 @@ def _render_template(self, init_config_tpl, instance_tpl, variables):
             for key in tpl:
                 # iterate over template variables found in the templates
                 for var in self.PLACEHOLDER_REGEX.findall(str(tpl[key])):
-                    if var.strip('%') in variables and variables[var.strip('%')]:
+                    var_value = variables[var.strip('%')]
```
masci: Should be `var_value = variables.get(var.strip('%'))`?

utils/service_discovery/sd_docker_backend.py
```diff
@@ -32,49 +32,100 @@ def __init__(self, agentConfig):
...
+                return ip_addr
+
+        # try to get the bridge IP address
+        log.debug("No network found for container %s (%s), trying with IPAddress field" % (c_id[:12], c_img))
+        ip_addr = c_inspect.get('NetworkSettings', {}).get('IPAddress')
+        if ip_addr:
+            return ip_addr
 
-        return ip_addr
+        else:
```
masci: I think you can remove the `else`

utils/service_discovery/sd_docker_backend.py
```diff
@@ -32,49 +32,100 @@ def __init__(self, agentConfig):
...
+            res = ip_dict.get(tpl_parts[-1])
+            if res is None:
+                log.warning("The key passed for template variable %s was not found." % tpl_var)
+                if 'bridge' in ip_dict:
+                    log.warning("Using the bridge network instead.")
+                    return ip_dict['bridge']
+                else:
+                    last_key = sorted(ip_dict.iterkeys())[-1]
+                    log.warning("Trying with the last key: '%s' instead." % last_key)
+                    return ip_dict[last_key]
```
masci: it could be worth removing this duplicated `if/else` code block to ease maintainance

utils/service_discovery/sd_docker_backend.py
```diff
@@ -32,49 +32,100 @@ def __init__(self, agentConfig):
...
+            else:
+                last_key = sorted(ip_dict.iterkeys())[-1]
+                log.debug("Trying with the last key: '%s' instead." % last_key)
+                return ip_dict[last_key]
+        else:
+            res = ip_dict.get(tpl_parts[-1])
+            if res is None:
+                log.warning("The key passed for template variable %s was not found." % tpl_var)
+                if 'bridge' in ip_dict:
+                    log.warning("Using the bridge network instead.")
```
masci: is the logging level discrepancy intentional? (above the same issue is logged as `debug`)

utils/service_discovery/sd_docker_backend.py
```diff
@@ -113,11 +184,15 @@ def get_tags(self, c_inspect):
 
         return tags
 
-    def _get_additional_tags(self, container_inspect):
+    def _get_additional_tags(self, container_inspect, tpl_var):
```
masci: `tpl_var` is not used, is it ok?

utils/service_discovery/sd_docker_backend.py
hkaj: If no key is passed because there's no need for it, there is probably no need for a warning whereas if a key was explicitly passed but not found it definitely requires a warning. But I'm tempted to log both as warning. Better be a little spammy than too obscure.

masci: I was referring to `log.warning("Using the bridge network instead.")`

hkaj:
thanks for the feedback @masci, PR updated.

masci:
👍  LGTM

PR #2574, opened 2016-06-07
[mongo] Updated the declaration of service check tags to include those input …

avaughan89:
…into the yaml config file

Tested locally and confirmed in app here:

http://cl.ly/300R3G3M3p14

### DIFF
--- a/checks.d/mongo.py
+++ b/checks.d/mongo.py
@@ -576,28 +576,27 @@ class MongoDb(AgentCheck):
         db_name = parsed.get('database')
         clean_server_name = server.replace(password, "*" * 5) if password is not None else server
 
-        additional_metrics = instance.get('additional_metrics', [])
+        if not db_name:
+            self.log.debug('No MongoDB database found in URI. Defaulting to admin.')
+            db_name = 'admin'
 
         tags = instance.get('tags', [])
+        # de-dupe tags to avoid a memory leak
+        tags = list(set(tags))
+        service_check_tags = [
+            "db:%s" % db_name
+        ]
+        service_check_tags.extend(tags)
+        # Add the `server` tag to the metrics' tags only (it's added in the backend for service checks)
         tags.append('server:%s' % clean_server_name)
 
         # Get the list of metrics to collect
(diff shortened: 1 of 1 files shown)


MrTravisB:
Any chance you do this for the memcache check as well?

https://github.com/DataDog/dd-agent/blob/master/checks.d/mcache.py#L325

    service_check_tags = ["host:%s" % server, "port:%s" % port] + custom_tags

avaughan89:
@MrTravisB For that check it would look like:

`service_check_tags.extend(custom_tags)` 

with similar placement.  I believe we will look into adding this into all our checks.

olivielpeau:
Just a note: this service check already has a `host` tag that gets moved to a `server` tag in the backend, so with this change each service check will end up with 2 `server` tags (they can be seen on the screenshot you provided actually).

Not sure if it's something that could make the service check harder to use, but it's definitely something we need to take into account before merging.

olivielpeau:
Just amended the commit so that service checks don't have a duplicate `server` tag.

avaughan89:
@olivielpeau Good catch.  Thanks for updating.

checks.d/mongo.py
```diff
@@ -576,29 +576,28 @@ def total_seconds(td):
         db_name = parsed.get('database')
         clean_server_name = server.replace(password, "*" * 5) if password is not None else server
 
-        additional_metrics = instance.get('additional_metrics', [])
+        if not db_name:
+            self.log.info('No MongoDB database found in URI. Defaulting to admin.')
```
remh: Let's log in debug. Otherwise this will spam the log for little information.

remh: Just realized it was in info in the past. We should still log this to debug.

truthbk:
@olivielpeau any consensus on the usability of the additional `server` tag? Personally I don't think it makes all that sense given there's already a matching server tag available for the host.

olivielpeau:
just had a quick chat with @truthbk: I did amend the commit so that there wouldn't be a duplicate `server` tag on the service check, so this is good to go!

Merging

PR #2575, opened 2016-06-07
[Postgres] hard limit on pg count metric

masci:
Counting the number of tables to provide the `postgresql.table.count` metric can be very expensive, especially on large clusters.

This PR contains two changes to mitigate the impact of a slow count:
1) Put an hard limit on the count, not user-configurable
2) Suggest to disable the metric in the `postgres.yaml` file

For backward compatibility, the metric is still collected by default

**Notice**: couldn't provide an integration test for the hard limit feature because the strings containing the SQL code are class fields evaluated at import time.

### DIFF
--- a/checks.d/postgres.py
+++ b/checks.d/postgres.py
@@ -197,7 +197,7 @@ SELECT schemaname, count(*) FROM
 (
   SELECT schemaname
   FROM %s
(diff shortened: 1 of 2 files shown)


remh:
👍

PR #2578, opened 2016-06-07
[ci] use travis cache only

degemer:
Travis now does one cache per matrix element, which was missing
previously and was the reason why we started using a custom cache.


degemer:
Travis doc: https://docs.travis-ci.com/user/caching#Caches-and-build-matrices

masci:
👍

PR #2580, opened 2016-06-08
GNU-Friendly sed

sha1sum:
For those with GNU utilities installed (`sed` in particular), the OS X "vanilla" `sed` command to set the API key into the configuration breaks. The proposed changes in this PR will check whether the vanilla or GNU version of `sed` is installed and use the appropriate command depending on the `sed` version.


degemer:
I tried it with both vanilla sed and GNU one, and it works with both.
Thanks for the contribution @sha1sum!

Issue #2576, opened 2016-06-07
[core] Service discovery strips important information from image name


geowa4:
According to the service discovery guide, it works by matching the name of the image to a folder in etcd. However, when looking at the code (https://github.com/DataDog/dd-agent/blob/fd22e3c57d4d690e8bda902ed4dccdfb926fa71b/utils/service_discovery/sd_docker_backend.py#L138), this is not the case. It is only taking a small portion of the image name. This is causing conflicts for me that block me from using the feature. As an example, let's assume I have images `myregistry.com/foo/web` and `myregistry.com/bar/web`. These two projects have very different checks. As another example, assume I have images `myregistry.com/foo:web` and `myregistry.com/foo:worker` which share all layers except the command used to execute them. Obviously, the worker check will be much different from the web check.

In both cases, the two images look identical to the linked line of code. To fix this, I propose using labels (https://docs.docker.com/engine/userguide/labels-custom-metadata/) or environment variables. Labels were added to attach extra metadata, which is perfect for this use case. I also propose environment variables since that's what a lot of other tools still use as an artifact of history. I would be willing to write the code necessary to do this, but I wanted to bring this up for discussion beforehand. Let me know your thoughts.

Thanks,
George

geowa4:
I realized I wasn't clear how the labels and environment variables would work. I'm not suggesting storing the check config data in labels. I'd like to use labels and environment variables to store the key the agent uses to march in etcd. For instance, `docker run -e DATADOG_ID=foo-web myregistry.com/foo/web`. This way, instead of deducing "web" as the key to check in etcd, the agent would pull the environment variable DATADOG_ID and check etcd for `/datadog/check_configs/foo-web`, or more generically, `/datadog/check_configs/$DATADOG_ID`. This strategy would allow the agent to fall back on the current scheme for those that changed their image naming scheme to match what the agent is looking for.

geowa4:
Running the REPL, I've found that the environment variables are not getting pulled into the container dictionary. Labels are. 

```
>>> docker_client.containers()
[...]

PR #2583, opened 2016-06-08
[core] Use datadog_id label for service discovery.

geowa4:
This still falls back to parsing the image name when this label is not
present.

I created this as a result of #2576 and my conversation with support.

### DIFF
--- a/utils/service_discovery/sd_docker_backend.py
+++ b/utils/service_discovery/sd_docker_backend.py
@@ -131,11 +131,17 @@ class SDDockerBackend(AbstractSDBackend):
                 if c_id == status.get('containerID', '').split('//')[-1]:
                     return pod.get(key, {})
 
+    def _get_config_id(self, container):
+        config_id = container.get('Labels', {}).get('datadog_id', None)
+        if config_id is None or config_id == '':
+            config_id = container.get('Image').split(':')[0].split('/')[-1]
+        return config_id
+
     def get_configs(self):
         """Get the config for all docker containers running on the host."""
         configs = {}
         containers = [(
-            container.get('Image').split(':')[0].split('/')[-1],
+            self._get_config_id(container),
             container.get('Id'), container.get('Labels')
         ) for container in self.docker_client.containers()]
 
--- a/tests/core/test_service_discovery.py
+++ b/tests/core/test_service_discovery.py
@@ -63,9 +63,17 @@ def client_read(path):
 class TestServiceDiscovery(unittest.TestCase):
     docker_container_inspect = {
         u'Id': u'69ff25598b2314d1cdb7752cc3a659fb1c1352b32546af4f1454321550e842c0',
-        u'Image': u'6ffc02088cb870652eca9ccd4c4fb582f75b29af2879792ed09bb46fd1c898ef',
+        u'Image': u'nginx',
         u'Name': u'/nginx',
-        u'NetworkSettings': {u'IPAddress': u'172.17.0.21', u'Ports': {u'443/tcp': None, u'80/tcp': None}}
+        u'NetworkSettings': {u'IPAddress': u'172.17.0.21', u'Ports': {u'443/tcp': None, u'80/tcp': None}},
+        u'Labels': {}
+    }
+    docker_container_inspect_with_label = {
+        u'Id': u'69ff25598b2314d1cdb7752cc3a659fb1c1352b32546af4f1454321550e842c0',
+        u'Image': u'private-registry.com/nginx:production',
+        u'Name': u'/nginx',
+        u'NetworkSettings': {u'IPAddress': u'172.17.0.21', u'Ports': {u'443/tcp': None, u'80/tcp': None}},
+        u'Labels': {u'datadog_id': u'custom-nginx'}
     }
     kubernetes_container_inspect = {
         u'Id': u'389dc8a4361f3d6c866e9e9a7b6972b26a31c589c4e2f097375d55656a070bc9',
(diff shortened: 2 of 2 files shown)


geowa4:
The unit test failure appears to be in installing Kong and unrelated to the pull request.

hkaj:
Hi @geowa4, thanks for submitting a PR. I understand your issue and it's definitely worth fixing, but I'm not sure this is the right approach. We should probably just update the path we query in the kv store to include repo and tag.

The problem with using something else is that we're already watching the config store for changes in keys and values to trigger a configuration reload in the agent. Moving this information somewhere else would require adding more calls to external APIs for this purpose and we don't want that.

Would it work for you if we move to `repository/path/image_name:image_tag` in the store instead of `image_name` only? Possibly replacing `/` with `-`, I didn't try it yet.

geowa4:
Hi @hkaj, I'm not sure I understand your opposition. This requires no additional calls to any other APIs.

hkaj:
@geowa4 sorry that wasn't clear.

What I'm thinking about is the case where you want to add or remove a check for a container. In this case you would modify the label to add/remove a key where the config template can be found.
That means the agent has to inspect containers to see if labels were added, removed or modified since its previous run and then query the new path or stop querying an old one.

That's more expensive to perform than just watching the canonical path (`/datadog/check_configs` by default). Especially because we already watch this path for changes in configuration templates (added tags, modified options, etc.).

Improving the path achieves the same goal without needing to inspect containers for updated labels.

geowa4:
@hkaj, I think I'm still missing something. All I've done with this change is offer an alternative to parsing the image name for a magic string. From what I gather, you take that parsed string and append it to a predefined path like so: `/datadog/check_configs/$PARSED_IMAGE_NAME`. What I have submitted in this PR is optionally using an explicit label to replace that parsing, making the path `/datadog/check_configs/$SPECIAL_LABEL_VALUE`.

There is no continuous checking for updates to any labels. The mode of operation is exactly the same as it is now, which is how you want it. Also, I don't think you can change the labels of a running container.

hkaj:
ooh my bad, I didn't get it correctly. It shouldn't be too expensive to run actually.

Let me give you some more context:

Right now at each run the agent looks for changes in the config store[0][1] and triggers a configuration reload[2] if changes happened (templates were edited, added, removed, or checks were added/removed for some containers).
So let's say you have an nginx container with the nginx check enabled through a config template. If you want to also enable the http_check against it you update the template and the agent will pick-up the new config without having to restart.

Now if we use container labels to specify which template to query and apply to each container we'll also need to look for changes in labels for every container.

That's not as expensive as I thought since we can simply keep a map of these labels somewhere and run `docker.Client().containers(filter=blah)` to get the current one at each run and update it, but fixing the path issue by using the full image name (repo and tag included) makes this part unnecessary and thus seems more efficient to me.

Let me know what you think.

[0]: https://github.com/DataDog/dd-agent/blob/55feac7d7549361e929f5586e05c8bd4ea591802/agent.py#L201-L210
[1]: https://github.com/DataDog/dd-agent/blob/25b130b05ce70f79353005515bb3cd29a4a0be93/utils/service_discovery/abstract_config_store.py#L155-L175
[2]: https://github.com/DataDog/dd-agent/blob/55feac7d7549361e929f5586e05c8bd4ea591802/agent.py#L212-L219

geowa4:
> Now if we use container labels to specify which template to query and apply to each container we'll also need to look for changes in labels for every container.

I don't understand why that's true. Each container has many labels just as it has exactly one image name. For a running container, labels change as often as images names, which is never. This PR introduces a special label so we can amend that statement to "each container has exactly one `datadog_id` label just as it has exactly one image name". Logically, these are now equivalent.

I do not support replacing characters in the image name. This practice still loses information provided by the user; it just loses less than the current implementation. While it is an improvement, it is not sufficient, especially when other more explicit solutions exist. For example, if someone is refactoring in a way requiring different checks and changes her images from `private-registry.com/foo/bar:web` to `private-registry.com/foo/bar-web`, you're going to make her angry. This is not acceptable.

geowa4:
The effect of my proposed change:
```
$ python
Python 2.7.11 (default, Dec  5 2015, 20:44:19)
[GCC 4.2.1 Compatible Apple LLVM 7.0.0 (clang-700.1.76)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from utils.dockerutil import DockerUtil
>>> docker_client = DockerUtil().client
>>> docker_client.containers()
[{u'Status': u'Up 45 hours', u'Created': 1465399125, u'Image': u'nginx', u'Labels': {u'datadog_id': u'custom-nginx'}, u'NetworkSettings': {u'Networks': {u'bridge': {u'NetworkID': u'', u'MacAddress': u'', u'GlobalIPv6PrefixLen': 0, u'Links': None, u'GlobalIPv6Address': u'', u'IPv6Gateway': u'', u'IPAMConfig': None, u'EndpointID': u'', u'IPPrefixLen': 16, u'IPAddress': u'172.17.0.2', u'Gateway': u'172.17.0.1', u'Aliases': None}}}, u'HostConfig': {u'NetworkMode': u'default'}, u'ImageID': u'sha256:6f8d099c3adc0f9c14bdc11d6c698a647c2f0457aa1ae76a592d6f7dc9bdb5b8', u'State': u'running', u'Command': u"nginx -g 'daemon off;'", u'Names': [u'/stoic_allen'], u'Mounts': [], u'Id': u'445272bc04e4ad5f0d0a480e7a56af36aeef57053601835ef3579d98cdf3e567', u'Ports': [{u'IP': u'0.0.0.0', u'Type': u'tcp', u'PublicPort': 32770, u'PrivatePort': 443}, {u'IP': u'0.0.0.0', u'Type': u'tcp', u'PublicPort': 32771, u'PrivatePort': 80}]}]
>>> docker_client.containers()[0].get('Image').split(':')[0].split('/')[-1]
u'nginx'
>>> docker_client.containers()[0].get('Labels', {}).get('datadog_id')
u'custom-nginx'
```

The `container` dictionary also has a `Labels` key that is a dictionary of all labels applied to the container. My change simply looks for a key `datadog_id` in that dictionary. If it does not exist, it falls back to the equivalent of the original code which uses the `Image` key.

utils/service_discovery/sd_docker_backend.py
```diff
@@ -131,11 +131,17 @@ def _get_kube_config(self, c_id, key):
                 if c_id == status.get('containerID', '').split('//')[-1]:
                     return pod.get(key, {})
 
+    def _get_config_id(self, container):
+        config_id = container.get('Labels', {}).get('datadog_id')
```
danroyo: If the first `get` call uses the fallback value, the second call here is going to throw a KeyError

geowa4: It doesn't, but I updated the code to be very explicit and avoid confusion.

From my REPL:
```
>>> {}.get('datadog_id') is None
True
```

geowa4:
Superseded by #2593.

PR #2586, opened 2016-06-10
[http_check] Add allow_redirects parameter

n0ts:
Hi, I add allow_redirects parameter http_check. Because I want to check http status only 301 or 302.

Regards,

### DIFF
--- a/checks.d/http_check.py
+++ b/checks.d/http_check.py
@@ -198,15 +198,16 @@ class HTTPCheck(NetworkCheck):
         weakcipher = _is_affirmative(instance.get('weakciphers', False))
         ignore_ssl_warning = _is_affirmative(instance.get('ignore_ssl_warning', False))
         skip_proxy = _is_affirmative(instance.get('no_proxy', False))
+        allow_redirects = _is_affirmative(instance.get('allow_redirects', True))
 
         return url, username, password, http_response_status_code, timeout, include_content,\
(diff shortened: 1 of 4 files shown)


truthbk:
Hi @n0ts thank you very much for this. This definitely looks like something we were missing. We'll review this in-depth soon (although it already looks good - thanks for the test), and we hope to merge it soon. Thanks!

checks.d/http_check.py
```diff
@@ -243,7 +244,8 @@ def _check(self, instance):
                 self.log.debug("Weak Ciphers will be used for {0}. Suppoted Cipherlist: {1}".format(
                     base_addr, WeakCiphersHTTPSConnection.SUPPORTED_CIPHERS))
 
-            r = sess.request('GET', addr, auth=auth, timeout=timeout, headers=headers, proxies=instance_proxy,
+            r = sess.request('GET', addr, auth=auth, timeout=timeout, headers=headers,
+                             proxies=instance_proxy, allow_redirects=allow_redirects,
```
truthbk: In `requests` `allow_redirects` defaults to `True` as well so 👍  for the the default value as it maintains backward compatibility.

truthbk:
CI failures are unrelated, and this looks good to me! Especially with that test you wrote @n0ts 

Thank you so much! :shipit:

PR #2588, opened 2016-06-10
[flare] Add sd_configcheck output to flare command

hkaj:
Right now the flare command doesn't collect service discovery configcheck command, we need to run `/etc/init.d/datadog-agent configcheck` explicitely. It's a very useful command for investigating config issues so let's add it.

### DIFF
--- a/agent.py
+++ b/agent.py
@@ -39,12 +39,12 @@ from util import (
     get_hostname,
     Watchdog,
 )
-from utils.flare import configcheck, Flare
+from utils.flare import Flare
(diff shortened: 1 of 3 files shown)


degemer:
LGTM, only one remark: why not merge the default `configcheck` and `get_sd_configcheck` in a `utils/configcheck.py` ?

hkaj:
good point @degemer ! Done, let me know what you think.

degemer:
Looks really nice now! 💯 
:shipit:

PR #2590, opened 2016-06-13
[haproxy] Add `collate_status_tags_per_host` option

olivielpeau:
Fixes #2315

- Refactored the aggregation method by implementing a counter by context
- Added a few mock tests
- Fixed a bug on aggregates when `count_status_by_service` is disabled

### DIFF
--- a/checks.d/haproxy.py
+++ b/checks.d/haproxy.py
@@ -26,6 +26,17 @@ class Services(object):
     ALL_STATUSES = (
         'up', 'open', 'down', 'maint', 'nolb'
     )
+    AVAILABLE = 'available'
+    UNAVAILABLE = 'unavailable'
+    COLLATED_STATUSES = (AVAILABLE, UNAVAILABLE)
+
+    STATUS_MAP = {
+        'up': AVAILABLE,
+        'open': AVAILABLE,
+        'down': UNAVAILABLE,
+        'maint': UNAVAILABLE,
+        'nolb': UNAVAILABLE
+    }
 
     STATUSES_TO_SERVICE_CHECK = {
         'UP': AgentCheck.OK,
@@ -87,6 +98,10 @@ class HAProxy(AgentCheck):
             instance.get('collect_status_metrics_by_host', False)
         )
 
+        collate_status_tags_per_host = _is_affirmative(
+            instance.get('collate_status_tags_per_host', False)
+        )
+
         count_status_by_service = _is_affirmative(
             instance.get('count_status_by_service', True)
         )
@@ -113,6 +128,7 @@ class HAProxy(AgentCheck):
             tag_service_check_by_host=tag_service_check_by_host,
             services_incl_filter=services_incl_filter,
             services_excl_filter=services_excl_filter,
+            collate_status_tags_per_host=collate_status_tags_per_host,
             count_status_by_service=count_status_by_service,
         )
 
@@ -133,7 +149,8 @@ class HAProxy(AgentCheck):
     def _process_data(self, data, collect_aggregates_only, process_events, url=None,
                       collect_status_metrics=False, collect_status_metrics_by_host=False,
                       tag_service_check_by_host=False, services_incl_filter=None,
-                      services_excl_filter=None, count_status_by_service=True):
+                      services_excl_filter=None, collate_status_tags_per_host=False,
+                      count_status_by_service=True):
         ''' Main data-processing loop. For each piece of useful data, we'll
         either save a metric, save an event or both. '''
 
@@ -190,6 +207,7 @@ class HAProxy(AgentCheck):
                 self.hosts_statuses, collect_status_metrics_by_host,
                 services_incl_filter=services_incl_filter,
                 services_excl_filter=services_excl_filter,
+                collate_status_tags_per_host=collate_status_tags_per_host,
                 count_status_by_service=count_status_by_service
             )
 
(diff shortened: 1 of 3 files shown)


checks.d/haproxy.py
```diff
@@ -303,14 +321,13 @@ def _process_backend_hosts_metric(self, hosts_statuses, services_incl_filter=Non
...
 
-        # use a counter unless we have a unique tag set to gauge
-        counter = defaultdict(int)
-        if count_status_by_service and collect_status_metrics_by_host:
-            # `service` and `backend` tags will exist
-            counter = None
+        reported_statuses = Services.ALL_STATUSES
+        if collate_status_tags_per_host:
+            reported_statuses = Services.COLLATED_STATUSES
+        counter = defaultdict(lambda: {status: 0 for status in reported_statuses})
```
yannmh: Although it was already named like that before those changes, can we use a more comprehensive varaible name than `counter` ?

checks.d/haproxy.py
```diff
@@ -319,51 +336,37 @@ def _process_status_metric(self, hosts_statuses, collect_status_metrics_by_host,
...
-                tags.append('service:%s' % service)
-
             if self._is_service_excl_filtered(service, services_incl_filter, services_excl_filter):
                 continue
 
+            tags = []
+            if count_status_by_service:
+                tags.append('service:%s' % service)
             if collect_status_metrics_by_host:
                 tags.append('backend:%s' % hostname)
```
yannmh:
```python
            try:
                service, hostname, status = host_status
            except Exception:
                service, status = host_status

```
Are we sure that the `hostname` variable is always defined when `collect_status_metrics_by_host ` ?

checks.d/haproxy.py
```diff
@@ -319,51 +336,37 @@ def _process_status_metric(self, hosts_statuses, collect_status_metrics_by_host,
...
-            self.gauge(metric_name, count, tags + ['status:%s' % status])
-
-        for state in Services.ALL_STATUSES:
-            if state != status:
-                self.gauge(metric_name, 0, tags + ['status:%s' % state.replace(" ", "_")])
+            # Compute aggregates
+            agg_tags = []
+            if count_status_by_service:
+                agg_tags.append('service:%s' % service)
+            agg_statuses_counter[tuple(agg_tags)][Services.STATUS_MAP.get(status, status)] += count
```
yannmh: Can we just use `counter_status` here instead of re-resolving its value (`Services.STATUS_MAP.get(status, status)`) ?

checks.d/haproxy.py
```diff
@@ -319,51 +336,37 @@ def _process_status_metric(self, hosts_statuses, collect_status_metrics_by_host,
...
+        for key, count_per_status in counter.iteritems():
+            key_tags = list(key)
+            for status, count in count_per_status.iteritems():
+                self.gauge('haproxy.count_per_status', count, tags=key_tags + ['status:%s' % status])
+
+        # Send aggregates
+        for service, service_agg_statuses in agg_statuses_counter.iteritems():
+            service_tags = list(service)
+            for status, count in service_agg_statuses.iteritems():
+                self.gauge("haproxy.count_per_status", count, tags=service_tags + ['status:%s' % status])
```
yannmh: Overriding `haproxy.count_per_status` would lead to unexpected behavior.

See
```python
from time import sleep
from statsd import statsd


# Samples
context_a = [u"foo:bar"]
context_b = [u"foo:qux"]

metric_name = u"metric.test.aggregate"


while True:
    # != contexts
    statsd.gauge(metric_name, -10, tags=context_a)
    statsd.gauge(metric_name, 10, tags=context_b)

    # Overwrite aggregate
    statsd.gauge(metric_name, 50, tags=[])

    sleep(10)
```
and
unspecified

tests/checks/mock/test_haproxy.py
```diff
@@ -0,0 +1,138 @@
...
+            self.assertMetric('haproxy.count_per_status', value=1, tags=['status:available', 'service:a'])
+            self.assertMetric('haproxy.count_per_status', value=4, tags=['status:available', 'service:b'])
+            self.assertMetric('haproxy.count_per_status', value=2, tags=['status:unavailable', 'service:b'])
+            self.assertMetric('haproxy.count_per_status', value=1, tags=['status:available', 'service:c'])
+            self.assertMetric('haproxy.count_per_status', value=1, tags=['status:unavailable', 'service:c'])
+        else:
+            self.assertMetric('haproxy.count_per_status', value=6, tags=['status:available'])
+            self.assertMetric('haproxy.count_per_status', value=3, tags=['status:unavailable'])
+
+    @mock.patch('requests.get', return_value=MockResponse())
```
yannmh: You can simply use `Mock` from the `mock` module instead of `MockResponse`.
```
@mock.patch('requests.get', return_value=Mock(content=MOCK_DATA))
```

checks.d/haproxy.py
```diff
@@ -319,51 +336,37 @@ def _process_status_metric(self, hosts_statuses, collect_status_metrics_by_host,
...
+            agg_statuses_counter[tuple(agg_tags)][Services.STATUS_MAP.get(status, status)] += count
+
+        for key, count_per_status in counter.iteritems():
+            key_tags = list(key)
+            for status, count in count_per_status.iteritems():
+                self.gauge('haproxy.count_per_status', count, tags=key_tags + ['status:%s' % status])
+
+        # Send aggregates
+        for service, service_agg_statuses in agg_statuses_counter.iteritems():
+            service_tags = list(service)
```
yannmh: This copy is unnecessary.

checks.d/haproxy.py
```diff
@@ -319,51 +336,37 @@ def _process_status_metric(self, hosts_statuses, collect_status_metrics_by_host,
...
-            if state != status:
-                self.gauge(metric_name, 0, tags + ['status:%s' % state.replace(" ", "_")])
+            # Compute aggregates
+            agg_tags = []
+            if count_status_by_service:
+                agg_tags.append('service:%s' % service)
+            agg_statuses_counter[tuple(agg_tags)][Services.STATUS_MAP.get(status, status)] += count
+
+        for key, count_per_status in counter.iteritems():
+            key_tags = list(key)
```
yannmh: This copy is unnecessary.

tests/checks/mock/test_haproxy.py
```diff
@@ -0,0 +1,138 @@
...
+
+class MockResponse(object):
+    def __init__(self):
+        self.content = MOCK_DATA
+
+    def raise_for_status(self):
+        pass
+
+
+class TestCheckHAProxy(AgentCheckTest):
```
yannmh: 🙇

yannmh:
@olivielpeau the logic behind `collate_status_tags_per_host` looks good to me.
However I realize that the current code as well as your changes might suffer from the same issue: the `haproxy.count_per_status` metric is sent with conflicting contexts leading to unexpected aggregation results 😢 , c.f. my comment above.

checks.d/haproxy.py
olivielpeau: good point, `hostname` is supposed to be there since it's extracted [here](https://github.com/DataDog/dd-agent/blob/5.8.2/checks.d/haproxy.py#L237-L240), but the extraction could potentially not work as expected if the collected data isn't valid.

It's kind of outside the scope of this PR but it should be handled in a better way, agreed, I'll work on that

checks.d/haproxy.py
olivielpeau: hmm, the `counter_status` depends on the `collate_status_tags_per_host` option, whereas this `status` for the aggregate metrics is always a "collated" status, but I can make the code look up the value only once indeed. I'll change this

tests/checks/mock/test_haproxy.py
olivielpeau: oh didn't know about this, thanks!

checks.d/haproxy.py
olivielpeau: I had forgotten about this issue, thanks for pointing it out.

I don't really know what to think: indeed it seems like it's better to send a metric with a consistent set of tag keys, but the check was already not entirely complying with this (when `collect_status_metrics_by_host` is enabled, the `backend:` tag would be set for the non-aggregate metrics only).

That said, to fix most of the issue we could either:
- change the metric name for the aggregate statuses to `haproxy.count_per_collated_status` for instance. Drawback: this would be a breaking change for current users of the check.
- use `status:up` and `status:down` instead of `status:available` and `status:unavailable` for the non-aggregate statuses when `collate_status_tags_per_host` is enabled. This would prevent breaking the existing monitors/graphs scoped to either `status:available` or `status:unavailable`. I'd prefer doing this: even though we're not entirely solving the issue, at least there's a tag value that separates the "aggregates" from the "non-aggregates".

I'd appreciate your thoughts on this :)

olivielpeau:
Thanks @yannmh for the review, I had overlooked this issue indeed. I've addressed the first comments you made in new commits.

Copy-pasting my inline response to your comment here (since it got "outdated" by Github because the line of code has changed):

I don't really know what to think: indeed it seems like it's better to send a metric with a consistent set of tag keys, but the check was already not entirely complying with this (for instance, when `collect_status_metrics_by_host` is enabled, the `backend:` tag is set for the non-aggregate metrics only).

That said, to fix most of the issue we could either:

- change the metric name for the aggregate statuses to `haproxy.count_per_collated_status` for instance. Drawback: this would be a breaking change for current users of the check.
- use `status:up` and `status:down` instead of `status:available` and `status:unavailable` for the non-aggregate statuses when `collate_status_tags_per_host` is enabled. This would prevent breaking the existing monitors/graphs scoped to either `status:available` or `status:unavailable`. I like this solution: even though we're not entirely solving the issue, at least there's a tag value that separates the "aggregates" from the "non-aggregates". And I think we can always assume that users query the metric with the `status:` tag (since that's the whole point of the metric).
I'd appreciate your thoughts on this :)

yannmh:
> That said, to fix most of the issue we could either:
* change the metric name for the aggregate statuses to haproxy.count_per_collated_status for instance. Drawback: this would be a breaking change for current users of the check.

I agree, let's not do that.

> * use status:up and status:down instead of status:available and status:unavailable for the non-aggregate statuses when collate_status_tags_per_host is enabled. This would prevent breaking the existing monitors/graphs scoped to either status:available or status:unavailable. I like this solution: even though we're not entirely solving the issue, at least there's a tag value that separates the "aggregates" from the "non-aggregates". And I think we can always assume that users query the metric with the status: tag (since that's the whole point of the metric).

I believe we should keep `status:available` and `status:unavailable` for *collated* statuses to keep a distinction between *collated* and *not-collated* statuses, and avoid context superpositions. On top of that, we could run the `agg_statuses_counter` logic only when  `collate_status_tags_per_host == False`.

To sum up we'd have something like:

1. `collate_status_tags_per_host == False`.
   ```
  +======================+==========================+==================+===================================+
| From variable        | Metric name              | Base tags        |                                   |
+----------------------+--------------------------+------------------+-----------------------------------+
| statuses_counter     | haproxy.count_per_status | service, backend | status ∈ [up, open, down ... ]    |
+----------------------+--------------------------+------------------+-----------------------------------+
| agg_statuses_counter | haproxy.count_per_status | service          | status ∈ [available, unavailable] |
+----------------------+--------------------------+------------------+-----------------------------------+
   ```
2. `collate_status_tags_per_host == True`
  ```
  +======================+==========================+==================+===================================+
| From variable        | Metric name              | Base tags        |                                   |
+----------------------+--------------------------+------------------+-----------------------------------+
| statuses_counter     | haproxy.count_per_status | service, backend | status ∈ [available, unavailable] |
+----------------------+--------------------------+------------------+-----------------------------------+
| agg_statuses_counter | -                        | -                | -                                 |
+----------------------+--------------------------+------------------+-----------------------------------+

  ```

olivielpeau:
@yannmh Your comment makes a lot of sense, I've implemented it and pushed the commit.

The PR is ready for a final review, could you have a look? :)

yannmh:
:shipit:

olivielpeau:
Thanks!

Rebased and squashed, will merge once the CI passes.

olivielpeau:
CI failure (Kong) is unrelated. Merging!

PR #2591, opened 2016-06-13
[win32_event_log] Fix small regression on `msg_text` selection

olivielpeau:
When the windows event `Message` evaluates to a "false" value, do not
select its value as the event's `msg_text`.

Same behavior for the `InsertionStrings`.

Fixes the integration test. This small regression was introduced in #2136.


yannmh:
:shipit:

PR #2592, opened 2016-06-13
[core] always default to ipv6, then fallback to ipv4.

truthbk:
## Why
On older `tornado` versions (including the `dd-agent` default) the `allow_ipv6` option would default to `True` when using the `curl_httpclient` (expected), but would actually default to `False` when using the `simple_httpclient`. Since our default http client is the simple one, this implies that by default our ipv6 support requires enabling the curl client. This PR should address this.

Note: on newer tornado's the weird behavior was corrected, but doing this explicitly will hurt no one.


gmmeyer:
This is surprisingly simple. 👍

PR #2593, opened 2016-06-13
[service discovery] Replace buggy image name extraction with get_identifier.

### LINKED ISSUE #2576: [core] Service discovery strips important information from image name

hkaj:
# Why
To make it possible to associate explicitly a configuration template with a container, and fix the image name extraction.

# What
- This PR makes it possible to assign a `datadog_id` label to containers that specify the config store key associated with the configuration template for said container. If such label is not applied, falls back to image name matching.
- The image name matching has also been fixed, it didn't play well with private image repositories serving on a custom port. To avoid breaking backward compatibility we still fall back to the old format. The new resolution order is: `configuration file --> container label --> full image name --> legacy image name --> auto-config`

# Related issues/PRs
- supersedes #2583 
- fixes #2576

### DIFF
--- a/utils/service_discovery/abstract_config_store.py
+++ b/utils/service_discovery/abstract_config_store.py
@@ -106,7 +106,7 @@ class AbstractConfigStore(object):
 
         if len(check_names) != len(init_config_tpls) or len(check_names) != len(instance_tpls):
             log.error('Malformed configuration template: check_names, init_configs '
-                      'and instances are not all the same length. Container with identifier {0} '
+                      'and instances are not all the same length. Container with identifier {} '
                       'will not be configured by the service discovery'.format(identifier))
             return []
 
@@ -125,7 +125,7 @@ class AbstractConfigStore(object):
                 source = CONFIG_FROM_TEMPLATE
                 check_names, init_config_tpls, instance_tpls = res
             else:
-                log.debug("Could not find directory {0} in the config store, "
+                log.debug("Could not find directory {} in the config store, "
                           "trying to convert to the old format...".format(identifier))
                 legacy_ident = identifier.split(':')[0].split('/')[-1]
                 res = self._issue_read(legacy_ident)
@@ -137,7 +137,7 @@ class AbstractConfigStore(object):
         except (KeyError, TimeoutError, json.JSONDecodeError) as ex:
             # this is kind of expected, it means that no template was provided for this container
             if isinstance(ex, KeyError):
-                log.debug("Could not find directory {0} in the config store, "
+                log.debug("Could not find directory {} in the config store, "
                           "trying to auto-configure a check...".format(identifier))
             # this case is not expected, the agent can't reach the config store
             if isinstance(ex, TimeoutError):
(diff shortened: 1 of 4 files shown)


geowa4:
From https://docs.docker.com/engine/userguide/labels-custom-metadata/, which I only recently realized: 
> To prevent naming conflicts, Docker recommends using namespaces to label keys using reverse domain notation.

I recommend using `com.datadoghq.service-discovery.identifier` as the label instead of `datadog_id`.

hkaj:
good point @geowa4

utils/service_discovery/abstract_config_store.py
```diff
@@ -113,45 +117,65 @@ def get_check_tpls(self, image, **kwargs):
...
-            source = CONFIG_FROM_TEMPLATE
-        except (KeyNotFound, TimeoutError, json.JSONDecodeError) as ex:
-            # first case is kind of expected, it means that no template was provided for this container
-            if isinstance(ex, KeyNotFound):
+            res = self._issue_read(identifier)
+            if res and len(res) == 3:
+                source = CONFIG_FROM_TEMPLATE
+                check_names, init_config_tpls, instance_tpls = res
+            else:
+                log.debug("Could not find directory {0} in the config store, "
```
masci: small notice: since we dropped 2.6 support, we can avoid using indexes like `{0}` when not needed, `{}` is fine.

utils/service_discovery/sd_docker_backend.py
```diff
@@ -172,13 +175,17 @@ def get_configs(self):
                               ' discovery failed, leaving it alone.' % (cid[:12], image))
         return configs
 
-    def _get_check_configs(self, c_id, image, trace_config=False):
+    def get_config_id(self, image, labels):
+        """Look for a DATADOG_ID label, return its value or the image name if missing"""
+        return labels.get(DATADOG_ID) if DATADOG_ID in labels else image
```
masci: also `return labels.get(DATADOG_ID) or image`

masci:
Not sure why tests are not passing, otherwise 👍  LGTM

hkaj:
This doesn't work in Kubernetes as we can't apply labels to individual containers here. We'll have to find an alternative way.

PR #2597, opened 2016-06-14
discovery: fix getting host ip for overlay network

awasilyev:
Getting container %%host%% variable is broken for overlay multihost docker network.
Here is example of NetworkSettings for container, launched in overlay network with swarm: https://gist.github.com/awasilyev/d1194c638a96667478998c0f74f53a2b

This pr tries to solve this case by checking ip address for the first network defined in Network section.

### DIFF
--- a/utils/service_discovery/sd_docker_backend.py
+++ b/utils/service_discovery/sd_docker_backend.py
@@ -42,6 +42,10 @@ class SDDockerBackend(AbstractSDBackend):
         """Extract the host IP from a docker inspect object, or the kubelet API."""
         ip_addr = container_inspect.get('NetworkSettings', {}).get('IPAddress')
         if not ip_addr:
+            # overlay network case
+            nets = container_inspect.get('NetworkSettings', {}).get('Networks')
+            if nets: ip_addr = nets[nets.keys()[0]].get('IPAddress')
+        if not ip_addr:
             if not is_k8s():
                 return
             # kubernetes case


hkaj:
Hi @awasilyev, thanks for contributing! This is a known issue and we already have a PR that fixes it (among other various improvements: #2573 
It also introduces the ability to choose explicitly the network to use through `%%host_network-name%%`.

We still appreciate that you took some time to contribute!
Let me know if you have any question/comment.

awasilyev:
good news, I will wait for new release.
do you have plans for supporting custom tags based on container labels?

hkaj:
we already support it for docker metrics: https://github.com/DataDog/dd-agent/blob/5.8.0/conf.d/docker_daemon.yaml.example#L123-L126 is it something you'd like to see somewhere else?

awasilyev:
I mean tags for discovered services. Say, for example, rabbitmq container labeled 'com.example.tenant=staging' was created, and all metrics, gathered from this container should have tag com.example.tenant:staging tag

awasilyev:
Another currently uncovered case - when dd-agent is not dockerized and running on the node.
In this case we need to check Ports dictionary and get HostIp/HostPort values, for example as %%hostip%% / %hostport%.

            "Ports": {
                "15672/tcp": [
                    {
                        "HostIp": "10.11.0.43",
                        "HostPort": "32768"
                    }
                ],
            },

hkaj:
Yes collecting container labels as tags is an idea we're playing with. I don't have any ETA for it yet though.

About your latest comment, service discovery is mostly focused on environments where the agent is containerized, but that's probably something we should cover too and shouldn't be too much work to implement. I'll add a feature request for it.

I'm going to close this PR for now, but feel free to re-open if you have any more question or remark.

Thanks again!

JustAdam:
Regarding when dd-agent is running on the node, another option is to adding the container to the node's bridge network.

PR #2600, opened 2016-06-15
[gohai][processes] Simplify payload

olivielpeau:
- Only send a processes payload when gohai actually returns data
- Remove the `format_description` and `format_version` fields

### DIFF
--- a/checks/collector.py
+++ b/checks/collector.py
@@ -378,7 +378,7 @@ class Collector(object):
                         payload['resources'] = {
                             'processes': processes_payload,
                             'meta': {
-                                'host': payload['internalHostname'],
+                                'host': self.hostname,
                             }
                         }
                 except Exception:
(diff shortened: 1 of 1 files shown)


checks/collector.py
```diff
@@ -383,19 +369,18 @@ def run(self, checksd=None, start_event=True, configs_reloaded=False):
...
+                    processes_snaps = gohai_processes_json.get('processes')
+                    if processes_snaps:
+                        processes_payload = {
+                            'snaps': [processes_snaps]
+                        }
+
+                        payload['resources'] = {
+                            'processes': processes_payload,
+                            'meta': {
+                                'host': payload['internalHostname'],
```
degemer: Why not use `self.hostname` ? I think it carries more sense than `paylaod['internalHostname]` (while being the same thing).

degemer:
LGTM apart from the small remark.
I trust you on the uselessness of `format_description` and `format_version`, I didn't check.

checks/collector.py
olivielpeau: For no reason other than my own laziness to change it ;) Thanks for spotting this!

degemer:
👍

PR #2602, opened 2016-06-15
[sqlserver] expose connection_string in debug logs

stephenlechner:
This will be helpful for troubleshooting certain connection issues we've seen.


stephenlechner:
We actually think we've found the cause of the issue we were trying to investigate, so this extra logging is probably not necessary at this point. Closing without merging.

Issue #2596, opened 2016-06-14
mysql.replication.slave_running returns Critical state on master node


DanielMuller:
In a MySQL RDS master-slave setup, setting _replication:true_ on both master and slave config result in an unwanted _critical_ check state. Setting it on slave, returns the correct OK check status.
Setting _replication:true_ on master allows to populate _mysql.replication.slaves_connected_

I am expecting an OK status from master when master has a running slave. Or at least no check at all.

Example mysql.yaml:
```yaml
instances:
  - server: my_master.rds.amazonaws.com
    port: 3306
    user: datadog
    pass: XXXXXX
    tags:
      - dbinstanceidentifier:my_master
    options:
      replication: true
  - server: my_slave.rds.amazonaws.com
    port: 3306
    user: datadog
    pass: XXXXXX
    tags:
      - dbinstanceidentifier:my_slave
    options:
      replication: true
```

truthbk:
@DanielMuller thanks for reporting on this. This looks like a bug - we'll try to move quickly on this one. Apologies for the inconvenience.

truthbk:
@DanielMuller PR with fix has been merged. Should be out with the next bugfix release.

PR #2603, opened 2016-06-15
[mysql] fix replication service check.

truthbk:
## Why
The service check wasn't doing the right thing when running on the master. Should address issue #2596.

## What
Couple things:
- We now default to collecting the number of slaves by looking at `SELECT * FROM INFORMATION_SCHEMA.PROCESSLIST WHERE COMMAND LIKE '%Binlog dump%'` because looking at the worker threads could be inaccurate (it's blocking though, but we should be fine).
 
- `Slave_running` will also be available on the master node, so we have to also check the number of slaves is 0 (or None). That should guarantee being on a slave, and not falling through that logic branch on a master (and thus report CRITICAL, when we're fine).

For now we're keeping the replication check on the master in case a customer would like to run the `dd-agent` just on that node and still get replication insights.

### DIFF
--- a/checks.d/mysql.py
+++ b/checks.d/mysql.py
@@ -525,7 +525,7 @@ class MySql(AgentCheck):
         if _is_affirmative(options.get('replication', False)):
             # Get replica stats
             results.update(self._get_replica_stats(db))
-            results.update(self._get_slave_status(db, performance_schema_enabled))
+            results.update(self._get_slave_status(db))
             metrics.update(REPLICA_VARS)
 
             # get slave running form global status page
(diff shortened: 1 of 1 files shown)


checks.d/mysql.py
```diff
@@ -535,17 +535,17 @@ def _collect_metrics(self, host, db, tags, options, queries):
             # slaves will only be collected iff user has PROCESS privileges.
             slaves = self._collect_scalar('Slaves_connected', results)
 
-            if slave_running is not None:
+            if slave_running and not slaves:  # slave
```
degemer: It think `slaves` can only be a ` float`, wouldn't it be better to do a `slaves == 0.` ?

checks.d/mysql.py
```diff
@@ -535,17 +535,17 @@ def _collect_metrics(self, host, db, tags, options, queries):
...
                 if slave_running.lower().strip() == 'on':
                     slave_running_status = AgentCheck.OK
                 else:
                     slave_running_status = AgentCheck.CRITICAL
-            elif slaves or binlog_running:
-                if slaves and slaves > 0 and binlog_running:
+            elif slaves or binlog_running:  # master
```
degemer: Same here, `slaves > 0.` ?

degemer:
👍

checks.d/mysql.py
truthbk: It can also be a `None`, so I'll fix that, good catch.

checks.d/mysql.py
truthbk: I think it's fine if we do the comparison below. Also, it looks like float(0) resolves to False on `2.7.11` anyways.

checks.d/mysql.py
```diff
@@ -535,17 +535,17 @@ def _collect_metrics(self, host, db, tags, options, queries):
...
-            elif slaves or binlog_running:
-                if slaves and slaves > 0 and binlog_running:
+                    slave_running_status = AgentCheck.WARNING
+            elif slave_running:  # slave (or standalone)
+                if slave_running.lower().strip() == 'on':
                     slave_running_status = AgentCheck.OK
                 else:
-                    slave_running_status = AgentCheck.WARNING
-            else:
+                    slave_running_status = AgentCheck.CRITICAL
```
truthbk: if on standalone, they should obviously disable `replication` on the YAML.

degemer:
Way easier to read, thanks @truthbk!
:shipit:

truthbk:
Thank you @degemer !

PR #2604, opened 2016-06-15
[sqlserver] [skip ci] warn users about % char in yaml

stephenlechner:
The '%' character must be escaped as '%%' for adodbapi to make a
correct connection_string. Adding a comment to warn users in the
sqlserver.yaml.example file.

Only a comment, so [skip ci] is appropriate.

### DIFF
--- a/conf.d/sqlserver.yaml.example
+++ b/conf.d/sqlserver.yaml.example
@@ -34,6 +34,8 @@ init_config:
     #   tag_by: db
 
 instances:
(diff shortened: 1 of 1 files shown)


degemer:
Could you remove the `Currently` ? I don't think it's needed (especially since it might never be fixed).

degemer:
👍

stephenlechner:
Thanks--sorry I forgot to add the [skip ci] on that :/

PR #2605, opened 2016-06-15
[util] remove LaconicFilter

degemer:
This filter may be useful, but it:
- complexifies debug when you don't know it exists ("Why don't I see
  this line multiple times in the logs"),
- adds a lot of noise when trying to track memory utilization,
- is completely unexpected.


olivielpeau:
Wholly agree with your reasons to remove the filter!

Went through the checks that are currently using the filter to check that they wouldn't start flooding the logs, and IMO there are 2 places, in `checks/system/unix.py`, where we should change the way we log errors (because we currently log an exception for _every_ line of output that we can't parse):
- https://github.com/DataDog/dd-agent/blob/5f481fc74201f7e7c5a892148ecc9da7d651c962/checks/system/unix.py#L356
- https://github.com/DataDog/dd-agent/blob/5f481fc74201f7e7c5a892148ecc9da7d651c962/checks/system/unix.py#L441

Apart from this the PR looks good to me! 🚀

remh:
While we are at it, it could be useful to increase the max logging size:
https://github.com/DataDog/dd-agent/blob/master/config.py#L43

As sometimes when investigating issues, logs already got rotated.

degemer:
Updated as per your comments. I guess `dogstream` is one other thing that could flood the logs, but...

olivielpeau:
Yeah... The custom dogstream parsers that log a lot could start flooding the collector logs, but I still think it's better to remove the filter.

Should be up to the custom parsers' implementations to log things responsibly IMO...

checks/system/unix.py
```diff
@@ -347,13 +347,16 @@ def check(self, agentConfig):
...
             for line in lines:
                 try:
                     match = re.search(regexp, line)
                     if match is not None:
                         meminfo[match.group(1)] = match.group(2)
                 except Exception:
-                    self.logger.exception("Cannot parse %s", proc_meminfo)
+                    parse_error = True
+            if parse_error:
+                    self.logger.error("Error parsing %s", proc_meminfo)
```
olivielpeau: this line is over-indented :stuck_out_tongue:

olivielpeau:
Had one nitpick, feel free to merge once it's fixed! :fireworks:

degemer:
Thanks for the review, waiting for the 💚  CI before merging.

PR #2609, opened 2016-06-16
Use docker hostname instead of default route to query cadvisor & kubelet

hkaj:
Using the default route works fine in most cases, but if the user has a virtual network set up (like Calico) it won't point to the real host.
The docker host name is more reliable.

`_get_default_router` is not used anywhere else, so I killed it.

### DIFF
--- a/checks.d/kubernetes.py
+++ b/checks.d/kubernetes.py
@@ -46,6 +46,7 @@ FUNC_MAP = {
     RATE: {True: HISTORATE, False: RATE}
 }
 
+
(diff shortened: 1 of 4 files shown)


hkaj:
@remh In case users are using the agent outside a container we could still fallback to `_get_default_router` actually. Adding it back

truthbk:
Looks good to me. Please squash before merging! Also, let's let CI go 💚 before merging. 👍

truthbk:
@hkaj 👍 for the test fixes! Nice.

PR #2610, opened 2016-06-20
[mysql] Fix slave_running for mysql 5.7 with slave binlog

truthbk:
When a MySQL 5.7 is a slave _and_ has a running binlog with no
downstream slaves attached, this check assumed that the instance was a
master with no running slaves.

Fix it so that it checks for Slave_IO_Running and Slave_SQL_Running
before checking for Binlog_enabled and Slaves_connected.

This PR should supersede #2493 - it is essentially no more than a rebase of it. We first look at the slave variables available in most 5.7+ versions, and if we don't yet have concluded on a status, we then look at possible master, slave combinations.

### DIFF
--- a/checks.d/mysql.py
+++ b/checks.d/mysql.py
@@ -535,35 +535,37 @@ class MySql(AgentCheck):
             # slaves will only be collected iff user has PROCESS privileges.
             slaves = self._collect_scalar('Slaves_connected', results)
 
-            if self._is_master(slaves, binlog_running):  # master
-                if slaves > 0 and binlog_running:
-                    slave_running_status = AgentCheck.OK
-                else:
-                    slave_running_status = AgentCheck.WARNING
-            elif slave_running:  # slave (or standalone)
-                if slave_running.lower().strip() == 'on':
-                    slave_running_status = AgentCheck.OK
-                else:
-                    slave_running_status = AgentCheck.CRITICAL
-            else:  # slave - with no `Slave_running`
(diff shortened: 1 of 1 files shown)


truthbk:
@nrw505 this would be the rebased PR. I believe it respect your changes to the logic and authorship.

Thanks for your contribution.

checks.d/mysql.py
```diff
@@ -535,35 +535,37 @@ def _collect_metrics(self, host, db, tags, options, queries):
...
+                        slave_running_status = AgentCheck.OK
+                    elif not slave_io_running and not slave_sql_running:
+                        slave_running_status = AgentCheck.CRITICAL
+                    else:
+                        # not everything is running smoothly
+                        slave_running_status = AgentCheck.WARNING
+
+            # if we don't yet have a status - inspect
+            if slave_running_status == AgentCheck.UNKNOWN:
+                if self._is_master(slaves, binlog_running):  # master
```
olivielpeau: On MySQL < 5.7, couldn't the same issue arise here? (i.e. if we have a slave with binlog running, we'd regard it as a master because of this condition and send a `WARNING` status)

olivielpeau:
@truthbk Added in one comment, let me know what you think :)

checks.d/mysql.py
olivielpeau: An optional setting in the conf file sounds good to me! (for instance an option to force the check to regard the server as a master even when binlog is running)

olivielpeau:
LGTM! 

(my comment about mysql < 5.7 can be addressed later in a separate PR, feel free to merge this PR)

PR #2614, opened 2016-06-21
service discovery: add hport variables

awasilyev:
this will add hport variable, containing container host port for service discovery templating.


degemer:
Thanks for the PR @awasilyev! Someone will review it in the coming days.

hkaj:
@awasilyev I'm not sure what's the use case for this, and it's a little confusing to have 2 similar variables. Why connecting to the host port instead of the container directly?

PR #2616, opened 2016-06-21
[ssh][testing] updated domain name to netgarage.org

masci:
Fixes integration test that was failing b/c `smashthestack.org` was deprecated.


degemer:
👍

masci:
ES is still failing, have another path on top of this to fix

PR #2617, opened 2016-06-21
[ES][testing] fix es integration tests

masci:
The previous task only attempted to re-download the package if the ES dir did not exist but if the download failed the cache remained dirty and the build stuck.
Now it checks for the ES bin, if not found cleans up the ES dir and download the package.


degemer:
🎉

Issue #2558, opened 2016-05-31
Document datadog.conf defaults better


tuukkamustonen:
Looking at reference `datadog.conf`, it's not clear what the defaults are.

There are some inconsistencies. For example:

    # Collect AWS EC2 custom tags as agent tags (requires an IAM role associated with the instance)
    # collect_ec2_tags: no

I assume default here is ‘no’. Fair enough.

But then there is:

    # use unique hostname for GCE hosts, see http://dtdg.co/1eAynZk
    gce_updated_hostname: yes

If I comment that, does it default to ‘no’? Why is it not commented like the rest of the settings?

    # If you need a proxy to connect to the Internet, provide the settings here
    # proxy_host: my-proxy.com
    # proxy_port: 3128
    # proxy_user: user
    # proxy_password: password

Yeah it’s understandable that proxy is by default disabled, but if interpret this with the same logic as before, one would understand that proxy defaults to `user@pass:my-proxy.com:3128`.

Also:

    # Force the hostname to whatever you want.
    #hostname: mymachine.mydomain

I assume hostname is detected by default, `mymachine.mydomain` is not the default…

    # collector_log_file: /var/log/datadog/collector.log
    # forwarder_log_file: /var/log/datadog/forwarder.log
    # dogstatsd_log_file: /var/log/datadog/dogstatsd.log

In reality, logging is enabled only if these are uncommented, but reader can’t know that it’s the logic (I first assumed they were defaults and that logging would be enabled by default).

It would be clearer if you documented these settings like (as an example):

    # Collect AWS EC2 custom tags as agent tags (requires an IAM role associated with the instance) (default: no)
    # collect_ec2_tags: no

    # use unique hostname for GCE hosts, see http://dtdg.co/1eAynZk (http://dtdg.co/1eAynZk) (default: no)
    # gce_updated_hostname: no

    # If you need a proxy to connect to the Internet, provide the settings here (default: disabled)
    # proxy_host: my-proxy.com
    # proxy_port: 3128
    # proxy_user: user
    # proxy_password: password

    # Force the hostname to whatever you want (default: auto-detected)
    #hostname: mymachine.mydomain

Or some other way to indicate default value clearly.

I actually sent this to Datadog support first (before realising I can file a ticket here) and their response was:

> -- Everything you've said there is correct and I understand the confusion. There definitely seem to be discrepancies between what is commented out and what is considered default for different parameters. In some cases the values are examples and in others they are default. --

[...]

PR #2618, opened 2016-06-21
[config] clean defaults

degemer:
Defaults are now commented.

### DIFF
--- a/datadog.conf.example
+++ b/datadog.conf.example
@@ -3,7 +3,7 @@
 # The host of the Datadog intake server to send Agent data to
 dd_url: https://app.datadoghq.com
 
-# If you need a proxy to connect to the Internet, provide the settings here
+# If you need a proxy to connect to the Internet, provide the settings here (default: disabled)
 # proxy_host: my-proxy.com
 # proxy_port: 3128
 # proxy_user: user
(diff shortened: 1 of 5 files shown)


hkaj:
Behold the sneakiness of those spaces you added: https://github.com/DataDog/docker-dd-agent/pull/100

degemer:
Fix https://github.com/DataDog/dd-agent/issues/2558

datadog.conf.example
```diff
@@ -36,8 +36,7 @@ api_key:
 # create_dd_check_tags: no
 
 # Collect AWS EC2 custom tags as agent tags (requires an IAM role associated with the instance)
-# collect_ec2_tags: no
-
+# collect_ec2_tags: yes
```
remh: Default is no here

datadog.conf.example
```diff
@@ -48,24 +47,23 @@ api_key:
 # collector_profile_interval: 20
 
 # use unique hostname for GCE hosts, see http://dtdg.co/1eAynZk
-gce_updated_hostname: yes
+# gce_updated_hostname: yes
```
remh: Unfortunately you can't do that. As we want to keep the old defaults for existing installations (for compatibility purposes) but we want new installations to have this flag enabled

degemer:
Thanks for the review, updated as per your comments @remh.

gmmeyer:
This looks good to me, it's a lot better than before! 👍

PR #2619, opened 2016-06-23
[core][dogstatsd] Added support for IP agnostic connections

masci:
### Why
Dogstatsd only listens to IPv4 connections, thus IPv6 clients aren't able to connect.

### What
Now dogstatsd listens to an IP agnostic socket connection. IPv4 addresses are mapped to IPv6 to keep configuration file compatible with older versions.

While at the job: fixed tests, added tests and fixed PEP8 where possible.


dogstatsd.py
```diff
@@ -131,6 +133,43 @@ def serialize_event(event):
...
+
+    return None
+
+
+def normalize_host(host):
+    """
+    Map the host to an IPv6 address, trying to resolve the name
+    """
+    try:
+        addr = socket.gethostbyname(host)
```
truthbk: Not that it matters much because you'll be converting the ipv4 to ipv6 anyways - but maybe we should favor the ipv6 record if it exists. I'm saying this because the DNS might have the same name resolve to an `A` (ipv4) and an `AAAA` (ipv6) record.... So `getaddrinfo` might be a better option for dual-stack resolution - otherwise if the hostname only has an ipv6 record this might fail.

dogstatsd.py
```diff
@@ -318,18 +357,16 @@ def __init__(self, metrics_aggregator, host, port, forward_to_host=None, forward
...
+        self.socket.setsockopt(IPPROTO_IPV6, IPV6_V6ONLY, 0)
         self.socket.setblocking(0)
-        try:
-            self.socket.bind(self.address)
-        except socket.gaierror:
-            if self.address[0] == 'localhost':
-                log.warning("Warning localhost seems undefined in your host file, using 127.0.0.1 instead")
-                self.address = ('127.0.0.1', self.address[1])
-                self.socket.bind(self.address)
+        self.socket.bind(self.address)
```
truthbk: question: we sure we don't need the `try...except` block?

dogstatsd.py
```diff
@@ -131,6 +133,43 @@ def serialize_event(event):
...
+
+def normalize_host(host):
+    """
+    Map the host to an IPv6 address, trying to resolve the name
+    """
+    try:
+        addr = socket.gethostbyname(host)
+    except socket.gaierror:
+        log.debug('Not an hostname or unable to resolve %s, defaulting to ::1', host)
+        addr = '::1'
```
truthbk: not convinced we should default to loopback if we can't resolve.... 😕

truthbk:
In general it looks good, just a few open questions, and the DNS issue.

Thanks a lot for splitting the tests, much, much cleaner and clear! 😍

dogstatsd.py
masci: for `gaierror` we are already covered b/c `self.address` handles that. We can add a more generic try/catch but what should we do in the exception handler?

truthbk: Yeah, maybe no need to do anything - doesn't seem like we'd be able to do much other than just print a meaningful message (which is probably already fine in the exception anyway).

tests/core/test_dogstatsd.py
```diff
@@ -1,892 +1,113 @@
...
+    @mock.patch('dogstatsd.select')
+    def test_start(self, select):
+        select.select.side_effect = [KeyboardInterrupt, SystemExit]
+        s1 = Server(mock.MagicMock(), '::1', '1234')
+        s1.start()
+        self.assertEqual(s1.socket.family, socket.AF_INET6)
+
+        s2 = Server(mock.MagicMock(), '127.0.0.1', '2345')
+        s2.start()
+        self.assertEqual(s2.socket.family, socket.AF_INET6)
```
truthbk: @masci not sure if this is because of the MagicMock()... shouldn't this be `socket.AF_INET`?

truthbk:
This looks good! Good job 👍

Issue #2612, opened 2016-06-20
Marathon URL can't contain a path


ensonik:
Inside Mesosphere the marathon url contains a path. This unfortunately doesn't work with the urljoin() found in the marathon check. 

For example, my url is: http://marathon.mesos/service/marathon

Running this through url join (as it is in the check), I get the following:
```
>>> urljoin("http://marathon.mesos/service/marathon", "/v2/apps")
'http://marathon.mesos/v2/apps'
>>> urljoin("http://marathon.mesos/service/marathon/", "/v2/apps")
'http://marathon.mesos/v2/apps'
```

Obviously, the check fails with a 404.

degemer:
You're right, I think there is an easy way to fix this, according to http://stackoverflow.com/a/1794540.
This works:
```
>>> urljoin("http://marathon.mesos/service/marathon/", "v2/apps")
'http://marathon.mesos/service/marathon/v2/apps'
```
Could you try this locally to confirm (replace `'/v2/apps` by `v2/apps` there https://github.com/DataDog/dd-agent/blob/master/checks.d/marathon.py#L50) ? You can edit the check in `/opt/datadog-agent/agent/checks.d/marathon.py` and then restart the agent.

ensonik:
It works. I updated the marathon.py script with `v2/apps` instead of `/v2/apps` and it works in all cases. 

The repl confirms. 
```
>>> urljoin("http://marathon.mesos/service/marathon", "/v2/apps")
'http://marathon.mesos/v2/apps'
>>> urljoin("http://marathon.mesos/service/marathon/", "v2/apps")
'http://marathon.mesos/service/marathon/v2/apps'
```

Do you want a PR?

degemer:
Thanks for the details!
We'll be happy to review your PR.

PR #2620, opened 2016-06-23
Fix a small problem that prevented marathon full path from being properly built 

ensonik:
…properly built when the base url included a path.

As tested in the REPL, the fix will properly join base url's and paths:
```
  >>> urljoin("http://marathon.mesos", "v2/apps")
  'http://marathon.mesos/v2/apps'

  >>> urljoin("http://marathon.mesos/service/marathon/", "v2/apps")
  'http://marathon.mesos/service/marathon/v2/apps'

  >>> urljoin("http://marathon.mesos", "v2/apps")
  'http://marathon.mesos/v2/apps'

  >>> urljoin("http://marathon.mesos/", "v2/apps")
  'http://marathon.mesos/v2/apps'

  >>> urljoin("http://marathon.mesos:8080", "v2/apps")
  'http://marathon.mesos:8080/v2/apps'

  >>> urljoin("http://marathon.mesos:8080/", "v2/apps")
  'http://marathon.mesos:8080/v2/apps'
```
Care needs to be taken to include a trailing slash in the base URL when there is a path included.
```
  >>> urljoin("http://marathon.mesos/service/marathon", "v2/apps")
  'http://marathon.mesos/service/v2/apps' # This is wrong. Include trailing slash to fix.
```

### DIFF
--- a/checks.d/marathon.py
+++ b/checks.d/marathon.py
@@ -47,7 +47,7 @@ class Marathon(AgentCheck):
         default_timeout = self.init_config.get('default_timeout', self.DEFAULT_TIMEOUT)
         timeout = float(instance.get('timeout', default_timeout))
 
-        response = self.get_json(urljoin(url, "/v2/apps"), timeout, auth)
+        response = self.get_json(urljoin(url, "v2/apps"), timeout, auth)
         if response is not None:
             self.gauge('marathon.apps', len(response['apps']), tags=instance_tags)
             for app in response['apps']:


degemer:
Thanks for the PR @ensonik!
Could you add a special case when there is no trailing slash in the URL ? (and add a trailing slash in the code)
Something like:
```
if url[-1] != '/':
    url += '/'
```

degemer:
Fix #2612

degemer:
Sorry, my comment is probably not needed, merging.
Thanks again for your contribution @ensonik !

PR #2623, opened 2016-06-24
[redis] use time delta to compute master_link_down_since_seconds

gzysk8:
The value for `master_link_down_since_seconds` returned from the Redis INFO command is the timestamp in seconds at which the master went down. We should take the difference between this and the current time before submitting the metric to Datadog.


gzysk8:
Actually going to keep this as is. The output normally shows the seconds the link has been down, but if the slave was never able to connect to the master, it shows a timestamp.

PR #2625, opened 2016-06-24
[mongo] Fix case where optimeDate is not available

### LINKED ISSUE #2581: MongoDB agent fails on arbier
The mongodb agent fails to complete monitoring when monitored instance is an arbiter:
This is due to the fact that arbiter does not have an optimeDate field (https://github.com/DataDog/dd-agent/blob/master/checks.d/mongo.py:703):

>                 if current is not None and primary is not None:
>                     lag = primary['optimeDate'] - current['optimeDate']
>                     data['replicationLag'] = total_seconds(lag)

Maybe need to cahgne to 
>                 if current is not None and primary is not None and optimeData in current:
>                     lag = primary['optimeDate'] - current['optimeDate']
>                     data['replicationLag'] = total_seconds(lag)
>                 else:
>                     data['replicationLag'] = 0

remh:
Fix #2581

### DIFF
--- a/checks.d/mongo.py
+++ b/checks.d/mongo.py
@@ -701,8 +701,9 @@ class MongoDb(AgentCheck):
 
                 # Compute a lag time
                 if current is not None and primary is not None:
-                    lag = primary['optimeDate'] - current['optimeDate']
-                    data['replicationLag'] = total_seconds(lag)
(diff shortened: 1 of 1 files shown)


degemer:
LGTM, :shipit:

PR #2626, opened 2016-06-24
[docker_daemon] fix disk metrics rounding issue

degemer:
Docker Status rounds the disk metrics, which means something like this
can happen:
```
[u'Data Space Used', u'5.238 GB']
[u'Data Space Available', u'18.12 GB']
[u'Data Space Total', u'23.35 GB']
```
So:
```
used = 5238000000
free = 18120000000
total = 23350000000
```

With `used + free > total` because of the rounding error.
In this case, this commits changes the default behavior to taking
`total` as `used + free` and calculating the disk percentage based on
that. (There should only be a small difference between `total` and `free + used`)


hkaj:
fair enough solution. LGTM

PR #2627, opened 2016-06-27
[openstack] nova/neutron extend name search

degemer:
Sometimes the check doesn't find the Neutron/Nova endpoint because the name is
not what is expected. This commit extends the current search by adding
one possible name for each component: 'Compute' for Nova and
'Networking' for Neuron.


truthbk:
Looks consistent with the customer provided json. I don't like that we have to do this, but unfortunately the API is not very consistent across versions so I understand the need for this. 👍

degemer:
Thanks @truthbk !

PR #2628, opened 2016-06-27
[core] clean debug logs and kill Dogstream-based DdForwarder

degemer:
## 	[collector] remove useless debug logs
This commit:
- removes the creation of a Ganglia object when not enabled
- "       "   "        "  " Dogstreams "   "    "   "
- removes the `Sample: ...` debug logs for WMI

### DIFF
--- a/checks/libs/wmi/sampler.py
+++ b/checks/libs/wmi/sampler.py
@@ -201,7 +201,6 @@ class WMISampler(object):
 
         try:
             if self.is_raw_perf_class and not self._previous_sample:
-                self.logger.debug(u"Querying for initial sample for raw performance counter.")
                 self._current_sample = self._query()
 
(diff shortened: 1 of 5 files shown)


checks/libs/wmi/sampler.py
```diff
@@ -215,7 +214,6 @@ def sample(self):
             raise
         else:
             self._sampling = False
-            self.logger.debug(u"Sample: {0}".format(self._current_sample))
 
     def __len__(self):
         """
```
yannmh: 👍

gmmeyer:
This looks pretty good, I don't see any issue with it. The build failed on an unrelated check. So, rerun it if you want but otherwise: LGTM.

degemer:
Thanks @gmmeyer, I re-run it, it passed! Merging.

PR #2631, opened 2016-06-28
[http_check] Bring back include_content option

hkaj:
This option was removed by mistake a while back, let's add it back.

### DIFF
--- a/checks.d/http_check.py
+++ b/checks.d/http_check.py
@@ -34,6 +34,8 @@ from util import headers as agent_headers
 from utils.proxy import get_proxy
 
 DEFAULT_EXPECTED_CODE = "(1|2|3)\d\d"
+CONTENT_LENGTH = 200
+
 
 class WeakCiphersHTTPSConnection(urllib3.connection.VerifiedHTTPSConnection):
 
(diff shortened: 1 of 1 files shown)


checks.d/http_check.py
```diff
@@ -288,8 +290,10 @@ def _check(self, instance):
             else:
                 expected_code = http_response_status_code
 
-            message = "Incorrect HTTP return code for url %s. Expected %s, got %s" % (
+            message = "Incorrect HTTP return code for url %s. Expected %s, got %s." % (
                 addr, expected_code, str(r.status_code))
+            if include_content:
+                message += '\nContent: {}'.format(r.content[:CONTENT_LENGTH])
 
             self.log.info(message)
```
yannmh: Shall we add it here too: https://github.com/DataDog/dd-agent/blob/master/checks.d/http_check.py#L315-L319 ?

yannmh:
Looks like I introduced the regression a while ago, with Datadog Agent 5.3.

Thanks for fixing it @hkaj !

checks.d/http_check.py
hkaj: yeah that would make sense, good call @yannmh. Adding it.

yannmh:
:shipit:

PR #2633, opened 2016-06-28
[iis] no warnings on 'Name' property 🔇

yannmh:
Do not log a warning about non digit property values for the 'Name'
property.
Test that the check completes with no warnings.

### DIFF
--- a/checks.d/iis.py
+++ b/checks.d/iis.py
@@ -102,8 +102,10 @@ class IIS(WinWMICheck):
             )
         except pythoncom.com_error as e:
             if '0x80041017' in str(e):
-                self.warning("You may be running IIS6/7 which reports metrics a \
-                             little differently. Try enabling the is_2008 flag for this instance.")
+                self.warning(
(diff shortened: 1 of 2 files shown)


degemer:
👍

checks.d/iis.py
```diff
@@ -137,7 +139,10 @@ def _extract_metrics(self, wmi_sampler, sites, tags):
 
             # Tag with `tag_queries` parameter
             for wmi_property, wmi_value in wmi_obj.iteritems():
-                # Tag with `tag_by` parameter
+                # No metric extraction on 'Name' property
+                if wmi_property == 'name':
```
degemer: Could there be a case issue (`name` vs `Name`) ?

yannmh: `wmi_obj` is a `CaseInsensitiveDict` 💥 , c.f. https://github.com/DataDog/dd-agent/blob/master/checks/libs/wmi/sampler.py#L36

degemer:
Thanks for the answer, :shipit: !

PR #2634, opened 2016-06-28
[status page] Remove proxy info from it

remh:
Reading the proxy settings is a bit costly as it reparse the whole
config file and we do it every time we persist the forwarder status.

This information is not that valuable as it will be collected by a
flare anyway. Let’s just remove it


gmmeyer:
LGTM 👍

PR #2636, opened 2016-06-29
ServiceDiscovery: Gather tags for non-k8s

awasilyev:
This will add gathering tags from labels during service discovery for non-k8s containers.
Tags are gathered from container labels, named as com.datadoghq.sd.check.tags.*
Also one additional tag fetched - container name.


olivielpeau:
Thanks @awasilyev for your contribution!

We have to make sure that your addition doesn't conflict/overlap a feature that we're currently developing, but we'll make sure to review this PR and keep you updated.

remh:
@hkaj can you have a look please ?

hkaj:
Thanks for the PR @awasilyev !
Container name change too often, it's not something we want to tag with by default. And now templates can be stored in docker labels or kubernetes annotations so there's no use for retrieving tags from a separate label.
Closing the PR, let me know if you have any question.

PR #2640, opened 2016-06-29
kernel 4+ won't error out on status now

gmmeyer:
On Kernel 3.18+ on some distributions that also use overlayfs, certain commands will fail. However, they'll work fine if they're not using overlayfs.

This is a stopgap solution, so that the commands will work on our docker testing environments.


degemer:
`if [ "$DOCKER_DD_AGENT" != "" ] && ( ( [ $major -eq 3 ] && [ $minor -ge 18 ] ) || [ $major -gt 3 ] ); then`

degemer:
👍

PR #2645, opened 2016-06-30
Add TcpExt metrics for Linux

agy:
- Includes a minor refactor to reuse the /proc/net/snmp parsing code

### DIFF
--- a/checks.d/network.py
+++ b/checks.d/network.py
@@ -215,8 +215,22 @@ class Network(AgentCheck):
                 self._submit_devicemetrics(iface, metrics)
 
         try:
-            proc_snmp_path = "{}/net/snmp".format(proc_location)
-            proc = open(proc_snmp_path, 'r')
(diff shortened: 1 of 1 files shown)


irabinovitch:
@truthbk @agy Should this be moved to Datadog/integrations-core?

olivielpeau:
This PR should be ported to [integrations-core](https://github.com/DataDog/integrations-core).

One concern that we may have with the present PR is that it collects a lot of new metrics by default. Some of the new metrics it collects were added in https://github.com/DataDog/integrations-core/pull/299. @agy could you have a look at https://github.com/DataDog/integrations-core/pull/299 and let us know if you think that PR omits important metrics that you'd want the check to collect? If so, could you port your changes over to [integrations-core](https://github.com/DataDog/integrations-core)?

Going to close the PR since it's no use keeping it open, but thanks for opening this :)

PR #2647, opened 2016-06-30
Implement consul self_leader_check

ross:
A mechanism for detecting leadership changes that can be enabled on all server nodes in a cluster. It will emit a single event from the new leader. This avoids the need to designate a single agent as the watcher for leadership changes, all agents can watch themselves.

### DIFF
--- a/checks.d/consul.py
+++ b/checks.d/consul.py
@@ -41,6 +41,7 @@ class ConsulCheck(AgentCheck):
         self._local_config = None
         self._last_config_fetch_time = None
         self._last_known_leader = None
+        self._we_are_leader = False
 
     def consul_request(self, instance, endpoint):
         url = urljoin(instance.get('url'), endpoint)
@@ -94,7 +95,7 @@ class ConsulCheck(AgentCheck):
     def _is_instance_leader(self, instance):
         try:
             agent_url = self._get_agent_url(instance)
-            leader = self._last_known_leader or self._get_cluster_leader(instance)
+            leader = self._get_cluster_leader(instance)
             self.log.debug("Consul agent lives at %s . Consul Leader lives at %s" % (agent_url,leader))
             return agent_url == leader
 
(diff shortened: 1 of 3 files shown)


checks.d/consul.py
```diff
@@ -94,15 +95,14 @@ def _get_agent_datacenter(self, instance):
     def _is_instance_leader(self, instance):
         try:
             agent_url = self._get_agent_url(instance)
-            leader = self._last_known_leader or self._get_cluster_leader(instance)
```
ross: I may look at seeing if it can be preserved, but the one change here is that _last_known_leader isn't managed when in `self_leader_check` mode. Thus we need to look to see who is the leader each time this is called.

checks.d/consul.py
```diff
@@ -122,23 +122,29 @@ def _check_for_leader_change(self, instance):
...
+            self._emit_leader_change(instance, leader)
 
         self._last_known_leader = leader
 
+    def _emit_leader_change(self, instance, leader):
+        agent_dc = self._get_agent_datacenter(instance)
+
+        tags = ["curr_consul_leader:{0}".format(leader), "consul_datacenter:{0}".format(agent_dc)]
+        if self._last_known_leader:
+            tags.append("prev_consul_leader:{0}".format(self._last_known_leader))
```
ross: As [mentioned above](https://github.com/DataDog/dd-agent/pull/2647/files#r69222053) `self_leader_check` doesn't keep track of `_last_known_leader` and thus can't emit it as part of the event.

ross:
Closing and re-opening to kick off another build. Not sure the failure is in my changes.

olivielpeau:
Thanks @ross for your contribution, we'll make sure to review it soon.

ross:
Testing this out in detail now. Realized that as it currently stands this fires a leadership change even when the datadog agent is restarted on the leader node. Should be easy enough to address. Will update once that's been solved.

ross:
Updated the PR with an improved version that involves less diff from the original and preserves the full event in both modes. This has been tested out with an actual cluster and so far so good.

PR #2649, opened 2016-07-02
[haproxy] Allow per-instance custom tags

theojulienne:
In some more complex haproxy installations like at GitHub, it's desirable to be able to group haproxy metrics by tags such that multiple instance URLs may represent the same logical service, but without all instances configured on a single host being part of the same service. Currently the only real grouping supported is host-level and based on `instance_url`, or using the frontend/backend names used in the configuration (which may overlap).

This PR adds an optional `tags` option to haproxy instance configurations, allowing custom grouping like the following:
```
instances:
  - url: http://localhost:1101/admin?stats
    tags:
      - "my-grouping:foo"
  - url: http://localhost:1102/admin?stats
    tags:
      - "my-grouping:foo"
  - url: http://localhost:2101/admin?stats
    tags:
      - "my-grouping:bar"
  - url: http://localhost:2102/admin?stats
    tags:
      - "my-grouping:bar"
```


gmmeyer:
Thanks for your PR @theojulienne! 

Could you also add this to the [example .yaml](https://github.com/theojulienne/dd-agent/blob/master/conf.d/haproxy.yaml.example) file? 

I think this is otherwise pretty reasonable!

theojulienne:
Thanks @gmmeyer - I've added the example as requested!

gmmeyer:
Great! I added this to the 5.10 milestone, we'll merge it for that and release it then.

Thank you very much! 😄

gmmeyer:
@theojulienne could you add a test for us as well? We want to be certain that this works properly, given the widespread use of haproxy and the number of tags that this could add if this functions improperly.

gmmeyer:
Thank you very much for your contribution! I'm going to merge this :smile: and it'll be released in the next version.

PR #2653, opened 2016-07-05
[dns_check] Allow querying a custom record type

olivielpeau:
New `record_type` optional parameter in the configuration.

Along with tests: runs of the check in `test_invalid_config` now reload
the config correctly and test that the expected exception is raised.

### DIFF
--- a/checks.d/dns_check.py
+++ b/checks.d/dns_check.py
@@ -51,12 +51,14 @@ class DNSCheck(AgentCheck):
         if nameserver is not None:
             resolver.nameservers = [nameserver]
 
+        record_type = instance.get('record_type', 'A')
+
         status = AgentCheck.CRITICAL
         start_time = time.time()
         try:
-            self.log.debug('Resolving hostname %s...' % hostname)
-            answer = resolver.query(hostname)
-            assert(answer.rrset.items[0].address)
+            self.log.debug('Querying "%s" record for hostname "%s"...' % (record_type, hostname))
+            answer = resolver.query(hostname, rdtype=record_type)
+            assert(answer.rrset.items[0].to_text())
             end_time = time.time()
         except dns.exception.Timeout:
             self.log.error('DNS resolution of %s timed out' % hostname)
--- a/conf.d/dns_check.yaml.example
+++ b/conf.d/dns_check.yaml.example
@@ -5,3 +5,7 @@ instances:
     - hostname: www.example.org
       nameserver: 127.0.0.1
       timeout: 8
+
+      # Specify an (optional) `record_type` to customize the record type
+      # queried by the check (default: "A")
+      # record_type: A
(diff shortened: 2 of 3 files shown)


conf.d/dns_check.yaml.example
```diff
@@ -5,3 +5,7 @@ instances:
     - hostname: www.example.org
       nameserver: 127.0.0.1
       timeout: 8
+
+      # Specify an (optional) `record_type` to customize the record type
+      # queried by the check
```
hkaj: maybe specify that the default is `A record` to avoid any confusuion?

hkaj:
one small comment and then :shipit:

olivielpeau:
Thanks @hkaj! Addressed your comment, I'll merge once the CI passes

checks.d/dns_check.py
```diff
@@ -51,12 +51,14 @@ def query_dns(self, instance, timeout):
         if nameserver is not None:
             resolver.nameservers = [nameserver]
 
+        record_type = instance.get('record_type', 'A')
+
         status = AgentCheck.CRITICAL
         start_time = time.time()
         try:
             self.log.debug('Resolving hostname %s...' % hostname)
```
masci: should we debug also `record_type`?

checks.d/dns_check.py
```diff
@@ -51,12 +51,14 @@ def query_dns(self, instance, timeout):
...
-            answer = resolver.query(hostname)
-            assert(answer.rrset.items[0].address)
+            answer = resolver.query(hostname, rdtype=record_type)
+            assert(answer.rrset.items[0].to_text())
             end_time = time.time()
         except dns.exception.Timeout:
```
masci: we should also catch `NoAnswer` in case the query fails for a record type that doesn't exist for the address

tests/checks/mock/test_dns_check.py
```diff
@@ -52,6 +55,18 @@ def test_success(self, mocked_query):
                                 tags=['resolved_hostname:www.example.org', 'nameserver:127.0.0.1'])
         self.coverage_report()
 
+    @mock.patch.object(Resolver, 'query', side_effect=success_query_mock)
+    def test_success_CNAME(self, mocked_query):
+        config = {
+            'instances': [{'hostname': 'www.example.org', 'nameserver': '127.0.0.1', 'record_type': 'CNAME'}]
```
masci: can we also test a wrong/invalid `record_type`?

olivielpeau: we do it in the `test_invalid_config` test (https://github.com/DataDog/dd-agent/pull/2653/files#diff-3254c25e4da8782409c2288e998ebc38R96)

masci: My bad, looking at the wrong part of the module

masci:
👍

olivielpeau:
Thanks! Merging

PR #2655, opened 2016-07-05
query leader.mesos by default for mesos master check

hkaj:
Only the leader report metrics so it makes sense, also this way we can use the default file in docker since localhost doesn't resolve to the mesos metrics api in a container if `--net=host` is not used.

### DIFF
--- a/conf.d/mesos_master.yaml.example
+++ b/conf.d/mesos_master.yaml.example
@@ -2,4 +2,4 @@ init_config:
   default_timeout: 10
 
 instances:
-  - url: "http://localhost:5050"
+  - url: "http://leader.mesos:5050"


olivielpeau:
Hmm, looks like the metrics from `/metrics/snapshot` are available from any master node (c.f. https://github.com/DataDog/dd-agent/blob/5.8.3/checks.d/mesos_master.py#L223-L233) and not only the leader, so I'm not sure changing the default to the leader makes sense in the general case.

hkaj:
heh I trusted https://github.com/DataDog/dd-agent/blob/5.8.3/checks.d/mesos_master.py#L7 but should have read more. Good catch @olivielpeau I'll do it another way.

PR #2657, opened 2016-07-05
Added an option to disable following redirects.

tschellhorn:
Some http_checks look to make sure that connecting to one link does actually result in a temporary or permanent redirect. This change allows disabling the python requests module from following redirects.


degemer:
Thanks for the contribution @tschellhorn !
Could you add this option in the configuration file: https://github.com/DataDog/dd-agent/blob/master/conf.d/http_check.yaml.example and add a short test: https://github.com/DataDog/dd-agent/blob/master/tests/checks/integration/test_http_check.py ?

Let us know if you need help to do this !

miketheman:
Appears to be superseded by #2586.
Recommend closure.

tschellhorn:
Closing out as duplicated PR pushed after this PR created.

PR #2658, opened 2016-07-06
[mongo] repairing of mongodb.can_connect check

cryptspirit:
Connection to database going in db.command function, but the call of this function happens outside of exception block. The check can't catch unreachable instances and integration check mongodb.can_connect doesn't work properly

### DIFF
--- a/checks.d/mongo.py
+++ b/checks.d/mongo.py
@@ -639,12 +639,20 @@ class MongoDb(AgentCheck):
                 message=message)
             raise Exception(message)
 
-        self.service_check(
-            self.SERVICE_CHECK_NAME,
-            AgentCheck.OK,
-            tags=service_check_tags)
+        try:
+            status = db.command('serverStatus', tcmalloc=collect_tcmalloc_metrics)
+        except Exception:
+            self.service_check(
+                self.SERVICE_CHECK_NAME,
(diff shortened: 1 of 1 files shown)


degemer:
Thanks for the PR @cryptspirit !
I think there may be an issue if the connection requires authentification though: `status = db.command('serverStatus', tcmalloc=collect_tcmalloc_metrics)` happens before `db.authenticate(username, password)`.

What do you think of something like this instead ?
```python
try:
     status = db.command('serverStatus', tcmalloc=collect_tcmalloc_metrics)
except Exception:
    self.service_check(
        self.SERVICE_CHECK_NAME,
        AgentCheck.CRITICAL,
        tags=service_check_tags)
    raise
else:
    self.service_check(
        self.SERVICE_CHECK_NAME,
        AgentCheck.OK,
        tags=service_check_tags)
```

cryptspirit:
@degemer right. there is need to add separate exception block for serverStatus execution

degemer:
@cryptspirit would you like to update your PR with this change then ? (if you don't have the time, we can open another one)

cryptspirit:
I will

degemer:
Thank you @cryptspirit ! (and sorry for the delay)
Merging.

PR #2660, opened 2016-07-06
[Event log] Unicode everything

remh:
Otherwise the check can fail with
```
  File "checks\__init__.pyc", line 750, in run
  File "C:\Program Files (x86)\Datadog\Datadog Agent\files\..\checks.d\win32_event_log.py", line 147, in check
  File "C:\Program Files (x86)\Datadog\Datadog Agent\files\..\checks.d\win32_event_log.py", line 240, in to_event_dict
  File "C:\Program Files (x86)\Datadog\Datadog Agent\files\..\checks.d\win32_event_log.py", line 210, in _msg_text
UnicodeEncodeError: 'ascii' codec can't encode character u'\u200e' in position 31: ordinal not in range(128)
```

### DIFF
--- a/checks.d/win32_event_log.py
+++ b/checks.d/win32_event_log.py
@@ -203,17 +203,17 @@ class LogEvent(object):
 
             msg_text_fields.append("```\n%%%")
 
-            msg_text = "\n".join(msg_text_fields)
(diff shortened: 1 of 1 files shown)


yannmh:
🙇

PR #2661, opened 2016-07-06
[core] don't use docker hostname if it's a EC2 one

degemer:
Docker info sometimes returns default EC2 hostnames and `get_hostname`
wasn't checking whether it was possible to use the instance id.
This commits change this behavior by allowing `get_hostname` to check
the docker proposed hostname before using it.

### DIFF
--- a/util.py
+++ b/util.py
@@ -204,7 +204,7 @@ def get_hostname(config=None):
     if hostname is None and docker_util.is_dockerized():
         docker_hostname = docker_util.get_hostname()
         if docker_hostname is not None and is_valid_hostname(docker_hostname):
-            return docker_hostname
+            hostname = docker_hostname
 
     # then move on to os-specific detection
     if hostname is None:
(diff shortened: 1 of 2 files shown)


util.py
```diff
@@ -355,6 +357,14 @@ class NoIAMRole(Exception):
         pass
 
     @staticmethod
+    def is_default(hostname):
+        hostname = hostname.lower()
+        for prefix in EC2.DEFAULT_PREFIXES:
+            if hostname.startswith(prefix):
+                return True
+        return False
```
yannmh: Can we add a short test ?

yannmh:
⚠️  This is a potential backward compatibility breaker, let's test these changes carefully.

- [ ] Run the dockerized Datadog Agent `5.8.x` on EC2  
- [ ] Apply the patch & restart
- [ ] No host is created, instead the existing host is matched using the previous hostname as an alias

degemer:
## Host with latest version (`5.8.4`)

### Datadog app
screen shot 2016-07-18 at 15 10 15

### `datadog-agent` info
screen shot 2016-07-18 at 15 11 26

We can see that the hostname is the bad one `ip-172-31-16-235`, but we sent the correct aliases, and Datadog is aware of its instance-id `i-662d48da`.

After ~15min, I applied the patch and restarted the agent.

## Same host with the patch
### Datadog app
screen shot 2016-07-18 at 15 28 47
screen shot 2016-07-18 at 15 28 55

### `datadog-agent` info
screen shot 2016-07-18 at 15 25 48

Same host, no duplication, this time with the correct hostname: `i-662d48da`.


@yannmh The only issue would be if the host never sent its AWS metadata before upgrading the agent. Which only happens sometimes at the start of the agent, but is fixed after ~4h. So very unlikely.

degemer:
Added a test for the EC2 default hostname.

PR #2662, opened 2016-07-06
upgrades ruby on travis

gmmeyer:
Travis's ruby is currently at an old version that many gems have dropped support for. This will upgrade it.


degemer:
👍 if the build passes, we don't care about the ruby version since it's only used internally for our tests.

gmmeyer:
The

PR #2666, opened 2016-07-07
Stop checks before reloading them

hkaj:
It's cleaner and it prevents the thread pool from growing with no limit after each configuration reload if a network check is enabled.


olivielpeau:
LGTM

PR #2667, opened 2016-07-07
[core] upgrades supervisor to avoid supervisor bug

gmmeyer:
Supervisor 1.3.1 has a bug in it: https://github.com/Supervisor/supervisor/issues/562

This will upgrade it to fix the bug.


degemer:
Changelog: https://github.com/Supervisor/supervisor/blob/master/CHANGES.txt

PR #2668, opened 2016-07-07
add docs for tags...

stensonb:
they appear to be supported: https://github.com/DataDog/dd-agent/blob/master/checks.d/etcd.py#L76

so, let's add them in the example yaml


olivielpeau:
Makes sense, thanks @stensonb!

Merging

PR #2670, opened 2016-07-08
[core][collector] Use flag to signal config reload

olivielpeau:
Reloading configurations and checks during collection runs exposes us
to race conditions, pain and unhappiness (for instance, thread leaks
from `NetworkCheck`s).

To fix that, we reload configurations synchronously before collections
runs. The SIGHUP signal handler sets a flag to let the collector know
that it needs to reload configurations before the next run, and that's
all the signal handler does.


hkaj:
looks good 👍

PR #2673, opened 2016-07-08
Add an HTTP timeout to many integrations.

gphat:
### What's this PR do?
Adds a timeout to many integration's HTTP request methods.

### Motivation
We recently a problem with a cluster that caused the HTTP response to not return for awhile. It ended up repeatedly being killed by the watchdog and at one point even did this somehow:
```
2016-07-08 15:06:28 UTC | INFO | dd.collector | checks.collector(collector.py:539) | Finished run #1. Collection time: 146.23s. Emit time: 0.04s
```

This could've been easily avoided with a timeout!

O_o

### Notes
I picked 9 because it's < 10. Adding this an instance variable on the class — as is done in some other integrations — seemed Hard because i'd have to thread it through all the various functions that are called. Open to suggestions on this.

I came along in 2nd commit and added this to every place I found it missing that we used. I'd love to see this added to _all_ the integrations, but I'll leave that for someone else.

### DIFF
--- a/checks/__init__.py
+++ b/checks/__init__.py
@@ -322,6 +322,7 @@ class AgentCheck(object):
         self.agentConfig = agentConfig
         self.in_developer_mode = agentConfig.get('developer_mode') and psutil
         self._internal_profiling_stats = None
+        self.default_integration_http_timeout = float(agentConfig.get('default_integration_http_timeout', 9))
 
         self.hostname = agentConfig.get('checksd_hostname') or get_hostname(agentConfig)
         self.log = logging.getLogger('%s.%s' % (__name__, name))
--- a/checks.d/haproxy.py
+++ b/checks.d/haproxy.py
@@ -141,7 +141,7 @@ class HAProxy(AgentCheck):
 
         self.log.debug("HAProxy Fetching haproxy search data from: %s" % url)
 
-        r = requests.get(url, auth=auth, headers=headers(self.agentConfig), verify=verify)
+        r = requests.get(url, auth=auth, headers=headers(self.agentConfig), verify=verify, timeout=self.default_integration_http_timeout)
         r.raise_for_status()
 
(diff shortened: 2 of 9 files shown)


gmmeyer:
I think this PR is a good idea! I need to look a bit closer to properly review it. 

But, I think we should find a way to ensure that it's not just a magic number. It's probably be better to put it in the global configuration file and then add it into the AgentCheck class init: https://github.com/DataDog/dd-agent/blob/d18f3cbd9049d80280caa72b073190ad82377795/checks/__init__.py#L306-L348

That way it'll be initialized in every check by default.

gphat:
Excellent. I'll adjust and ping back!

gphat:
Hey @gmmeyer,

How's that look? Python isn't my first language and I have no real ego about this fix, so please advise if you'd like this shaped another way.

checks/__init__.py
```diff
@@ -322,6 +322,7 @@ def __init__(self, name, init_config, agentConfig, instances=None):
         self.agentConfig = agentConfig
         self.in_developer_mode = agentConfig.get('developer_mode') and psutil
         self._internal_profiling_stats = None
+        self.default_integration_http_timeout = int(agentConfig.get('default_integration_http_timeout', 9))
```
remh: Nitpick but let's cast this to a float if someone wants to use a non integer timeout

remh:
Thanks looks good beside the nitpick. Could you also squash your commits please ?

gphat:
Updated, squashed and all that jazz. Lemme know if I can do anything else!

truthbk:
Merging @gphat! Thanks a bunch for your contribution, should make it into the next release. 👍

gphat:
Marvelous. Thanks @truthbk, @gmmeyer and @remh!

PR #2677, opened 2016-07-11
[iis] Fix metrics tagging when multiple sites are specified on instance

olivielpeau:
Before this commit, the check would tag metrics from sites of the same instance with multiple `site:` tags on each metric.

Fix this behavior and improve existing tests to catch the issue.


gmmeyer:
lgtm 👍

olivielpeau:
Thanks @gmmeyer!

Merging. Travis is failing on an unrelated test (`network`), I'll work on fixing the build through a separate PR.

PR #2678, opened 2016-07-12
adding /health to etcd check

stensonb:
Hi all -

The built-in check for etcd neglects to hit the "/health" endpoint.  That endpoint is used by the `etcdctl cluster-health` command-line utility for monitoring cluster health.  It seemed natural to collect this as state and push as a gauge.

Details on what the "/health" endpoint does are here: https://github.com/coreos/etcd/blob/master/etcdserver/api/v2http/client.go#L366-L391

TODO: write some sort of tests...probably can use some help on this...or a pointer or two...my python is very weak (if you couldn't already tell).

Any suggestions?

### DIFF
--- a/checks.d/etcd.py
+++ b/checks.d/etcd.py
@@ -101,7 +101,7 @@ class Etcd(AgentCheck):
         if self_response is not None:
             for key in self.SELF_HEALTH:
                 if key in self_response:
-                    val = 1 if self_response[key]=='true' else 0
+                    val = 1 if self_response[key] == 'true' else 0
                     self.gauge(self.SELF_HEALTH[key], val, tags=instance_tags)
                 else:
                     self.log.warn("Missing key {0} in stats.".format(key))
--- a/checks.d/etcd.py
+++ b/checks.d/etcd.py
@@ -40,6 +40,10 @@ class Etcd(AgentCheck):
         'watchers': 'etcd.store.watchers'
     }
 
+    SELF_HEALTH = {
+        'health': 'etcd.self.health'
+    }
+
     SELF_GAUGES = {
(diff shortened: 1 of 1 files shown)


gmmeyer:
Thank you @stensonb for your contribution! We'll be sure to review this soon.

checks.d/etcd.py
```diff
@@ -92,6 +96,16 @@ def check(self, instance):
         timeout = float(instance.get('timeout', self.DEFAULT_TIMEOUT))
         is_leader = False
 
+        # Gather self health status
+        self_response = self._get_self_health(url, ssl_params, timeout)
+        if self_response is not None:
+            for key in self.SELF_HEALTH:
+                if key in self_response:
+                    val = 1 if self_response[key] == 'true' else 0
```
degemer: I think a service check would actually make more sense, something like `etcd.is_healthy` (first name I found, there's probably a better one) which would do something like `etcd.can_connect`. What do you think ?

(short documentation about service checks: http://docs.datadoghq.com/guides/agent_checks/#sending-service-checks)

degemer:
Hey @stensonb, thanks again for the PR! I added a comment, and I also have another question: is this `/health` endpoint available for all the `etcd` version or was it added recently ?

degemer:
Hey @stensonb, do you need any help with this PR? Not sure if you saw my comment about the service check?

irabinovitch:
Should this be moved to integrations-core?

masci:
@irabinovitch on it

masci:
@stensonb moved your work to https://github.com/DataDog/integrations-core/pull/917 and closing this, thanks for the contribution!

PR #2679, opened 2016-07-12
[core] better dogstatsd server error message

gmmeyer:
The dogstatsd server error message does not currently include enough information. It should include the serveraddr as well, especially because, when it errors out, the message comes directly after the message `2016-07-12 19:34:47 UTC | INFO | dd.dogstatsd | dogstatsd(dogstatsd.py:165) | Reporting to http://localhost:17123 every 10s` this can easily be confusing.

### DIFF
--- a/dogstatsd.py
+++ b/dogstatsd.py
@@ -440,7 +440,8 @@ class Dogstatsd(Daemon):
             try:
                 self.server.start()
             except Exception as e:
-                log.exception('Error starting server')
+                log.exception(
+                    'Error starting dogstatsd server on %s', self.server.sockaddr)
                 raise e
         finally:
             # The server will block until it's done. Once we're here, shutdown


dogstatsd.py
```diff
@@ -440,7 +440,8 @@ def run(self):
             try:
                 self.server.start()
             except Exception as e:
-                log.exception('Error starting server')
+                log.exception(
+                    'Error starting dogstatsd server on {0}'.format(self.server.sockaddr))
```
yannmh: Ultra nitpick 👮 : it's usually preferable to let the `logging` module do the formatting, i.e.
```python
log.exception('Error starting dogstatsd server on %s', self.server.sockaddr)
```

It avoids to perform unnecessary or prematurely formatting (although in your case you're logging with the `EXCEPTION` level so it makes no difference).

yannmh:
:shipit:

gmmeyer:
I changed it to align with your nit, @yannmh.

PR #2681, opened 2016-07-13
[haproxy] Fix `KeyError` when an unknown status is found

olivielpeau:
An unknown status (i.e. one that's not listed in
`Services.ALL_STATUSES`) would make the check fail with a `KeyError`
because that status wasn't in the initialized `dict` of statuses.

With this commit the check now sends unknown statuses as:
- an `UNAVAILABLE` status if `collate_status_tags_per_host` is enabled,
in order to keep the number of contexts low
- the unknown status as-is if `collate_status_tags_per_host` is left
disabled, in order to give maximum visibility in the actual statuses

Also, address the regression reported in #2685 with a slightly different approach to make it play well with the other parts of the check: suffixes to the statuses are handled correctly again. Thanks a bunch to @alexism for reporting this and suggesting a fix (see #2685 for more details).


checks.d/haproxy.py
```diff
@@ -338,11 +362,12 @@ def _process_status_metric(self, hosts_statuses, collect_status_metrics_by_host,
                     self.warning('`collect_status_metrics_by_host` is enabled but no host info\
                                  could be extracted from HAProxy stats endpoint for {0}'.format(service))
                 service, status = host_status
-            status = status.lower()
 
             if self._is_service_excl_filtered(service, services_incl_filter, services_excl_filter):
                 continue
 
+            status = self._normalize_status(status)
```
alexism: why not applying this normalization in `_line_to_dict` ?

checks.d/haproxy.py
```diff
@@ -287,9 +295,21 @@ def _tag_match_patterns(self, tag, filters):
                 return True
         return False
 
+    @staticmethod
+    def _normalize_status(status):
+        """
+        Try to match the HAProxy status with one of the statuses defined in `ALL_STATUSES`,
+        if it can't be matched return the status as-is in a tag-friendly format
+        """
+        formatted_status = status.lower().replace(" ", "_")
```
alexism: This is not needed.

tests/checks/mock/test_haproxy.py
```diff
@@ -19,6 +19,7 @@
 b,BACKEND,0,0,1,2,0,421,1,0,0,0,,0,0,0,0,UP,6,6,0,,0,1,0,,1,3,0,,421,,1,0,,1,,,,,,,,,,,,,,0,0,
 c,i-1,0,0,0,1,,1,1,0,,0,,0,0,0,0,UP,1,1,0,0,1,1,30,,1,3,1,,70,,2,0,,1,1,,0,,,,,,,0,,,,0,0,
```
alexism: It would be good to cover the `UP %d/%d` cases

checks.d/haproxy.py
olivielpeau: replacing spaces with underscores is needed when the raw status is `no check` for instance, but we could do it after we've tried matching the status with the "known" statuses

checks.d/haproxy.py
olivielpeau: applying it in `_line_to_dict` would also require changing the status comparisons that are done in `_process_event`. Not necessarily complex though, I'll have a closer look to see what could be done there

tests/checks/mock/test_haproxy.py
olivielpeau: agreed, I'll change the test to cover that

checks.d/haproxy.py
anroots: Not sure if relevant, but I found this page because I'm failing with
```
  File "/opt/datadog-agent/agent/checks.d/haproxy.py", line 355, in _process_status_metric
    statuses_counter[tuple(tags)][counter_status] += count
KeyError: 'no check'
```

- so status can definitely be "no check".

tests/checks/mock/test_haproxy.py
olivielpeau: ok, I've changed the test to cover this case

alexism:
looks good to me. Thanks @olivielpeau !

tests/checks/mock/test_haproxy.py
```diff
@@ -135,7 +140,68 @@ def test_count_per_status_collate_per_host(self, mock_requests):
...
+        self.assertMetric('haproxy.count_per_status', value=1, tags=['backend:i-3', 'status:unavailable'])
         self.assertMetric('haproxy.count_per_status', value=1, tags=['backend:i-4', 'status:unavailable'])
         self.assertMetric('haproxy.count_per_status', value=1, tags=['backend:i-5', 'status:unavailable'])
 
         self._assert_agg_statuses(count_status_by_service=False, collate_status_tags_per_host=True)
+
+    # This mock is only useful to make the first `run_check` run w/o errors (which in turn is useful only to initialize the check)
+    @mock.patch('requests.get', return_value=mock.Mock(content=MOCK_DATA))
+    def test_count_hosts_statuses(self, mock_requests):
+        from collections import defaultdict
```
yannmh: Nitpick ✏️ :  this should be placed in the header.

checks.d/haproxy.py
yannmh: > replacing spaces with underscores is needed when the raw status is no check for instance

I am not sure to follow. Is it for tag normalization ?

yannmh:
:shipit:

olivielpeau:
Thanks @alexism and @yannmh for the reviews!

I'll merge once the CI passes

olivielpeau:
CI failure is unrelated (Kong). Merging!

PR #2685, opened 2016-07-15
fix parsing of the haproxy status field.

alexism:
As per the documentation, the status might have information after the documented enumeration (UP/DOWN/...).

See https://cbonte.github.io/haproxy-dconv/configuration-1.5.html#9.1 :

     `17. status [LFBS]: status (UP/DOWN/NOLB/MAINT/MAINT(via)...)`

(mind the 3 trailing dots)

More precisely, the haproxy code shows that 2 suffixes might be used: ` %d/%d` and ` (agent)`
https://github.com/haproxy/haproxy/blob/a5de024d42c4113fc6e189ea1d0ba6335219e151/src/dumpstats.c#L4117-L4129

When these suffixes are included in the haproxy stats (for example, `up 1/2`, the check will fail with:

```
2016-07-15 01:23:05 UTC | WARNING | dd.collector | checks.haproxy(__init__.py:654) | ALEX service=geo hostname=None status=up 1/2
2016-07-15 01:23:05 UTC | ERROR | dd.collector | checks.haproxy(__init__.py:767) | Check 'haproxy' instance #0 failed
  File "/opt/datadog-agent/agent/checks/__init__.py", line 750, in run
  File "/opt/datadog-agent/agent/checks.d/haproxy.py", line 132, in check
    count_status_by_service=count_status_by_service,
  File "/opt/datadog-agent/agent/checks.d/haproxy.py", line 211, in _process_data
    count_status_by_service=count_status_by_service
  File "/opt/datadog-agent/agent/checks.d/haproxy.py", line 357, in _process_status_metric
    statuses_counter[tuple(tags)][counter_status] += count
KeyError: 'up 1/2'
```

In previous version of the agent, receiving `UP 1/2` would not throw an error because the value wasn't used has a key. It would however cause the count of available services to be wrong.

https://github.com/DataDog/dd-agent/blob/5.6.x/checks.d/haproxy.py#L267


remh:
Thanks a lot @alexism . I think it's similar to https://github.com/DataDog/dd-agent/pull/2681 ?

@olivielpeau can you make sure of this ?

alexism:
@remh it seems to be the same issue indeed.
However I don't think the proposed solution is correct as it only ignores the unrecognized status and [defaults](https://github.com/DataDog/dd-agent/pull/2681/files#diff-5fdb6bd235209777fea94f2dd5c5a7b2R371) to `UNAVAILABLE`. This could lead to incorrect information. For example, when the status is `UP 1/2`, the service is indeed up, not unavailable.

The proper solution is to improve the status parser to handle all the documented values. This is the approach I took in this PR (https://github.com/DataDog/dd-agent/pull/2685)

alexism:
duplicate of https://github.com/DataDog/dd-agent/pull/2681

PR #1918, opened 2015-09-15
[jenkins] postpone Jenkins check deprecation

yannmh:
Postpone Jenkins AgentCheck deprecation until Datadog Jenkins plugin is considered ready.

PR #2688, opened 2016-07-18
[jenkins] Deprecate check

olivielpeau:
Its replacement is ready now, so we can go ahead and deprecate it.

(See #1918 which reverted the deprecation because the replacement wasn't deemed ready at the time)


hkaj:
:shipit:

degemer:
`[skip ci]` 🔴

olivielpeau:
Merging, CI failures are unrelated

PR #2690, opened 2016-07-18
[tests] lower flakiness of test_no_parallelism

degemer:


masci:
Can we try to re-enable the test under Win?

degemer:
@masci Trying it now. It is still relying on `time.sleep` though.

degemer:
I restarted it twice on AppVeyor (so in total 6 runs) and it didn't fail on Windows, so I think we can let it enabled.

PR #2691, opened 2016-07-20
[dogstatsd] fix server address when non_local_traffic is passed

masci:
We used to pass an empty string `''` meaning `any available interface` when `non_local_traffic` is present in the config file.

This doesn't play well with the new code since we now use a unique, fullstack socket so we explicitly set the IPv4 meta address `0.0.0.0` for retro compat.

This leaves an open issue: it's not possible to start the agent with `non_local_traffic` on an IPv6 only environment. For standalone Dogstatsd instances this is not a real issue, one can pass `::` as `bind_host`.

### DIFF
--- a/dogstatsd.py
+++ b/dogstatsd.py
@@ -484,6 +484,7 @@ def init(config_path=None, use_watchdog=False, use_forwarder=False, args=None):
     forward_to_port = c.get('statsd_forward_port')
     event_chunk_size = c.get('event_chunk_size')
     recent_point_threshold = c.get('recent_point_threshold', None)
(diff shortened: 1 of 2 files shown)


remh:
Looks good to me

bchen421:
What about multi-home systems that are on many networks?  We want to explicitly set bind_host to listen on trusted networks and allow non_local_traffic, but now we have to work around this by using iptables or some set of security group rules.  Shouldn't a setting called bind_host actually set the bind_host?  We upgraded to agent 5.10 and was hit with this unpleasant surprise

PR #2694, opened 2016-07-20
[sdk] minor tweaks - sdk env detection, check location

truthbk:
`AgentCheckTest.load_check()` and `load_check()` in `tests.checks.common` are both needed, one by integration tests, the other by unit tests (respectively). We could maybe get rid of the former but it would involve too many changes at the moment. The proposed changes will allow both unit tests (if added in the SDK) and integration tests to correctly load the checks in the testing infrastructure.

Also, because integrations-core would've been dropped into `3rd-party`, and that is misleading, renaming to `dd-integrations`.


olivielpeau:
LGTM! :shipit:

olivielpeau:
`integrations` ❤️

degemer:
Default, core_integration and checks_mock are 💚 . Merging.

PR #2695, opened 2016-07-20
[disk] `use_df` parameter to force use `df`

yannmh:
`psutil.disk_usage` method uses the `statvfs` POSIX API to resolve the
disks' usage. **Problem?** This call can hang "forever" when a file
system is listed but not available. Worse, it can’t get interrupted or
killed, making the Watchdog helpless.

I attached more references to this problem below.

> Basically if you have a network filesystem listed in /proc/mounts and
  it is unreachable (e.g. because there is no network), statvfs will
  hang on stat’ing the network directory, even if you called statvfs on
  a completely different directory

To workaround this issue, add an optional `use_df` parameter to instruct
the check to use the `df` command instead of the `psutil` module. `df`
uses the `statvfs` POSIX API too, but prints an appropriate error message
when it fails instead of hanging.

**References**
https://sourceware.org/ml/libc-hacker/2003-09/msg00056.html
http://ilostmynotes.blogspot.com/2014/05/cross-platform-python-method-for.html
https://www.redhat.com/archives/rhl-list/2004-March/msg01647.html

cc @gmmeyer

### DIFF
--- a/checks.d/disk.py
+++ b/checks.d/disk.py
@@ -36,9 +36,11 @@ class Disk(AgentCheck):
 
     def check(self, instance):
         """Get disk space/inode stats"""
+        use_df = Platform.is_unix() and _is_affirmative(instance.get('use_df', False))
+
         # Windows and Mac will always have psutil
         # (we have packaged for both of them)
-        if self._psutil():
+        if self._psutil() and not use_df:
             if Platform.is_linux():
                 procfs_path = self.agentConfig.get('procfs_path', '/proc').rstrip('/')
                 psutil.PROCFS_PATH = procfs_path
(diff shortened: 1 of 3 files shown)


checks.d/disk.py
```diff
@@ -36,9 +36,11 @@ def __init__(self, name, init_config, agentConfig, instances=None):
 
     def check(self, instance):
         """Get disk space/inode stats"""
+        use_df = Platform.is_unix() & _is_affirmative(instance.get('use_df', False))
```
yannmh: I wonder if we should narrow the scope to Linux only. Thoughts?

olivielpeau:
Changes are 👍 

Let's discuss if the issue occurs often enough to justify a new flag in the Agent-packaged check though :)

yannmh:
This doesn't work. Closing.

PR #2697, opened 2016-07-21
Added catching of single job id exceptions to deal broken redirect in resource manager

johan-g:
Since the redirect for finished queries in resource manager to mapreduce history server is broken (faulty addresses are generated), a quick workaround to still collect metrics from the jobs still running is to catch the exceptions within the loop


gmmeyer:
Thank you for your contribution @johan-g! We'll review it soon and either merge it or get back to you with some additional feedback! 😄

johan-g:
Thanks, I pushed one more update to this as sometimes JSONDecodeError is raised due to this race condition too

PR #2698, opened 2016-07-22
[supervisord] add custom tags

Ricky-Thomas:
Hack Day 7/22 adding custom tagging to supervisor

### DIFF
--- a/checks.d/supervisord.py
+++ b/checks.d/supervisord.py
@@ -51,14 +51,15 @@ class SupervisordCheck(AgentCheck):
 
     def check(self, instance):
         server_name = instance.get('name')
+        custom_tags = instance.get('tags', [])
 
         if not server_name or not server_name.strip():
             raise Exception("Supervisor server name not specified in yaml configuration.")
 
-        server_service_check_tags = ['%s:%s' % (SERVER_TAG, server_name)]
+        server_service_check_tags = ['%s:%s' % (SERVER_TAG, server_name)] + custom_tags
         supe = self._connect(instance)
         count_by_status = defaultdict(int)
-
+        
         # Gather all process information
         try:
             processes = supe.getAllProcessInfo()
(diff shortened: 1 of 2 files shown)


checks.d/supervisord.py
```diff
@@ -51,14 +51,15 @@ class SupervisordCheck(AgentCheck):
...
 
         if not server_name or not server_name.strip():
             raise Exception("Supervisor server name not specified in yaml configuration.")
 
-        server_service_check_tags = ['%s:%s' % (SERVER_TAG, server_name)]
+        server_service_check_tags = ['%s:%s' % (SERVER_TAG, server_name)] + custom_tags
         supe = self._connect(instance)
         count_by_status = defaultdict(int)
-
+
```
degemer: Nit: can you remove these extra spaces ?

conf.d/supervisord.yaml.example
```diff
@@ -42,6 +42,8 @@ instances:
 #      - apache2       #           If not specified, the check will monitor all processes.
 #      - webapp
 #      - java
+#     tags:
+#      - processgroup:groupname
```
degemer: Nit: I think there is only a one space diff between `tags` and `-processgroup:groupname`, we usually use 2, can you update ?

degemer:
Thanks @Ricky-Thomas ! I think you forgot this one: https://github.com/DataDog/dd-agent/blob/master/checks.d/supervisord.py#L146

Could you also update the tests ? https://github.com/DataDog/dd-agent/blob/master/tests/checks/integration/test_supervisord.py or https://github.com/DataDog/dd-agent/blob/master/tests/checks/mock/test_supervisord.py

truthbk:
Hi @Ricky-Thomas,

Thanks again for your contribution, always very much appreciated. We're currently transitioning our checks to a new SDK-based model for the agent. An unfortunate side-effect is that we have to impose a freeze on the dd-agent repo. The SDK consists of two repositories https://github.com/DataDog/integrations-core and https://github.com/DataDog/integrations-extras, where all of our checks will reside. This PR would belong in the extras repository. Would you be so kind as to re-open the PR there? We have facilitated some tools to help you with the SDK development that should help you get up and running in no time: sdk guide datadog sdk gem. We are very sorry for the inconvenience caused.

PR #2702, opened 2016-07-25
Improve service discovery to only reload checks that need it

hkaj:
# What does this PR do?

It makes configuration reload smarter.
The gist of it is that the agent now maintains a mapping `container_identifier <--> checks` and only reloads checks for which a docker event associated with an identifier in this mapping happened since the last collector run.

# Motivation

In environments where a lot of docker events happen, full config reload on any container event is not a good design. We noticed several cases where the agent triggered a reload at each run, making rate collection impossible (calculating rates requires values from several runs, and these are forgotten on each check reload).
Individual reload is a better option.

# Additional notes

In more details:
- the `identifier_to_checks` cache is built based on the content of the config store and the `auto_conf` folder. This cache is invalidated and refreshed when a template is modified in the K/V store.
- a full reload is triggered on two events: a `SIGHUP` is received by the collector or the agent detects a change in the config store.
- Changes to `auto_conf` files are not detected yet so an agent restart or a manual `SIGHUP` is still needed here.
- Incidentally this PR adds a `load_check` function that does the same thing as `load_check_directory` but for a single check which name was passed as a third parameter.

### DIFF
--- a/agent.py
+++ b/agent.py
@@ -19,6 +19,7 @@ import os
 import signal
 import sys
 import time
+from copy import copy
 
 # For pickle & PID files, see issue 293
 os.umask(022)
@@ -31,6 +32,7 @@ from config import (
     get_parsed_args,
     get_system_stats,
     load_check_directory,
+    load_check
 )
 from daemon import AgentSupervisor, Daemon
 from emitter import http_emitter
@@ -76,6 +78,7 @@ class Agent(Daemon):
         self._checksd = []
         self.collector_profile_interval = DEFAULT_COLLECTOR_PROFILE_INTERVAL
         self.check_frequency = None
+        # this flag can be set to True, False, or a list of checks (for partial reload)
         self.reload_configs_flag = False
         self.sd_backend = None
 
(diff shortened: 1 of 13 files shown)


util.py
```diff
@@ -200,7 +200,7 @@ def get_hostname(config=None):
                 return gce_hostname
 
     # Try to get the docker hostname
-    docker_util = DockerUtil()
+    docker_util = DockerUtil(agentConfig=config)
```
olivielpeau: Passing `agentConfig` doesn't change anything here, or am I missing something?

utils/service_discovery/consul_config_store.py
```diff
@@ -46,10 +46,13 @@ def get_client(self, reset=False):
 
     def client_read(self, path, **kwargs):
         """Retrieve a value from a consul key."""
-        recurse = kwargs.get('recursive', False)
+        recurse = kwargs.get('recursive', False or kwargs.get('all, False'))
```
olivielpeau: this looks up the `'all, False'` key. The correct lookup should be:

```py
recurse = kwargs.get('recursive') or kwargs.get('all', False)
```
or
```py
recurse = kwargs.get('recursive', kwargs.get('all', False))
```

depending on whether a "falsey" value of `kwargs['recursive']` should default to a lookup of the `all` key or not

utils/dockerutil.py
```diff
@@ -111,7 +122,11 @@ def get_events(self):
                 # [0]: https://github.com/docker/docker-py/pull/1082
                 log.debug('Unable to parse Docker event: %s', event)
 
-        return self.events, should_reload_conf
+        return self.events, conf_reload_set
+
+    def _get_checks_from_image(self, image):
+        """Get the list of checks applied to an image from the image_to_checks cache in the config store"""
+        return self.config_store.image_to_checks[image]
```
olivielpeau: Here we're relying on `image_to_checks` to be a `defaultdict` in case the image isn't in the cache. I think we should probably make that explicit either here or at the initialization of `self.image_to_checks` in `abstract_config_store`

utils/service_discovery/abstract_config_store.py
```diff
@@ -61,6 +65,28 @@ def client_read(self, path, **kwargs):
     def dump_directory(self, path, **kwargs):
         raise NotImplementedError()
 
+    def _populate_image_to_checks(self):
+        """Populate the image_to_checks cache with templates pulled
+        from the config store and from the auto-config folder"""
+        images_to_checks = defaultdict(set)
```
olivielpeau: Nitpick: we could call the variable `image_to_checks` to be consistent with `self.image_to_checks` :)

utils/dockerutil.py
```diff
@@ -93,16 +101,19 @@ def is_ecs(self):
...
         event_generator = self.client.events(since=self._latest_event_collection_ts,
                                              until=now, decode=True)
         self._latest_event_collection_ts = now
         for event in event_generator:
             try:
-                if event.get('status') in CONFIG_RELOAD_STATUS:
-                    should_reload_conf = True
+                if self.config_store and event.get('status') in CONFIG_RELOAD_STATUS:
+                    image = event.get('from', '')
+                    checks = self._get_checks_from_image(image)
```
olivielpeau: Are we still able to identify the correct checks if the `com.datadoghq.sd.check.id` label is used on the container?

utils/service_discovery/abstract_config_store.py
```diff
@@ -208,8 +237,18 @@ def crawl_config_template(self):
...
             return True
         return False
+
+    def _update_image_to_checks(self, image, check_names):
+        """Try to insert in the image_to_checks cache the mapping between an image and its check names"""
+        if image not in self.image_to_checks:
+            self.image_to_checks[image] = set(check_names)
+        elif self.image_to_checks[image] != set(check_names):
+            log.warning("Trying to cache check names for image %s but a different value is already there. "
+                        "This should not happen. Not updating." % image)
```
olivielpeau: I don't understand why this branch of the conditional (i.e. already having a stored value in the cache) should not happen whereas the other branch (`image not in self.image_to_checks `) should 😕

agent.py
```diff
@@ -98,26 +101,82 @@ def _handle_sighup(self, signum, frame):
...
+                # if the check previously failed to load, pop it from init_failed_checks
+                checksd['init_failed_checks'].pop(check_name)
+
+            fresh_check = load_check(self._agentConfig, hostname, check_name)
+
+            # this is an error dict
+            # checks that failed to load are added to init_failed_checks
+            if isinstance(fresh_check, dict) and 'error' in fresh_check.keys():
+                checksd['init_failed_checks'][fresh_check.keys()[0]] = fresh_check.values()[0]
+
```
olivielpeau: here, if the check failed to initialize but was previously loaded, don't we want to remove it from `checksd['initialized_checks']` too?

olivielpeau:
Added in a few comments, but looks nice overall! 📈

agent.py
hkaj: welp, that's a good point

util.py
hkaj: it's used [in DockerUtil](https://github.com/DataDog/dd-agent/blob/haissam/fine-grained-reload/utils/dockerutil.py#L55-L58) to get the instance of the config_store. We could pass a subset of it instead since we basically just need the `sd_config_backend` field, but that would be more confusing than useful imho.

utils/service_discovery/abstract_config_store.py
hkaj: Well none should happen, but we can cope silently with the first one. However for the second one there's no way to know which value to keep. Do you think we should add a warning for the other branch too?

hkaj:
Thanks @olivielpeau, that was a kick-ass review!
Updated according to your comments.

utils/service_discovery/abstract_config_store.py
olivielpeau: hmm I don't think we need to add a warning, but to be really nitpicky I'd like the `This should not happen` part to be left out of the message (since it's a bit scary imho) and add a comment on the function like `This should very rarely happen, but we try to cope with it as much as possible` :)

olivielpeau:
Responded to one of your comments, but other than that LGTM

Feel free to merge once that's addressed 🎆

olivielpeau:
👍  looking good, feel free to merge! (once the commit history is cleaned up a bit :) )

hkaj:
It's failing on a cassandra timeout. Merging as it's unrelated.

PR #2704, opened 2016-07-25
[http_check] log exceptions 🔊

yannmh:
When the HTTP check is misconfigured, the thread pool used for network
checks woud restart over and over without providing any usefull
information, i.e.

```
Running check http_check
Stopping Thread Pool
Starting Thread Pool
Stopping Thread Pool
Starting Thread Pool
received 0 payloads since last flush
```

Instead, exceptions should be logged so it is easier to troubleshoot.


degemer:
LGTM

PR #2707, opened 2016-07-26
[ES] Allow path in elasticsearch url

masci:
## Why

When users provide an url like this in `elastic.yaml`:

```
instances:
    - url: http://example.com/elasticsearch/
```

due to a misuse of `urljoin`, the resulting url for (example) stats is this:

```

'http://example.com/_stats'
```
## How

Get rid of `urljoin`, strip trailing slashes from configurations and concatenate the final url with `+`

### DIFF
--- a/checks.d/elastic.py
+++ b/checks.d/elastic.py
@@ -322,25 +322,28 @@ class ESCheck(AgentCheck):
         health_url, nodes_url, stats_url, pshard_stats_url, pending_tasks_url, stats_metrics, \
             pshard_stats_metrics = self._define_params(version, config.cluster_stats)
 
+        # Strip trailing slash from the ES url
+        es_url = config.url.strip('/')
+
         # Load clusterwise data
         if config.pshard_stats:
-            pshard_stats_url = urlparse.urljoin(config.url, pshard_stats_url)
+            pshard_stats_url = es_url + pshard_stats_url
             pshard_stats_data = self._get_data(pshard_stats_url, config)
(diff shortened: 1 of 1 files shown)


Martiflex:
After further review, the fix offered doesn't take into account that prior to the `urljoin`s that were changed, the **get_instance_config** function modifies the url and removes the path part of the url:

https://github.com/DataDog/dd-agent/blob/2787f95069fc22c3530d8f88281bf6f76a67d9b1/checks.d/elastic.py#L277-L283

```
import urlparse
url = "http://internal.dev3.master.us-west-2.prod.aws.intellify.io/elasticsearch/"
parsed = urlparse.urlparse(url)
if parsed[2] != "":
 url = "%s://%s" % (parsed[0], parsed[1])
print url

# result: http://internal.dev3.master.us-west-2.prod.aws.intellify.io
```

It seems actually _intended_ to remove the path from the URL, so there may be good reasons for that?

truthbk:
Hi @masci,

Thanks again for your contribution, always very much appreciated. We're currently transitioning our checks to a new SDK-based model for the agent. An unfortunate side-effect is that we have to impose a freeze on the dd-agent repo. The SDK consists of two repositories https://github.com/DataDog/integrations-core and https://github.com/DataDog/integrations-extras, where all of our checks will reside. This PR would belong in the **[core](https://github.com/DataDog/integrations-core)** repository. Would you be so kind as to re-open the PR there? We have facilitated some tools to help you with the SDK development that should help you get up and running in no time: 

PR #2708, opened 2016-07-26
[ssh] Be more specific when logging ssh errors

masci:
## Why
If the ssh check has trouble loading a private key file, we log a generic error message so that it's impossible to determine the cause.

## What
Catch specific exceptions and log different warning messages.


degemer:
👍

PR #2711, opened 2016-07-27
[RabbitMQ] new health check

masci:
### What does this PR do?

A new status check was added, `rabbitmq.status` that reflects the status of the RabbitMQ service (up/down), which is different from the outcome of the `aliveness` strategy that is still implemented by the existing `rabbitmq.aliveness` status check.

### Motivation

The aliveness check is misleading when the RabbitMQ service is down - an exception is raised and the check simply provides _no data_ whilst the _info_ command, along with the UI correctly report an issue for the integration.

### DIFF
--- a/tests/checks/mock/test_rabbitmq.py
+++ b/tests/checks/mock/test_rabbitmq.py
@@ -1,0 +1,62 @@
+import sys
+
+import mock
+import requests
+
+from checks import AgentCheck
+from tests.checks.common import AgentCheckTest, get_checksd_path, get_os
+
+
+class TestRabbitMQ(AgentCheckTest):
+
+    CHECK_NAME = 'rabbitmq'
+
+    @classmethod
+    def setUpClass(cls):
+        sys.path.append(get_checksd_path(get_os()))
+
+    @classmethod
+    def tearDownClass(cls):
+        sys.path.pop()
+
+    def test__get_data(self):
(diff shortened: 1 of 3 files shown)


tests/checks/mock/test_rabbitmq.py
```diff
@@ -0,0 +1,62 @@
...
+    def tearDownClass(cls):
+        sys.path.pop()
+
+    def test__get_data(self):
+        with mock.patch('rabbitmq.requests') as r:
+            from rabbitmq import RabbitMQ, RabbitMQException  # pylint: disable=import-error
+            check = RabbitMQ('rabbitmq', {}, {"instances": [{"rabbitmq_api_url": "http://example.com"}]})
+            r.get.side_effect = [requests.exceptions.HTTPError, ValueError]
+            self.assertRaises(RabbitMQException, check._get_data, '')
+            self.assertRaises(RabbitMQException, check._get_data, '')
```
truthbk: twice?

truthbk:
Love the tests ❤️ !

This looks great to me, just one comment. Otherwise :shipit:

tests/checks/mock/test_rabbitmq.py
masci: It's to consume both the `side_effect`s so we test it both with `HTTPError` (or any requests exception) and `ValueError`

PR #2713, opened 2016-07-27
[mongodb] authenticate with X.509

yannmh:
To authenticate using a X.509 certificate, one must use the `db.auth()`
method specifying `MONGODB-X509` for the `mechanism` field, and the user
that corresponds to the client certificate for the user field.

More information:
https://api.mongodb.com/python/current/examples/authentication.html
https://docs.mongodb.com/manual/tutorial/configure-x509-client-authentication/

### DIFF
--- a/checks.d/mongo.py
+++ b/checks.d/mongo.py
@@ -550,25 +550,37 @@ class MongoDb(AgentCheck):
             metric_prefix=metric_prefix, metric_suffix=metric_suffix
         )
 
-    def _authenticate(self, database, username, password, ssl_params):
+    def _authenticate(self, database, username, password, use_x509):
         """
         Authenticate to the database.
+
+        Available mechanisms:
+        * Username & password
+        * X.509
+
+        More information:
+        https://api.mongodb.com/python/current/examples/authentication.html
         """
-        # X.509 authentication
+        authenticated = False
         try:
-            if ssl_params:
+            # X.509
+            if use_x509:
                 self.log.debug(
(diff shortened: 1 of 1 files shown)


degemer:
Could you also add a `try ... except` around the `authenticate` method ?
According to the [docs](https://api.mongodb.com/python/current/api/pymongo/database.html?highlight=authenticate#pymongo.database.Database.authenticate) it can raise a subclass of `PyMongoError`, and we don't catch it for now (thus not sending the SC).

degemer:
Apart from that, LGTM

checks.d/mongo.py
```diff
@@ -533,6 +534,38 @@ def _normalize(self, metric_name, submit_method, prefix):
...
+                authenticated = database.authenticate(username, mechanism='MONGODB-X509')
+
+            # Username & password
+            else:
+                authenticated = database.authenticate(username, password)
+
+        except pymongo.errors.PyMongoError as e:
+            self.log.error(
+                u"Authentication failed due to invalid credentials or configuration issues. %s", e
+            )
```
degemer: I think there's a `raise` missing.

degemer: Could this log.error leak the `password` if there was an auth failure ?

degemer: For posterity's sake: raise is not needed as it is dealt with once the function returns `False`.
And @yannmh is working in another branch on avoiding credentials leak in all logs.

olivielpeau:
This was released with 5.9.0, removing 5.10 milestone

PR #2714, opened 2016-07-27
[disk] timeout on disk usage

yannmh:
Timeout disk usage queries after 5 seconds with a warning and no data.
Recover at next iteration retrieving the existing query.


truthbk:
looks good! 👍

PR #2715, opened 2016-07-27
[core] remove noisy logs

degemer:
### What does this PR do?

Remove noisy logs.

### Motivation

Not see them anymore.

### Additional Notes

:fire:

### DIFF
--- a/daemon.py
+++ b/daemon.py
@@ -24,6 +24,7 @@ from utils.process import is_my_process
 
 log = logging.getLogger(__name__)
 
+
 class AgentSupervisor(object):
     ''' A simple supervisor to keep a restart a child on expected auto-restarts
     '''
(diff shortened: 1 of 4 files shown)


daemon.py
```diff
@@ -154,7 +154,6 @@ def daemonize(self):
 
 
     def start(self, foreground=False):
-        log.info("Starting")
```
olivielpeau: I like this line, it lets you search easily in the logs for when the component started.

olivielpeau:
Just added one small 🚒  comment, let me know what you think.

Other than that 🚢 🔥 🔥 🔥

daemon.py
degemer: Indeed, it works for both the collector and dogstatsd, I'm going to let this one and remove the duplicate in the collector (`Running in foreground`).

degemer:
Done, thanks @olivielpeau !
Will merge once it's 💚 .

PR #2716, opened 2016-07-28
[CI] run multiple instances of pylint

masci:
### What does this PR do?

Adds the `-P` option to `xargs` to run more than one `pylint` process in parallel. The `--jobs` options offered by `pylint` was the first choice but it's buggy, so the fallback to `xargs`.

### Motivation

The execution is sequential so on Travis one core is sitting during the checks. Moreover, `pylint` has a huge overhead at startup that in our case is multiplied by the number of files to process, ~270. 

The first problem is addressed by the `-P` option passed to `xargs`, the second one is addressed by letting each `pylint` instance inspect 100 files per run (`-n 100` passed to `xargs`).


degemer:
😍  👍

PR #2724, opened 2016-08-01
Add CPU throttling stats to Docker metrics.

ejholmes:
### What does this PR do?

This adds a `docker.cpu.throttled` metric to the metrics generated by `docker_daemon.py`.

### Motivation

This metric is easily obtained from `docker stats` but missing from DataDog, and can be useful in determining if a container is getting CPU throttled.

### DIFF
--- a/checks.d/docker_daemon.py
+++ b/checks.d/docker_daemon.py
@@ -69,6 +69,13 @@ CGROUP_METRICS = [
         },
     },
     {
+        "cgroup": "cpu",
+        "file": "cpu.stat",
+        "metrics": {
+            "nr_throttled": ("docker.cpu.throttled", GAUGE)
+        },
+    },
+    {
         "cgroup": "blkio",
         "file": 'blkio.throttle.io_service_bytes',
         "metrics": {
--- a/checks.d/docker_daemon.py
+++ b/checks.d/docker_daemon.py
@@ -72,7 +72,7 @@ CGROUP_METRICS = [
         "cgroup": "cpu",
         "file": "cpu.stat",
         "metrics": {
-            "nr_throttled": ("docker.cpu.throttled", GAUGE)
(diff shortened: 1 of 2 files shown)


truthbk:
@ejholmes thank you for the PR, we'll review soon and either merge or get back to you with some feedback. Thanks again! 👍

parkr:
@truthbk Hey! This would be useful to me as well as some other fixes on master. Any idea when you might cut another release?

checks.d/docker_daemon.py
```diff
@@ -69,6 +69,13 @@
         },
     },
     {
+        "cgroup": "cpu",
+        "file": "cpu.stat",
+        "metrics": {
+            "nr_throttled": ("docker.cpu.throttled", GAUGE)
```
truthbk: I believe this is always increasing value, in which case a `rate` might be a better data type... not totally sure though.

truthbk:
@parkr we will try to merge this in soon, and hopefully make available in the next release.

truthbk:
@ejholmes just one comment... totally open to discussion obviously! :)

truthbk:
@ejholmes quick ping regarding the question about the nature of `nr_throttled`. I'd like to merge this for our next release but we'd like to tackle/clarify that small issue. Thanks!

ejholmes:
Hey @truthbk. Yeah, this should be an ever increasing number, so a rate probably makes the most sense here. I'll go ahead and update this PR.

truthbk:
Thanks a ton @ejholmes! I think we're good to go! Thanks so much! :shipit:

truthbk:
Tests are failing, it's unrelated really. I'm going to hold off before merging though just to make sure we get to the failed test before doing so.

ejholmes:
I can also rebase this on current master if that'll help.

truthbk:
@ejholmes no worries, it's broken on master as well. You're good.

PR #2725, opened 2016-08-02
[ci][win32] Add basic integration tests

olivielpeau:
### What does this PR do?

This adds basic integration tests for the following checks on Windows:
* `iis`
* `sqlserver`
* `windows_service`
* `wmi_check`

### Motivation

Right now we don't have any integration tests on the windows checks, so our CI can't even detect some basic regressions. These tests should help avoid some of these regressions.

### Additional Notes

* For simplicity's sake, these tests make heavy use of the environment provided by AppVeyor (in particular the services that are pre-installed on AppVeyor VMs). This means that they would most likely fail if run on a local dev environment on Windows.
* The new tests all run under the `windows` flavor
* The tests run sequentially after the core tests of the Agent have run because the setup phase of the whole environment (install of `pywin32`, etc) still takes a lot of time compared to the actual tests, so it makes sense to do it once for all the tests.


tests/checks/integration/test_iis.py
```diff
@@ -0,0 +1,77 @@
+# 3p
+from nose.plugins.attrib import attr
+
+# project
+from checks import AgentCheck
```
degemer: Not needed, we can avoid using it by calling the `assertServiceCheckOK`, `assertServiceCheckCritical`, etc. Let's 🔥  it!

degemer:
Nitpicks: 
- it could be nice to add tests cases for not existing hosts for IIS and Windows Service (to check the exception returned)
- same with WMI, what happens when we query a non-existing metric ?

👏 🍰 💯 👍  for this PR!

olivielpeau:
Thanks @degemer for the review! I've addressed your comments, could you give this another 👀 ?

tests/checks/integration/test_windows_service.py
```diff
@@ -0,0 +1,29 @@
...
+
+# project
+from tests.checks.common import AgentCheckTest
+
+INSTANCE = {
+    'host': '.',
+    'services': ['EventLog', 'Dnscache', 'NonExistingService'],
+}
+
+INVALID_HOST_INSTANCE = {
```
degemer: I think you forgot to add the corresponding test case 😄

olivielpeau: :facepalm: will fix, thanks for spotting this

olivielpeau: Ok, should be good now :)

degemer:
:shipit:

PR #2727, opened 2016-08-02
[process] Use configured proc path when getting pagefault stats

rdxiang:
### What does this PR do?

Uses the configured proc path (`psutil.PROCFS_PATH` which is set during the check's initialization) when getting pagefault stats instead of `/proc`. 

### Motivation

I found https://github.com/DataDog/dd-agent/pull/2482 (makes proc path configurable) after wanting to run the process integration in docker with, and noticed the page fault stats still used '/proc' which means the integration wouldn't be able to get accurate stats if proc path was something else.

### DIFF
--- a/checks.d/process.py
+++ b/checks.d/process.py
@@ -276,9 +276,10 @@ class ProcessCheck(AgentCheck):
 
         # http://man7.org/linux/man-pages/man5/proc.5.html
         try:
-            data = file_to_string('/proc/%s/stat' % pid)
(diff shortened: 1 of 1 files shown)


degemer:
Thanks for the PR @rdxiang ! I think you're right and we forgot this, but I'll ask @truthbk who worked on this to be sure.

truthbk:
@rdxiang @degemer totally an oversight on my part. Thank you so much for addressing this 👏 - merging! :shipit:

Issue #2722, opened 2016-08-01
[kubernetes] add more metrics related to k8s deployments


onbjerg:
Right now we can only filter based on replication controllers using `kube_replication_controller`. It is my understanding that ReplicaSet are also included in this.

It would be a lot better if you could filter based on `kube_replica_set` and/or `kube_deployment`. Specific metrics for these could prove valuable for monitoring, e.g. the number of pods running for a specific ReplicaSet/Deployment vs. the desired count. All of these values can be extracted from the k8s API.

Another nice addition would be the number of available and unavailable pods. A pod _can_ be running, but not be available. A pod is only available when it's readiness probe succeeds, otherwise it is unavailable. These metrics could be called `kubernetes.pods.available` and `kubernetes.pods.unavailable` respectively.

My proposal is to add the following:
- `kubernetes.pods.desired` (for ReplicaSet, Deployments and ReplicationControllers)
- `kubernetes.pods.available`
- `kubernetes.pods.unavailable`
- `kube_replica_set` or/and `kube_deployment` tags

The addition of these metrics and tags would improve the overall experience with the k8s integration.

onbjerg:
I might work on this if I have time later this week and if the proposal is 👍

onbjerg:
Maybe we could also pull some actual data from the pods, like their mem/CPU requests and limits.

onbjerg:
The deployment statuses can be found in ``/apis/extensions/v1beta1/deployments``.

jonmoter:
+1 on kube_deployment tags

onbjerg:
Waiting for merge #2728

PR #2728, opened 2016-08-02
[kubernetes] add metrics for k8s deployments

onbjerg:
### What does this PR do?

Adds the following metrics for k8s deployments
- `kubernetes.pods.desired`
- `kubernetes.pods.available`
- `kubernetes.pods.unavailable`

And tags these metrics with 
- `kube_deployment:{name}`

Additionally also tags pods with their deployment name, if any (as described in #2722)
### Motivation

I wanted to be able to check the health of my deployments in Datadog using the number of available vs desired pods in my k8s cluster. At the moment, this is not possible.
### Additional Notes

Not that I can think of 🙈

### DIFF
--- a/tests/checks/mock/test_kubernetes.py
+++ b/tests/checks/mock/test_kubernetes.py
@@ -70,7 +70,7 @@ class TestKubernetes(AgentCheckTest):
             with mock.patch('utils.kubeutil.KubeUtil.retrieve_deployments_list', side_effect=lambda: json.loads(Fixtures.read_file("deployments_list.json", string_escape=False))):
                 with mock.patch('utils.dockerutil.DockerUtil.get_hostname', side_effect=lambda: 'foo'):
                     # Can't use run_check_twice due to specific metrics
-                    self.run_check(config, mocks=mocks, force_reload=True)
+                    self.run_check(config, force_reload=True)
                     self.assertServiceCheck("kubernetes.kubelet.check", status=AgentCheck.CRITICAL, tags=None, count=1)
 
     @mock.patch('utils.kubeutil.KubeUtil.retrieve_json_auth')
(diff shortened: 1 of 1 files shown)


therc:
This is great. Should it also export availability and unavailability as a percentage? Sometimes you have deployments that are very different in size, but you want to create one alerting rule that covers them all (e.g. no more than 10% unavailable). 

I'm not very familiar with DD alerting yet, but there are further additions that could be made to improve things, e.g. to reduce false alarms.

onbjerg:
@therc You can create more complex metrics in DD using multiple metrics, so in essence you can get the percentage figure yourself. I did this with memory in my k8s cluster.

onbjerg:
Tests are failing because some dependencies won't be installed in one of the runs. Mocks test is fine, so we should be good?

irabinovitch:
CC: @remh @hkaj @masci for feedback.

@onbjerg Thanks for putting this together, this looks like a fantastic addition to our kubernetes support. Especially when combined with what we're adding in #2551. We have a few of our k8s focused folks out over the next few days, but we'll jump on this upon their return.

onbjerg:
@irabinovitch Cool! Any chance you could tell me why the tests are failing?

therc:
@onbjerg I see what you mean, but I'm curious which memory metrics you used for numerator and denominator. I can't find a pair of good ones, e.g. Kubernetes usage vs limit per container. Unless you used system.*. ...

And I see that you can use that for alerts, too. Cool.

irabinovitch:
@onbjerg tests look to be passing.

onbjerg:
@therc Hi! I didn't choose the memory metrics, I just tagged the existing ones. AFAIK the memory metrics are pulled from cAdvisor on the ``/api/v1.3/subcontainers/`` endpoint.

therc:
Is this going to be rebased?

onbjerg:
Sure, give me a moment.

onbjerg:
Alright, I've only just now had time to actually do it. It might take a while, there's loads of merge conflicts because of #2551..

onbjerg:
Uhh

onbjerg:
My [branch](https://github.com/onbjerg/dd-agent/commits/feature/k8s-deployments) looks clean, very weird.

onbjerg:
Tests are failing now. My bad, I'll look at it when I get home.

onbjerg:
Now the tests are failing because of a ``Connection aborted`` during ``test_events``. Doesn't seem like this is due to the PR. Any insights @masci or @irabinovitch?

therc:
I'd be happy to test an updated version of this PR. We're getting very close, with all the recent improvements in Kubernetes monitoring...

onbjerg:
@therc Damn it, I already rebased once 😁  I'll have a look.

therc:
Actually, I'm wondering if #2890 makes this redundant...

masci:
Somehow redundant but notice #2890 doesn't work if you don't run `kube-state-metrics` within your cluster. We're trying to keep things modular so one can choose the best strategy according to their own environment

hkaj:
I still think it's a good one as it doesn't rely on the kube state api that is optional. Being able to get this info only based on core components is good.

irabinovitch:
@hkaj Thoughts on next steps here?

PR #2729, opened 2016-08-03
Allow `connection_timeout` to be set for MySQL checks

scottgeary:
### What does this PR do?

Allow a connection timeout to be used on when connecting to MySQL instances.

### Motivation

When monitoring more than one instance of MySQL, if the host becomes unavailable, it'll block all the other MySQL instance checks. This block can be longer than the polling period, and you get false alerts and service checks.

Use case: Monitoring many RDS instances from a single host. A failover event occurs, causing connection issues to a single box temporarily. This connection hang (while the Master is being rotated), also blocks all the other monitored instances. 

### Testing Guidelines

Unsure how to a reasonable test case. Other than pull a remote host down, or disable networking.

### Additional Notes

Defaults to the original no-timeout '[None](https://github.com/PyMySQL/PyMySQL/blob/pymysql-0.6.6/pymysql/connections.py#L822)' behaviour. When the original behaviour is used, the OS socket defaults on [sock.settimeout()](https://docs.python.org/2/library/socket.html#socket.socket.settimeout) are used. Which could be a long time.

Also, `read_timeout` + `write_timeout` could also occur (once connected, and issuing SQL), but the bundled version of `pymysql` (v0.6.6) isn't new enough to contain these parameters. ([v0.7.3](https://github.com/PyMySQL/PyMySQL/blob/397fb74cf47232d2f388d013778cf0ff967618e1/CHANGELOG#L34) required).

### DIFF
--- a/checks.d/mysql.py
+++ b/checks.d/mysql.py
@@ -375,7 +375,11 @@ class MySql(AgentCheck):
             ssl = dict(ssl) if ssl else None
 
             if defaults_file != '':
-                db = pymysql.connect(read_default_file=defaults_file, ssl=ssl)
+                db = pymysql.connect(
+                    read_default_file=defaults_file,
+                    ssl=ssl,
+                    connect_timeout=connect_timeout
+                )
             elif mysql_sock != '':
                 self.service_check_tags = [
                     'server:{0}'.format(mysql_sock),
--- a/checks.d/mysql.py
+++ b/checks.d/mysql.py
@@ -294,7 +294,7 @@ class MySql(AgentCheck):
             procfs_path = self.agentConfig.get('procfs_path', '/proc').rstrip('/')
             psutil.PROCFS_PATH = procfs_path
 
-        host, port, user, password, mysql_sock, defaults_file, tags, options, queries, ssl = \
+        host, port, user, password, mysql_sock, defaults_file, tags, options, queries, ssl, connect_timeout = \
             self._get_config(instance)
 
         self._set_qcache_stats()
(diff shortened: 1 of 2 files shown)


checks.d/mysql.py
```diff
@@ -330,9 +330,10 @@ def _get_config(self, instance):
         options = instance.get('options', {})
         queries = instance.get('queries', [])
         ssl = instance.get('ssl', {})
+        connect_timeout = instance.get('connect_timeout', None)
```
degemer: Should `connect_timeout`  be cast as an int ?

degemer:
Thank you for the PR @scottgeary ! I added one comment, and I was also wondering why you didn't add the `connect_timeout` to `pymysql.connect(read_default_file=defaults_file, ssl=ssl)`: is it already part of the config file ?

checks.d/mysql.py
scottgeary: If this is a question of what the default setting _should_ be:
- Setting it to `None` preserves the original check behaviour (leaving it to system socket defaults).
- Explicitly setting it to `1 second` changes the current behaviour, and I'm not sure that assuming 1 second would be appropriate for everyone.

The [actual code](https://github.com/PyMySQL/PyMySQL/blob/pymysql-0.6.6/pymysql/connections.py#L815) that changes the [socket timeout](https://docs.python.org/2/library/socket.html#socket.socket.settimeout) accepts any value types `integer/float` as well as `None`.

degemer: I was wondering if we had to cast the `instance.get('connect_timeout')` to an int when it wasn't `None`, but it doesn't seem to be needed.

We have to keep the default to `None` for backward compatibility, even if it would probably better to set one at 5 seconds or so. Thanks for the details!

degemer:
Thanks again for the PR, it's merged now and will be part of the next minor release!

Issue #2732, opened 2016-08-04
disk.py reports missing attribute


kangelos:
Stock installation under Centos 6.7
disk collector breaks with
File "/root/.datadog-agent/agent/checks.d/disk.py", line 270, in _list_devices#012    flattened_devices = self._flatten_devices(raw_devices)#012  File "/root/.datadog-agent/agent/checks.d/disk.py", line 250, in _flatten_devices#012    elif previous and self._is_number(parts[0]):#012AttributeError: 'Disk' object has no attribute '_is_number'

degemer:
Thank you for the report @kangelos !

Would you mind running `sudo -u dd-agent df -Tk` and either post it there or send it to support at datadoghq.com ?

We think we have a solution, but we'd like to be sure before merging it. 
Let us know if you have any question!

kangelos:
Just in case you miss the email	

```

df -Tk
Filesystem Type 1K-blocks Used Available Use% Mounted on
/dev/sda3 ext4 93421512 70963180 17706052 81% /
tmpfs tmpfs 32969616 0 32969616 0% /dev/shm
/dev/sda1 ext4 512752 90460 395412 19% /boot
10.71.27.253:/vol/ads
nfs 1020054752 56080768 963973984 6% /ads
```

degemer:
Thanks @kangelos ! I updated my PR thanks to this, it should work now.

If you want to give it a try, you can download the `disk.py` on my branch: https://raw.githubusercontent.com/DataDog/dd-agent/cf9928120c36318b0f597dacab2c937c8b08f70c/checks.d/disk.py, put it in place of ` /root/.datadog-agent/agent/checks.d/disk.py`, and restart the agent.

I used your `df -Tk` output for a test in https://github.com/DataDog/dd-agent/pull/2733/files (I changed the ip address and the mountpoint), is it okay with you ?

kangelos:
Hi Quentin 

your patch is in and it looks good, 

no more complaints from disk.py 

Regards 

.A.

On 2016-08-04 23:36, Quentin Madec wrote:

> Thanks @kangelos [1] ! I updated my PR thanks to this, it should work now. 
> 
> If you want to give it a try, you can download the disk.py on my branch: https://raw.githubusercontent.com/DataDog/dd-agent/cf9928120c36318b0f597dacab2c937c8b08f70c/checks.d/disk.py, put it in place of /root/.datadog-agent/agent/checks.d/disk.py, and restart the agent. 
> 
[...]

PR #2733, opened 2016-08-04
[disk] handle multilines df output

degemer:
### What does this PR do?

remove missing method `_is_number` in `disk` check.

### Motivation

Fix https://github.com/DataDog/dd-agent/issues/2732

### Additional Notes

The previous method was:
```python
try:
    float(my_var)
except:
    return False
return True
```

### DIFF
--- a/tests/checks/fixtures/disk/centos-df-Tk
+++ b/tests/checks/fixtures/disk/centos-df-Tk
@@ -1,0 +1,6 @@
+Filesystem Type   1K-blocks   Used      Available  Use%  Mounted on
+/dev/sda3  ext4   93421512    70963180  17706052   81%   /
+tmpfs      tmpfs  32969616    0         32969616   0%    /dev/shm
+/dev/sda1  ext4   512752      90460     395412     19%   /boot
+10.1.5.223:/vil/cor
+           nfs     1020054752 56080768  963973984  6%    /cor
--- a/tests/checks/mock/test_disk.py
+++ b/tests/checks/mock/test_disk.py
@@ -168,6 +168,20 @@ class TestCheckDisk(AgentCheckTest):
 
         self.coverage_report()
 
+    @mock.patch('utils.subprocess_output.get_subprocess_output',
+                return_value=(Fixtures.read_file('centos-df-Tk'), "", 0))
+    @mock.patch('os.statvfs', return_value=MockInodesMetrics())
+    def test_no_psutil_centos(self, mock_df_output, mock_statvfs):
+        self.run_check({'instances': [{'use_mount': 'no',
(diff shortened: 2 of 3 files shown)


degemer:
This is not the right way to do it. I'll update once we know more.

degemer:
Updated with a test, it works now!

tests/checks/fixtures/disk/centos-df-Tk
```diff
@@ -0,0 +1,6 @@
+Filesystem Type 1K-blocks Used Available Use% Mounted on
+/dev/sda3 ext4 93421512 70963180 17706052 81% /
+tmpfs tmpfs 32969616 0 32969616 0% /dev/shm
+/dev/sda1 ext4 512752 90460 395412 19% /boot
+10.1.5.223:/vil/cor
+nfs 1020054752 56080768 963973984 6% /cor
```
olivielpeau: Nitpick: the output seems to be aligned by column, it might be good to make sure the fixture has all the extra white spaces this implies (see https://github.com/DataDog/dd-agent/blob/a6812459c9894378a4e5a1a3252cd49c3e9f6679/tests/test_system.py#L39-L51)

tests/checks/mock/test_disk.py
```diff
@@ -168,6 +168,20 @@ def test_no_psutil_freebsd(self, mock_df_output, mock_statvfs):
 
         self.coverage_report()
 
+    @mock.patch('utils.subprocess_output.get_subprocess_output',
+                return_value=(Fixtures.read_file('centos-df-Tk'), "", 0))
+    @mock.patch('os.statvfs', return_value=MockInodesMetrics())
+    def test_no_psutil_centos(self, mock_df_output, mock_statvfs):
+        self.run_check({'instances': [{'use_mount': 'no',
+                                       'excluded_filesystems': ['devfs'],
+                                       'excluded_disk_re': 'zroot/.+'}]},
```
olivielpeau: we could use instance options that make sense for this specific test for `excluded_filesystems` and `excluded_disk_re` (I could think of either having none of these options or the same options as the debian test)

olivielpeau:
Added 2 nitpicks, but overall this looks 💚  !

Feel free to merge once they're addressed

degemer:
Thanks for the review, I addressed both comments, waiting for 🍏 .

tests/checks/fixtures/disk/centos-df-Tk
degemer: Nice 👀

olivielpeau:
💚  (CI failure is caused by an elasticsearch test that fails during setup, unrelated)

Issue #2361, opened 2016-03-17
[postgresql] using peer auth through unix socket


zerc:
Hi, everyone!

It seems like currently this feature doesn't work or I can't understand how I can use it. Look:

* [Strange call](https://github.com/DataDog/dd-agent/blob/master/checks.d/postgres.py#L571-L573) of `pg.connect`:

```python
if host == 'localhost' and password == '':
    # Use ident method
    connection = pg.connect("user=%s dbname=%s" % (user, dbname))
```

So this string goes into the [user kwarg](https://github.com/mfenniak/pg8000/blob/master/pg8000/__init__.py#L42-L44):

```python
def connect(
        user=None, host='localhost', unix_sock=None, port=5432, database=None,
        password=None, ssl=False, timeout=None, **kwargs):
```

Also according to `pg8000` we need to provide a path to `unix_sock` instead of `host`:

```python
:keyword unix_sock:
    The path to the UNIX socket to access the database through, for
    example, ``'/tmp/.s.PGSQL.5432'``.  One of either ``host`` or
    ``unix_sock`` must be provided.
```

And I can't find a way to do this through the agent.

So anyone knows how I can make `peer` connection with the agent or is it really not working now?

Thanks

DylanFrese:
I would also like to be able to use a UNIX Socket to monitor a postgres instance, it would simplify authentication as I wouldn't have to put credentials in the DataDog configuration file (the DataDog configuration is managed by SaltStack, but the Postgres configuration is deployed from another project and differs between server environments). Ideally, `postgres.yaml` would be able to take a `unix_sock` field in place of `host`/`port`.

If nobody wants to pick this up, I'll try creating a PR in the next few days, I just imagine it'll take me longer than someone who's more familiar with the project/Python.

PR #2734, opened 2016-08-04
[postgres] Allow 'host' config value be UNIX socket path

DylanFrese:
The PostgreSQL check has been changed to try to connect to the PostgreSQL database via a UNIX socket if the hostname provided in the configuration begins with a forward-slash ('/').

This behaviour is similar to the 'psql' command-line utility. From `psql(1)`:

```
  --host=hostname
      Specifies the host name of the machine on which the server is
      running. If the value begins with a slash, it is used as the
      directory for the Unix-domain socket.
```

With the difference being that we interpret it as an exact path, whereas psql interprets it as a directory to look for a socket with a particularly formatted name.

This addresses issue #2361. Arguably, a new configuration parameter, `unix_sock` or similar, should be added. I chose to do it this way as it touches less code, and is, as mentioned above, similar to how `psql` does things.

The documentation at `Datadog/documentation/content/integrations/postgresql.md` should also be updated with the new content of `postgres.yaml.example`.

### DIFF
--- a/checks.d/postgres.py
+++ b/checks.d/postgres.py
@@ -584,6 +584,11 @@ SELECT s.schemaname,
                 elif port != '':
                     connection = pg.connect(host=host, port=port, user=user,
                         password=password, database=dbname, ssl=ssl)
+                elif host.startswith('/'):
+                    # If the hostname starts with /, it's probably a path
+                    # to a UNIX socket. This is similar behaviour to psql
+                    connection = pg.connect(unix_sock=host, user=user,
+                        password=password, database=dbname)
                 else:
                     connection = pg.connect(host=host, user=user, password=password,
                         database=dbname, ssl=ssl)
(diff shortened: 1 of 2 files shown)


truthbk:
Thanks for this @DylanFrese code looks good, we'll talk over whether introducing a new `unix_sock` parameter makes sense, or if if this is - as it seems - more than enough. Thanks a bunch!

antifuchs:
I just ran into the problem that this fixes - any chance this could be pulled in sometime? (:

olivielpeau:
Thanks @DylanFrese, just gave your changes a test run and everything works nicely! Keeping `host` makes sense to me too, thanks for the thorough explanation.

Merging, this will be part of the next minor release, `5.10.0`

Travis is failing but I think the current master fixes the failures. I'll keep an eye on the CI once this is merged

PR #2735, opened 2016-08-07
Log URL in kubelet check failures

therc:
This might have saved me a lot of time debugging issues, making it more obvious that I was trying to access "localhost:10255" from within the dd-agent container.


truthbk:
Makes sense - logs are meant to provide actionable information, and this does precisely that. Thanks! 👍

PR #2736, opened 2016-08-08
Network check corrections

tschellhorn:
When looking over the Network Check class, I noticed a few issues in the constructor for the class. This PR enables the duplication check and stops overwriting another method parameter.

### DIFF
--- a/checks/network_checks.py
+++ b/checks/network_checks.py
@@ -74,13 +74,14 @@ class NetworkCheck(AgentCheck):
         # to keep track of statuses
         names = []
         for inst in instances:
-            name = inst.get('name', None)
-            if not name:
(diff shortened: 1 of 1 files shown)


truthbk:
@tschellhorn indeed! Good find, we should've spotted this a long time ago 😊 thank you for the PR.   Could you please squash the commits before we merge?

tschellhorn:
I haven't added test cases yet as I don't have a working test environment setup for running the test cases.

tschellhorn:
After looking at the tests, I didn't see any specific ones for the NetworkCheck class. Let me know if  test cases should be added to the other classes that subclass this class.

truthbk:
Thanks @tschellhorn! Very much appreciated.

PR #2739, opened 2016-08-08
[mongo] Adds metrics for collections

gmmeyer:
### What does this PR do?

This PR adds some additional metrics from collections. It will now collect the following metrics from each collection (from the database that you're collecting from):

```
mongodb.collection.indexSizes
mongodb.collection.size
mongodb.collection.avgObjSize
mongodb.collection.count
mongodb.collection.capped
mongodb.collection.max
mongodb.collection.maxSize
mongodb.collection.storageSize
mongodb.collection.nindexes
```

### Motivation

Collection level metrics, like the amount of space that a collection is taking up, or the average object size, scoped to a specific collection, can be very important. If a certain collection is too large, maybe it should be split up. If the indexes on a collection are getting much, much larger than the others, maybe you want to split the collection. There are a lot of these things that can be important.

### Testing Guidelines

It tests to see whether it works in the most basic sense.

### DIFF
--- a/checks.d/mongo.py
+++ b/checks.d/mongo.py
@@ -335,6 +335,18 @@ class MongoDb(AgentCheck):
         "writeLock.time": GAUGE,
     }
 
+    COLLECTION_METRICS = {
+        'collection.size': GAUGE,
+        'collection.avgObjSize': GAUGE,
+        'collection.count': GAUGE,
+        'collection.capped': GAUGE,
+        'collection.max': GAUGE,
+        'collection.maxSize': GAUGE,
+        'collection.storageSize': GAUGE,
+        'collection.nindexes': GAUGE,
+        'collection.indexSizes': GAUGE,
+    }
+
     """
     Mapping for case-sensitive metric name suffixes.
 
@@ -364,6 +376,7 @@ class MongoDb(AgentCheck):
         'metrics.commands': COMMANDS_METRICS,
         'tcmalloc': TCMALLOC_METRICS,
         'top': TOP_METRICS,
+        'collection': COLLECTION_METRICS,
     }
 
     # Replication states
@@ -393,6 +406,10 @@ class MongoDb(AgentCheck):
 
         # List of metrics to collect per instance
         self.metrics_to_collect_by_instance = {}
+
+        self.collection_metrics_names = []
+        for (key, value) in self.COLLECTION_METRICS.iteritems():
+            self.collection_metrics_names.append(key.split('.')[1])
 
     def get_library_versions(self):
         return {"pymongo": pymongo.version}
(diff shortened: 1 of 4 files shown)


gmmeyer:
The apache failure is not mine, and that's the only failure :)

checks.d/mongo.py
```diff
@@ -355,6 +367,7 @@ class MongoDb(AgentCheck):
         'durability': DURABILITY_METRICS,
         'locks': LOCKS_METRICS,
         'wiredtiger': WIREDTIGER_METRICS,
+        'collection': COLLECTION_METRICS,
```
remh: Does it need to be in the default metrics as well ? Looks like it's processed differently, no ?

checks.d/mongo.py
```diff
@@ -875,3 +887,49 @@ def total_seconds(td):
...
+        # get collection level stats
+        try:
+            # Ensure that you're on the right db
+            db = cli[db_name]
+            # grab the collections from the database
+            coll_names = db.collection_names()
+            collection_metrics = []
+            for m in self.COLLECTION_METRICS:
+                # grab all the collections
+                collection_metrics.append(m.split('.')[1:])
```
remh: Do we need to generate that list at each iteration ? It seems to be a constant to me.

gmmeyer: Yup, I think you're right. I can generate it elsewhere.

checks.d/mongo.py
```diff
@@ -875,3 +887,49 @@ def total_seconds(td):
...
+
+            # loop through the collections
+            for coll_name in coll_names:
+                # grab the stats from the collection
+                stats = db.command("collstats", coll_name)
+                # loop through the metrics
+                for m in collection_metrics:
```
remh: Looking at the the content of COLLECTION_METRICS, collection_metrics will be a list of lists with a single element each.

checks.d/mongo.py
```diff
@@ -875,3 +887,49 @@ def total_seconds(td):
...
+                    coll_tags = tags + ["db:%s" % db_name, "collection:%s" % coll_name]
+                    value = None
+                    try:
+                        # descend into the dict to grab the value we're looking for
+                        for c in m:
```
remh: Looking at the the content of COLLECTION_METRICS, collection_metrics will be a list of lists with a single element each.

So there will be just one element in m

ci/mongo.rb
```diff
@@ -70,6 +70,16 @@ def mongo_rootdir
...
+
+      bar_cmd = 'db.bar.save({1: []}); db.bar.save({})'
+      foo_cmd = 'db.foo.save({1: []}); db.foo.save({}); db.foo.save({1: []});'
+      sh %(#{mongo_rootdir}/bin/mongo\
+           --eval "#{bar_cmd}" 'localhost:37017/test'\
+           >> $VOLATILE_DIR/mongo.log)
+      sh %(#{mongo_rootdir}/bin/mongo\
+          --eval "#{foo_cmd}" 'localhost:37017/test'\
+          >> $VOLATILE_DIR/mongo.log)
+      sleep_for 30
```
remh: why do we need a 30s sleep ?

checks.d/mongo.py
gmmeyer: Yea, with the ones I added, that's correct, but there's others that we could grab from the dict that would have more.

remh:
We should not do it for every collection by default. Can you do something similar to the postgres check where we list the tables we want more metrics on (here it would be the collections not the tables)

degemer:
👍

PR #2740, opened 2016-08-09
[docker_daemon] Add 5 opt-in checks for container & volume counts

parkr:
### What does this PR do?

This pull request adds 5 new metrics:
- `docker.containers.count` – equivalent to `docker ps -a | wc -l`
- `docker.containers.state_dead.count` – count of all dead containers
- `docker.containers.state_exited.count` – count of all exited containers
- `docker.volumes.count` – count of all volumes
- `docker.volumes.dangling.count` – count of all dangling volumes

Each metric has its own toggle in the settings so you can opt in to each on its own.
### Motivation

We've been battling a bit of weirdness with containers and volumes lying around. We use ephemeral docker containers for some operations but found that volumes would be lying around in the "dangling" state, and that some containers weren't properly cleaned up.
### Testing Guidelines

### Additional Notes

I went with one monolithic commit here because the code was too close together for Git to play nicely with cherry-picking the commits. Let me know if that wasn't the right way to go and I can make 5 distinct commits.

### DIFF
--- a/checks.d/docker_daemon.py
+++ b/checks.d/docker_daemon.py
@@ -207,6 +207,11 @@ class DockerDaemon(AgentCheck):
             # Other options
             self.collect_image_stats = _is_affirmative(instance.get('collect_images_stats', False))
             self.collect_container_size = _is_affirmative(instance.get('collect_container_size', False))
+            self.collect_container_count = _is_affirmative(instance.get('collect_container_count', False))
+            self.collect_dead_container_count = _is_affirmative(instance.get('collect_dead_container_count', False))
+            self.collect_exited_container_count = _is_affirmative(instance.get('collect_exited_container_count', False))
+            self.collect_volume_count = _is_affirmative(instance.get('collect_volume_count', False))
+            self.collect_dangling_volume_count = _is_affirmative(instance.get('collect_dangling_volume_count', False))
             self.collect_events = _is_affirmative(instance.get('collect_events', True))
             self.collect_image_size = _is_affirmative(instance.get('collect_image_size', False))
             self.collect_disk_stats = _is_affirmative(instance.get('collect_disk_stats', False))
@@ -260,6 +265,21 @@ class DockerDaemon(AgentCheck):
 
         if self.collect_container_size:
             self._report_container_size(containers_by_id)
+
+        if self.collect_container_count:
+            self._report_container_count_by_state(containers_by_id)
+
+        if self.collect_dead_container_count:
+            self._report_container_count_by_state(containers_by_id, state="Dead")
+
+        if self.collect_exited_container_count:
+            self._report_container_count_by_state(containers_by_id, state="Exited")
+
+        if self.collect_volume_count:
+            self._report_volume_count()
+
+        if self.collect_dangling_volume_count:
+            self._report_volume_count(filters={'dangling': True})
 
         # Collect disk stats from Docker info command
         if self.collect_disk_stats:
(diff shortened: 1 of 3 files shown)


olivielpeau:
Thanks @parkr! We'll review your changes in details soon.

No problem at all about the monolithic commit, it looks clean enough as it is 😃

parkr:
@olivielpeau Thanks! It's been a while since I have done any Python work so pardon the lack of idiomatic Python. I tried to fix it up by using `filter` and so forth where I can, but still wearing my Ruby hat. 😄 I wasn't able to get the tests fixed up yet, but will try to take a look soon. I'll be on vacation next week so it may have to wait until after that.

checks.d/docker_daemon.py
```diff
@@ -479,14 +499,34 @@ def _report_container_size(self, containers_by_id):
...
+        count = 0
+        tags = {}
+        filterlambda = lambda x: not self._is_container_excluded(x) and state is "Any" or container["State"] is state
+        filtered = list(filter(filterlambda, containers_by_id))
+        tags = self._get_tags(filtered[0], PERFORMANCE)
+
+        m_func = FUNC_MAP[GAUGE][self.use_histogram]
+        # Report docker.container.count if state is "Any", otherwise
+        # report docker.container.state_STATE.count
+        suffix = ".state_{}".format(state.lower()) if state is not "Any" else ""
```
remh: any reason not to use a tag instead for the state ?

checks.d/docker_daemon.py
```diff
@@ -249,6 +254,21 @@ def check(self, instance):
         if self.collect_container_size:
             self._report_container_size(containers_by_id)
 
+        if self.collect_container_count:
+            self._report_container_count_by_state(containers_by_id)
+
+        if self.collect_dead_container_count:
```
remh: Would it be better to just report the number of containers by state for all states ? It could even be enabled by default as:
 - I believe the number of different possible states is bounded and small
 - We've already collected all the needed information (in `containers_by_id`)

checks.d/docker_daemon.py
```diff
@@ -479,14 +499,34 @@ def _report_container_size(self, containers_by_id):
...
+        suffix = ".state_{}".format(state.lower()) if state is not "Any" else ""
+        m_func(self, 'docker.container{}.count'.format(suffix), len(filtered), tags=tags)
+
+    def _report_volume_count(self, filters={}):
+        volumes = self.docker_client.volumes(filters=filters)
+        count = len(volumes['Volumes'])
+
+        m_func = FUNC_MAP[GAUGE][self.use_histogram]
+        suffix = '.' + '-'.join(sorted(filters.keys())) if len(filters) is not 0 else ''
+        m_func(self, 'docker.volumes{}.count'.format(suffix), count)
```
remh: Same thing here, should we use a tag instead ?

checks.d/docker_daemon.py
```diff
@@ -249,6 +254,21 @@ def check(self, instance):
...
+            self._report_container_count_by_state(containers_by_id, state="Dead")
+
+        if self.collect_exited_container_count:
+            self._report_container_count_by_state(containers_by_id, state="Exited")
+
+        if self.collect_volume_count:
+            self._report_volume_count()
+
+        if self.collect_dangling_volume_count:
+            self._report_volume_count(filters={'dangling': True})
```
remh: Looks like, dangling volumes will already be collected when you call collect_volume_count line 266.

We should likely reuse that so we don't do 2 API calls to the docker API.

remh:
Thanks a lot @parkr ! That's pretty cool! Could you also add tests for the metrics you are collecting please ?

checks.d/docker_daemon.py
parkr: Hey @remh! No particular reason! I am used to a much more basic version of graphite and tags are new. How would you like to proceed here?

checks.d/docker_daemon.py
parkr: AFAICT, the resulting JSON we receive back does not include information about whether the volume is dangling. We cannot therefore filter based on the response of the first function call.

parkr:
> Could you also add tests for the metrics you are collecting please ?

I just went looking through other PR's to see how this is best accomplished and didn't find recent examples of PR's which add functionality to this check and add tests. I'll take a stab but it might be terrible.

checks.d/docker_daemon.py
hkaj: Hi @parkr 
The way we usually do this is to report `docker.containers.count` several times, each time with a different state tag. This could look like:

```python
# const
STATES = ['created', 'restarting', 'running', 'paused', 'exited', 'dead']
...
def _report_container_count_by_state(self, containers_by_id):
    tags = self._get_tags(filtered[0], PERFORMANCE)
    m_func = FUNC_MAP[GAUGE][self.use_histogram]
    for state in STATES:
        # the filter changes for each state
        filterlambda = lambda x: not self._is_container_excluded(x) and container["State"] is state
        filtered = list(filter(filterlambda, containers_by_id))
        # add the state tag
        tags.append('state:{}'.format(state)
        m_func(self, 'docker.container.count'.format(suffix), len(filtered), tags=tags)
```
This way you can sum the reported values of `docker.container.count` on a graph to get the total count, or split the graph over the `state` tag and have the detailed count.

Let me know if it's still unclear.

(this snippet is not tested and could be optimized)

checks.d/docker_daemon.py
```diff
@@ -491,14 +511,34 @@ def _report_container_size(self, containers_by_id):
...
-
                 m_func(self, 'docker.container.size_rw', container['SizeRw'],
                        tags=tags)
             if "SizeRootFs" in container:
                 m_func(
                     self, 'docker.container.size_rootfs', container['SizeRootFs'],
                     tags=tags)
 
+    def _report_container_count_by_state(self, containers_by_id, state="Any"):
+        count = 0
```
hkaj: looks like it's not used, you can remove it safely

checks.d/docker_daemon.py
```diff
@@ -491,14 +511,34 @@ def _report_container_size(self, containers_by_id):
...
+        tags = {}
+        filterlambda = lambda x: not self._is_container_excluded(x) and state is "Any" or container["State"] is state
```
hkaj: `container["State"]` --> `x["State"]`, or rename x to container and `self._is_container_excluded(x)` --> `self._is_container_excluded(container)`

hkaj:
Hi @parkr 
Do you have the bandwidth to finish this up? Otherwise I can take over. Not rushing you, I just want to make sure this PR gets all the ❤️ it deserves (these are some interesting metrics!).

parkr:
> Do you have the bandwidth to finish this up?

I got stuck trying to write tests for this. Couldn't seem to mock the response from the docker API.

parkr:
#3077, which supersedes this PR, was merged.

PR #2741, opened 2016-08-09
[mysql] Robust innodb engine status check

scottgeary:
### What does this PR do?

Fail gracefully when issuing the query for `SHOW ENGINE INNODB STATUS`: An empty set _can_ be returned, and produces an ERROR.

### Motivation

When monitoring RDS Aurora read instances, although the InnoDB engine is certainly enabled (passing the relevant checks), the engine status query returns an empty set. Presumably only the current Write instance maintains this data. 

The existing code path, issues & attempts to use the result set in the same `try` block. `fetchone()` produces an error on an empty set:

```
      - instance #13 [ERROR]: "'NoneType' object has no attribute '__getitem__'"
          File "/opt/datadog-agent/agent/checks/__init__.py", line 750, in run
          File "/etc/dd-agent/checks.d/mysql.py", line 323, in check
            raise e
        TypeError: 'NoneType' object has no attribute '__getitem__'
```

This PR splits these responsibilities, allowing for a seperate rowcount test and failing gracefully with a WARN. Service health checks then don't alert.

```
      - instance #13 [WARNING]
          Warning: 'SHOW ENGINE INNODB STATUS' returned no data.
```

As you don't really know what instance is currently in which state, it's likely you'll want to always check for these metrics on every host within your Aurora cluster. I can't see any clear way to detect the current instance type from within MySQL: Would have been nice to do it within `_is_innodb_engine_enabled()`.

### Testing Guidelines

Aurora RDS Write + Read instance required. Manually testing query:

Primary/Write instance:
```
mysql> select @@aurora_version;
+------------------+
| @@aurora_version |
+------------------+
| 1.6              |
+------------------+
1 row in set (0.01 sec)

mysql> show engine innodb status;
<ACUTAL ROW DATA REMOVED>
1 row in set (0.00 sec)
```

Read instance:
```
mysql> select @@aurora_version;
+------------------+
| @@aurora_version |
+------------------+
| 1.6              |
+------------------+
1 row in set (0.00 sec)

mysql> show engine innodb status;
Empty set (0.00 sec)
```

### Additional Notes

- [RDS Aurora overview](http://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_Aurora.html)
- [Which DB instance you are connected to](http://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Aurora.BestPractices.html#Aurora.BestPractices.DeterminePrimaryInstanceConnection). `innodb_read_only` _could_ be a preferred approach for detection, but unsure how this is used on regular MySQL.

### DIFF
--- a/checks.d/mysql.py
+++ b/checks.d/mysql.py
@@ -901,7 +901,7 @@ class MySql(AgentCheck):
                          tables (must grant PROCESS): %s" % str(e))
             return {}
 
-        if (cursor.rowcount < 1)
+        if cursor.rowcount < 1:
             # No data from SHOW ENGINE STATUS, even though the engine is enabled.
             # EG: This could be an Aurora Read Instance
             self.warning("'SHOW ENGINE INNODB STATUS' returned no data.")
--- a/checks.d/mysql.py
+++ b/checks.d/mysql.py
@@ -896,13 +896,19 @@ class MySql(AgentCheck):
         try:
             with closing(db.cursor()) as cursor:
                 cursor.execute("SHOW /*!50000 ENGINE*/ INNODB STATUS")
-                innodb_status = cursor.fetchone()
-                innodb_status_text = innodb_status[2]
         except (pymysql.err.InternalError, pymysql.err.OperationalError, pymysql.err.NotSupportedError) as e:
             self.warning("Privilege error or engine unavailable accessing the INNODB status \
                          tables (must grant PROCESS): %s" % str(e))
             return {}
 
+        if (cursor.rowcount < 1)
+            # No data from SHOW ENGINE STATUS, even though the engine is enabled.
+            # EG: This could be an Aurora Read Instance
+            self.warning("'SHOW ENGINE INNODB STATUS' returned no data.")
+            return {}
+
+        innodb_status = cursor.fetchone()
+        innodb_status_text = innodb_status[2]
         results = defaultdict(int)
 
         # Here we now parse InnoDB STATUS one line at a time


checks.d/mysql.py
```diff
@@ -896,13 +896,19 @@ def _get_stats_from_innodb_status(self, db):
...
             with closing(db.cursor()) as cursor:
                 cursor.execute("SHOW /*!50000 ENGINE*/ INNODB STATUS")
-                innodb_status = cursor.fetchone()
-                innodb_status_text = innodb_status[2]
         except (pymysql.err.InternalError, pymysql.err.OperationalError, pymysql.err.NotSupportedError) as e:
             self.warning("Privilege error or engine unavailable accessing the INNODB status \
                          tables (must grant PROCESS): %s" % str(e))
             return {}
 
+        if cursor.rowcount < 1:
```
truthbk: I'm wondering if you get an empty output because the user we're querying with does not have enough privileges to read the DB. I believe `SHOW /*!50000 ENGINE*/ INNODB STATUS` requires the user to have `PROCESS` privileges... That doesn't necessarily change the scope of this PR but is probably worth knowing.

truthbk:
Thanks a lot for this @scottgeary . Overall it makes a lot of sense, I'm just wondering if it's a privileges issue, and if so, if it's best to behave in the manner the PR implements (quite possibly this is still the desired behavior).

checks.d/mysql.py
scottgeary: I've just double checked the privileges that the datadog user has, and can confirm `PROCESS` is included (and no different from the other Aurora cluster members):

```
mysql> show grants;
+------------------------------------------------------------------------------------------------------------------------------+
| Grants for datadog@IP                                                                                                        |
+------------------------------------------------------------------------------------------------------------------------------+
| GRANT PROCESS, REPLICATION CLIENT ON *.* TO 'datadog'@'IP' IDENTIFIED BY PASSWORD '*HASH'                                    |
| GRANT SELECT ON `performance_schema`.* TO 'datadog'@'IP'                                                                     |
+------------------------------------------------------------------------------------------------------------------------------+
2 rows in set (0.01 sec)

mysql> show engine innodb status;
Empty set (0.00 sec)
```

olivielpeau:
Tested these changes on a local non-aurora mysql server and they don't affect the behavior of the check.

I'm all for merging this for 5.10. Later we may want to spend some time to experiment with the privileges of the mysql user/the mysql instance type to see if things could be improved.

remh:
Closing in favor of #2971

PR #2744, opened 2016-08-09
Add elastic search cluster name to the list of tags for all data from…

derekwbrown:
if you have not yet done so.*

### What does this PR do?

Adds the tag cluster_name, with the value retrieved from the Elastic Search cluster.

### Motivation

Feature request.

### Testing Guidelines

### Additional Notes

Adds an order dependency to the code that's not ideal.

### DIFF
--- a/checks.d/elastic.py
+++ b/checks.d/elastic.py
@@ -322,16 +322,25 @@ class ESCheck(AgentCheck):
         health_url, nodes_url, stats_url, pshard_stats_url, pending_tasks_url, stats_metrics, \
             pshard_stats_metrics = self._define_params(version, config.cluster_stats)
 
+        # Load stats data.
+        # This must happen before other URL processing as the cluster name
+        # is retreived here, and added to the tag list.
+
+        stats_url = urlparse.urljoin(config.url, stats_url)
+        stats_data = self._get_data(stats_url, config)
+        if stats_data['cluster_name']:
+            # retreive the cluster name from the data, and append it to the
+            # master tag list.
+            config.tags.append("cluster_name:{}".format(stats_data['cluster_name']))
+            config.service_check_tags.append("cluster_name:{}".format(stats_data['cluster_name']))
+        self._process_stats_data(nodes_url, stats_data, stats_metrics, config)
+
         # Load clusterwise data
         if config.pshard_stats:
             pshard_stats_url = urlparse.urljoin(config.url, pshard_stats_url)
             pshard_stats_data = self._get_data(pshard_stats_url, config)
             self._process_pshard_stats_data(pshard_stats_data, config, pshard_stats_metrics)
 
-        # Load stats data.
-        stats_url = urlparse.urljoin(config.url, stats_url)
-        stats_data = self._get_data(stats_url, config)
-        self._process_stats_data(nodes_url, stats_data, stats_metrics, config)
 
         # Load the health data.
         health_url = urlparse.urljoin(config.url, health_url)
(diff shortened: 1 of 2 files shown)


checks.d/elastic.py
```diff
@@ -504,6 +508,9 @@ def _process_pending_tasks_data(self, data, config):
             self._process_metric(node_data, metric, *desc, tags=config.tags)
 
     def _process_stats_data(self, nodes_url, data, stats_metrics, config):
+        # retreive the cluster name from the data, and append it to the
+        # master tag list.	
+        config.tags.append("cluster_name:{}".format(data['cluster_name']))
```
degemer: Should we also update `config.service_check_tags` ? (or did you exclude them on purpose ?)

degemer: Also: is there always a `cluster_name` key in the `data` dict ?

checks.d/elastic.py
```diff
@@ -322,16 +322,20 @@ def check(self, instance):
         health_url, nodes_url, stats_url, pshard_stats_url, pending_tasks_url, stats_metrics, \
             pshard_stats_metrics = self._define_params(version, config.cluster_stats)
 
+        # Load stats data.
+        # This must happen before other URL processing as the cluster name
+        # is retreived here, and added to the tag list.
+
+        stats_url = urlparse.urljoin(config.url, stats_url)
+        stats_data = self._get_data(stats_url, config)
+        self._process_stats_data(nodes_url, stats_data, stats_metrics, config)
```
degemer: I'm not a huge fan either of modifying `config` after its creation, but I don't think there is another way to do this.
Maybe we could put the config update in the `check` function ?
Something like:
```python
stats_data = self._get_data(stats_url, config)
config.tags.append("cluster_name:{}".format(stats_data['cluster_name']))
self._process_stats_data(nodes_url, stats_data, stats_metrics, config)
```
This way it would be more obvious that it's being changed. What do you think ?

degemer:
Tests are failing:
- flake8 (lint) is not happy about a whitespace: https://travis-ci.org/DataDog/dd-agent/jobs/151064107#L324
- https://github.com/DataDog/dd-agent/blob/master/tests/checks/integration/test_elastic.py has to be updated with the new tags

checks.d/elastic.py
derekwbrown: There should always be the cluster name, but I will add the check.

checks.d/elastic.py
derekwbrown: That would work.

checks.d/elastic.py
```diff
@@ -322,16 +322,24 @@ def check(self, instance):
...
+        # Load stats data.
+        # This must happen before other URL processing as the cluster name
+        # is retreived here, and added to the tag list.
+
+        stats_url = urlparse.urljoin(config.url, stats_url)
+        stats_data = self._get_data(stats_url, config)
+        if stats_data['cluster_name']:
+            # retreive the cluster name from the data, and append it to the
+            # master tag list.
+            config.tags.append("cluster_name:{}".format(stats_data['cluster_name']))
```
degemer: Do you know if it makes sense to add it to  `service_check_tags` ?

degemer:
LGTM apart from the last comment!

checks.d/elastic.py
olivielpeau: My 2 cents: worked a bit on the integration, and to me it makes sense to add it to `service_check_tags` too.

truthbk:
@derekwbrown added the service check tags, I think we're good to merge.

yannmh:
Tests are failing: https://travis-ci.org/DataDog/dd-agent/jobs/158142800

The service checks tags are not added when status is `CRITICAL`.

PR #2747, opened 2016-08-10
[http_check] log error when tags aren't formatted correctly

gzysk8:
### What does this PR do?

Logs an error message when the tags specified in the `http_check.yaml` are not string types
### Motivation

I worked on a support case where the customer's http check was not reporting any metrics or checks. It was unclear why at first since there were no errors and the endpoints were accessible. Upon further investigation, it was determined that the tags were not formatted correctly causing the metrics submission to silently fail. The problem was an extra space after the colon in the `key:value` tag pair as follows:

```
tags:
      - foo: bar
```

The yaml config here is still valid but the extra space causes each element of the tags list to be treated as a dictionary instead of a string, and thus the metrics can't be submitted with these tags. The tags should be formatted with no space between key and value:

```
tags:
      - foo:bar
```

Working on adding a test

### DIFF
--- a/checks.d/http_check.py
+++ b/checks.d/http_check.py
@@ -262,6 +262,17 @@ class HTTPCheck(NetworkCheck):
             running_time = time.time() - start
             # Store tags in a temporary list so that we don't modify the global tags data structure
             tags_list = list(tags)
+
+            # Check for ill-formatted tags that aren't string types and log an error
+            bad_tags = []
+            for tag in tags_list:
+                if type(tag) is not str:
+                    bad_tags.append(tag)
+
+            if bad_tags:
+                self.log.error('Tags not formatted correctly for %s. %s not of type string'
+                               % (addr,str(bad_tags)))
+
             tags_list.append('url:%s' % addr)
             self.gauge('network.http.response_time', running_time, tags=tags_list)
 
--- a/tests/checks/integration/test_http_check.py
+++ b/tests/checks/integration/test_http_check.py
@@ -146,6 +146,14 @@ CONFIG_HTTP_REDIRECTS = {
 
 FAKE_CERT = {'notAfter': 'Apr 12 12:00:00 2006 GMT'}
 
+BAD_TAGS_CONFIG = {
+    'instances': [{
+        'name': 'bad_tags_config',
+        'url': 'http://httpbin.org',
+        "tags": [{"env": "production"}]
+    }]
+}
+
 
 @attr(requires='network')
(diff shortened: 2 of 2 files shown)


truthbk:
@gzysk8 the test that failed is flaky and unrelated to your changed. It may have been fixed in `master` already, so if you rebase (although there are no conflicts) it might improve things and get the CI running. If you could also squash the commites, that'd be great.

checks.d/http_check.py
```diff
@@ -262,6 +262,17 @@ def _check(self, instance):
             running_time = time.time() - start
             # Store tags in a temporary list so that we don't modify the global tags data structure
             tags_list = list(tags)
+
+            # Check for ill-formatted tags that aren't string types and log an error
+            bad_tags = []
+            for tag in tags_list:
+                if type(tag) is not str:
+                    bad_tags.append(tag)
```
truthbk: So up for debate here is what should we do with the bad tags? Should we remove bad tags from the tag list - log the broken tags and submit the metric with the OK tags. Or should we identify the broken tags, log them and not send anything? We're currently identifying the "broken" tags, and submitting `tags_list` regardless of the fact it might contain bad tags. So we should either not make the call, or make the call without the broken tags.

I personally think we should only attempt the submission if there are no broken tags, to guarantee the customer _knows_ that if his check is working it's because the config is fine (and not partly fine).

checks.d/http_check.py
```diff
@@ -262,6 +262,17 @@ def _check(self, instance):
...
+
+            if bad_tags:
+                self.log.error('Tags not formatted correctly for %s. %s not of type string'
```
truthbk: I think this log line will be more meaningful if we print one line per bad tag, in this case we'd print a single line for all bad tags, which might confuse the user.

truthbk:
Hi @gzysk8,

Thanks again for your contribution, always very much appreciated. We're currently transitioning our checks to a new SDK-based model for the agent. An unfortunate side-effect is that we have to impose a freeze on the dd-agent repo. The SDK consists of two repositories https://github.com/DataDog/integrations-core and https://github.com/DataDog/integrations-extras, where all of our checks will reside. This PR would belong in the **[core](https://github.com/DataDog/integrations-core)** repository. Would you be so kind as to re-open the PR there? We have facilitated some tools to help you with the SDK development that should help you get up and running in no time: 
- [sdk guide](http://docs.datadoghq.com/guides/integration_sdk/) 
- [datadog sdk gem]( https://github.com/DataDog/datadog-sdk-testing). 

PR #2749, opened 2016-08-10
[redis] additional Stats metrics

gzysk8:
### What does this PR do?

Adds a couple metrics exposed in the output of the Redis info command.
### Motivation

Feature request from customer.

### DIFF
--- a/checks.d/redisdb.py
+++ b/checks.d/redisdb.py
@@ -53,6 +53,10 @@ class Redis(AgentCheck):
 
         # stats
         'latest_fork_usec':             'redis.perf.latest_fork_usec',
+        'bytes_received_per_sec':       'redis.bytes_received_per_sec',
+        'bytes_sent_per_sec':           'redis.bytes_sent_per_sec',
+        # Note: 'bytes_received_per_sec' and 'bytes_sent_per_sec' are only
+        # available on Azure Redis
 
         # pubsub
         'pubsub_channels':              'redis.pubsub.channels',


olivielpeau:
@gzysk8 Have you been able to check your changes? (in particular, have you been able to confirm that these metrics actually exist in redis' `info` output?)

I've just checked on the latest release of redis (`3.2.4`) and the new keys you've added aren't there. The closest I could find were `total_net_input_bytes` and `total_net_output_bytes`

gzysk8:
@olivielpeau The case I had, the client was using Azure Redis so perhaps they're specific to that? He modified the check himself and got it working - https://datadog.zendesk.com/agent/tickets/60580

Appears those metrics exist in the info from here too - http://stackoverflow.com/questions/25416562/stackexchange-redis-with-azure-redis-is-unusably-slow-or-throws-timeout-errors

checks.d/redisdb.py
```diff
@@ -53,6 +53,8 @@ class Redis(AgentCheck):
 
         # stats
         'latest_fork_usec':             'redis.perf.latest_fork_usec',
+        'bytes_received_per_sec':       'redis.bytes_received_per_sec',
+        'bytes_sent_per_sec':           'redis.bytes_sent_per_sec',
```
olivielpeau: could you add comments next to these 2 lines as a reminder that these metrics are only available on Azure Redis? :)

truthbk:
The metadata should be updated in `dogweb` as well.

PR #2750, opened 2016-08-10
Michael/kafka log all consumers

imdark:
If no consumer_groups is defined in configuration it
will ask zookeeper and get all consumer_groups, topics
and partitions offset and lag.

If the consumer_groups are defined as none:
all offsets in kafka will be used, otherwise none will be used

sim

### DIFF
--- a/checks.d/kafka_consumer.py
+++ b/checks.d/kafka_consumer.py
@@ -32,11 +32,11 @@ class KafkaCheck(AgentCheck):
     def _get_all_consumers_offsets(zk_conn, zk_prefix):
         consumer_offsets = {}
         topics = defaultdict(set)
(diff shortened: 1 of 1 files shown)


gmmeyer:
Thanks a lot for your contribution @imdark! We'll review it soon and either merge it or get back to you with some feedback! 👍

jeffwidman:
Similar to #2781

irabinovitch:
@imdark Do you have any thoughts on the approach in #2781 vs the one taken here?

degemer:
Sorry for the late response @imdark, closing in favour of https://github.com/DataDog/dd-agent/pull/2880.
Thanks again for your contribution, please let us know what you think of #2880!

PR #2751, opened 2016-08-11
[wmi_check][test] Fix flaky `tag_queries` test

olivielpeau:
### What does this PR do?

Fix flaky `tag_queries` test in wmi check integration test.

`CommandLine` is generally empty for the `svchost` processes, which
makes the check fail when used as a target property of `tag_queries`
(for good reasons IMO). Let's use a property that we're sure exists
and is not empty.

Also, make the error message clearer when the target property exists
but is empty.

### Motivation

Fix a flaky test, and have a better error message for missing/empty target properties


degemer:
:shipit:

olivielpeau:
Thanks @degemer!

PR #2745, opened 2016-08-09
[docker_daemon] [ecs] introspection resilience

truthbk:
*Note: Please remember to review the Datadog [Contribution Guidelines](https://github.com/DataDog/dd-agent/blob/master/CONTRIBUTING.md)
if you have not yet done so.*

### What does this PR do?

ECS agent container no longer has the IP address or ports available when we inspect its container to infer the introspection point. The docker gateway, typically at `172.17.0.1` is fine for constructing the url to query the API. We can set it as a parameter in the config file, or we can extract it from the agent's own information.

Also, although rare, if the agent is not run dockerized, we can grab it `@localhost`

### Motivation

Our ECS task reporting broke due to our inability to query the endpoint.

### Testing Guidelines

An overview on [testing](https://github.com/DataDog/dd-agent/blob/master/tests/README.md)
is available in our contribution guidelines.

### Additional Notes

Anything else we should know when reviewing?

PR #2753, opened 2016-08-12
[docker] Fall back to default route if docker hostname is not resolvable

remh:
### What does this PR do?
In case, the agent cannot resolve the docker host hostname, we fall back on the default gateway.

### Motivation

Customer ticket on kubernetes where the agent was unable to resolve the docker host hostname.

### DIFF
--- a/utils/dockerutil.py
+++ b/utils/dockerutil.py
@@ -5,6 +5,8 @@
 # stdlib
 import logging
 import os
+import socket
+import struct
 import time
 
 # 3rd party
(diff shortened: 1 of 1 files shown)


remh:
@truthbk @hkaj to be merged before https://github.com/DataDog/dd-agent/pull/2702 (as it's merging into Haissam's branch)

truthbk:
Related: #2745

utils/dockerutil.py
```diff
@@ -138,11 +140,26 @@ def _get_checks_from_inspect(self, inspect):
...
         except Exception:
-            log.critical("Unable to find docker host hostname")
+            log.critical("Unable to find docker host hostname. Trying default route")
+
+        try:
+            with open('/proc/net/route') as f:
+                for line in f.readlines():
+                    fields = line.strip().split()
+                    if fields[1] == '00000000':
+                        return socket.inet_ntoa(struct.pack('<L', int(fields[2], 16)))
```
truthbk: We might need some fallback for ipv6 only systems...

truthbk:
Other than the ipv6 concern, this looks good to me.

utils/dockerutil.py
remh: This is to be run within a Docker container which doesn't support ipv6 only networks

PR #2757, opened 2016-08-12
Consul net latency

ross:
### What does this PR do?

This PR adds a new option to the consul check to include calculated network latencies based on https://www.consul.io/docs/internals/coordinates.html. Both inter and intra dc metrics are included. The dc to dc metrics are under `consul.net.dc-latency`, include `source_datacenter`/`dest_datacenter` tags, and are host-less. The node to node metrics are under `consul.net.latency`, are emitted per-host, and include a `consul_datacenter` tag.

### Motivation

Previously had access to this data in graphite and found it interesting/useful at times. It's nice to be able to see network latencies between hosts, racks, availability-zones, roles, etc using the host-level tags. 

### Additional Notes

This PR is a follow on to https://github.com/DataDog/dd-agent/pull/2647 which should be merged prior to merging this.

### DIFF
--- a/checks.d/consul.py
+++ b/checks.d/consul.py
@@ -378,31 +378,31 @@ class ConsulCheck(AgentCheck):
         datacenters = self._get_coord_datacenters(instance)
         for datacenter in datacenters:
             name = datacenter['Datacenter']
-            if name != agent_dc:
-                # Not us, skip
-                continue
-            # Inter-datacenter
-            for other in datacenters:
-                other_name = other['Datacenter']
-                if name == other_name:
-                    # Ignore ourself
-                    continue
-                latencies = []
-                for node_a in datacenter['Coordinates']:
-                    for node_b in other['Coordinates']:
-                        latencies.append(distance(node_a, node_b))
-                latencies.sort()
-                tags = main_tags + ['source_datacenter:{}'.format(name),
-                                    'dest_datacenter:{}'.format(other_name)]
-                n = len(latencies)
-                half_n = int(floor(n / 2))
-                if n % 2:
-                    median = latencies[half_n]
-                else:
-                    median = (latencies[half_n - 1] + latencies[half_n]) / 2
-                self.gauge('consul.net.dc.latency.min', latencies[0], hostname='', tags=tags)
-                self.gauge('consul.net.dc.latency.median', median, hostname='', tags=tags)
-                self.gauge('consul.net.dc.latency.max', latencies[-1], hostname='', tags=tags)
+            if name == agent_dc:
+                # This is us, time to collect inter-datacenter data
+                for other in datacenters:
+                    other_name = other['Datacenter']
+                    if name == other_name:
+                        # Ignore ourself
+                        continue
+                    latencies = []
+                    for node_a in datacenter['Coordinates']:
+                        for node_b in other['Coordinates']:
+                            latencies.append(distance(node_a, node_b))
+                    latencies.sort()
+                    tags = main_tags + ['source_datacenter:{}'.format(name),
+                                        'dest_datacenter:{}'.format(other_name)]
+                    n = len(latencies)
+                    half_n = int(floor(n / 2))
+                    if n % 2:
+                        median = latencies[half_n]
+                    else:
+                        median = (latencies[half_n - 1] + latencies[half_n]) / 2
+                    self.gauge('consul.net.dc.latency.min', latencies[0], hostname='', tags=tags)
+                    self.gauge('consul.net.dc.latency.median', median, hostname='', tags=tags)
+                    self.gauge('consul.net.dc.latency.max', latencies[-1], hostname='', tags=tags)
+                # We've found ourself, we can move on
+                break
 
         # Intra-datacenter
         nodes = self._get_coord_nodes(instance)
--- a/checks.d/consul.py
+++ b/checks.d/consul.py
@@ -16,6 +16,8 @@ from checks import AgentCheck
 import requests
 
 
+# More information in https://www.consul.io/docs/internals/coordinates.html,
+# code is based on the snippet there.
 def distance(a, b):
     a = a['Coord']
     b = b['Coord']
--- a/checks.d/consul.py
+++ b/checks.d/consul.py
@@ -365,9 +365,15 @@ class ConsulCheck(AgentCheck):
         if perform_network_latency_checks:
             self.check_network_latency(instance, agent_dc, main_tags)
 
+    def _get_coord_datacenters(self, instance):
+        return self.consul_request(instance, '/v1/coordinate/datacenters')
+
+    def _get_coord_nodes(self, instance):
+        return self.consul_request(instance, 'v1/coordinate/nodes')
+
     def check_network_latency(self, instance, agent_dc, main_tags):
 
-        datacenters = self.consul_request(instance, '/v1/coordinate/datacenters')
+        datacenters = self._get_coord_datacenters(instance)
         for datacenter in datacenters:
             name = datacenter['Datacenter']
             if name != agent_dc:
@@ -397,7 +403,7 @@ class ConsulCheck(AgentCheck):
                 self.gauge('consul.net.dc-latency.max', latencies[-1], hostname='', tags=tags)
 
         # Intra-datacenter
-        nodes = self.consul_request(instance, 'v1/coordinate/nodes')
+        nodes = self._get_coord_nodes(instance)
         for node in nodes:
             node_name = node['Node']
             latencies = []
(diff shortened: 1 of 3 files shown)


ross:
Example of min, median, and 75th percentile latencies for a single node in AWS:

screen shot 2016-08-12 at 1 42 13 pm

truthbk:
Hi @ross, thanks a lot for this. We have something similar shipping with the agent https://github.com/DataDog/go-metro. But this is great if customers have already deployed consul and don't want any of the sniffing overhead.

truthbk:
It's a pretty big contribution, so we'll do our best to review it as diligently as possible. Thanks!

ross:
> Hi @ross, thanks a lot for this. We have something similar shipping with the agent https://github.com/DataDog/go-metro. But this is great if customers have already deployed consul and don't want any of the sniffing overhead.

That looks pretty interesting. Yeah, nice thing about the consul stuff is that it comes for free (once you're running consul.)

> It's a pretty big contribution, so we'll do our best to review it as diligently as possible. Thanks!

Most of the changes are actually part of https://github.com/DataDog/dd-agent/pull/2647 as mentioned above. This one is a entirely new code that's only turned on if configured. I've been running both for a couple weeks now without any problems.

The 2 travis ci failures seem to be unrelated bootstrapping stuff. Majority seemed to pass. I assume you all can re-run them if need be.

truthbk:
Hi @ross there's a small flake issue here: `./checks.d/consul.py:31:7: E111 indentation is not a multiple of four`.

checks.d/consul.py
```diff
@@ -323,3 +361,69 @@ def check(self, instance):
...
+    def check_network_latency(self, instance, agent_dc, main_tags):
+
+        datacenters = self._get_coord_datacenters(instance)
+        for datacenter in datacenters:
+            name = datacenter['Datacenter']
+            if name != agent_dc:
+                # Not us, skip
+                continue
+            # Inter-datacenter
+            for other in datacenters:
```
truthbk: We can probably `break` after this for loop

checks.d/consul.py
```diff
@@ -323,3 +361,69 @@ def check(self, instance):
...
+                                    'dest_datacenter:{}'.format(other_name)]
+                n = len(latencies)
+                half_n = int(floor(n / 2))
+                if n % 2:
+                    median = latencies[half_n]
+                else:
+                    median = (latencies[half_n - 1] + latencies[half_n]) / 2
+                self.gauge('consul.net.dc-latency.min', latencies[0], hostname='', tags=tags)
+                self.gauge('consul.net.dc-latency.median', median, hostname='', tags=tags)
+                self.gauge('consul.net.dc-latency.max', latencies[-1], hostname='', tags=tags)
```
truthbk: I'm going to have to check how this `hostname=''` parameter might play with our backend. I get why you're overriding (because they're not host metrics), but `key:value` tags with an empty value might not be desireable.

truthbk: should we maybe namespace as `consul.net.dc.latency.*` and `consul.net.node.latency`?

tests/checks/mock/test_consul.py
```diff
@@ -370,3 +474,68 @@ def test_new_leader_event(self):
         self.assertEqual(event['event_type'], 'consul.new_leader')
         self.assertIn('prev_consul_leader:My Old Leader', event['tags'])
         self.assertIn('curr_consul_leader:My New Leader', event['tags'])
+
+    def test_self_leader_event(self):
```
truthbk: 💯  great stuff, thank you so much for the tests.

conf.d/consul.yaml.example
```diff
@@ -23,11 +23,25 @@ instances:
       # Whether to perform checks against the Consul service Catalog
       catalog_checks: yes
 
+      # Whether to enable self leader checks. Each agent with this enabled will
+      # watch for itself to become the leader and will emit an event when that
+      # happens. It is safe/expected to enable this on all nodes in a consul
+      # cluster since only the new leader will emit the (single) event. This
+      # flag takes precedence over new_leader_checks.
+      self_leader_check: yes
```
truthbk: As mentioned earlier, I like this a lot better than what we were previously doing. However this would modify the check behavior and could throw some people off. Don't make any changes to this (or the code) just yet - but we might have to (maybe in a quick separate PR).

checks.d/consul.py
```diff
@@ -118,24 +151,30 @@ def _check_for_leader_change(self, instance):
...
-            })
+            # There was a leadership change
+            if perform_new_leader_checks or (perform_self_leader_check and agent == leader):
+                # We either emit all leadership changes or emit when we become the leader and that just happened
+                self.log.info(('Leader change from {0} to {1}. Sending new leader event').format(
+                    self._last_known_leader, leader))
+
+                self.event({
+                    "timestamp": int(datetime.now().strftime("%s")),
+                    "event_type": "consul.new_leader",
```
truthbk: Again, not your fault at all, but `consul.new_leader` should probably go in a constant. It's not your job to cleanup this, so feel free to ignore.

checks.d/consul.py
```diff
@@ -118,24 +151,30 @@ def _check_for_leader_change(self, instance):
...
-                "msg_text": "The Node at {0} is the new leader of the consul datacenter {1}".format(
-                    leader,
-                    agent_dc
-                ),
-                "tags": ["prev_consul_leader:{0}".format(self._last_known_leader),
-                         "curr_consul_leader:{0}".format(leader),
-                         "consul_datacenter:{0}".format(agent_dc)]
-            })
+            # There was a leadership change
+            if perform_new_leader_checks or (perform_self_leader_check and agent == leader):
```
truthbk: If we have `perform_self_leader_check` should we also send an event when we relinquish leadership? Just trying to cover the scenario where perhaps the agent on the new leader, for whatever reason, fails to submit the new leadership event....

checks.d/consul.py
```diff
@@ -102,7 +124,18 @@ def _is_instance_leader(self, instance):
             return False
 
     def _check_for_leader_change(self, instance):
-        agent_dc = self._get_agent_datacenter(instance)
+        perform_new_leader_checks = instance.get('new_leader_checks',
+                                                 self.init_config.get('new_leader_checks', False))
+        perform_self_leader_check = instance.get('self_leader_check',
+                                                 self.init_config.get('self_leader_check', False))
+
+        if perform_new_leader_checks and perform_self_leader_check:
```
ross: Yeah, this was all an attempt to preserve the existing behavior as-is. There would be other options, but it wasn't immediately clear how I'd preserve the old behavior for people who might be relying on it.

checks.d/consul.py
ross: My assumption was that you'd want to run the new way on all leaders so that if another picked up leadership you'd see the message from there. If no leader was elected you'd probably find out about that another way. It's easy enough for the former leader to emit an event when it changes. Downside would be that you'd get two events. Not the end of the world so up to you.

checks.d/consul.py
ross: Not sure what you mean here.

checks.d/consul.py
ross: > I'm going to have to check how this hostname='' parameter might play with our backend. I get why you're overriding (because they're not host metrics), but key:value tags with an empty value might not be desireable.

I've done it quite frequently in any case where the metric doesn't belong to a host, it seems to avoid having the metric tagged with things that make no sense (all the other host-level tags.) Until we did so those tags seemed to be confusing people at times. I tested things out and it all seems to work from the outside. I guess it might be blowing up/causing problems internally or something :grin:





> should we maybe namespace as consul.net.dc.latency.* and consul.net.node.latency

I can make that change if you like.

checks.d/consul.py
truthbk: So, because we're only interested in `agent_dc == name`, I think that after this for loop completes, in line 406, we can probably `break` the outer for loop as there's no reason to waste any more cycles there - the linear search is done. `datacenters` is probably going to be small so there's no big performance gain, that's true.

ross: Ah. It's the other way around. This is saying if the the other datacenter is ourself `continue` and skip it. Otherwise process the the timing (between us and this round's other)

ross: Sorry, too many for loops there. I think I see what you're saying in the first one. I'll rework this in the morning.

checks.d/consul.py
truthbk: Thanks for modifying the namespacing. I checked regarding the blank of the hostname tags. We should be fine. :)

truthbk:
CI failures are a known issue to start the SQL server and unrelated. 👍

PR #2761, opened 2016-08-15
Make sure the filter gets updated with the last timestamp on each

derekwbrown:
if you have not yet done so.*

### What does this PR do?

PR fixes a customer problem whereby the CPU consumed by the Win32 event log check would increase linearly with time.

### Motivation

Customer request

### Testing Guidelines

### Additional Notes

Reset WMI query filters before each run.  Otherwise, all checks are done relative to when the agent was started, resulting in an ever-growing list of events to be parsed.

The WMI filter was attached to the (cached) sampler object.  So, even though the filter was recomputed each time with a new "TimeGenerated >= " string, that was never actually used by the WMI query.  So, each successive query returned any new entries, PLUS any entries that had already been retrieved.  Processing the list then had a linearly increasing time (and space).  Compounding it is the check to make sure the duplicate entries aren't reported, which then caused the (ever growing) list to be parsed again.


truthbk:
This makes sense... I'm wondering if we might have some of the other wmi checks also affected by the caching of the filters. 

This looks good to me. Could you please squash into a single commit? 👍

derekwbrown:
Redoing without additional merges

PR #2763, opened 2016-08-16
[packaging] revert to old hostname sed

degemer:
Since the config file is not the good one yet.
This needs to be reverted once 5.9 is out.


olivielpeau:
@degemer Let's open the PR that reverts this now, on the `5.9.0` milestone, so that we don't forget about it when we release `5.9.0` :)

degemer:
Indeed, I added it, thanks Olivier!

PR #2764, opened 2016-08-16
[core] Uses our own tmp directory

gmmeyer:
### What does this PR do?

Currently, we are using `/tmp` for most of our temporary files, there are situations in which this can cause a problematic permissions issue, as python wants executable privileges on the files in `/tmp`. Instead we're going to use `./run`

### DIFF
--- a/agent.py
+++ b/agent.py
@@ -44,7 +44,8 @@ from config import (
     load_check,
     generate_jmx_configs,
     _is_affirmative,
-    SD_PIPE_NAME
+    SD_PIPE_NAME,
+    setup_temp_dir
 )
 from daemon import AgentSupervisor, Daemon
 from emitter import http_emitter
@@ -471,6 +472,8 @@ def main():
     autorestart = agentConfig.get('autorestart', False)
     hostname = get_hostname(agentConfig)
     in_developer_mode = agentConfig.get('developer_mode')
+
+    setup_temp_dir(agentConfig)
 
     COMMANDS_AGENT = [
         'start',
--- a/agent.py
+++ b/agent.py
@@ -473,7 +473,7 @@ def main():
     hostname = get_hostname(agentConfig)
     in_developer_mode = agentConfig.get('developer_mode')
 
-    setup_temp_dir(agentConfig)
+    setup_temp_dir(agentConfig, proc_name='agent')
 
     COMMANDS_AGENT = [
         'start',
--- a/agent.py
+++ b/agent.py
@@ -44,8 +44,7 @@ from config import (
     load_check,
     generate_jmx_configs,
     _is_affirmative,
-    SD_PIPE_NAME,
-    setup_temp_dir
+    SD_PIPE_NAME
 )
 from daemon import AgentSupervisor, Daemon
 from emitter import http_emitter
@@ -64,6 +63,7 @@ from utils.service_discovery.config_stores import get_config_store
 from utils.service_discovery.sd_backend import get_sd_backend
 from utils.watchdog import Watchdog
 from utils.windows_configuration import get_sdk_integration_paths
+from utils.tmp_files import setup_temp_dir
 
 # Constants
 PID_NAME = "dd-agent"
@@ -473,7 +473,7 @@ def main():
     hostname = get_hostname(agentConfig)
     in_developer_mode = agentConfig.get('developer_mode')
 
-    setup_temp_dir(agentConfig, proc_name='agent')
+    setup_temp_dir(agentConfig, proc_name='collector')
 
     COMMANDS_AGENT = [
         'start',
--- a/config.py
+++ b/config.py
@@ -21,7 +21,6 @@ import string
 import sys
 import traceback
 from urlparse import urlparse
-import tempfile
 
 # 3p
 import simplejson as json
@@ -1387,37 +1386,3 @@ def initialize_logging(logger_name):
     # re-get the log after logging is initialized
     global log
     log = logging.getLogger(__name__)
-
-def setup_temp_dir(agentConfig, proc_name=None):
-    # if it has neither of these, then just drop back
-    if not agentConfig["use_dd_temp_dir"] and not agentConfig["custom_temp_dir"]:
-        return
-
-    if agentConfig["custom_temp_dir"]:
-        temp_dir = agentConfig["custom_temp_dir"]
-
-    if agentConfig["use_dd_temp_dir"]:
-        temp_dir = os.path.join(__file__, '..', 'tmp')
-
-    if proc_name:
-        temp_dir = os.path.join(temp_dir, proc_name)
-
-    temp_dir = os.path.abspath(temp_dir)
-
-    # make sure that it exists
-    if not os.path.exists(temp_dir):
-        # This is being run by every process,
-        # creating a race condition which will throw an OSError
-        # when one tries ot make the folder after the other
-        # We should just pass on an OSError
-        # This should be lessened by adding the proc_name to each tempdir
-        # But, the error should be caught regardless
-        try:
-            # the default is 0777, it should probably be 0666
-            os.makedirs(temp_dir, '0666')
-        except OSError:
-            pass
-
-    # Python will look at this variable first to determine what tempdir to use,
-    # it only uses system tempdirs or looks in envvars if it is not set
-    tempfile.tempdir = temp_dir
--- a/config.py
+++ b/config.py
@@ -1397,7 +1397,7 @@ def setup_temp_dir(agentConfig, proc_name=None):
         temp_dir = agentConfig["custom_temp_dir"]
 
     if agentConfig["use_dd_temp_dir"]:
-        temp_dir = os.path.join(__file__, '..', 'run', 'temp')
+        temp_dir = os.path.join(__file__, '..', 'tmp')
 
     if proc_name:
         temp_dir = os.path.join(temp_dir, proc_name)
--- a/config.py
+++ b/config.py
@@ -21,6 +21,7 @@ import string
 import sys
 import traceback
 from urlparse import urlparse
+import tempfile
 
 # 3p
 import simplejson as json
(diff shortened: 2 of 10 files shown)


truthbk:
This looks good, have you been able to test it out?

degemer:
The only issue I see with that approach is that to really reproduce the agent behaviour when running `dd-agent check` or other commands/utilities we'll have to set the env.
It's not really important, but we'll have to keep this in mind if we see discrepancies between the agent and `dd-agent check`.
I agree with @truthbk, did you have time to test them ?

gmmeyer:
Yea, it works. I got a new CentOS VM, I took away execution permissions from the tmp directory recursively, and it failed. Then, I replaced the supervisor config with the one I have here, and it worked.

gmmeyer:
@degemer I'll be honest, I have no idea why it requires the executable permissions and I don't feel that it should. I have tried to figure it out and have not been able to do so. Do you have any ideas?

It's not a bad idea to keep `run` controlled. Where do you think we should create the temp directory? (I think `./tmp` makes sense.)

We are going to need to be careful with cleaning it each time the agent starts up, because each proc could be using the temp directory. I think it's a good idea, but I'm not sure how to resolve that. Do you have any ideas? We might just need to have different ones for each proc.

degemer:
I didn't find a reason for executable permissions (well, apart from creating creating directories).

Does this happen often ? Making sure this change (creating our own temp directory) doesn't break anything is going to take some time. Couldn't we just suggest the customer to set another temp directory for the `dd-agent` user ? (with writeable rights, and `/opt/datadog-agent/run` could even be used) Not sure if it's easy to do or not, but it'd be way easier for us than adding our own temp directory. What do you think ?

gmmeyer:
I think this is a very reasonable objection. And, additionally, I think that there are some real drawbacks of making our own tmp directory. (Among the biggest is that it won't get auto cleaned when the computer boots.)

I think having fallback error instructions and a supervisor.conf for them to use makes sense. Do you think we should not do this? (Honestly I don't necessarily disagree.)

degemer:
I think it makes sense to put it in a KB, since it's definitely helpful (and required) for some people. Not sure we should even put something in `dd-agent` about it (unless a lot of people encounter the same issue), since it requires an edit of `supervisor.conf` and not `datadog.conf`.

gmmeyer:
That's a good solution. I'll close this PR.

remh:
Reopening

gmmeyer:
We should also add the logging into here as well.

degemer:
If we want to do this by default, could we use `/opt/datadog-agent/tmp` ? This way `/opt/datadog-agent/run` stays "clean" and there's no risk of interfering with the current agent behaviour.

gmmeyer:
Yes, we can do that

gmmeyer:
Should we clean the tmpdir on agent start?

degemer:
It would be nice, but there's a risk of race condition (since we start the 4 processes at the same time). If you find an easy solution, go for it!

gmmeyer:
I'll put some thought into it

gmmeyer:
@degemer what about having supervisord start a small script that cleans it before the agent starts and only starts the agent afterwards. 

The reason I think it's important to clean this out is that most distros clean out the normal tmp directory at various times, usually on reboot and sometimes with a cron script. So, it might make sense for us to do something like this, too.

gmmeyer:
It's possible that we use so few tmp files that this will never be an issue, so we can just check that out, too.

gmmeyer:
The temp directory is now dynamic and is set through the conf file. 

They can set a temp directory via `custom_temp_dir`, or they can use ours (located at `run/temp`) by setting `use_dd_temp_dir` to affirmative.

config.py
```diff
@@ -1299,3 +1308,37 @@ def initialize_logging(logger_name):
     # re-get the log after logging is initialized
     global log
     log = logging.getLogger(__name__)
+
+def setup_temp_dir(agentConfig, proc_name=None):
```
gmmeyer: That would be fine with me

config.py
```diff
@@ -1299,3 +1308,37 @@ def initialize_logging(logger_name):
...
+        # But, the error should be caught regardless
+        try:
+            # the default is 0777, it should probably be 0666
+            os.makedirs(temp_dir, '0666')
+        except OSError:
+            pass
+
+    # Python will look at this variable first to determine what tempdir to use,
+    # it only uses system tempdirs or looks in envvars if it is not set
+    tempfile.tempdir = temp_dir
```
gmmeyer: I think that also makes sense

agent.py
```diff
@@ -465,6 +466,8 @@ def main():
     hostname = get_hostname(agentConfig)
     in_developer_mode = agentConfig.get('developer_mode')
 
+    setup_temp_dir(agentConfig, proc_name='agent')
```
gmmeyer: yes we can

config.py
```diff
@@ -1299,3 +1308,37 @@ def initialize_logging(logger_name):
...
+    if not os.path.exists(temp_dir):
+        # This is being run by every process,
+        # creating a race condition which will throw an OSError
+        # when one tries ot make the folder after the other
+        # We should just pass on an OSError
+        # This should be lessened by adding the proc_name to each tempdir
+        # But, the error should be caught regardless
+        try:
+            # the default is 0777, it should probably be 0666
+            os.makedirs(temp_dir, '0666')
```
gmmeyer: good idea

config.py
gmmeyer: The rationale for keeping it this way is because we don't want to have to handle the branching config logic outside the function. It makes sense to keep it contained here so we don't have to repeat ourselves 4 times. However, it also does make sense to make the function signature clearer.

I think that it's better to have it contained, but if you can think of a way to make it better I'm down to hear that

utils/tmp_files.py
```diff
@@ -0,0 +1,40 @@
...
+        # creating a race condition which will throw an OSError
+        # when one tries ot make the folder after the other
+        # We should just pass on an OSError
+        # This should be lessened by adding the proc_name to each tempdir
+        # But, the error should be caught regardless
+        try:
+            # the default is 0777, it should probably be 0755
+            os.makedirs(temp_dir, 0755)
+        except OSError:
+            pass
```
degemer: Should we instead revert to the default behavior if the path still doesn't exist ? (i.e. using system default)

utils/tmp_files.py
```diff
@@ -0,0 +1,40 @@
...
+def setup_temp_dir(agentConfig, proc_name=None):
+    # if it has neither of these, then just drop back
+    if not agentConfig["use_dd_temp_dir"] and not agentConfig["custom_temp_dir"]:
+        return
+
+    if agentConfig["custom_temp_dir"]:
+        temp_dir = agentConfig["custom_temp_dir"]
+
+    if agentConfig["use_dd_temp_dir"]:
+        temp_dir = os.path.join(__file__, '..', 'tmp')
```
degemer: Maybe you meant `os.path.join(os.path.basename(__file__), '..', '..', 'tmp')` ? (since this was in util.py previously?)

config.py
degemer: Yep, keeping the branching inside is fine, I'm just not comfortable with passing the whole agentConfig (I find it harder to understand at a glance what setup_temp_dir does).
What do you think of `def setup_temp_dir(use_dd_temp_dir, custom_temp_dir, proc_name=None):` and pass it as `setup_temp_dir(agentConfig['use_dd_temp_dir'], agentConfig['custom_temp_dir'], proc_name='forwarder')` ?

utils/tmp_files.py
gmmeyer: This happens if the path does exist. makedirs is mkdir -p

utils/tmp_files.py
gmmeyer: yes, I did

config.py
```diff
@@ -560,6 +561,14 @@ def get_config(parse_args=True, cfg_path=None, options=None, can_query_registry=
         if config.has_option("Main", "gce_updated_hostname"):
             agentConfig["gce_updated_hostname"] = _is_affirmative(config.get("Main", "gce_updated_hostname"))
 
+        agentConfig["use_dd_temp_dir"] = False
+        if config.has_option("Main", "use_dd_temp_dir"):
+            agentConfig["use_dd_temp_dir"] = _is_affirmative(config.get("Main", "use_dd_temp_dir"))
+
+        agentConfig["custom_temp_dir"] = None
+        if config.has_option("Main", "custom_temp_dir"):
+            agentConfig["custom_temp_dir"] = config.get("Main", "custom_temp_dir")
```
gmmeyer: We do this pattern in other places, though. Is it useful anywhere?

gmmeyer:
@degemer that's not good. We should figure out what's making those files

gmmeyer:
@degemer do you know where those calls came from?

degemer:
I don't know, I didn't track them down.
I still find this solution over-engineered (it's my fault), so I'd like to take a step back and rethink it.

What we want:
- being able to run on Cloudfoundry and systems with restricted access to `/tmp`.

What we (well, I) don't want:
- use a custom temp directory by default. Reasons: it's new, it's not really recommended, and we don't really need for 99% of our use cases (well, even 100% currently).

So given what we don't want, we have two solutions:
- put it in datadog.conf, what we (well, you)'re currently doing. It's a pain, adds a lot of code and seems hard to do well.
- put it in supervisor.conf. We could have a separate supervisor.conf with this setting for Cloudfoundry, and make it available to other customers. It could become annoying for them as this file is overwritten every time the agent is upgraded.

I'm obviously heavily biased towards the second solution, but I thought that before trying to chase down those calls we could check another solution.

What do you think?

gmmeyer:
So, honestly, we no longer need this for cloud foundry.

degemer:
Should we close this PR then?

If people stumble on this PR looking for a way to change the temp dir, easy (and initial) solution is to add `,TMPDIR=/path/to/custom/temp/dir` in `/etc/dd-agent/supervisor.conf` at the end of this line https://github.com/DataDog/dd-agent/blob/bd3f366f23807d8f3eeaf2654269e7b5c98205d2/packaging/supervisor.conf#L31

gmmeyer:
Yeah we can. Let's do that.

PR #2772, opened 2016-08-19
[collector][logging] Log check starts with info level on first run only

olivielpeau:
### What does this PR do?

Lower log level of the `Running check xxxx` message after the first run of the collector.

### Motivation

Since we removed the LaconicFilter (see #2605) the collector is filled
with `Running check xxxx` messages that get logged for every check
at every run. It fills up the logs too quickly for an info-level logging.

### Testing

`rake run` on my dev env, looked at the collector logs: it works

### DIFF
--- a/checks/collector.py
+++ b/checks/collector.py
@@ -380,12 +380,15 @@ class Collector(object):
             if res:
                 metrics.extend(res)
 
+        # Use `info` log level for some messages on the first run only, then `debug`
+        log_method = log.info if self._is_first_run() else log.debug
+
         # checks.d checks
         check_statuses = []
         for check in self.initialized_checks_d:
             if not self.continue_running:
                 return
-            log.info("Running check %s" % check.name)
+            log_method("Running check %s" % check.name)
             instance_statuses = []
             metric_count = 0
             event_count = 0
--- a/checks/collector.py
+++ b/checks/collector.py
@@ -381,14 +381,14 @@ class Collector(object):
                 metrics.extend(res)
 
         # Use `info` log level for some messages on the first run only, then `debug`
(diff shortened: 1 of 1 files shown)


checks/collector.py
```diff
@@ -380,12 +380,15 @@ def run(self, checksd=None, start_event=True, configs_reloaded=False):
...
+        # Use `info` log level for some messages on the first run only, then `debug`
+        log_method = log.info if self._is_first_run() else log.debug
+
         # checks.d checks
         check_statuses = []
         for check in self.initialized_checks_d:
             if not self.continue_running:
                 return
-            log.info("Running check %s" % check.name)
+            log_method("Running check %s" % check.name)
```
degemer: Could you also fix the `%` and use the log interpolation ? `log_method("Running check %s, check.name)`

checks/collector.py
```diff
@@ -380,12 +380,15 @@ def run(self, checksd=None, start_event=True, configs_reloaded=False):
             if res:
                 metrics.extend(res)
 
+        # Use `info` log level for some messages on the first run only, then `debug`
+        log_method = log.info if self._is_first_run() else log.debug
```
degemer: We should probably have a method `log_at_first_run`, unless this is the only thing doing this ?

degemer:
Nice catch ! 🍰

checks/collector.py
olivielpeau: it's the only thing doing this right now so having a method might be overkill, but `log_at_first_run` is definitely a better name, I'll change that

olivielpeau:
Thanks @degemer for the review! Addressed your comments, let me know what you think :)

degemer:
LGTM, :shipit:

PR #2773, opened 2016-08-19
[docker_daemon] filters were including filtering  into regex compiled…

truthbk:
if you have not yet done so.*

### What does this PR do?

There's an issue with the way we include/exclude containers. The current logic would include the tag key in the regex by doing `re.compile(rule)` and therefore not match.

### Motivation

Bugfix.

### Testing Guidelines

### Additional Notes

… expression.


truthbk:
Ill advised PR. The extractors were doing their thing, and the logic as is is _not_ broken. Closing PR.

PR #2646, opened 2016-06-30
[flare] hide multiple endpoints api_keys

degemer:

PR #2774, opened 2016-08-19
[core] easier config for multiple endpoints

degemer:
### What does this PR do?

Change multiple endpoints config.

### Motivation

Clearer config.

### Additional Notes

The previous configuration was done this way because of legacy
issues/conflicts that have been fixed since.

### DIFF
--- a/datadog.conf.example
+++ b/datadog.conf.example
@@ -18,6 +18,8 @@ dd_url: https://app.datadoghq.com
 # The Datadog api key to associate your Agent's data with your organization.
 # Can be found here:
 # https://app.datadoghq.com/account/settings
+# This can be a comma-separated list of api keys.
+# (default: None, the agent doesn't start without it)
 api_key:
 
 # Force the hostname to whatever you want. (default: auto-detected)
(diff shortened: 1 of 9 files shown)


datadog.conf.example
```diff
@@ -97,17 +99,6 @@ gce_updated_hostname: yes
...
-# Multiple endpoints/api_keys
-# If you want to send your data to multiple accounts,
-# add your extra api_keys here in a comma-separated list
-# (default: disabled)
-# other_api_keys: my_apikey, my_other_apikey
-# If you want to send your data to multiple endpoints,
-# add your extra api_keys here in a comma-separated list.
-# For each extra dd_url, an api_key must be specified in
-# other_api_keys (default: disabled)
-# other_dd_urls: https://app.example.com
```
olivielpeau: Should we have the same explanation under `dd_url` now?

olivielpeau:
Looks nice and clean! 👍 

FYI `flare` also needs to be updated accordingly (c.f. #2646)

Also, not directly related to this PR, but the multiple endpoints feature introduced a small change in the behavior of the forwarder: previously the forwarder didn't submit any payload if the  `api_key` were not set (i.e. empty string), whereas now the forwarder still tries to submit the payloads with an empty api key (even though it's useless to do so). Was the change made on purpose? I don't think it's really an issue but it'd be nice to not submit payloads at all when the `api_key` for the endpoint is empty.

datadog.conf.example
degemer: I don't think so since we're the only ones likely to use it. I only put it there because it was illogical to explain `other_api_keys` without explaining `other_dd_urls`

PR #2776, opened 2016-08-22
included queue name as a tag in mapreduce check and made including ap…

johan-g:
Added queuename as a tag to metric and made including app/job name configurable as including it makes datadog interface useless due to too many tags


masci:
Howdy @johan-g and thanks for your contribution!
While waiting for someone to review the PR would you mind to fix the tests? It should be a matter of making the testsuite aware of the new tag.
Thanks!

johan-g:
will do

PR #2779, opened 2016-08-22
[php5_fpm] allow Host header override

allixsenos:
### What does this PR do?

This PR allows you to override the Host HTTP header sent during requests to the fpm status and ping URLs.

### Motivation

In load-balanced configurations it's possible an FPM pool is only mapped to http://production.url/status and not http://localhost. But production.url points to some load balancer which distributes load across many different servers.

To be able to test local servers you need to be able to say "Connect to localhost, but tell it you're after production.url"

### Additional Notes

This change adds the Host HTTP header if defined, otherwise the check works as is. The check has integration tests, and from them I was unable to figure out how to make a check that demonstrates a failing and a working check based on differing Host headers.

### DIFF
--- a/checks.d/php_fpm.py
+++ b/checks.d/php_fpm.py
@@ -42,6 +42,7 @@ class PHPFPMCheck(AgentCheck):
         password = instance.get('password')
 
         tags = instance.get('tags', [])
+        http_host = instance.get('http_host')
 
         if user and password:
             auth = (user, password)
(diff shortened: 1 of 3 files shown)


hkaj:
Howdy @allixsenos 
Thank you for the contribution!
It's a legit use case and the change set is reasonable, I'll merge it and it _should_ go out with 5.9.
Thanks again

PR #2780, opened 2016-08-23
Add support for sum histogram_aggregates type

ross:
### What does this PR do?

Support for a new histogram aggregate type, sum, which records the sum of all
values seen during the sampling period. Matches up with the [timing](https://github.com/etsy/statsd/blob/master/docs/metric_types.md#timing) option of the same name in the original statsd. 

### Motivation

Heavy users of statsd and have used `sum` there and went to enable it here and noticed it wasn't an option. 

/cc @jnunemaker

### DIFF
--- a/aggregator.py
+++ b/aggregator.py
@@ -258,13 +258,15 @@ class Histogram(Metric):
         min_ = self.samples[0]
         max_ = self.samples[-1]
         med = self.samples[int(round(length/2 - 1))]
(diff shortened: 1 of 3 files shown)


remh:
Looks good to me. Thanks @ross . Don't forget that this won't be enabled by default and you'll need to enable it in the configuration.

We are about to release our 5.9 agent so this will go out with 5.10.

Thanks again!

PR #1313, opened 2015-01-26
[kafka] Added support for getting all consumer groups

egut:
If no consumer_groups is defined in configuration it
will ask zookeeper and get all consumer_groups, topics
and partitions offset and lag.

If the consumer_groups are defined kafka_consumer will
work as the old way and only view selected consumer_group,
topic and partions offset and lag.

Issue #2611, opened 2016-06-20
Cannot see metric "kafka.consumer_lag" for consumers using "offsets.storage=kafka"


sbrnunes:
Hi

We're trying to set-up the integration between Datadog and Kafka and report metrics for a few consumers that commit the offsets into Kafka (we use "offsets.storage=kafka").

We're able to see metrics for consumers using "offsets.storage=zookeeper", but not for the ones commiting the offsets into Kafka.

We're particularly interested in knowing the consumer lag, which, as far as I know, is reported as "kafka.consumer_lag".

In the logs we can only see the following warning:

`2016-06-20 17:03:48 UTC | WARNING | dd.collector | checks.kafka_consumer(kafka_consumer.py:58) | No zookeeper node at /kafka/consumers/<consumer_group>/offsets/<topic>/<partition>`

Any idea what we might be doing wrong?

degemer:
Hi @sbrnunes !

It's probably a configuration issue in `kafka_consumer.yaml`, we have a blog post which explain in details  how to configure our Kafka check: https://www.datadoghq.com/blog/monitor-kafka-with-datadog/.
Would you mind checking it out to see if it helps you solve your issue ?

jalaziz:
Actually, I believe this is an issue with the check. Unless something changed recently, the agent only checks consumer offsets from ZK and not Kafka. We have the same issue. I've been meaning to fix the check to support Kafka-based offsets, but haven't gotten around to it yet.

degemer:
My bad, you're right @jalaziz. Sorry @sbrnunes, I misunderstood the issue. Our `kafka_consumer`  check was initially written when consumers offsets were only available in Zookeeper, and we didn't update it to support offsets in Kafka.
I added it to our backlog, but feel free to open a PR!

sbrnunes:
Oh, that makes sense now @degemer. I was getting confused about this as I couldn't see, in the codebase, any code pulling the offsets from Kafka.

seeingidog:
Using the combination of [Burrow](https://github.com/linkedin/Burrow) to monitor offsets and [this plugin](https://github.com/packetloop/datadog-agent-burrow) is working for me.

jeffwidman:
The easiest solution would be to add support upstream in `kafka-python`, as tracked here: https://github.com/dpkp/kafka-python/issues/819

alexef:
Also having this problem.

jeffwidman:
[...]

PR #2781, opened 2016-08-23
Make consumer_groups settings optional with kafka_consumer check

dan-blanchard:
### What does this PR do?

Modifies the `kafka_consumer` check to make `consumer_groups` and everything under it optional.

If `consumer_groups` is not in the config file, fall back to reading from ZK.

`consumer_groups` can also just be a list of consumer groups, in which case we will read topics and partitions from ZK.

If topics are specified for a consumer group, partitions are now optional.
### Motivation

Enumerating all of our consumers and partitions was really onerous.
### Additional Notes

This builds on the work in #1313.

### DIFF
--- a/checks.d/kafka_consumer.py
+++ b/checks.d/kafka_consumer.py
@@ -29,15 +29,34 @@ class KafkaCheck(AgentCheck):
         self.kafka_timeout = int(
             init_config.get('kafka_timeout', DEFAULT_KAFKA_TIMEOUT))
 
+    def _get_partitions_for_topic(self, zk_conn, zk_path_partition_tmpl, consumer_group, topic):
+        """Returns all partitions for the topic as found in zookeeper"""
+        zk_path = zk_path_partition_tmpl % (consumer_group, topic)
+        return self._get_children(zk_conn, zk_path, 'partitions')
+
+    def _get_children(self, zk_conn, zk_path, name_for_error):
+        children = []
+        try:
+            children = zk_conn.get_children(zk_path)
+        except NoNodeError:
+            self.log.warn('No zookeeper node at %s' % zk_path)
+        except:
+            self.log.exception('Could not read %s from %s' % (name_for_error, zk_path))
+        return children
+
     def check(self, instance):
-        consumer_groups = self.read_config(instance, 'consumer_groups',
-                                           cast=self._validate_consumer_groups)
+        # Only validate consumer_groups if specified; fallback to zk otherwise
+        if 'consumer_groups' in instance:
+            consumer_groups = self.read_config(instance, 'consumer_groups',
+                                               cast=self._validate_consumer_groups)
         zk_connect_str = self.read_config(instance, 'zk_connect_str')
         kafka_host_ports = self.read_config(instance, 'kafka_connect_str')
 
         # Construct the Zookeeper path pattern
         zk_prefix = instance.get('zk_prefix', '')
-        zk_path_tmpl = zk_prefix + '/consumers/%s/offsets/%s/%s'
+        zk_path_consumer = zk_prefix + '/consumers/'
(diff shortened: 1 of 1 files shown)


dan-blanchard:
A coworker just informed me that you can actually store the offsets in Kafka itself now, so I will expand this to check there as well tomorrow.

jeffwidman:
Big fan of having this merged.

We have thousand of topics / consumers and both are changing frequently. We are in the process of upgrading to Kafka 10, and once that's complete we'll switch to storing the offsets in Kafka since that's far more scalable than storing them in ZK.

dan-blanchard:
> We have thousand of topics / consumers and both are changing frequently. We are in the process of upgrading to Kafka 10, and once that's complete we'll switch to storing the offsets in Kafka since that's far more scalable than storing them in ZK.

Unfortunately I haven't actually gotten a chance to update this PR to fetch the offsets stored directly in Kafka yet.

jeffwidman:
> Unfortunately I haven't actually gotten a chance to update this PR to fetch the offsets stored directly in Kafka yet.

See also https://github.com/DataDog/dd-agent/issues/2611 and https://github.com/dpkp/kafka-python/issues/819. Also, there's another PR here adding kafka support (according to the PR description, I haven't tested it myself): https://github.com/DataDog/dd-agent/pull/2750

checks.d/kafka_consumer.py
```diff
@@ -47,20 +66,37 @@ def check(self, instance):
             # Query Zookeeper for consumer offsets
             consumer_offsets = {}
             topics = defaultdict(set)
+            if not consumer_groups:
```
jeffwidman: This throws:

    UnboundLocalError: local variable 'consumer_groups' referenced before assignment

I fixed by changing to `if not 'consumer_groups'  in instance:` although I'd probably prefer to collapse the check together with the check in line 49 so it's only defined in one place.

checks.d/kafka_consumer.py
```diff
@@ -47,20 +66,37 @@ def check(self, instance):
...
+                    topics = topic_partitions
+                    topic_partitions = {topic: self._get_partitions_for_topic(zk_conn,
+                                                                              zk_path_partition_tmpl,
+                                                                              consumer_group,
+                                                                              topic)
+                                        for topic in topics}
                 for topic, partitions in topic_partitions.iteritems():
                     # Remember the topic partitions that we've see so that we can
                     # look up their broker offsets later
                     topics[topic].update(set(partitions))
```
jeffwidman: This is now throwing:

      File "/opt/datadog-agent/agent/checks/__init__.py", line 771, in run
      File "/opt/datadog-agent/agent/checks.d/kafka_consumer.py", line 90, in check
        topics[topic].update(set(partitions))
    TypeError: list indices must be integers, not str

I'm guessing when the topics are built automatically from ZK above they're built as a list when they should actually be a dict. Afraid I haven't time right now to further debug, but hopefully @dan-blanchard this can put you on the right track.

Looks like something above is passing

jeffwidman:
Hit some errors testing with the following for `kafka_consumer.yml`:

    instances:
      - kafka_connect_str: kafka01:9092
        zk_connect_str: zookeeper01:2181
        zk_prefix: ''

jeffwidman:
I spent most of yesterday and today on this--while the API and general structure are correct, there's several bugs in this implementation. We need this at my workplace, so I refactored it into #2880. Credit still goes to @dan-blanchard for most of the initial work on this.

degemer:
Thank you very much for your contribution @dan-blanchard !

We'll go with https://github.com/DataDog/dd-agent/pull/2880 which is an extension of your work (if I understand correctly).
I'm going to close this PR, please let us know if you have any remark (on https://github.com/DataDog/dd-agent/pull/2880 or on this PR)!

Issue #2755, opened 2016-08-12
pg8000 client doesn't work for our postgres/pgbouncer db setup


sethcleveland:
We'd like to use the postgres custom metrics feature with our postgres db setup, but can't because of exceptions. To resolve this, swapping the pg8000 client with the psycopg2 one works. See below. I basically made a custom checker and overrode the pg.connect method.

```python
# pull in datadog's postgresql AgentCheck
import imp
imp.load_source("datadog_checksd_postgres", "/opt/datadog-agent/agent/checks.d/postgres.py")

import sys
import pprint
import psycopg2
import datadog_checksd_postgres
from datadog_checksd_postgres import PostgreSql

###
# replace pg8000 with psycopg2 because pg8000 does not work with our arch.
# the postgres agent fails with the following exception:
#
# 2016-08-12 15:50:30 UTC | ERROR | dd.collector | checks.postgres(__init__.py:767) | Check 'postgres' instance #0 failed
# Traceback (most recent call last):
#   File "/opt/datadog-agent/agent/checks/__init__.py", line 750, in run
#     self.check(copy.deepcopy(instance))
#   File "/opt/datadog-agent/agent/checks.d/postgres.py", line 674, in check
#     version = self._get_version(key, db)
#   File "/opt/datadog-agent/agent/checks.d/postgres.py", line 306, in _get_version
#     cursor.execute('SHOW SERVER_VERSION;')
#   File "/opt/datadog-agent/embedded/lib/python2.7/site-packages/pg8000/core.py", line 567, in execute
#     self._c.execute(self, "begin transaction", None)
#   File "/opt/datadog-agent/embedded/lib/python2.7/site-packages/pg8000/core.py", line 1684, in execute
#     self.handle_messages(cursor)
#   File "/opt/datadog-agent/embedded/lib/python2.7/site-packages/pg8000/core.py", line 1761, in handle_messages
#     raise self.error
# ProgrammingError: ('ERROR', '26000', 'prepared statement "pg8000_statement_0" does not exist')
datadog_checksd_postgres.pg = psycopg2
# need to rewire these -- InterfaceError routes to reconnect, ProgrammingError logs a message
datadog_checksd_postgres.InterfaceError = InterfaceError
datadog_checksd_postgres.ProgrammingError = ProgrammingError

###
# remove ssl from the kwargs, otherwise PostgreSql's get_connection fails with the following exception
#
[...]

PR #2782, opened 2016-08-23
[postgres] Allow to use psycopg2

remh:
### What does this PR do?

This PR adds a flag in the postgres check to use psycopg2 instead of pg8000

### Motivation

See discussion at #2755 

### Testing Guidelines

Added a test that enables this flag. Also modified existing test to make sure it keeps using pg8000 by default.

### Additional Notes
Some other approaches more "global" could be tempting, but they wouldn't let the possibility to do it on a per instance basis.


gmmeyer:
LGTM 👍

PR #2783, opened 2016-08-23
[mongo] Fixes a mistake in the Mongo Collections Check

gmmeyer:
### What does this PR do?

The previous Pull Request for the Mongo Collections Check, #2739, has a small bug in it. I had forgotten to take out something after I stopped making the collection name an array and made it a string.

### DIFF
--- a/checks.d/mongo.py
+++ b/checks.d/mongo.py
@@ -922,11 +922,11 @@ class MongoDb(AgentCheck):
                             # we tag the index
                             idx_tags = coll_tags + ["index:%s" % idx]
                             submit_method, metric_name_alias = \
-                                self._resolve_metric('collection.%s' % m, metrics_to_collect)
+                                self._resolve_metric('collection.%s' % m, self.COLLECTION_METRICS)
                             submit_method(self, metric_name_alias, val, tags=idx_tags)
                     else:
                         submit_method, metric_name_alias = \
-                            self._resolve_metric('collection.%s' % m, metrics_to_collect)
+                            self._resolve_metric('collection.%s' % m, self.COLLECTION_METRICS)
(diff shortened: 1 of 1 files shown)


degemer:
Missed it when reviewing, LGTM!

checks.d/mongo.py
```diff
@@ -922,11 +924,11 @@ def total_seconds(td):
                             # we tag the index
                             idx_tags = coll_tags + ["index:%s" % idx]
                             submit_method, metric_name_alias = \
-                                self._resolve_metric('collection.%s' % m, metrics_to_collect)
+                                self._resolve_metric('collection.%s' % m, self.COLLECTION_METRICS)
```
degemer: It's always the same metric, but the tag is changing, isn't it ? Because you could put it outside of the `for` loop if it's the case 😛

checks.d/mongo.py
```diff
@@ -913,7 +913,9 @@ def total_seconds(td):
                 # loop through the metrics
                 for m in self.collection_metrics_names:
                     coll_tags = tags + ["db:%s" % db_name, "collection:%s" % coll_name]
-                    value = stats[m]
+                    value = stats.get(m, None)
+                    if not value:
+                        continue
```
degemer: 🍰

degemer:
Small remark, but looks good otherwise!

gmmeyer:
I resolved that nit @degemer

degemer:
:bow:, let's :shipit: then ! (maybe wait for the 💚  CI, as you want)

gmmeyer:
The failures are rabbit and elastic, so, it's just the random stuff we usually get.

PR #2784, opened 2016-08-24
[mysql] Check for null key_buffer_size before dividing by it

### LINKED ISSUE #2777: [mysql] float division by zero
mysql monitoring problem in case of `performance_schema=off` in `[mysqld]` settings
```
    mysql
    -----
      - instance #0 [ERROR]: 'float division by zero'
      - Collected 0 metrics, 0 events & 3 service checks
```
instance options is
```
    options:
        replication: false
        galera_cluster: false
        extra_performance_metrics: false
```
Both agent and mysqld running on same `CoreOS` server (official `mysql:5.7` used)
(on neighbor server same set, but without `performance_schema=off` working fine)

hkaj:
### What does this PR do?

Work around an issue in mysql where if the unit is not specified by the user in the config file for `key_buffer_size` (eg. 4 instead of 4G) the value will be `0` in `GLOBAL VARIABLES`.

### Motivation

fix #2777 
Of course the config should be corrected on the user side but we can avoid failing on that rather easily so let's do it.

### DIFF
--- a/checks.d/mysql.py
+++ b/checks.d/mysql.py
@@ -490,13 +490,15 @@ class MySql(AgentCheck):
         results['Key_buffer_size'] = key_buffer_size
 
         try:
-            key_cache_utilization = 1 - ((key_blocks_unused * key_cache_block_size) / key_buffer_size)
+            # can be null if the unit is missing in the user config (4 instead of 4G for eg.)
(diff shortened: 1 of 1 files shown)


truthbk:
Thanks @hkaj, LGTM! 👍

PR #2752, opened 2016-08-12
Add events Spark Standalone support to the spark agent check

wjsl:
### What does this PR do?

Enabled Spark metrics to be gathered when clusters are provisioned using Spark's Standalone cluster mode. Currently the agent checks works when Spark is running with Apache YARN to provision its resources. This PR re-uses the bulk of the collection logic, but retrofits an adapter to locate application servers via Spark Standalone's APIs.

Events have also been added. The agent can now raise events when both jobs and stages have status transitions (ie, moving from WAITING to RUNNING to COMPLETED or FAILED).

### Testing Guidelines

Run the mock test using `nosetests`. Note that the test fit was only modified to continue force the agent to pretend it is using YARN.

PR #2786, opened 2016-08-25
Minerkasch/spark standalone

zachradtka:
This PR is a copy of #2752 because no one has access to that fork.

### What does this PR do?

Enabled Spark metrics to be gathered when clusters are provisioned using Spark's Standalone cluster mode. Currently the agent checks works when Spark is running with Apache YARN to provision its resources. This PR re-uses the bulk of the collection logic, but retrofits an adapter to locate application servers via Spark Standalone's APIs.

Events have also been added. The agent can now raise events when both jobs and stages have status transitions (ie, moving from WAITING to RUNNING to COMPLETED or FAILED).

### DIFF
--- a/checks.d/spark.py
+++ b/checks.d/spark.py
@@ -222,8 +222,9 @@ class SparkCheck(AgentCheck):
         Figures out what mode we're in and fetches running apps
         '''
         cluster_mode = instance.get(SPARK_CLUSTER_MODE)
-        if not cluster_mode:
-            raise Exception('No cluster mode specified. %s must be set to %s or %s', (SPARK_CLUSTER_MODE, SPARK_STANDALONE_MODE, SPARK_YARN_MODE))
+        if cluster_mode is None:
+            self.log.warning('The value for `spark_cluster_mode` was not set in the configuration. Defaulting to "%s"' % SPARK_YARN_MODE)
+            cluster_mode = SPARK_YARN_MODE
 
         if cluster_mode == SPARK_STANDALONE_MODE:
             return self._standalone_init(instance)
--- a/checks.d/spark.py
+++ b/checks.d/spark.py
@@ -89,18 +89,16 @@ SPARK_CLUSTER_MODE = 'spark_cluster_mode'
 SPARK_STANDALONE_MODE = 'spark_standalone_mode'
 SPARK_YARN_MODE = 'spark_yarn_mode'
 
-SPARK_STANDALONE_MASTER = 'spark_standalone_master_uri'
-SPARK_STANDALONE_SERVICE_CHECK = 'spark_standalone_master'
-
-# Service Check Names
+# Service Checks
+SPARK_STANDALONE_SERVICE_CHECK = 'spark.standalone_master.can_connect'
 YARN_SERVICE_CHECK = 'spark.resource_manager.can_connect'
 SPARK_SERVICE_CHECK = 'spark.application_master.can_connect'
 
 # URL Paths
 YARN_APPS_PATH = 'ws/v1/cluster/apps'
 SPARK_APPS_PATH = 'api/v1/applications'
-SPARK_MASTER_STATE_PATH = "/json/"
-SPARK_MASTER_APP_PATH = "/app/"
+SPARK_MASTER_STATE_PATH = '/json/'
+SPARK_MASTER_APP_PATH = '/app/'
 
 # Application type and states to collect
 YARN_APPLICATION_TYPES = 'SPARK'
@@ -219,7 +217,7 @@ class SparkCheck(AgentCheck):
 
     def _get_running_apps(self, instance, tags):
         '''
-        Figures out what mode we're in and fetches running apps
+        Determine what mode was specified
         '''
         cluster_mode = instance.get(SPARK_CLUSTER_MODE)
         if cluster_mode is None:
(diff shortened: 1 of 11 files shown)


truthbk:
Awesome! Thanks @zachradtka

truthbk:
@zachradtka I made some comments - for some minor changes in #2752 (overall the code was really good). Can you take a look at those and let me know if you can make those changes or if you disagree for whatever reason? Thanks a ton!

zachradtka:
I have made most of the changes. I believe the one change I have left is "refactor _event_for_stage_status_change() and _event_for_job_status_change()". I just didn't have time to get to it tonight.

One of the larger changes I made was changing `resourcemanager_uri` to `spark_url` in the config file. I was wondering if this would break compatibility for users who are currently using for YARN? 

I did this because I didn't like having two separate variables to get a URL from the user. I found it confusing to have to change the variable name to use a different manager (YARN, Spark Standalone).

If you have a better solution, I am open to ideas. I guess the simple solution would be to just use `resourcemanager_uri` for everything.

truthbk:
Hi @zachradtka, thanks for those changes. I'll review them in more detail later today, but I did want to warn you beforehand about the dreaded `resourcemanager_uri` as opposed to `spark_url`. Unfortunately those changes would bring in backward compatibility issues because users normally don't want to update their config files after an upgrade (we should seamlessly accept what's already there). This does bring issues sometimes, when we don't get a config variable name right, or we introduce some design changes, but we need to live with it. I'm fine with having `resourcemanager_uri` be an alias of `spark_url`, but we need to be able to handle the old-school configuration variables too.

zachradtka:
Thanks for confirming that. My latest commit handles users who still use `resourcemanager_uri`. It will first check for `spark_url`, then try to get a value for `resourcemanager_uri`. If `resourcemanager_uri` is used a warning logged, but the agent still runs.

checks.d/spark.py
```diff
@@ -228,32 +232,48 @@ def _get_running_apps(self, instance, tags):
...
+
+            if master_address:
+                self.log.warning('The use of `%s` is deprecated. Please use `%s` instead.' %
+                    (DEPRECATED_MASTER_ADDRESS, MASTER_ADDRESS))
+            else:
+                raise Exception('A URL for `%s` must be specified in the instance '
+                    'configuration' % MASTER_ADDRESS)
+
+        # Get the cluster name from the instance configuration
+        cluster_name = instance.get('cluster_name')
```
truthbk: Would a cluster name need to be defined in standalone mode too?

conf.d/spark.yaml.example
```diff
@@ -12,7 +12,8 @@ instances:
   # ResourceManager hostname can be found in the yarn-site.xml conf file
   # under the property yarn.resourcemanager.address The ResourceManager port
   # can be found in the yarn-site.xml conf file under the property
-  # yarn.resourcemanager.webapp.address
+  # yarn.resourcemanager.webapp.address. This is http://localhost:8088 by default
+  #
```
truthbk: could we add a small note about how `resourcemanager_uri` is deprecated but still functional?

checks.d/spark.py
```diff
@@ -546,6 +494,53 @@ def _set_metric(self, metric_name, metric_type, value, tags=None):
...
+        '''
+        Create an event for a stage changing status
+        '''
+        stage_name = current_stage.get('name')
+        stage_id = current_stage.get('stageId')
+        current_status = current_stage.get('status')
+
+        self._set_event(current_status, previous_status, tags, STAGE_EVENT, stage_name, stage_id)
+
+    def _set_event(self, current_status, previous_status, tags, event_type, state_name, state_id):
```
truthbk: nice! Thank you! 💯

truthbk:
@zachradtka just a couple small questions. Other than that this looks great, thanks a lot for the refactor!

checks.d/spark.py
zachradtka: The reason I decided to require it is because Spark Standalone is still running on a cluster it is just a Spark only cluster. It is also a good way for anyone who may be running multiple instances that are executing a job with the same name to differentiate metrics. Also with my next PR (#2797) it is required, so I felt it was better to make it required for any mode of spark you chose: Standalone, YARN, or Mesos.

zachradtka:
Thanks for the review! Let me know if you wanted to address anything else 😄

checks.d/spark.py
truthbk: Sounds good to me. Thanks for the detailed explanation! :)

truthbk:
@zachradtka thank you! Great job. We (not necessarily you!) should add some tests to this in the future. The mocking or infra-setup for them will be painful... hehe.

Merging! :shipit:

buryat:
I got this error while trying to run dd-agent:

```
2016-09-02 19:08:48 UTC | ERROR | dd.collector | config(config.py:801) | Unable to import check module spark.py from checks.d
  File "/root/dd-agent/config.py", line 797, in _get_check_class
    check_module = imp.load_source('checksd_%s' % check_name, check_path)
  File "/root/dd-agent/checks.d/spark.py", line 81, in <module>
    from bs4 import BeautifulSoup
ImportError: No module named bs4
```

zachradtka:
Yes, that is because BeautifulSoup needs to be installed. I am not sure how DataDog handles dependencies, but I included the following in requirements.txt

```
...
# checks.d/spark.py
beautifulsoup4==4.5.1
```

Is this how you require a new dependency? If not, please let me know and I will update.

zachradtka:
@truthbk: PR #2797 contains the unit tests for Spark Standalone and Mesos.

truthbk:
beautifulsoup issue will be addressed in build by:
- https://github.com/DataDog/omnibus-software/pull/73
- https://github.com/DataDog/dd-agent-omnibus/pull/91

PR #2788, opened 2016-08-25
[marathon] Count running/pending deployments

bradhe:
### What does this PR do?

Adds a new metric for to Marathon checks to count the number of deployments that are pending.

### Motivation

Basically, we'd like to track when that number stays high (e.g. > 1) so that we can alert on stuck deployments.

### Testing Guidelines

~~An overview on [testing](https://github.com/DataDog/dd-agent/blob/master/tests/README.md)
is available in our contribution guidelines.~~

Added some tests, there was no real coverage for this guy.

### Additional Notes

I added a new helper method for loading a JSON fixture. I'm happy to pull that out, though, as it feels like it's extending this a little much.

### DIFF
--- a/checks.d/marathon.py
+++ b/checks.d/marathon.py
@@ -16,70 +16,70 @@ from checks import AgentCheck
 
 class Marathon(AgentCheck):
 
-	DEFAULT_TIMEOUT = 5
-	SERVICE_CHECK_NAME = 'marathon.can_connect'
+    DEFAULT_TIMEOUT = 5
+    SERVICE_CHECK_NAME = 'marathon.can_connect'
 
-	APP_METRICS = [
-		'backoffFactor',
-		'backoffSeconds',
-		'cpus',
-		'disk',
-		'instances',
-		'mem',
-		'taskRateLimit',
-		'tasksRunning',
-		'tasksStaged'
-	]
+    APP_METRICS = [
+        'backoffFactor',
+        'backoffSeconds',
+        'cpus',
+        'disk',
+        'instances',
+        'mem',
+        'taskRateLimit',
+        'tasksRunning',
+        'tasksStaged'
+    ]
 
-	def check(self, instance):
-		if 'url' not in instance:
-			raise Exception('Marathon instance missing "url" value.')
+    def check(self, instance):
+        if 'url' not in instance:
+            raise Exception('Marathon instance missing "url" value.')
 
-		# Load values from the instance config
-		url = instance['url']
-		user = instance.get('user')
-		password = instance.get('password')
-		if user is not None and password is not None:
-			auth = (user,password)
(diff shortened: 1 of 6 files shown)


checks.d/marathon.py
```diff
@@ -16,66 +16,70 @@
...
+				message='%s returned a status of %s' % (url, r.status_code),
+				tags = ["url:{0}".format(url)])
+			raise Exception("Got %s when hitting %s" % (r.status_code, url))
+
+		else:
+			self.service_check(self.SERVICE_CHECK_NAME, AgentCheck.OK,
+				tags = ["url:{0}".format(url)]
+			)
+
+		return r.json()
```
bradhe: Converted spaces to tabs here for consistency.

checks.d/marathon.py
```diff
@@ -56,6 +56,10 @@ def check(self, instance):
                     if attr in app:
                         self.gauge('marathon.' + attr, app[attr], tags=tags)
 
+        response = self.get_json(urljoin(url, "v2/deployments"), timeout, auth)
```
masci: Do you think we can get this behind a config flag? I imagine this will slow down the check a bit, being not a Marathon expert was wondering if this is something any Marathon user would trade for the new metric or not.

tests/checks/common.py
```diff
@@ -130,6 +131,9 @@ def read_file(file_name, string_escape=True):
                 contents = contents.decode('string-escape')
             return contents.decode("utf-8")
 
+    @staticmethod
+    def read_json_file(file_name, string_escape=True):
```
masci: Looks good to me, let's keep it in this PR

tests/checks/mock/test_marathon.py
```diff
@@ -0,0 +1,56 @@
...
+        'default_timeout': 5
+    },
+    'instances': [
+        {
+            'url': 'http://localhost:8080'
+        }
+    ]
+}
+
+class MarathonCheckTest(AgentCheckTest):
```
masci: 🙇

checks.d/marathon.py
bradhe: I'm cool with tossing this behind a config flag. I'll push up a commit for that momentarily.

bradhe:
@masci I've pushed some commits that address your feedback.

bradhe:
It'd be really great to get this merged so that we can start monitoring this! I'd like to avoid having to start sending my own metrics via an external script :(

masci:
@bradhe I'll push this up on our queue but at the time you submitted the PR the agent was already in feature freeze for `5.9`, let me see what we can do.
Thanks for your patience!

bradhe:
Thanks @masci, no problem-o.

PR #2789, opened 2016-08-25
[ci] reduce code duplication

degemer:


ci/nginx.rb
```diff
@@ -63,23 +63,8 @@ def nginx_rootdir
...
-      end
-      if ENV['SKIP_CLEANUP']
-        puts 'Skipping cleanup, disposable environments are great'.yellow
-      else
-        puts 'Cleaning up'
-        Rake::Task["#{flavor.scope.path}:cleanup"].invoke
-      end
-      raise exception if exception
+      Rake::Task['ci:common:execute'].invoke(flavor)
+      endception if exception
```
masci: `exception` is not defined here

ci/common.rb
```diff
@@ -178,6 +178,29 @@ def self.for(smth, max_timeout = DEFAULT_TIMEOUT)
       sh %(#{path}nosetests -s -v -A "#{nose}" #{tests_directory})
       t.reenable
     end
-    task execute: [:before_install, :install, :before_script, :script]
+
+    task :execute, :flavor do |_t, attr|
+      flavor = attr[:flavor]
+      # Check if the check uses the cache
+      cached = flavor.tasks.any? { |task| task.name == (flavor.scope.path + ':before_cache') }
```
masci: We should proably double check flavors that have the `before_cache` tasks but actually don't use it.
I think this is why Win tests are failing during the `before_cache` task  for `core_integrations`: it was not executed before but now it is since it exists and `cached` becomes true.
Haven't had the opportunity to verify my thoughts, let me know if this makes sense.

degemer: You're right, we should probably have a better way to do this. I think that right now `before_cache` is always executed.
Since the only issue is Windows, I added a `if Gem.win_platform?` before executing the `find` in `ci:common:before_cache`.
We should probably clean this code once again later.

degemer: Hum, `unless Gem.win_platform?` is better.

Issue #2671, opened 2016-07-08
[wmi] More friendly error for missing/incorrect WMI class


mausch:
Currently if you try to query a non-existing class you get a rather obscure `OLE error`. https://github.com/DataDog/dd-agent/issues/2648#issuecomment-230834999
Instead, it would be nice to see a message such as "Invalid WMI class xxx".

remh:
Thanks @mausch . It got fixed in an earlier version of the agent.

PR #2790, opened 2016-08-25
[wmi] better `com_errors` handling

yannmh:
**[wmi] friendly `com_errors` 💑**

Maintain a list of "known" WMI `com_errors`, and translate them to user
intelligible errors.
Add a `mute` parameter to `WMISampler`.
- If set to `True` (default for system checks), `com_error` are
  "translated" (when possible), logged and do not cause any
  interruption.
- If not (user checks), "translated" exceptions raise.

**[wmi_check][iis] unmute various exceptions 🔊**

Unmute WMI `com_errors` exceptions related to an invalid `provider`
option or WMI class.

### DIFF
--- a/checks/libs/wmi/exceptions.py
+++ b/checks/libs/wmi/exceptions.py
@@ -1,0 +1,43 @@
+"""
+List WMI "known" `com_errors` errors.
+
+Translate to user intelligible exceptions.
+"""
+
+_user_exception_by_com_errors = {}
+
+
+def com_error(error_id):
+    """
+    A decorator that assigns an `error_id` to an intelligible exception.
+    """
+    def set_exception(exception):
+        _user_exception_by_com_errors[error_id] = exception
+        return exception
+    return set_exception
+
+
+def raise_on_com_error(error):
+    """
+    Raise the user exception associated with the given `com_error` or
+    fall back to the original exception.
+    """
+    raise _user_exception_by_com_errors.get(error[0], error)
+
+
+# List of user exceptions
+class WMIException(Exception):
+    """
+    Abtract exception for WMI.
+    """
+    def __init__(self):
+        """
+        Use the class docstring as an exception message.
+        """
+        super(WMIException, self).__init__(self.__doc__)
+
+
+@com_error(-2147217392)
+class WMIInvalidClass(WMIException):
+    """WMI class is invalid."""
+    pass
--- a/checks/libs/wmi/sampler.py
+++ b/checks/libs/wmi/sampler.py
@@ -30,6 +30,7 @@ from win32com.client import Dispatch
 
 # project
 from checks.libs.wmi.counter_type import get_calculator, get_raw, UndefinedCalculator
+from checks.libs.wmi.exceptions import raise_on_com_error, WMIException
 from utils.timeout import timeout, TimeoutException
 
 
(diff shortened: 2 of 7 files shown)


yannmh:
Fix #2671

checks/libs/wmi/exceptions.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+    A decorator that assigns an `error_id` to an intelligible exception.
+    """
+    def set_exception(exception):
+        _user_exception_by_com_errors[error_id] = exception
+        return exception
+    return set_exception
+
+
+def raise_on_com_error(error, silent=True):
```
truthbk: What does `silent` do? I can't see where we use it...

yannmh: Oops this is a leftover. Thanks for catching.

utils/debug.py
```diff
@@ -52,7 +52,7 @@ def get_check(name, config_str):
     check_class = None
     classes = inspect.getmembers(check_module, inspect.isclass)
     for name, clsmember in classes:
-        if AgentCheck in clsmember.__bases__:
+        if issubclass(clsmember, AgentCheck) and clsmember != AgentCheck:
```
truthbk: 💯

checks/libs/wmi/sampler.py
```diff
@@ -91,8 +92,10 @@ class WMISampler(object):
 
     def __init__(self, logger, class_name, property_names, filters="", host="localhost",
                  namespace="root\\cimv2", provider=None,
-                 username="", password="", and_props=[], timeout_duration=10):
+                 username="", password="", and_props=[],
+                 mute=True, timeout_duration=10):
```
truthbk: I get why you have implemented the `mute` feature. But do we really need it? None of the checks initiate the sampler with a different value here, no?

truthbk:
Overall beautiful, as always, just have a question for you...

truthbk:
Looks good to me! 👍

checks/libs/wmi/sampler.py
yannmh: The Datadog integration checks relying on `wmi` have `mute` set `False`.  This allows the check to raise, and the exception message to be used by the `dd-agent` info page.

truthbk: Yeah, I had looked at the top of the PR yesterday, and forgot about `mute` being `False` there. Thanks for the explanation though.

yannmh:
Tests are failing on AppVeyor. Need to fix it.

yannmh:
Partially (`BUGFIX`) merged with https://github.com/DataDog/dd-agent/pull/2812. 

Other improvements remain open in this PR. The logic need to be improved in order to differentiate errors with the same WMI COM error id.

truthbk:
Hi @yannmh,

Thanks again for your contribution, always very much appreciated. We're currently transitioning our checks to a new SDK-based model for the agent. An unfortunate side-effect is that we have to impose a freeze on the dd-agent repo. The SDK consists of two repositories https://github.com/DataDog/integrations-core and https://github.com/DataDog/integrations-extras, where all of our checks will reside. This PR would belong in the `core` repository. Would you be so kind as to re-open the PR there? We have facilitated some tools to help you with the SDK development that should help you get up and running in no time: 

Also, in your case, you also have an agent core component in the PR, so please feel free to open a new PR just with the changes relevant to the core. 

PR #2791, opened 2016-08-26
[ci] do not run integration tests if not relevant

degemer:
### What does this PR do?

This PR tries to avoid running a integration test for an agent check (and
only this, core integration tests, linter and core mocked tests are
always run) when it is not required.

To do this, we first lists all modified files, and then
- check if a check file was modified (in `checks.d`)
- check if a check test file was modified
- ignore fixtures and configuration

If any file other than the one listed above are modified, all tests will
be run. Otherwise, a list of modfied checks is built, and only these
tests will be run (with the core tests).

### Motivation

Faster Travis.

### DIFF
--- a/ci/common.rb
+++ b/ci/common.rb
@@ -38,6 +38,26 @@ def install_requirements(req_file, pip_options = nil, output = nil, use_venv = n
   end
 end
 
+def can_skip?
+  return false, [] if Gem.win_platform?
+
+  modified_checks = []
+  `git diff-tree --no-commit-id --name-only -r FETCH_HEAD origin/master`.each_line do |filename|
+    filename.strip!
+    if filename.start_with? 'checks.d'
+      check_name = File.basename(filename, '.py')
+    elsif filename.start_with?('tests/checks/integration', 'tests/checks/mock')
+      check_name = File.basename(filename, '.py').slice 'test_'
+    elsif filename.start_with?('tests/checks/fixtures', 'conf.d')
+      next
+    else
+      return false, []
+    end
+    modified_checks << check_name unless modified_checks.include? check_name
+  end
+  [true, modified_checks]
+end
+
 # helper class to wait for TCP/HTTP services to boot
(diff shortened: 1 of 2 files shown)


ci/common.rb
```diff
@@ -183,12 +203,22 @@ def self.for(smth, max_timeout = DEFAULT_TIMEOUT)
...
+      # flavor.scope.path is ci:cassandra
+      # flavor.scope.path[3..-1] is cassandra
+      check_name = flavor.scope.path[3..-1]
+
+      can_skip, checks = can_skip?
+      puts 'CAN_SKIP RESULT'.red
+      puts can_skip, checks
+      can_skip &&= !%w(default core_integration checks_mock).include?(check_name)
+      if can_skip && !checks.include?(check_name)
+        puts "Skipping #{check_name} tests, not affected by the change".yellow
```
truthbk: 💯

ci/common.rb
```diff
@@ -38,6 +38,26 @@ def install_requirements(req_file, pip_options = nil, output = nil, use_venv = n
   end
 end
 
+def can_skip?
+  return false, [] if Gem.win_platform?
+
+  modified_checks = []
+  `git diff-tree --no-commit-id --name-only -r FETCH_HEAD origin/master`.each_line do |filename|
```
truthbk: Maybe as a future improvment, it could be nice to grab the branch the PR is going to merge into with `TRAVIS_PULL_REQUEST` and the github API.

degemer: You're right, it would be better, but since in our case we either run our tests against master or a release branch. For master, it's already in the source code, and I think that comparing 5.{y}.x against master will trigger all builds (which is what we want on a release branch).

So yeah, something to do when we have more time! 😄

degemer:
Everything if working, merging. I tested locally for changes only about one integration, and it worked. But I cannot test it more, it needs to be on master.

PR #2793, opened 2016-08-26
[ci] better cleanup

degemer:
### What does this PR do?

Better cleaning before caching.

### Motivation

Uploading the cache takes ~10s, we could win 10 * 55 / 5 = 110s of build time.


degemer:
Not working, closing. It's better anyway to lose 10s uploading a cache to be able to really cache the python dependencies. (right now, installing the python dependencies takes ~10s ; with the standard `pip` cache of Travis, it takes ~1min).

PR #2797, opened 2016-08-31
Minerkasch/spark mesos

zachradtka:
### What does this PR do?

Enabled Spark metrics to be gathered when clusters are provisioned using Apache Mesos. This is an extension of #2786. This PR also includes unit test for Mesos and Spark Standalone.


truthbk:
@zachradtka reviewing this now... since it took me so long I refreshed the `base` so I had only the mesos changes in the `diff` ;)

checks.d/spark.py
```diff
@@ -299,6 +307,33 @@ def _standalone_init(self, spark_master_address):
...
+        Return a dictionary of {app_id: (app_name, tracking_url)} for running Spark applications.
+        '''
+        running_apps = {}
+
+        metrics_json = self._rest_request_to_json(master_address,
+            MESOS_MASTER_APP_PATH,
+            MESOS_SERVICE_CHECK)
+
+        if metrics_json.get('frameworks'):
+            for app_json in metrics_json.get('frameworks'):
```
truthbk: you could probably get rid of the `if` above and write this as `for app_json in metrics_json.get('frameworks', {}):` - if it's an empty dict we just won't iterate.

conf.d/spark.yaml.example
```diff
@@ -2,24 +2,30 @@ init_config:
 
 instances:
   #
```
truthbk: thanks for the instructions, very clear 👍

tests/checks/mock/test_spark.py
```diff
@@ -13,16 +13,24 @@
 CLUSTER_NAME = 'SparkCluster'
 APP_NAME = 'PySparkShell'
```
truthbk: I can't begin to thank you enough for these tests!! Thank you so much! 💯

checks.d/spark.py
```diff
@@ -262,6 +265,11 @@ def _get_running_apps(self, instance, tags):
         if cluster_mode == SPARK_STANDALONE_MODE:
             return self._standalone_init(master_address)
 
+        elif cluster_mode == SPARK_MESOS_MODE:
+            running_apps = self._mesos_init(master_address)
+            return self._get_spark_app_ids(running_apps)
+
+
```
truthbk: nit: could we remove this blank line?

zachradtka:
Did you still want me to make these code changes?

truthbk:
@zachradtka if you don't mind that would be great. I approved because it's just two dumb nitpicks, but if you have the time it would be nice. 😄

truthbk:
No need to wait for the nitpick fixes. Merging.

PR #2798, opened 2016-08-31
[core] avoid calls to service discovery from dogstatsd

hkaj:
### What does this PR do?

Dogstatsd calls util.get_hostname that in turn was making calls to service discovery related code. This PR remove this behavior.

### Motivation

Although it was just calling the constructor of the config store class as a side effect, it's better to keep things separated.

### Testing Guidelines

We should make sure the hostname is still resolved the same way to avoid breaking things. This was tested locally with and without docker and the results are satisfying.

### Additional Notes

🙈


hkaj:
CI failures are unrelated (timeouts)

olivielpeau:
LGTM!

hkaj:
thanks @olivielpeau

PR #2799, opened 2016-08-31
Allow expected address to be checked.

gphat:
### What does this PR do?

Allows the DNS check to verify the resolved address against a list of expected values.

### Motivation

We have a number of DNS sanity checks that we check for known IPs. Think of public-facing ingestion points for which IP addresses are published.

### Testing Guidelines

None yet. Working on it.

### Additional Notes

I like your new PR format. 😸 

cc @rhwlo


checks.d/dns_check.py
```diff
@@ -52,21 +52,27 @@ def query_dns(self, instance, timeout):
...
         start_time = time.time()
         try:
             self.log.debug('Querying "%s" record for hostname "%s"...' % (record_type, hostname))
             answer = resolver.query(hostname, rdtype=record_type)
+            resolved_value = answer.rrset.items[0].to_text()
             assert(answer.rrset.items[0].to_text())
             end_time = time.time()
+            if len(expected_results) > 0:
+                if resolved_value not in expected_results:
+                    raise Exception('DNS resolution of %s resulted in unexpected address %s.' % (hostname, resolved_value))
```
degemer: There might be a better way than raising an exception, since this verification can only take place if the DNS resolution was done correctly: put this block after the `else`, before the `if end_time - start_time > 0`, and log an error + service critical. What do you think ?

conf.d/dns_check.yaml.example
```diff
@@ -9,3 +9,9 @@ instances:
       # Specify an (optional) `record_type` to customize the record type
       # queried by the check (default: "A")
       # record_type: A
+
+      # Specify expected results if you want to check that the returned IP
+      # matches. If you supply multiple records then a match to any one will
+      # be considered success
+      # addressses:
```
degemer: Only two `s` needed 😃

degemer: Could you also specify that it's disabled by default ?

checks.d/dns_check.py
degemer: Just noticed, this is also a way to fix https://travis-ci.org/DataDog/dd-agent/jobs/156839184#L608

checks.d/dns_check.py
```diff
@@ -52,21 +52,27 @@ def query_dns(self, instance, timeout):
...
 
         record_type = instance.get('record_type', 'A')
+        expected_results = instance.get('addresses', [])
 
         status = AgentCheck.CRITICAL
         start_time = time.time()
         try:
             self.log.debug('Querying "%s" record for hostname "%s"...' % (record_type, hostname))
             answer = resolver.query(hostname, rdtype=record_type)
+            resolved_value = answer.rrset.items[0].to_text()
```
degemer: What about multiple DNS records ? We should probably takes all the addresses and try to see if one of them matches (all of them maybe makes more sense). What do you think ?

gphat:
Looks like this might be right now? I need to verify this on my end as well…

checks.d/dns_check.py
```diff
@@ -52,23 +52,30 @@ def query_dns(self, instance, timeout):
...
-        except Exception:
-            self.log.exception('DNS resolution of %s has failed.' % hostname)
-            self.service_check(self.SERVICE_CHECK_NAME, status, tags=self._get_tags(instance))
+        except Exception as err:
+            self.log.exception(err)
+            self.service_check(self.SERVICE_CHECK_NAME, status, tags=self._get_tags(instance), message=err)
             raise
         else:
+            if len(expected_results) > 0:
+                missing_values = list(set(expected_results, resolved_results))
```
degemer:
```python
missing_values = [r for r in expected_results if r not in resolved_results]
```
maybe ? Unless you have a better way to do it.

checks.d/dns_check.py
```diff
@@ -52,23 +52,30 @@ def query_dns(self, instance, timeout):
...
+                if len(missing_values) > 0:
+                    self.log.error('DNS resolution of %s did not contain expected address(es) %s.' % (hostname, ", ".join(missing_values)))
+                    self.service_check(self.SERVICE_CHECK_NAME, status, tags=self._get_tags(instance))
```
degemer: A `OK` service check might be sent L74 even when this CRITICAL is sent.
One way to solve this would be to handle the case `end_time < start_time` first (and maybe send a service check with WARNING ?), then the check for the missing values and finally the OK service check.
What do you think @hkaj ?

checks.d/dns_check.py
```diff
@@ -52,23 +52,30 @@ def query_dns(self, instance, timeout):
...
         record_type = instance.get('record_type', 'A')
+        expected_results = instance.get('addresses', [])
 
         status = AgentCheck.CRITICAL
         start_time = time.time()
         try:
             self.log.debug('Querying "%s" record for hostname "%s"...' % (record_type, hostname))
             answer = resolver.query(hostname, rdtype=record_type)
-            assert(answer.rrset.items[0].to_text())
+            resolved_results = map(lambda x: x.to_text(), answer.rrset.items)
```
degemer: This won't fail when the first item is the empty string, whereas it was failing previously. Maybe filter this resolved_results to remove empty strings and later check that resolved_results != [] ?

checks.d/dns_check.py
hkaj: We can probably remove the `if end_time - start_time > 0:` condition and just do:
```python
if len(missing_values) > 0:
    self.log.error('DNS resolution of %s did not contain expected address(es) %s.' % (hostname, ", ".join(missing_values)))
    self.service_check(self.SERVICE_CHECK_NAME, status, tags=self._get_tags(instance))
else:
    self.service_check(self.SERVICE_CHECK_NAME, AgentCheck.OK, tags=self._get_tags(instance))
self.gauge('dns.response_time', end_time - start_time, tags=tags)
```

checks.d/dns_check.py
rhwlo: I’d prefer `list(set(expected_results) - set(resolved_results))` for efficiency unless there’s a reason to prefer the comprehension

degemer: Your method is also way easier to read/understand. 👍

cory-stripe:
Oops, kinda forgot about this. :)

sjenriquez:
Hey @cory-stripe, my most recent PR #2924 impacts your work here. I'm moving the DNS check over to the [integrations-core repo](https://github.com/DataDog/integrations-core). The changes here are great, we hope you'll move this PR to the new repo!

cory-stripe:
Ok, I'll check in to this soon. Thanks!

PR #2803, opened 2016-09-02
[redis] Fix tags on redis connect check.

gphat:
### What does this PR do?

Adjusts the `redis.can_connect` tags to include the those specified in the config file.

### Motivation
At present the `redis.can_connect` service check does not include the tags specified in the config file. I don't see why this is the case, so I figured I'd send a PR as a way of asking. That way if it's an oversight we've already got a fix lined up. :)

### Testing Guidelines

Added a test assertion for the tag.

### Additional Notes

Let me know if the tag is intended to be left out!

### DIFF
--- a/checks.d/redisdb.py
+++ b/checks.d/redisdb.py
@@ -157,12 +157,12 @@ class Redis(AgentCheck):
 
         tags = sorted(tags.union(tags_to_add))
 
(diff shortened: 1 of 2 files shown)


gphat:
Sure thing @degemer. Updated!

PR #2805, opened 2016-09-02
Vagelim/ceph jewel update

vagelim:
### What does this PR do?

Updates the Ceph integration for version 10.2.2

### Motivation

Gobstopper is working on a writeup of Ceph. After spinning up a 3-node cluster with the latest Jewel release, some metrics were not being reported. Additionally, we wanted to expand the number of metrics reported to include information on full and near full OSDs. Lastly, there was an inconsistency in the `pool_name` tag, which has been resolved.

### Testing

A test for the new metrics has been added, with additional test data pulled from a running 10.2.2 cluster.

### Additional Notes

Added default value of zero for operation rate metrics.
Fixed failing test, unrelated to these changes.

### DIFF
--- a/checks.d/ceph.py
+++ b/checks.d/ceph.py
@@ -37,7 +37,7 @@ class Ceph(AgentCheck):
         args = ceph_args + ['version']
         try:
             output,_,_ = get_subprocess_output(args, self.log)
-        except Exception, e:
+        except Exception as e:
             raise Exception('Unable to run cmd=%s: %s' % (' '.join(args), str(e)))
 
         raw = {}
@@ -46,7 +46,7 @@ class Ceph(AgentCheck):
                 args = ceph_args + cmd.split() + ['-fjson']
                 output,_,_ = get_subprocess_output(args, self.log)
                 res = json.loads(output)
-            except Exception, e:
+            except Exception as e:
                 self.log.warning('Unable to parse data from cmd=%s: %s' % (cmd, str(e)))
                 continue
 
--- a/checks.d/ceph.py
+++ b/checks.d/ceph.py
@@ -37,16 +37,16 @@ class Ceph(AgentCheck):
         args = ceph_args + ['version']
         try:
             output,_,_ = get_subprocess_output(args, self.log)
-        except Exception as e:
+        except Exception, e:
             raise Exception('Unable to run cmd=%s: %s' % (' '.join(args), str(e)))
 
         raw = {}
-        for cmd in ('mon_status', 'status', 'df detail', 'osd pool stats', 'osd perf'):
+        for cmd in ('mon_status', 'status', 'df detail', 'osd pool stats', 'osd perf', 'health detail'):
             try:
                 args = ceph_args + cmd.split() + ['-fjson']
                 output,_,_ = get_subprocess_output(args, self.log)
                 res = json.loads(output)
-            except Exception as e:
+            except Exception, e:
                 self.log.warning('Unable to parse data from cmd=%s: %s' % (cmd, str(e)))
                 continue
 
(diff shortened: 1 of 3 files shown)


vagelim:
Failed tests are unrelated to my changes

checks.d/ceph.py
```diff
@@ -82,20 +82,71 @@ def _extract_metrics(self, raw, tags):
...
+            # Health summary will be empty if no bad news
+            if raw['health_detail']['summary'] == []:
+                health = {'num_near_full_osds' : 0, 'num_full_osds' : 0}
+                self._publish(health, self.count, ['num_near_full_osds'], tags)
+                self._publish(health, self.count, ['num_full_osds'], tags)
+            else:
+                for osdhealth in raw['health_detail']['detail']:
+                    # Full looks like: osd.2 is full at 95%
+                    # Near full: osd.1 is near full at 94%
+                    pct = int(osdhealth[osdhealth.find('at ') + len('at '):osdhealth.find('%')])
```
masci: I'd use a (compiled since we're within a cycle) regexp here, `\d+%` should work. If you could provide a separate function/method only responsible for parsing this line it would be extremely easy to test.

checks.d/ceph.py
```diff
@@ -82,20 +82,71 @@ def _extract_metrics(self, raw, tags):
...
+                    local_tags = tags + ['ceph_osd:%s' % osdhealth[:osdhealth.find(' is')].replace('.', '')]
```
masci: Same as above, putting this on a separate function can ease testing. Can we assume `osd.num` is an invariant or users can specify an arbitrary name? The parsing is currently based on a substring of the error message (` is`), if we could use something more robust it'd be better.

checks.d/ceph.py
```diff
@@ -82,20 +82,71 @@ def _extract_metrics(self, raw, tags):
...
+        except KeyError:
+            self.log.debug('Error retrieving health metrics')
+
+        for osdinfo in raw['osd_pool_stats']:
+            name = osdinfo['pool_name']
+            local_tags = tags + ['ceph_pool:%s' % name]
+            ops = 0
+            try:
+                self._publish(osdinfo, self.gauge, ['client_io_rate', 'read_op_per_sec'], local_tags)
+                ops += osdinfo['client_io_rate']['read_op_per_sec']
```
masci: if `osdinfo` raises a `KeyError` after the `self._publish` succeeded, you publish again in the except branch, is this on purpose?

checks.d/ceph.py
```diff
@@ -82,20 +82,71 @@ def _extract_metrics(self, raw, tags):
...
+                        health = {'num_near_full_osds' : pct}
+                        self._publish(health, self.count, ['num_near_full_osds'], local_tags)
+                    else:
+                        health = {'num_full_osds' : pct}
+                        self._publish(health, self.count, ['num_full_osds'], local_tags)
+        except KeyError:
+            self.log.debug('Error retrieving health metrics')
+
+        for osdinfo in raw['osd_pool_stats']:
+            name = osdinfo['pool_name']
```
masci: You might want to use a `get` here?

tests/checks/mock/test_ceph.py
```diff
@@ -46,6 +46,36 @@ def test_tagged_metrics(self):
         for pool in ['pool0', 'rbd']:
             expected_tags = ['ceph_fsid:e0efcf84-e8ed-4916-8ce1-9c70242d390a',
                              'ceph_mon_state:peon',
-                             'ceph_pool_name:%s' % pool]
+                             'ceph_pool:%s' % pool]
             for metric in ['ceph.read_bytes', 'ceph.write_bytes', 'ceph.pct_used', 'ceph.num_objects']:
                 self.assertMetric(metric, count=1, tags=expected_tags)
+
+    def test_osd_status_metrics(self):
```
masci: Bonus points: test `_extract_metrics` method alone so you can pass in bogus payloads and see it the code is resilient (fixtures are good but here we are only testing the working case)

checks.d/ceph.py
vagelim: Yes; it seems more intuitive to report a value of zero when there are no operations, rather than report nothing at all.

checks.d/ceph.py
vagelim: Very good idea, will do.

checks.d/ceph.py
vagelim: This was unmodified from the original check: https://github.com/DataDog/dd-agent/blob/master/checks.d/ceph.py#L86

Not sure the benefit of using `get` in this case—we don't want to specify a default. But am willing to change if you see fit.

masci: It's just b/c it could raise an unhandled exception, `osdinfo.get('pool_name')` always succeeds even if we don't need a default

checks.d/ceph.py
```diff
@@ -82,20 +83,70 @@ def _extract_metrics(self, raw, tags):
...
-                local_tags = tags + ['ceph_pool:%s' % name]
+            # Health summary will be empty if no bad news
+            if raw['health_detail']['summary'] == []:
+                health = {'num_near_full_osds' : 0, 'num_full_osds' : 0}
+                self._publish(health, self.count, ['num_near_full_osds'], tags)
+                self._publish(health, self.count, ['num_full_osds'], tags)
+            else:
+                for osdhealth in raw['health_detail']['detail']:
+                    osd, pct = self._osd_pct_used(osdhealth)
+                    local_tags = tags + ['ceph_osd:%s' %  osd.replace('.','')]
```
masci: flake8 is not happy with the extra space afther the `%`: `./checks.d/ceph.py:94:57: E222 multiple spaces after operator`

masci:
A minor issue with flake8, let's fix it then LGTM 👍

PR #2809, opened 2016-09-06
[mesos] make ssl check optional

hkaj:
if you have not yet done so.*

### What does this PR do?

Allow disabling the ssl verification in mesos master and slave checks. This is mostly useful for self-signed certificates. This new option is disabled by default. When it's enabled the agent raises a warning in the collector.

### Motivation

It was requested by a user.

### Testing Guidelines

🙈 

### Additional Notes

👯

### DIFF
--- a/checks.d/mesos_master.py
+++ b/checks.d/mesos_master.py
@@ -6,11 +6,15 @@
 
 Collects metrics from mesos master node, only the leader is sending metrics.
 """
+# stdlib
+from urlparse import urlparse
+
 # 3rd party
 import requests
 
 # project
 from checks import AgentCheck, CheckException
+from config import _is_affirmative
 
 
 class MesosMaster(AgentCheck):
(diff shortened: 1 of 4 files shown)


checks.d/mesos_master.py
```diff
@@ -190,7 +194,13 @@ def check(self, instance):
         default_timeout = self.init_config.get('default_timeout', 5)
         timeout = float(instance.get('timeout', default_timeout))
 
-        state_metrics = self._check_leadership(url, timeout)
+        no_ssl = _is_affirmative(instance.get('disable_ssl_validation', False))
+        ssl_verify = False if no_ssl else True
+        parsed_url = urlparse(url)
+        if no_ssl and parsed_url.scheme == 'https':
+            self.log.warning('Skipping SSL cert validation for %s based on configuration.' % url)
```
yannmh: I don't think this warning is necessary: based on the default option value, the user explicitly set this behavior.

I am concerned this adds too much noise to the logs (the warning would be repeated at every check run).

hkaj: I could add a `ignore_ssl_warning` setting like in the http_check (which has the exact same behavior)? It's something important to remember so we should probably log something.

yannmh: Your choice. 

IMO the `http_check` validates the endpoint health itself so it makes sense in this case.

degemer: I agree with Yann, logging it once (first check run) would make sense, but it would otherwise spam the logs for nothing. 😢  (not saying skipping SSL cert is not a big deal, but I hate it when programs remind every 5min (in this case every 20s) of a decision I made, fully aware of the consequences).

checks.d/mesos_master.py
```diff
@@ -190,7 +194,13 @@ def check(self, instance):
         default_timeout = self.init_config.get('default_timeout', 5)
         timeout = float(instance.get('timeout', default_timeout))
 
-        state_metrics = self._check_leadership(url, timeout)
+        no_ssl = _is_affirmative(instance.get('disable_ssl_validation', False))
+        ssl_verify = False if no_ssl else True
+        parsed_url = urlparse(url)
```
yannmh: Or
```python
ssl_verify = not _is_affirmative(instance.get('disable_ssl_validation', False))
```

yannmh:
Overall, It looks good to me. I simply added a few nitpicks.

checks.d/mesos_master.py
hkaj: I feel the same way, but I wish all users made decisions fully aware of the consequences like you 😛 
Anyway yeah I'll log it in __init__ so it doesn't spam but still warns the user once in a while

hkaj:
@yannmh thanks for having a look! Updated according to your comments

yannmh:
👍

PR #2811, opened 2016-09-06
[ci] we know, flare is broken

degemer:
No need to keep the CI broken.
It's being worked on.

### DIFF
--- a/tests/core/test_flare.py
+++ b/tests/core/test_flare.py
@@ -1,5 +1,5 @@
 # stdlib
-import os.path
+import os
 import unittest
 import re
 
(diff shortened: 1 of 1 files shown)


yannmh:
👍 
Can you prepare a PR with a revert please ?

tests/core/test_flare.py
```diff
@@ -168,21 +168,6 @@ def test_upload_with_case_proxy(self, mock_config, mock_tempdir, mock_stfrtime,
         assert kwargs['proxies']['https'] == 'http://proxy_user:proxy_pass@proxy.host.com:3128'
         assert not kwargs['verify']
 
-    @attr(requires='core_integration')
```
truthbk: why don't we just add a `skip` attribute for when these things happen?

tests/core/test_flare.py
```diff
@@ -1,11 +1,12 @@
 # stdlib
-import os.path
+import os
 import unittest
 import re
 
 # 3p
 import mock
 from nose.plugins.attrib import attr
+from unittest.case import SkipTest
```
yannmh: #stdlib ?

degemer: 😵

PR #2815, opened 2016-09-06
Reset generator if dogstreams can't read a file

### LINKED ISSUE #2261: Dogstream fails to monitor log files if they don't exist at startup
If a log file is not preset at startup, dogstreams will fail to monitor the file if it ever appears.

The issue appears to be that `self._gen` in `datadog.py` is never unset on an exception. `TailFile.tail()` raises `StopIteration` if the file files to open, but this is thrown on the first call to `self._gen.next()` instead of when `tail()` is called. Future calls to `self._gen.next()` will just raise `StopIteration`.

I'm not sure if this is by design or not. I'm also not sure if unsetting the generator on an exception is always the right thing to do. However, a file that appears (or re-appears -- e.g. after log rotation) after datadog starts seems like a scenario that should be supported.

If this is not by design. I can prepare a pull request.

jalaziz:
### What does this PR do?

It resets the dogstream tail generator when an exception occurs. This allows the generator to be recreated on the next check.
### Motivation

Previously, once the generator failed, it would never be re-created. This would mean that if a file did not exist when the generator was created, it would never be checked.

Fixes #2261

### DIFF
--- a/checks/datadog.py
+++ b/checks/datadog.py
@@ -186,6 +186,8 @@ class Dogstream(object):
             except StopIteration as e:
                 self.logger.exception(e)
                 self.logger.warn("Can't tail %s file" % self.log_path)
+                # reset generator to try again during the next check interval
+                self._gen = None
 
             check_output = self._aggregate(self._values)
             if self._events:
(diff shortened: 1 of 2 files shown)


kyontan:
@truthbk How is the status of this?
Looks travis-ci failes not related to this pr tried to changes.

nagas:
https://github.com/DataDog/dd-agent/pull/3480

jalaziz:
I've rebased this to master (nothing changed). The intermittent test failures from before seems to be gone now.

PR #2817, opened 2016-09-07
[network] Add monotonic counts for some metrics

### LINKED ISSUE #2630: Should system.net.tcp.retrans_segs actually use a monotonic count?
The documentation seems to indicate that `system.net.tcp.retrans_segs` and similar metrics are measured as gauges. However, it seems that it's actually being measured as a rate in the `network` check. I understand that rates are stored as gauges, and that a rate is simply storing the difference between the previous and current value, but since it's time normalized it can result in odd values.

For example, we typically see things like 10.62 retransmitted segments even when using a `rollup` with `sum`.  This could also simply be to due to the fact that the measured interval using for normalization could be different than the flush interval pass into the `flush` method.

Using a monotonic count seems more appropriate here. Although, I could also see the rate being useful as well.

jalaziz:
### What does this PR do?

Add monotonic counts for tcp segments and udp datagrams. This allows more
precise counting of incoming and outgoing segments and datagrams for a given
time period.

The following metrics are affected:
- system.net.tcp.retrans_segs
- system.net.tcp.in_segs
- system.net.tcp.out_segs
- system.net.udp.in_datagrams
- system.net.udp.no_ports
- system.net.udp.in_errors
- system.net.udp.out_datagrams
- system.net.udp.rcv_buf_errors
- system.net.udp.snd_buf_errors
- system.net.tcp.retrans_packs
- system.net.tcp.sent_packs
- system.net.tcp.rcv_packs

For backwards compatibility, the new metrics are suffixed with `_count`.
### Motivation

This fixes #2630. It allows graphing and alerting on the precise number of segments, datagrams, or errors in a set time period.

### DIFF
--- a/checks.d/network.py
+++ b/checks.d/network.py
@@ -169,7 +169,9 @@ class Network(AgentCheck):
             for regex, metric in regex_list:
                 value = re.match(regex, line)
                 if value:
-                    self.rate(metric, self._parse_value(value.group(1)))
+                    parsed_value =  self._parse_value(value.group(1))
+                    self.rate(metric, parsed_value)
+                    self.monotonic_count('{}_count'.format(metric), parsed_value)
 
     def _check_linux(self, instance):
         proc_location = self.agentConfig.get('procfs_path', '/proc').rstrip('/')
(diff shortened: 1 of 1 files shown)


jalaziz:
Moved to DataDog/integrations-core#1551

PR #2820, opened 2016-09-07
Fix regression test; can't get tag when the connection's not actually…

derekwbrown:
### What does this PR do?

Fixes unit test that was checking for the new cluster tag on the test to a bad port.


yannmh:
We are not using the same tag set for service checks as a function of the `status`. This is actually an issue.
For instance, the following monitor would only return `OK` (best case  a `NO DATA` if set):
```
elasticsearch.can_connect by {cluster_name}
```

In short: let's remove this tag from all service checks. 

cc @degemer

degemer:
Tests are 💚 , sky is 🔵  , let's merge this!

PR #2822, opened 2016-09-07
[core] Hostname logic refactoring

remh:
### What does this PR do?

This PR:
- Moves hostname logic in it’s own util module
- Moves cloud metadata logic in it’s own util module
- Add a new way to determine the hostname using the kubelet
- Some code cleaning/consolidating

### Motivation

Needed to have another way to find the hostname when running kubernetes but i realized that this logic could benefit from refactoring so i did.

### DIFF
--- a/utils/hostname.py
+++ b/utils/hostname.py
@@ -1,0 +1,116 @@
+# (C) Datadog, Inc. 2010-2016
+# All rights reserved
+# Licensed under Simplified BSD License (see LICENSE)
+
+# stdlib
+import logging
+import re
+import socket
+
+# project
+from utils.cloud_metadata import EC2, GCE
+from utils.dockerutil import DockerUtil
+from utils.kubeutil import KubeUtil
+from utils.platform import Platform
+from utils.subprocess_output import get_subprocess_output
+
+VALID_HOSTNAME_RFC_1123_PATTERN = re.compile(r"^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])$")
+MAX_HOSTNAME_LEN = 255
+
+log = logging.getLogger(__name__)
+
+def is_valid_hostname(hostname):
+    if hostname.lower() in set([
+        'localhost',
+        'localhost.localdomain',
+        'localhost6.localdomain6',
+        'ip6-localhost',
+    ]):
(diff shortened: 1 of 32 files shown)


remh:
Forgot to mention that it also switched the metadata collection from GCE and AWS to `requests` instead of `urllib`

utils/hostname.py
```diff
@@ -0,0 +1,117 @@
...
+    # first, try the config
+    if config is None:
+        from config import get_config
+        config = get_config(parse_args=True)
+    config_hostname = config.get('hostname')
+    if config_hostname and is_valid_hostname(config_hostname):
+        return config_hostname
+
+    # Try to get GCE instance name
+    if hostname is None:
```
degemer: `hostname` will always be None

utils/hostname.py
```diff
@@ -0,0 +1,117 @@
...
+        gce_hostname = GCE.get_hostname(config)
+        if gce_hostname is not None:
+            if is_valid_hostname(gce_hostname):
+                return gce_hostname
+
+    # Try to get the docker hostname
+    if hostname is None and Platform.is_containerized():
```
degemer: Same

utils/hostname.py
remh: ha good point, it was the old logic though that just got moved but yeah

utils/hostname.py
```diff
@@ -0,0 +1,117 @@
...
+
+    # then move on to os-specific detection
+    if hostname is None:
+        if Platform.is_unix() or Platform.is_solaris():
+            unix_hostname = _get_hostname_unix()
+            if unix_hostname and is_valid_hostname(unix_hostname):
+                hostname = unix_hostname
+
+    # if we have an ec2 default hostname, see if there's an instance-id available
+    if (Platform.is_ecs_instance()) or (hostname is not None and EC2.is_default(hostname)):
```
degemer: I really don't like the `hostname = ...` parts, it makes this function way harder to understand than it needs to be.
Maybe not the scope of this PR, but it should be cleaned at some point.

degemer:
It looks way better now ! 🍰 

Only nitpicks, looks good otherwise.

utils/hostname.py
remh: Agreed, although i don't want to spend time changing the logic here as it's not the scope of this PR.
-> This would need to be way more tested
-> We'll have to do it for agent 6 anyway

remh:
Fixed the nitpicks

remh:
Tests are failing on the bad ES tests, merging.

PR #2823, opened 2016-09-08
[disk] another timeout on disk usage

gmmeyer:
### What does this PR do?

In #2714, we solved a problem with disk usage where nfs mounts would hang. In some cases, we're still seeing them hanging, because `os.statvfs` is being called in another place as well.

This adds a timeout wrapper to that second call, as well.

### DIFF
--- a/checks.d/disk.py
+++ b/checks.d/disk.py
@@ -187,7 +187,19 @@ class Disk(AgentCheck):
 
     def _collect_inodes_metrics(self, mountpoint):
         metrics = {}
-        inodes = os.statvfs(mountpoint)
+        # we need to timeout this, too.
+        try:
+            inodes = timeout(5)(os.statvfs)(mountpoint)
+        except TimeoutException:
+            self.log.warn(
+                u"Timeout while retrieving the disk usage of `%s` mountpoint. Skipping...",
+                mountpoint
+            )
+            return metrics
+        except Exception as e:
+            self.log.warn("Unable to get disk metrics for %s: %s", mountpoint, e)
+            return metrics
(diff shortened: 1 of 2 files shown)


checks.d/disk.py
```diff
@@ -187,7 +187,19 @@ def _collect_part_metrics(self, part, usage):
...
+        except TimeoutException:
+            self.log.warn(
+                u"Timeout while retrieving the disk usage of `%s` mountpoint. Skipping...",
+                mountpoint
+            )
+            return metrics
+        except Exception as e:
+            self.log.warn("Unable to get disk metrics for %s: %s", mountpoint, e)
+            return metrics
+
```
yannmh: I believe returning an empty dictionary would cause a `KeyError` exception [here](https://github.com/DataDog/dd-agent/blob/greg/statvfs-fix/checks.d/disk.py#L130).

Alternatively we could catch the exception within the `collect_metrics_psutil` method.

gmmeyer: @yannmh I think you're looking at the wrong thing. These are the `system.fs.inodes.*` metrics. They're not referenced anywhere else.

truthbk:
I think @gmmeyer is right, this should not have any problems with an empty dictionary: https://github.com/DataDog/dd-agent/blob/greg/statvfs-fix/checks.d/disk.py#L184

truthbk:
Tests broke though, let's get CI back to 💯

truthbk:
Thanks @gmmeyer, this looks good to me. Can you please squash before merging? Thanks a lot.

PR #2795, opened 2016-08-30
[ci] only skip tests for Travis PR

degemer:
### What does this PR do?

Only tries to skip tests for Travis PR.
And a smarter commit comparison.

### Motivation

Don't skip tests on master.

PR #2828, opened 2016-09-08
[ci] use proper names for badly named checks

degemer:
### What does this PR do?

Rename badly named checks to their correct Travis names when trying to determine whether a check should be skipped.

### Motivation

Tests did not run for https://github.com/DataDog/dd-agent/pull/2820, because of https://github.com/DataDog/dd-agent/pull/2791 and https://github.com/DataDog/dd-agent/pull/2795, breaking master tests. It cannot happen again.


ci/common.rb
```diff
@@ -42,6 +42,26 @@ def travis_pr?
   !ENV['TRAVIS'].nil? && ENV['TRAVIS_EVENT_TYPE'] == 'pull_request'
 end
 
+BAD_CITIZENS = {
+  'couch' => 'couchdb',
+  'disk' => 'sysstat',
```
olivielpeau: this specific exception is confusing IMO, and could be avoided... How about we add a `disk` flavor to the travis test matrix instead?

olivielpeau:
Looks good to me apart from one comment!

@truthbk: you might want to have a look too since you reviewed the original PRs :)

ci/common.rb
degemer: You're right, but but I'm not sure that adding the overhead of a new test in the matric is worth it: the disk check doesn't require any install. https://github.com/DataDog/dd-agent/blob/master/tests/checks/integration/test_disk.py

It could event be part of `core_integration`: thoughts ?

olivielpeau: I agree on the overhead of adding a new flavor, we should avoid that. Not sure about adding it to `core_integration` though since the check is not really part of the core of the Agent (and we always make the distinction between core of the agent and checks in our test flavors)

Another possibility is to consolidate under the same flavor all the check integration tests that don't need any specific install steps. AFAIK that would include the `disk`, `sysstat`, `network` and `docker_daemon` tests. This could have the drawback of adding confusion to the flavor names though (especially for the `docker_daemon` one, it's not intuitive to have it grouped with these other tests)

OR, we could just keep the `sysstat` flavor as it is but add the `network` check test to it, so that the `sysstat` flavor aggregates all the tests on the "system" checks that don't need any install. Thoughts?

degemer:
@olivielpeau finally implemented this change, sorry for the delay. Could you take one more look please ?

PR #2829, opened 2016-09-08
[dockerutil] cpu subsys will be mistaken for cpuacct, this fixes that.

truthbk:
### What does this PR do?

If we attempt to grab metrics from the `cpu` subsystem as is the case with the `throttled` metric introduced here #2724, we will currently fail to identify the correct `cpu.stat` file because the `cpu` subsystem will be mistaken for the `cpuacct` subsystem or similar. This PR attempts to address the issue for this and future metrics.

Also fixes badly indented function.

### Motivation
Found the bug while testing #2724.

### Testing Guidelines

### Additional Notes

### DIFF
--- a/utils/dockerutil.py
+++ b/utils/dockerutil.py
@@ -224,43 +224,43 @@ class DockerUtil:
         return mountpoints
 
     def find_cgroup(self, hierarchy):
-            """Find the mount point for a specified cgroup hierarchy.
+        """Find the mount point for a specified cgroup hierarchy.
 
-            Works with old style and new style mounts.
+        Works with old style and new style mounts.
 
-            An example of what the output of /proc/mounts looks like:
+        An example of what the output of /proc/mounts looks like:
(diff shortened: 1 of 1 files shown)


utils/dockerutil.py
```diff
@@ -279,7 +279,7 @@ def find_cgroup_from_proc(cls, mountpoints, pid, subsys, docker_root='/'):
         if subsys in subsystems:
             for mountpoint in mountpoints.itervalues():
                 stat_file_path = os.path.join(mountpoint, subsystems[subsys])
-                if subsys in mountpoint and os.path.exists(stat_file_path):
+                if subsys == mountpoint.split('/')[-1] and os.path.exists(stat_file_path):
```
hkaj: good catch, you saved us hours of wondering why the metrics don't match! 🍪

hkaj:
LGTM

PR #2830, opened 2016-09-08
[proc] fix to search by pid.

truthbk:
### What does this PR do?

While testing for #2119, I found a few problems. `pids` should be a set as that's what is expected by `get_process_state()`. It also expects a set of integers, not `psutil.Process`es. This short PR addresses the two.

### Motivation
Found the bug while testing for the feature.


degemer:
Nice catch! LGTM

PR #2833, opened 2016-09-09
[forwarder] drop payloads when no valid endpoint

degemer:
### What does this PR do?

Forwarder drops payloads when it receives empty endpoints.

### Motivation

This was the previous behaviour, let's do the same. See https://github.com/DataDog/dd-agent/pull/2564.

### DIFF
--- a/config.py
+++ b/config.py
@@ -325,6 +325,10 @@ def clean_dd_url(url):
     return url[:-1] if url.endswith('/') else url
 
 
+def remove_empty(string_array):
+    return filter(lambda x: x, string_array)
+
+
 def get_config(parse_args=True, cfg_path=None, options=None):
     if parse_args:
         options, _ = get_parsed_args()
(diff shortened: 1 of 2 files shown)


ddagent.py
```diff
@@ -398,6 +398,8 @@ def __init__(self, port, agentConfig, watchdog=True,
         self._metrics = {}
         AgentTransaction.set_application(self)
         AgentTransaction.set_endpoints(agentConfig['endpoints'])
+        if agentConfig['endpoints'] == {}:
```
olivielpeau: it'd be good to check the case when there is an endpoint but no attached api_keys and log the same warning when that happens

(otherwise in that case the forwarder drops all incoming payloads but doesn't really explain why)

config.py
```diff
@@ -325,6 +325,14 @@ def clean_dd_url(url):
     return url[:-1] if url.endswith('/') else url
 
 
+def remove_empty(string_array):
+    emptyless = []
+    for el in string_array:
+        if el:
+            emptyless.append(el)
+    return emptyless
```
olivielpeau: we could shorten the function to `return filter(lambda x: x, string_array)`, but it might be less readable, up to you

olivielpeau:
Had a few small comments but overall I like this approach! 🍰

ddagent.py
degemer: 🚀

degemer: Done in `config.py`.

olivielpeau:
:shipit: 

Travis failure is unrelated, having a look into fixing it

PR #2835, opened 2016-09-12
[dockerutil] cpu stats might not always fall into the expected "subsystem" match

truthbk:
### What does this PR do?

This PR fixes a problem that raises as a product of #2724. There are two approaches to this problem. One is what we're doing here, the other is to modify https://github.com/DataDog/dd-agent/blob/master/checks.d/docker_daemon.py#L73 such that the `cgroup` is set to `cpuacct` instead of `cpu`. 

The second approach is a simple one word change and it actually matches the behavior I have seen with Xenial and CentOS7 (still untested on other distros). However #2724 does come from a contributor PR and it also happens to be undeniable that `cpu` is a _legit_ cgroup. So I'm unsure about what may happen in other distributions or with people that run custom cgroups.

I have found the `cpu.stat` file to be available in the `cpuacct` (or `cpu,cpuacct` cgroup stats). I'd love to hear any feedback regarding this.

### Motivation

During testing for the upcoming release I found this to be problematic on Xenial as well as CentOS7. So this comes as a bugfix to the behavior met.

### Testing Guidelines

### DIFF
--- a/utils/dockerutil.py
+++ b/utils/dockerutil.py
@@ -276,6 +276,14 @@ class DockerUtil:
                     subsys = _subsys
                     break
 
+        # In Ubuntu Xenial, we've encountered containers with no `cpu`
+        # cgroup in /proc/<pid>/cgroup
+        if subsys == 'cpu' and subsys not in subsystems:
+            for sub, mountpoint in subsystems.iteritems():
+                if 'cpuacct' in sub:
+                    subsystems['cpu'] = mountpoint
+                    break
+
         if subsys in subsystems:
             for mountpoint in mountpoints.itervalues():
                 stat_file_path = os.path.join(mountpoint, subsystems[subsys])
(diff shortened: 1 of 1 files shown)


utils/dockerutil.py
```diff
@@ -276,6 +276,14 @@ def find_cgroup_from_proc(cls, mountpoints, pid, subsys, docker_root='/'):
                     subsys = _subsys
                     break
 
+        # In Ubuntu Xenial, we've encountered containers with no `cpu`
+        # cgroup in /proc/<pid>/cgroup
+        if subsys == 'cpu' and subsys not in subsystems:
+            for s, m in subsystems.iteritems():
```
remh: nitpick but `s` and `m` are not great variable names

utils/dockerutil.py
```diff
@@ -284,7 +292,7 @@ def find_cgroup_from_proc(cls, mountpoints, pid, subsys, docker_root='/'):
 
                 # CentOS7 will report `cpu,cpuacct` and then have the path on
                 # `cpuacct,cpu`
-                if 'cpuacct' in mountpoint and 'cpuacct' in subsys:
+                if 'cpuacct' in mountpoint and any(k in subsys for k in ['cpuacct', 'cpu']):
```
remh: tiny nitpick but it's better to use a set instead of a list for that

remh: also as you have just two options it would be more readable to just do
```python
if 'cpuacct' in mountpoint and ('cpuacct' in subsys or 'cpu' in subsys):
```

remh:
That looks good besides the nitpick

utils/dockerutil.py
truthbk: I could see this coming :blush: first because I think `s` is already reserverd in `pdb`, and also well, because they're single characters. But it had such a limited scope, and I couldn't think of of great alternatives to `subsys` and `subsystem` at the time... Fixing.

PR #2836, opened 2016-09-12
Filter containers consistently in docker and kubernetes

hkaj:
### What does this PR do?

This PR moves the container exclusion logic of the docker check to dockerutil so it can be used for other docker related checks. It also uses it in the Kubernetes check and excludes k8s pause containers by default.
### Motivation

Pause containers in Kubernetes are numerous and not interesting to monitor. Making the exclusion system consistent across all container checks allows to filter them out easily (also made them excluded by default).

### Additional Notes

It would be nice to not collect anything about excluded containers. Right now the docker check still collects some things, we should discuss that.


therc:
Thanks for doing this. Pause containers are so annoying.

remh:
```
./checks.d/docker_daemon.py:21:1: F401 'get_config_store' imported but unused
Failed task: RuntimeError Command failed with status (1): [flake8...]
```

remh:
Looks good to me beside the small flake8 error. I fixed it

PR #2839, opened 2016-09-13
Add timeout for fluentd collector

mattrobenolt:
### What does this PR do?

Adds a timeout argument to fluentd collector.

### Motivation

Our datadog-agent process hung when fluentd was having issues since the socket never timed out.

### DIFF
--- a/checks.d/fluentd.py
+++ b/checks.d/fluentd.py
@@ -16,6 +16,7 @@ from util import headers
 
 
 class Fluentd(AgentCheck):
+    DEFAULT_TIMEOUT = 5
     SERVICE_CHECK_NAME = 'fluentd.is_ok'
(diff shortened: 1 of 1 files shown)


yannmh:
Thanks for bringing the change @mattrobenolt.

One nitpick, can we move the default timeout evaluation to  the`__init__` method (similar https://github.com/DataDog/dd-agent/blob/master/checks.d/dns_check.py#L28-L30 to for example)? i.e.

```python
        self.default_timeout = init_config.get('default_timeout', self.DEFAULT_TIMEOUT)
```

yannmh:
@mattrobenolt I made the small change and opened a new PR: https://github.com/DataDog/dd-agent/pull/2978. This will be part of the Datadog Agent `5.10.0`.

PR #2843, opened 2016-09-13
Initial test code for support case where agent crashes with unicode i…

derekwbrown:
Intermediate fix for customer where some data is causing collector to crash due to incorrect type passed to 'return_control_chars'.

This code does a deep search to translate strings one by one (it is very expensive).

Also logs the offending string(s) so we can determine where the offending strings are coming from.


emitter.py
```diff
@@ -31,6 +31,24 @@
...
+        return newdict
+    if isinstance(item, list):
+        newlist = []
+        for listitems in item:
+            newlist.append(remove_control_chars_from(listitems))
+        return newlist
+    if(isinstance(item, str)):
+        newstr = remove_control_chars(item)
+        if item != newstr:
+            log.warning('changed string: ' + newstr)
```
remh: shouldn't you return `newstr` here ?

remh:
Looks nice but:

- You should add unit tests on that function
- You should catch possible `RuntimeError` in (the unlikely) case the max recursion depth is reached

tests/core/test_emitter.py
```diff
@@ -17,3 +18,23 @@ def test_remove_control_chars(self):
...
+            ]
+
+        def is_converted_same(msg):
+            new_msg = remove_control_chars_from(msg, None)
+            if str(new_msg) == str(msg):
+                return True
+            return False
+
+        for msg in bad_messages:
+            self.assertFalse(is_converted_same(msg))
```
remh: you should also test for an expected result i.e. in this case, the function could return something completely wrong and this test would still pass.

You should also add a few more cases but otherwise 👍

truthbk:
I added a commit to this PR so that we also handle actual unicode control characters and not just ASCII ones. Current test suite passes, but I haven't really bothered adding new test cases - just tweaked the existing ones so that unicode is accounted for.

emitter.py
```diff
@@ -29,7 +30,13 @@
 
 
 def remove_control_chars(s):
-    return control_char_re.sub('', s)
+    if isinstance(s, str):
+        sanitized = control_char_re.sub('', s)
+    elif isinstance(s, unicode):
+        sanitized = ''.join(['' if unicodedata.category(c) == 'Cc' else c
```
degemer: Is `Cc` the only category failing ? http://www.fileformat.info/info/unicode/category/index.htm

emitter.py
```diff
@@ -29,8 +30,34 @@
...
-
+    if isinstance(s, str):
+        sanitized = control_char_re.sub('', s)
+    elif isinstance(s, unicode):
+        sanitized = ''.join(['' if unicodedata.category(c) == 'Cc' else c
+                            for c in u'{}'.format(s)])
+
+    return sanitized
+
+def remove_control_chars_from(item, log):
```
degemer: Why not set the `log=None` by default ?

emitter.py
```diff
@@ -29,8 +30,34 @@
...
+    if isinstance(item, dict):
+        newdict = {}
+        for k, v in item.iteritems():
+            newval = remove_control_chars_from(v, log)
+            newkey = remove_control_chars(k)
+            newdict[newkey] = newval
+        return newdict
+    if isinstance(item, list):
+        newlist = []
+        for listitems in item:
```
degemer: Nit: `listitems` => `listitem`.

emitter.py
```diff
@@ -40,10 +67,22 @@ def http_emitter(message, log, agentConfig, endpoint):
...
-    except UnicodeDecodeError:
-        message = remove_control_chars(message)
-        payload = json.dumps(message)
+        try:
+            payload = json.dumps(message)
+        except UnicodeDecodeError:
+            newmessage = remove_control_chars_from(message, log)
+            payload = json.dumps(newmessage)
+    except UnicodeDecodeError as ude:
+        log.error('http_emitter: Unable to convert message to json %s ' % str(ude))
```
degemer: What do you think of `log.exception('http_emitter: Unable to convert message to json %s', ude)` - https://www.simonmweber.com/2014/11/24/python-logging-traps.html ?

emitter.py
truthbk: I'm not sure if it's the only category that might cause `json.dumps()` to fail, I do know it's the equivalent of the ASCII portion of remove_control_chars()... I could also remove `Cf`, which are likely also dangerous. I do that in the `github` and `bitbucket` integrations.

truthbk:
flaky tests - failures are unrelated. Tried a couple times, but couldn't get it all green. Merging.

PR #2844, opened 2016-09-13
[cloud metadata] Fail on non 2xx responses

remh:
### What does this PR do?

To keep the same behavior as before we need to make sure not to put bad
data in the metadata if the metadata endpoint returns a non 2xx
response code (that was already the case with urllib)

This fixes a bug that was introduced with https://github.com/DataDog/dd-agent/pull/2822

### Testing Guidelines

Confirmed that it fixed the issue i saw on Google Compute nodes.

### DIFF
--- a/utils/cloud_metadata.py
+++ b/utils/cloud_metadata.py
@@ -34,12 +34,15 @@ class GCE(object):
             return GCE.metadata
 
         try:
-            GCE.metadata = requests.get(
+            r = requests.get(
                 GCE.URL,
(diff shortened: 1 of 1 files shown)


truthbk:
This looks good. You don't want to log anything so that it fails less silently?

remh:
I can log something in debug, but in all cases some of these calls are expected to fail. I'll add something in debug. Thanks for the feedback

remh:
@truthbk done

truthbk:
Nice! :shipit:

PR #2848, opened 2016-09-15
[kubernetes] Add ability to specify service discovery via kubernetes annotations

### LINKED ISSUE #2794: [kubernetes] Use pod annotations for service discovery
Service discovery looks perfect for monitoring things running in kube. However, setting up another etcd cluster and keeping it up-to-date with deployments would get annoying.

Kubernetes lets you store arbitrary metadata in the `annotations` on a pod, which could include the datadog check config. That would be a pretty lean and powerful setup - you could switch monitoring configurations for your `Deployment` and get a rolling deploy of it. The metadata is stored in etcd at the end, so it's essentially the same setup. The nice part is there are less moving parts - you can reuse your existing "push to prod" logic to push monitoring logic.

mikekap:
### What does this PR do?

This change makes the docker service discovery read the kubernetes annotations
to discover how to monitor a pod. This behavior is only triggered when a
service discovery backend isn't set. The 3 annotations looked for are:
 - `com.datadoghq.sd/check_names`
 - `com.datadoghq.sd/init_configs`
 - `com.datadoghq.sd/instances`

The semantics are exactly the same as that of a KV store.

Fixes #2794

### Motivation

#2794 has more motivation. This allows not running a KV store and having multiple versions of monitoring configuration rolled out via standard deployment mechanisms.

### Testing Guidelines

Added a unit test.

### Additional Notes

I don't think there are any integration tests for service discovery, especially on kube. Either way, I'll be running this on my own kube cluster until this is upstreamed.

### DIFF
--- a/checks.d/docker_daemon.py
+++ b/checks.d/docker_daemon.py
@@ -18,7 +18,6 @@ from utils.dockerutil import DockerUtil, MountException
 from utils.kubeutil import KubeUtil
 from utils.platform import Platform
 from utils.service_discovery.sd_backend import get_sd_backend
-from utils.service_discovery.config_stores import get_config_store
 
 
 EVENT_TYPE = 'docker'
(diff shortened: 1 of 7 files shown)


checks.d/docker_daemon.py
```diff
@@ -161,13 +160,7 @@ def init(self):
             instance = self.instances[0]
 
             # if service discovery is enabled dockerutil will need a reference to the config store
```
hkaj: you can remove this comment now

hkaj:
Hey @mikekap 
This actually fixes an issue we have in the current RC so I went ahead and merged it. Thanks again!

PR #2849, opened 2016-09-15
[dns_check] support monitoring the performance of NXDOMAIN queries

jnewland:
### What does this PR do?

Adds support to `dns_check` for a new `record_type`: `NXDOMAIN`. If specified, an NXDOMAIN result is required from the query to return "success", and response time data is reported.
### Motivation

In many DNS systems, NXDOMAIN results are uncached. Further, a query for a unqualified domain name that one expects to return an NXDOMAIN result can result in many dns queries, depending on the resolver's configured search domain. For these reasons, queries that are expected to return NXDOMAIN are a good candidate for simulating the worst-case performance of a DNS lookup on any given instance. The worst-case performance of a DNS lookup is something we'd like to monitor, and this change provides a way to do so with Datadog.
### Testing Guidelines

Tests have been added for two cases:
- Confirming that a check instance configured with `record_type: NXDOMAIN` generates metrics in the case of an `NXDOMAIN` response.
- Confirming that a check instance configured with `record_type: NXDOMAIN` returns a critical check value in the event of a non-`NXDOMAIN` response.

### DIFF
--- a/conf.d/dns_check.yaml.example
+++ b/conf.d/dns_check.yaml.example
@@ -18,6 +18,15 @@ instances:
       #   - tag:one
       #   - tag:two
 
-    - hostname: nxdomain.example.org
-      nameserver: 127.0.0.1
-      record_type: NXDOMAIN
+    # If you use NXDOMAIN as the `record_type`, an NXDOMAIN result is expected from the query,
+    # and the check instance will report response time data.
+    # In many DNS systems, NXDOMAIN results are uncached. Further a query for a unqualified domain
+    # name that one expects to return an NXDOMAIN result can result in many dns queries, depending
+    # on the resolver's configured search domain.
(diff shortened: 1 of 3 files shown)


conf.d/dns_check.yaml.example
```diff
@@ -9,3 +9,7 @@ instances:
       # Specify an (optional) `record_type` to customize the record type
       # queried by the check (default: "A")
       # record_type: A
+
+    - hostname: nxdomain.example.org
```
olivielpeau: Could you add here a short explanation of why/how one would use an `NXDOMAIN` `record_type`, since its behavior is quite different from regular `record_type`s? (a summarized version of your PR comment would be perfect)

olivielpeau:
Rebased these changes on top of master and added the explanation to the yaml example file. Looks ready for a merge now.

Thanks again @jnewland!

PR #2851, opened 2016-09-16
[ceph] Fix check failure when health detail contains non-osd info

olivielpeau:
### What does this PR do?

Avoid check failure when the `detail` key of `health detail` contains info on non-osd components. The bug was introduced in https://github.com/DataDog/dd-agent/pull/2805

### Motivation

Improve check resilience.

### Testing

PR includes a test case that reproduces the issue.

### Additional Notes

The `detail` key of `health detail` can contain info on the health of
non-osd units, for example:

```
mon.vagrant-ubuntu-trusty-64 low disk space -- 18% avail
```

In that case the check would previously fail with an IndexError because
the regex we use wouldn't match this string.

Ignore the entry instead.

cc @vagelim

### DIFF
--- a/checks.d/ceph.py
+++ b/checks.d/ceph.py
@@ -91,14 +91,15 @@ class Ceph(AgentCheck):
             else:
                 for osdhealth in raw['health_detail']['detail']:
                     osd, pct = self._osd_pct_used(osdhealth)
-                    local_tags = tags + ['ceph_osd:%s' % osd.replace('.','')]
+                    if osd:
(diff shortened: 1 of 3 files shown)


masci:
Sorry for not spotting the issue on #2805 😿 
LGTM 👍

olivielpeau:
Thanks @masci! Merging

vagelim:
Sorry for not recognizing this issue when I modified the check! I had not encountered a situation like the one in your test data :(

PR #2852, opened 2016-09-16
[ceph] Improve health metrics

olivielpeau:
### What does this PR do?

* Catch `KeyError` on `osd_pool_stats`
* Improve health metrics: the `ceph.num_full_osds` and `ceph.num_near_full_osds` metrics report values that make more sense and add `ceph.osd.pct_used` metric (see explanation below).

### Motivation

* Check resilience
* Improved usability of the health metrics

### Testing

Updated the existing mock tests, and made them more detailed

#### health metrics

Previous behavior: the `ceph.num_near_full_osds` and
`ceph.num_full_osds` would either:
* take the value `0` and be tagged with no `osd` tag if no osd is
reporting health issues
* take a value representing the usage percentage, tagged by `osd`

This doesn't really make sense with respect to the name of the metrics
(`num_*`). To solve this, replace these metrics with the following:
* `ceph.num_near_full_ods` and `ceph.num_full_osds` report the total
number of osds that are respectively near full and full. Not tagged
by osd.
* when some osds report health issues, the check sends a
`ceph.osd.pct_used` metric which reports the usage
percentage, tagged by osd. Unfortunately we can't send `0` values
on this metric when no osd reports health issues since we can't tag by
osd in that case.

This should make these metrics more usable. Also, use `gauge` since
there's no reason to use `count`.

### DIFF
--- a/checks.d/ceph.py
+++ b/checks.d/ceph.py
@@ -103,44 +103,47 @@ class Ceph(AgentCheck):
         except KeyError:
             self.log.debug('Error retrieving health metrics')
 
-        for osdinfo in raw['osd_pool_stats']:
-            name = osdinfo.get('pool_name')
-            local_tags = tags + ['ceph_pool:%s' % name]
-            ops = 0
-            try:
-                self._publish(osdinfo, self.gauge, ['client_io_rate', 'read_op_per_sec'], local_tags)
-                ops += osdinfo['client_io_rate']['read_op_per_sec']
-            except KeyError:
-                osdinfo['client_io_rate'].update({'read_op_per_sec' : 0})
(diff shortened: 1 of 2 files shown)


olivielpeau:
cc @vagelim: Let me know what you think of these changes

vagelim:
This is a much better solution! I was unsure how to send this data to DD in a useful way. Please ping me when these changes have made it to prod, I will need to update the default screenboard accordingly.

olivielpeau:
Merging, thanks for the review!

@vagelim: this will be released with `5.9.0`, we'll ping you when it's out. Could you also update the metrics metadata with these new metrics if you get a chance? :)

vagelim:
The metadata: https://github.com/DataDog/dogweb/pull/15018

PR #2720, opened 2016-07-29
Advanced networking metrics

joewilliams:
*Note: Please remember to review the Datadog [Contribution Guidelines](https://github.com/DataDog/dd-agent/blob/master/CONTRIBUTING.md)
if you have not yet done so.*

### What does this PR do?

This changes and extends the network metrics that the dd-agent collects.

### Motivation

At GitHub we need more specific and low level network metrics collected.

### Testing Guidelines

An overview on [testing](https://github.com/DataDog/dd-agent/blob/master/tests/README.md)
is available in our contribution guidelines.

### Additional Notes

It's worth noting that this changes behavior of how the dd-agent counts TCP connection states. It splits them up into their original form rather than opening, closing, established. I believe the rest of this PR simply adds metrics without modifying behavior.

PR #2856, opened 2016-09-20
dont combine connection states

joewilliams:
### What does this PR do?

This changes the networking check to use the individual connection states rather than combine them. This PR is a more manageable version of https://github.com/DataDog/dd-agent/pull/2720. We have our own `nstat` based plugin that does the extended TCP metrics included in that PR but would like to split connection states into their actual buckets.
### Motivation

It bugs me that the states are added together.


gmmeyer:
Thanks a lot for the contribution @joewilliams!

First off, can you make sure to fix the tests that this breaks? 

Second, can you put this change behind a flag in the config file? We want to make sure to keep backwards compatibility. (I think the change is correct, but not everyone cares about the difference and we don't want to break anyone's setup.)

Can you clarify whether this would supercede your changes from #2720 or if they're both needed?

Thank you very much, again! I think this would be easily merged after these changes. 😄

gmmeyer:
Closing because this has been superseded by this PR: https://github.com/DataDog/dd-agent/pull/3158

PR #2857, opened 2016-09-20
[fixtures] restoring original datadog-agent-1.tar.bz2 tarball.

truthbk:
### What does this PR do?

Restores the tarball fixture to what it was originally.

### Motivation

Noticed the fixture had gone down to zero bytes after merging #2843 - not sure how that happened. Surprisingly no tests broke in the process. This PR brings back the original fixture.


gmmeyer:
👍

PR #2859, opened 2016-09-21
[docker_deamon] docker healthcheck as service checks.

truthbk:
### What does this PR do?

This PR allows us to send service checks according to the healthcheck reported by docker for containers. This is a new feature introduced with Docker `1.12`, and not all containers will implement it. Currently we will send an `UNKNOWN` status in that case. Maybe we should only submit the service check when available.

Because grabbing the health status requires inspecting the containers, and that's a potentially expensive call, for now this feature will be hid behind a feature flag `health_service_checks`.

Also added a simple test. More work could be done on this front.
### Motivation

Provides yet another point of view regarding a container's health, as intended by the container's maintainers. 

I have also noticed a couple CI/tooling issues - we can sneak them in here. Or address them in a separate PR which probably makes more sense. One of the issues affects contributors as there's a problem with the old `setuptools` and `rake setup_env`.

### DIFF
--- a/checks.d/docker_daemon.py
+++ b/checks.d/docker_daemon.py
@@ -22,6 +22,7 @@ from utils.service_discovery.sd_backend import get_sd_backend
 
 EVENT_TYPE = 'docker'
 SERVICE_CHECK_NAME = 'docker.service_up'
+HEALTHCHECK_SERVICE_CHECK_NAME = 'docker.container_health'
 SIZE_REFRESH_RATE = 5  # Collect container sizes every 5 iterations of the check
 MAX_CGROUP_LISTING_RETRIES = 3
 CONTAINER_ID_RE = re.compile('[0-9a-f]{64}')
@@ -218,8 +219,11 @@ class DockerDaemon(AgentCheck):
         # containers running with custom cgroups?
         custom_cgroups = _is_affirmative(instance.get('custom_cgroups', False))
 
+        # submit healtcheck service checks?
+        health_service_checks = _is_affirmative(instance.get('health_service_checks', False))
+
         # Get the list of containers and the index of their names
-        containers_by_id = self._get_and_count_containers(custom_cgroups)
+        containers_by_id = self._get_and_count_containers(custom_cgroups, health_service_checks)
         containers_by_id = self._crawl_container_pids(containers_by_id, custom_cgroups)
 
         # Send events from Docker API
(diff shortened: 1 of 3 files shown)


conf.d/docker_daemon.yaml.example
```diff
@@ -51,6 +51,12 @@ instances:
     #       enable if absolutely necessary.
     # custom_cgroups: false
 
+    # Report docker container healthcheck events as service checks
+    # Note: enabling this option modifies the way in which we inspect the containers and causes
+    #       some overhead - if you run a high volume of containers we may timeout. Please only
+    #       enable if absolutely necessary.
```
hkaj: I think you can remove the last sentence. The slow part of inspecting a container is the lock acquisition over said container, and this is also done in `docker ps`, so we may timeout the check but it could also happen without the inspect part.
Same for the `custom_cgroups` option actually, the last sentence should be dropped.

hkaj: There should also be a mention that this option is available starting from Docker 1.12 and enabling it with an older version will result in an UNKNOWN state for the service check. It should be obvious but I'd rather be extra explicit here.

hkaj:
Actually I wonder if we shouldn't ask for a list of container name instead and only enable service checks for those listed. This could get spammy, and health checks in dynamic clusters where containers cease to exist and appear very often this may not have a lot of value. What do you think?

checks.d/docker_daemon.py
```diff
@@ -268,6 +272,11 @@ def check(self, instance):
         if self.collect_disk_stats:
             self._report_disk_stats()
 
+        # Report docker healthcheck SC's where available
+        if health_service_checks:
+            health_scs_containers = instance.get('health_service_check_whitelist', [])
```
truthbk: Sorry forgot to push.

truthbk:
@hkaj so I tested this out... my only lingering question would be. Should we make the whitelist by image name (current behavior) or container name or both? Taking into account tags or not? Should we support wildcards? Let me know your thoughts - as you can see right now it's a straight match.

checks.d/docker_daemon.py
```diff
@@ -250,8 +251,11 @@ def check(self, instance):
         # containers running with custom cgroups?
         custom_cgroups = _is_affirmative(instance.get('custom_cgroups', False))
 
+        # submit healtcheck service checks?
+        health_service_checks = _is_affirmative(instance.get('health_service_checks', False))
```
hkaj: I think we don't need this parameter anymore. If the list is there, enable health checks, otherwise disable them. So you can look for the white list param directly, maybe check if it isn't empty.

joshenders:
Hi there, thanks a lot for this change. It looks like in the default `docker_daemon.yaml.example` config in version 5.19, events of type `['exec_start', 'exec_create']` are filtered which prevent healthcheck service check events from being created.

When I removed these filters I was able to see the healthchecks in the datadog

PR #2864, opened 2016-09-27
Extended networking metrics

ejholmes:
### What does this PR do?

Adds additional networking related metrics from /proc/net/netstat, /proc/net/snmp and /proc/net/udp under the `system.net` and `docker.net` namespaces.
### Motivation

The built-in network.py check doesn't give an accurate representation of TCP connection states, since it only represents the state at the time the metrics were collected. The raw metrics from /proc/net provide a much more granular and accurate view of networking.

For example, `/proc/net/netstat` gives a much more accurate representation for things like number of TCP connections opened, since it's a monotonically increasing counter within the Kernel.
### Additional Notes

There's a similar PR open at https://github.com/DataDog/dd-agent/pull/2720, but it's not Docker aware, which is a necessity for us at Remind. We've also chosen to use monotonic counters instead of rates, since that gives us a more accurate representation of how many events occurred in a time frame.

Hopefully we can combine ideas from both PR's, and get a more releasable version of this into `network.py`.


utils/network.py
```diff
@@ -0,0 +1,282 @@
...
+from checks import AgentCheck
+
+GAUGE = AgentCheck.gauge
+MONOTONIC_COUNT = AgentCheck.monotonic_count
+
+SNMP_METRICS = {
+    'Tcp': {
+        # The algorithm used to determine the timeout value used for
+        # retransmitting unacknowledged octets.
+        'RtoAlgorithm': ('tcpx.rto_algorithm', GAUGE),
```
ejholmes: We use `tcpx` only because these metrics are not in a release of dd-agent yet. Our preference would be for these to just be under `system.net.tcp` and `docker.net.tcp`.

checks.d/network_ext.py
```diff
@@ -0,0 +1,21 @@
...
+"""
+# stdlib
+import re
+
+# project
+from checks import AgentCheck
+from utils import network
+
+
+class NetworkExt(AgentCheck):
```
ejholmes: We have this as a separate check for now, but I think this would actually belong in the network.py check, possibly with a flag to enable extended metrics.

checks.d/docker_daemon.py
```diff
@@ -596,6 +597,8 @@ def _report_net_metrics(self, container, tags):
             # It is possible that the container got stopped between the API call and now
             self.warning("Failed to report IO metrics from file {0}. Exception: {1}".format(proc_net_file, e))
 
+        network.check_all(self, "docker.net", container['_proc_root'], tags=tags)
```
ejholmes: I think there's also a good chunk of code that could be shared between `network.py` and `docker_daemon.py` for extracting metrics from /proc/net.

utils/network.py
```diff
@@ -0,0 +1,282 @@
...
+def parse_proc_net_udp(check, netns, lines, ipv6=False, tags=[]):
+    columns = lines[0].strip().split()
+    for line in lines[1:len(lines)]:
+        parts = line.strip().split()
+        values = dict(zip(columns, parts[:4] + parts[4].split(':') + parts[5].split(':') + parts[6:]))
+        drops = int(values.get('drops', 0))
+        proto = "udpx"
+        if ipv6:
+            proto = "udpx6"
+        check.monotonic_count('%s.%s.drops' % (netns, proto), drops, tags=tags + ["inode:%s" % values.get('inode', 'unknown')])
```
ejholmes: This can produce a lot of individual metrics.

truthbk:
@ejholmes, I've taken the liberty of opening a PR here to help you transition to the SDK DataDog/integrations-extras#44

I created a patch with your work and dropped it that PR. Please feel free to continue your work there. 🙇

PR #2866, opened 2016-09-27
[docker] set alert_type to error when for oom and kill events

sjenriquez:
### What does this PR do?

This PR sets the event `alert_type` (Status in app) to `error` if a container reports `die` or `oom`.

### Motivation

Currently all Docker events are set to the default `alert_type: info`, this change will allow for additional filtering in the event stream and event monitoring for event types that may not be expected.

### DIFF
--- a/checks.d/docker_daemon.py
+++ b/checks.d/docker_daemon.py
@@ -120,6 +120,7 @@ IMAGE = "image"
 
 ECS_INTROSPECT_DEFAULT_PORT = 51678
 
+ERROR_ALERT_TYPE = ['oom', 'kill']
 
 def get_filters(include, exclude):
     # The reasoning is to check exclude first, so we can skip if there is no exclude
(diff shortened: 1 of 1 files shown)


checks.d/docker_daemon.py
```diff
@@ -646,6 +646,7 @@ def _format_events(self, aggregated_events, containers_by_id):
             status = defaultdict(int)
             status_change = []
             container_tags = set()
+            error_alert_type = ['die', 'oom']
```
hkaj: This should be a module constant imo.

checks.d/docker_daemon.py
```diff
@@ -672,6 +673,12 @@ def _format_events(self, aggregated_events, containers_by_id):
                 status_changes="\n".join(
                     ["%s \t%s" % (change[1].upper(), change[0]) for change in status_change])
             )
+
+            if any(e in status_text for e in error_alert_type):
```
hkaj: nitpicking because in this case it's not a real problem, but in general be careful with single-char variable names as some characters are used by [pdb](https://docs.python.org/2/library/pdb.html#debugger-commands), and it makes debugging difficult.

hkaj:
Thanks, LGTM now. You can rebase your changes to have only one commit and merge.

sjenriquez:
@hkaj Included `kill` as type `error`, can you have a look?

hkaj:
LGTM

PR #2868, opened 2016-09-28
[service_discovery] Get tags from docker, provides ability to drill i…

gregdurham:
### What does this PR do?

Provides the ability to pull tags from the container level, using the same API that the docker_daemon check uses
### Motivation

When using container schedulers other than kubernetes, the lack of ability to drill into issues at a container level when utilizing the service discovery plugin was frustrating. In order to provide a better insight for my development team, we added the ability to gather tags from docker. This allows us to specify a specific container/type of container/image/etc instead of only on the host level. When running N containers of the same type on the same host, it was difficult to get insight into what was happening. 
### Testing Guidelines

Use service discovery with docker, see tags being set for the container.

### DIFF
--- a/checks.d/docker_daemon.py
+++ b/checks.d/docker_daemon.py
@@ -945,3 +945,4 @@ class DockerDaemon(AgentCheck):
                 self.warning("Cannot parse %s content: %s" % (path, str(e)))
                 continue
         return container_dict
+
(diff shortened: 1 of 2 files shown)


gregdurham:
It seems that appveyor may be failing across the board

hkaj:
Hi @gregdurham 
Thanks for contributing!
This is a good idea, and actually something we plan to do. However the way it's done here is something that will be quite painful for our backend because of the lifecycle of container tags.
We have some work in progress internally to consolidate container metadata and integration and custom metrics and the tagging you propose here will be part of that.

I'm going to close this PR, but the feature will be implemented in a future version.
Thanks again

Issue #2872, opened 2016-09-29
Event dates collected by the Kubernetes agent have the wrong timezone


yannrouillard:
There is little mistake in the new events forwarder part of the kubernetes checker in the current tree: the UTC time of the even is incorrectly analyzed as a local time.

As a result each event is incorrectly timeshifted in past or future (unless living in the UTC timezone :) )

PR #2873, opened 2016-09-29
Fix incorrect time conversion for Kubernetes events

yannrouillard:
This a pull request proposal for issue #2872.

We simply replace time.mktime by calendar.gmtime which takes as input struct time in UTC like the ones used in Kubernetes events.


remh:
You are completely right, merging it. Thanks a lot @yannrouillard !

PR #2880, opened 2016-09-29
[kafka_consumer] Make args optional in kafka_consumer.py

jeffwidman:
### What does this PR do?

Makes the `consumer_groups`, `topics`, and `partitions` arguments optional in `kafka_consumer.py`. Missing config values will be fetched from Zookeeper. If a value is omitted, the parent value must still be it's expected type (typically a dict).
### Motivation

In production we have a large number of topics and consumer groups, which change frequently. It's impossible to keep them updated manually. 
### Additional Notes

There are several outstanding pull requests attempting to add this functionality: #1313, #2750, and #2781. I reviewed them, and preferred the API and general code structure implemented in #2781.

However, when I tested #2781, I ran into a number of bugs. Some were obvious typos, some were not so obvious edge cases. In addition to fixing those bugs, I refactored the code that fetched missing args from zookeeper to clarify the logic.

While refactoring, I found several bits of code confusing, so I added some clarifying comments.

This PR does not add the ability to fetch consumer offsets that are stored in Kafka (as opposed to Zookeeper). I looked at doing that, but I didn't see a straightforward way to do that using the `kafka-python` library. So I punted on that for now, and hopefully it can be added in a later PR.

I tested this code on both a Kafka 0.8.2 cluster and a Kafka 0.10.0.1 cluster each using Zookeeper 3.4.6.

### DIFF
--- a/checks.d/kafka_consumer.py
+++ b/checks.d/kafka_consumer.py
@@ -2,12 +2,9 @@
 # All rights reserved
 # Licensed under Simplified BSD License (see LICENSE)
 
-# stdlib
-from collections import defaultdict
-
 # 3p
-from kafka import KafkaClient
-from kafka.common import OffsetRequestPayload as OffsetRequest
+from kafka import SimpleClient
+from kafka.structs import OffsetRequestPayload
 from kazoo.client import KazooClient
 from kazoo.exceptions import NoNodeError
 
@@ -19,6 +16,18 @@ DEFAULT_ZK_TIMEOUT = 5
 
 
 class KafkaCheck(AgentCheck):
+    """
+    Check Consumer Lag for Kafka consumers that store their offsets in Zookeeper.
+
+    Modern Kafka consumers store their offset in Kafka rather than Zookeeper,
+    but support for this has not been added yet. It's tricky because this check
+    is much simpler if it assumes a single place can be queried for all consumer
+    offsets, but currently there's no easy way to do that. Once KIP-88 is
+    implemented, it will be much easier to add this functionality, although it
+    would only work for Kafka brokers >= 0.10.2.0. In the meantime, you can
+    instrument your individual kafka consumers to submit their offsets to
+    Datadog.
+    """
 
     SOURCE_TYPE_NAME = 'kafka'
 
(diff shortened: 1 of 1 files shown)


jeffwidman:
WIP, do NOT merge... 

`kafka_conn.send_offset_request()` method is throwing `FailedPayploadsError`.

I'm not sure if I'm using `kafka_conn.send_offset_request()` incorrectly, or if there's something within `kafka-python`. I googled, and looks like that exception often indicates a flaky network connection to the Kafka cluster. In this case, that shouldn't be relevant because I'm running this check from within a broker, so it's hitting localhost, and I verified that the address is routable. 

My one thought was that sending too many topics/partitions in the offset_request, but I tried it with just a couple of partitions and still hit problems. 

Perhaps @dpkp can take a quick look at line https://github.com/DataDog/dd-agent/pull/2880/files#diff-a67408d94dd8701e685782eb2e68db18R131 and see if I'm missing anything obvious.

jeffwidman:
This is now verified to be working on our clusters and should be ready for review/merging.

checks.d/kafka_consumer.py
```diff
@@ -29,38 +29,92 @@ def __init__(self, name, init_config, agentConfig, instances=None):
...
+        # consumer offset so that we can lookup the partition's newest message
+        # in the Kafka cluster to calculate consumer lag.
+        # This is tricky because different consumer groups may track different
+        # partitions of a topic. So when we modify a topic's list of partitions,
+        # we don't want to overwrite, only add missing partitions. And only want
+        # uniques, because even though each partition may have multiple consumer
+        # group offsets, it will only have one broker offset. Hence the
+        # defaultdict(set) usage.
+        # {topic1: set[(partition1, partition2)], topic2: set[partition1]}
+        tracked_topic_partitions = defaultdict(set)
```
remh: You don't need a defaultdict here, a regular dictionary will do AFAICS

checks.d/kafka_consumer.py
```diff
@@ -29,38 +29,92 @@ def __init__(self, name, init_config, agentConfig, instances=None):
...
-                        zk_path = zk_path_tmpl % (consumer_group, topic, partition)
+                        zk_path = (zk_path_partition_tmpl + '%s/') % (consumer_group, topic, partition)
                         try:
                             consumer_offset = int(zk_conn.get(zk_path)[0])
                             key = (consumer_group, topic, partition)
                             consumer_offsets[key] = consumer_offset
                         except NoNodeError:
                             self.log.warn('No zookeeper node at %s' % zk_path)
-                        except Exception:
+                        except:
```
remh: Same as above, you should avoid bare `except:` 
https://docs.python.org/2/howto/doanddont.html#except

checks.d/kafka_consumer.py
```diff
@@ -73,56 +127,75 @@ def check(self, instance):
...
             broker_offsets = {}
-            for topic, partitions in topics.items():
+            for topic, partitions in tracked_topic_partitions.iteritems():
                 offset_responses = kafka_conn.send_offset_request([
-                    OffsetRequest(topic, p, -1, 1) for p in partitions])
+                    OffsetRequestPayload(topic, p, -1, 1) for p in partitions])
 
                 for resp in offset_responses:
-                    broker_offsets[(resp.topic, resp.partition)] = resp.offsets[0]
+                    key = (resp.topic, resp.partition)
```
remh: Not that useful ;)

checks.d/kafka_consumer.py
```diff
@@ -73,56 +127,75 @@ def check(self, instance):
...
 
-        # Report the broker data
-        for (topic, partition), broker_offset in broker_offsets.items():
+        # Report the broker offset
+        # Easiest to do in separate loop from consumer metrics because each
+        # topic/partition combo will have one broker offset, but may have
+        # multiple consumer offsets
+        for (topic, partition), broker_offset in broker_offsets.iteritems():
             broker_tags = ['topic:%s' % topic, 'partition:%s' % partition]
-            broker_offset = broker_offsets.get((topic, partition))
```
remh: Hu, nice cleaning. Thanks!

checks.d/kafka_consumer.py
```diff
@@ -73,56 +127,75 @@ def check(self, instance):
...
+                    # topics are optional
+                    assert isinstance(topics, dict) or topics is None
+                    if isinstance(topics, dict):
+                        for topic, partitions in topics.iteritems():
+                            assert isinstance(topic, (str, unicode))
+                            # partitions are optional
+                            assert isinstance(partitions, (list, tuple)) or partitions is None
+                            if isinstance(partitions, (list, tuple)):
+                                for partition in partitions:
+                                    assert isinstance(partition, int)
```
remh: that's overkill. It will break if people use strings instead of int. Let's remove that last assert.

checks.d/kafka_consumer.py
jeffwidman: Yeah, I'm totally familiar with the rationale, just missed it when cleaning up #2781

checks.d/kafka_consumer.py
jeffwidman: I did it for consistency with L#113:

    key = (consumer_group, topic, partition)

Would you prefer I collapse both of them, collapse one, or leave both?

checks.d/kafka_consumer.py
jeffwidman: So I actually did try this originally, as I had the exact same thought (`defaultdict` came from the current code).

However, I ran into problems without it. My memory is a little hazy as it's been a few weeks, but IIRC it's needed to pre-populate the set in the dict so that L#106 updates rather than overwrites:

    tracked_topic_partitions[topic].update(set(partitions))

checks.d/kafka_consumer.py
jeffwidman: I wish I was just being overly pedantic here. :-)

It actually does need to be an `int` because these partitions are passed directly to `OffsetRequestPayload` in L#135, and if they're passed as strings then the underlying `kafka-python` throws `FailedPayloadsError`. I actually was hitting this for quite a while up in L#98 where partitions are pulled from ZK when they aren't specified in the YAML until I realized what was happening and cast them to `int`. 

I could solve this by just casting to `int` when passing them to `OffsetRequestPayload` in L#135, but that felt more like solving the symptom rather than the root cause. And if this code gets extended later down the road somehow, there's less risk of unexpected side effects if the `int` is forced as early as possible. 

Your thoughts?

jeffwidman:
I responded to the comments--had a few questions, once you let me know what you want, I'll push an updated patch. 

> Can you add a warning that it's better to list the actual consumer groups / topic / partitions that need to be monitored? As this algorithm will be in O(# consumer groups X topic X partitions) this can have a performance impact on the agent / zookeeper / kafka.

Yes, I'm not sure how to do this though... do I just print the warning? Or is there something in the stdlib I can hook into or something within the dd-agent framework that I can call to throw the warning? 

Along those lines, I was wondering about caching the ZK consumer groups/topics/partitions so we don't have to walk the entire ZK tree every time. We still need to hit ZK for each consumer group / partition combo to get the offset, but this would save on stat'ing the list of consumer groups, topics, and partitions... not a huge savings, but it's something.

I just wasn't sure how caching and custom checks work within the dd-agent framework. Normally for something this lightweight, I'd just save to a python var and re-access it. But is the custom check code reloaded from scratch on every run?

jeffwidman:
Got any spare cycles to look at my responses to your review @remh ?

jeffwidman:
Pushed an updated version of this. 

We hit some problems in production with our ZK offsets being in a corrupted state due to another issue, so this adds additional error checking to handle those scenarios.

jeffwidman:
Any updates here?

jeffwidman:
Any update?

jeffwidman:
Amended the commit to fix a potential race condition. Previously broker offset was fetched before consumer offset. If looping through consumer offsets took a while, a consumer offset for an up-to-date consumer may be newer than our cached broker offset. Better to fetch the consumer offsets first and slightly overstate consumer lag because the alternative of understating consumer lag could result in a negative consumer lag which is really confusing because it's theoretically impossible.

irabinovitch:
@jeffwidman Thank you for your contributions here.  This is currently flagged for being merged into the 5.12 milestone of Datadog agent.  

Starting in 5.12 we will also be moving all integrations out of this repository and into either Datadog/Integrations-core or Datadog/integrations-extras.  By splitting the checks for the core agent we will be able to release checks independently of agent releases, so that we can merge enhancements such as this one much more quickly. We believe this will make for a better contributor and user experience.

remh:
Hi @jeffwidman 
I tested this check on our existing installation and it started collecting tens of thousands of metrics without changing the existing configuration. This shouldn't be the default as it can lead to overcharges and performance issue.

I think we should:
- By default, just honor whitelists
- Enable this mode with a flag and a warning

jeffwidman:
Sure, that makes sense to default to whitelist.

For warning, what are you thinking? Like a warning note in the config or `warnings.warn`? Or is there a specific hook in the datadog AgentCheck framework I should be calling for the warnings?

remh:
@jeffwidman 
a message in the yaml.example file would be enough. There are other mechanisms but there a bit too visible. Thanks!

jeffwidman:
After looking at this more, I'm actually going to hold off on adding the whitelist flag as I'm working on extending this check to support offsets stored in Kafka rather than just the ones stored in Zookeeper, and that may change the interface slightly... 

I'm still researching the Kafka API protocol but it looks like the list of consumer groups / topics may not be so easy to fetch... I'd hate to add a flag to this interface and then later deprecate/tweak it in a later PR. It might be a month or two until I finish this as I'm tackling in-between other tasks, but it is something we will need internally, so it will get done eventually.

In the meantime, this works perfectly for anyone else looking to fetch offsets from zookeeper for all their consumer groups/partitions.

jeffwidman:
Migrated this to https://github.com/DataDog/integrations-core/pull/271

PR #2884, opened 2016-09-30
[install_script][yum] Enable all the repos when installing `datadog-agent`

olivielpeau:
### What does this PR do?

Fixes the install script when the `initscripts` package is not installed.

### Testing Guidelines

Tested the resulting install script on CentOS 5/6/7

### Additional Notes

The yum package of `datadog-agent` has a dependency on `initscripts`.
If we disable all the repos when installing `datadog-agent` _and_ if
`initscripts` is not installed `yum` won't be able to find the dep,
and the install will fail.

To fix this we fall back to installing the agent with all the repos enabled when
the install without them fails.


remh:
👍

PR #2885, opened 2016-09-30
Allow SSL connects to pgbouncer

dblackdblack:
### What does this PR do?
Rather than reconstructing the postgres connection string from its constituent parts (username, password, host, port), pass through connection strings authored by the user to psycopg2. This lets us use all of psycopg2's features, including SSL encryption.

### Motivation

pgbouncer v1.7 added SSL support, which is currently missing from the pgbouncer check.

### DIFF
--- a/checks.d/pgbouncer.py
+++ b/checks.d/pgbouncer.py
@@ -165,6 +165,10 @@ class PgBouncer(AgentCheck):
                 pg.extensions.ISOLATION_LEVEL_AUTOCOMMIT)
             self.log.debug('pgbouncer status: %s' % AgentCheck.OK)
 
+        # re-raise the ChceckExcpetions raised by _get_connect_kwargs()
+        except CheckException:
+            raise
+
         except Exception:
             message = u'Cannot establish connection to pgbouncer://%s:%s/%s' % (host, port, self.DB_NAME)
             self.service_check(self.SERVICE_CHECK_NAME, AgentCheck.CRITICAL,
--- a/checks.d/pgbouncer.py
+++ b/checks.d/pgbouncer.py
@@ -7,6 +7,9 @@
 Collects metrics from the pgbouncer database.
 """
 # 3p
+import re
+import urlparse
+
 import psycopg2 as pg
 
 # project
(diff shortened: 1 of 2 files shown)


remh:
Thanks for doing this ! We'll review it soon.

checks.d/pgbouncer.py
```diff
@@ -162,6 +168,7 @@ def check(self, instance):
         user = instance.get('username', '')
         password = instance.get('password', '')
         tags = instance.get('tags', [])
+        database_url = instance.get('database_url')
```
remh: Can you document that in the pgbouncer.yaml.example file please ?

dblackdblack: added

dblackdblack:
@remh I rejiggered the connection code to be cleaner and to make it more obvious as when things are getting cached

checks.d/pgbouncer.py
```diff
@@ -120,38 +120,62 @@ def _collect_stats(self, db, instance_tags):
...
+                host=host, port=port, user=user,
+                password=password, database_url=database_url
+            )
+
+            connection = pg.connect(**connect_kwargs)
+            connection.set_isolation_level(
+                pg.extensions.ISOLATION_LEVEL_AUTOCOMMIT)
+            self.log.debug('pgbouncer status: %s' % AgentCheck.OK)
+
+        # re-raise the ChceckExceptions raised by _get_connect_kwargs()
```
remh: Small typo here

checks.d/pgbouncer.py
```diff
@@ -120,38 +120,62 @@ def _collect_stats(self, db, instance_tags):
...
+        except CheckException:
+            raise
+
+        except Exception:
+            message = u'Cannot establish connection to pgbouncer://%s:%s/%s' % (host, port, self.DB_NAME)
```
remh: I think if someoneuses the `database_url` to connect this message will be come
`Cannot establish connection to pgbouncer://:/` as it doesn't use the url at all.
Unless i'm missing something ?

checks.d/pgbouncer.py
remh: Thanks 👍

dblackdblack:
@remh pong!

remh:
Thanks there is one last issue regarding the service check being sent when everything is alright https://github.com/plangrid/dd-agent/blob/0f5276b17725329803513cd7fb9e02ee4f44f0ed/checks.d/pgbouncer.py#L222 but i'll fix it.
Thanks again!

PR #2888, opened 2016-10-03
[service_discovery] Reload config on container change event from dock…

gregdurham:
### What does this PR do?

Reloads the configuration on call to `update_checks`

### Motivation

When running a container orchestrator i.e. nomad/kubernetes/etc, new containers (new versions as well as new containers with new images) are deployed onto that host. We found that when datadog starts on a fresh host with no containers, all appears to be good. However, we then found that when new containers were scheduled on the host, no metrics were gathered about these containers from the service discovery plugin, however docker_daemon stats were still flowing. On reloading the datadog-agent service, we found that service discovery was now collecting metrics, again adding new containers, no metrics were sent through for the newly scheduled containers, however old containers were flowing. After digging deeper, a single call to get_configs() on container changes, resolved this issue.

### Testing Guidelines

On old version: Configure service discovery, with no containers on the host, start a container, see that no plugins configured from service discovery are being run (no metrics in dashboard). Add this plugin, and see that the config from service_discovery backend are being reloaded on new containers being started on the host.

### DIFF
--- a/utils/service_discovery/sd_docker_backend.py
+++ b/utils/service_discovery/sd_docker_backend.py
@@ -48,6 +48,7 @@ class SDDockerBackend(AbstractSDBackend):
         AbstractSDBackend.__init__(self, agentConfig)
 
     def update_checks(self, changed_containers):
+        self.get_configs()
         conf_reload_set = set()
         for id_ in changed_containers:
             try:


gregdurham:
Appears to be failing due to something unrelated to this PR

hkaj:
Hi @gregdurham 
Thank you for the feedback. This call shouldn't be necessary, and it's actually quite damaging as it will reload the config of every check instead of only those that need it.

I couldn't reproduce your issue (I created a node with no container, started the agent, started a container and it was picked up correctly). I suppose there is an issue with your setup or a problem in the agent that makes your setup fail. Could you contact our support team and [send us a flare](https://help.datadoghq.com/hc/en-us/articles/204991415-Send-logs-and-configs-to-Datadog-via-flare-command) from a misbehaving agent? We'll have a look at what's failing and fix it, but this change is likely not the right solution.

Thanks again

gregdurham:
Thanks for trying! Very interesting that you couldn't repro. I just sent over a flare, and escalated the case. Very interested to see the outcome.

gregdurham:
Thank you for the assistance, it appears that in newer versions you need to use docker labels rather than the image part in order to get the config reloading. I am going to test this out and if all is good, close this PR out.

hkaj:
Anytime! As we mentioned on your support ticket you can either use docker labels, annotations if using kubernetes, or the full image name as the key name in consul - including repository and tag. So `gcr.io/custom-nginx:latest` instead of just `custom-nginx`  for example.

Let me know how it goes.

gregdurham:
Will do.

PR #2891, opened 2016-10-04
[service discovery] cache templates and reduce # of kv store calls

hkaj:
### What does this PR do?

Cache config templates to reduce the # of calls to the config store.
### Motivation

I was benchmarking service discovery against a node with 500 containers and noticed the high volumes of reads. It's not blocking or anything but we can do better.
### Testing Guidelines

Updated the relevant unit tests.


olivielpeau:
Moving to `5.11.0` milestone since the merge window for 5.10 is closing and this is not a blocker.

hkaj:
Working on a revamp, new PR coming soon.

PR #2894, opened 2016-10-07
[mongo] fixing hostname detection when username/password auth are part of the uri

ibwhite:
### What does this PR do?

When sending MongoDB events back to DataDog, the agent parses the MongoDB connection string URI and attempts to extract the hostname (in the `hostname_for_event` function).

However, if the URI contains username/password auth (`mongodb://username:password@hostname:port`), the existing code will incorrectly extract the username and report that as the hostname.

This PR fixes that and should correctly extract the hostname, whether or not auth is used. It has been tested in production.

### Motivation

I didn't like seeing incorrect **datadog is PRIMARY** messages in my event stream.


yannmh:
Thanks for fixing the issue @ibwhite ! 🙇

PR #2897, opened 2016-10-07
[win32] retry on process restart failures

yannmh:
Datadog Agent on Windows can potentially hit "random" `WindowsError`
exceptions while restarting a process. The Windows agent's supervisor
(`win32/agent`) crashes and processes are not monitored, or restarted
anymore.

Catch `WindowsError` exceptions and retry later.

References:
http://stackoverflow.com/questions/17076679/windowserror-access-is-denied-on-calling-process-terminate?answertab=votes#tab-top

### DIFF
--- a/win32/agent.py
+++ b/win32/agent.py
@@ -112,7 +112,9 @@ class AgentSvc(win32serviceutil.ServiceFramework):
             # Restart any processes that might have died.
             for name, proc in self.procs.iteritems():
                 if not proc.is_alive() and proc.is_enabled():
-                    servicemanager.LogInfoMsg("%s has died. Restarting..." % name)
+                    servicemanager.LogErrorMsg(
+                        u"`{}` has died. Restarting...".format(name)
+                    )
                     proc.restart()
 
             self._check_collector_blocked()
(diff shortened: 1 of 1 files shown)


yannmh:
@degemer can you take a pass on this one please?

win32/agent.py
```diff
@@ -171,28 +174,37 @@ def _can_restart(self):
...
+            )
+            self._process.start()
+        except WindowsError as e:
+            servicemanager.LogErrorMsg(
+                u"Fail to restart `{process}`: {err}".format(
+                    process=self._name, err=e
+                )
+            )
+        else:
+            self._restarts.append(time.time())
```
degemer: This means that the failed restarts are not counted. Not sure we want that, since it could create a small "start loop" (1 process creation try per second). What do you think ?

yannmh: Good call!

Additionally, what would you think about increasing the `SERVICE_SLEEP_INTERVAL` to 2 seconds instead of 1? I want to give more time to the processes to restart,  to avoid "illegitimate" disabling.

PR #2901, opened 2016-10-08
[kubernetes] Only use kube annotations for sd once per pod

mikekap:
### What does this PR do?

Currently the kube annotations are used for service discovery once per docker container. This causes "duplicate" checks to be instantiated if you have sidecar containers in your kubernetes pod.

This PR changes the annotations to annotate a specific container to monitor (by name).
### Motivation

Seeing dup checks in my cluster :X.
### Testing Guidelines

Unit tests. I will also be deploying this to my internal cluster and monitoring.
### Additional Notes

The various tuples flying around between config & backend are getting a bit old. AFAICT there are 4 distinct sets:
- Config->Backend: `[(source, (check_name, init_config, instance_tpl))]`
- `_get_config_templates` -> `_get_check_configs`: `[(source, (check_name, init_config, instance_tpl, variables))]`
- `_get_check_configs` -> `get_configs`: `[(source, (check_name, init_config, instance))]`

It's getting a bit weird. I removed `trace_configs` as a variation because that was getting to be a bit too much. I think there was also a bug where some code didn't handle this case.

### DIFF
--- a/utils/service_discovery/abstract_config_store.py
+++ b/utils/service_discovery/abstract_config_store.py
@@ -26,6 +26,7 @@ CHECK_NAMES = 'check_names'
 INIT_CONFIGS = 'init_configs'
 INSTANCES = 'instances'
 KUBE_ANNOTATIONS = 'kube_annotations'
+KUBE_POD_NAME = 'kube_pod_name'
 KUBE_ANNOTATION_PREFIX = 'com.datadoghq.sd/'
 
 
@@ -141,29 +142,26 @@ class AbstractConfigStore(object):
 
     def get_check_tpls(self, identifier, **kwargs):
         """Retrieve template configs for an identifier from the config_store or auto configuration."""
-        trace_config = kwargs.get(TRACE_CONFIG, False)
-
         # this flag is used when no valid configuration store was provided
         # it makes the method skip directly to the auto_conf
         if kwargs.get('auto_conf') is True:
             # When not using a configuration store on kubernetes, check the pod
             # annotations for configs before falling back to autoconf.
             kube_annotations = kwargs.get(KUBE_ANNOTATIONS)
+            kube_pod_name = kwargs.get(KUBE_POD_NAME)
             if kube_annotations:
                 kube_config = self._get_kube_config(identifier, kube_annotations)
                 if kube_config is not None:
                     check_names, init_config_tpls, instance_tpls = kube_config
                     source = CONFIG_FROM_KUBE
-                    return [(source, vs) if trace_config else vs
-                            for vs in zip(check_names, init_config_tpls, instance_tpls)]
+                    return [(source, '{}:{}'.format(kube_pod_name, i), vs)
+                            for i, vs in enumerate(zip(check_names, init_config_tpls, instance_tpls))]
 
             # in auto config mode, identifier is the image name
             auto_config = self._get_auto_config(identifier)
             if auto_config is not None:
                 source = CONFIG_FROM_AUTOCONF
-                if trace_config:
-                    return [(source, auto_config)]
-                return [auto_config]
+                return [(source, identifier, auto_config)]
             else:
                 log.debug('No auto config was found for image %s, leaving it alone.' % identifier)
                 return []
@@ -183,8 +181,8 @@ class AbstractConfigStore(object):
         # Try to update the identifier_to_checks cache
         self._update_identifier_to_checks(identifier, check_names)
 
-        return [(source, values) if trace_config else values
-                for values in zip(check_names, init_config_tpls, instance_tpls)]
+        return [(source, '{}:{}'.format(identifier, i), values)
+                for i, values in enumerate(zip(check_names, init_config_tpls, instance_tpls))]
 
     def read_config_from_store(self, identifier):
         """Try to read from the config store, falls back to auto-config in case of failure."""
--- a/utils/service_discovery/abstract_config_store.py
+++ b/utils/service_discovery/abstract_config_store.py
@@ -158,14 +158,14 @@ class AbstractConfigStore(object):
                 if kube_config is not None:
                     check_names, init_config_tpls, instance_tpls = kube_config
                     source = CONFIG_FROM_KUBE
-                    return [(source, '{}:{}'.format(kube_pod_name, i), vs)
+                    return [(source, vs)
                             for i, vs in enumerate(zip(check_names, init_config_tpls, instance_tpls))]
 
             # in auto config mode, identifier is the image name
             auto_config = self._get_auto_config(identifier)
             if auto_config is not None:
                 source = CONFIG_FROM_AUTOCONF
-                return [(source, identifier, auto_config)]
+                return [(source, auto_config)]
             else:
                 log.debug('No auto config was found for image %s, leaving it alone.' % identifier)
                 return []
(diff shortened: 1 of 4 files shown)


hkaj:
thanks @mikekap i'll review it shortly

hkaj:
Hi @mikekap 
Thanks for contributing (again!) 🎉 

That's an issue I should have caught before. Our lack of integration testing for k8s is getting punishing.
Also, thank you for removing the `trace_configs` thing. It really didn't age well.

Your approach in this PR works, but I wonder if we shouldn't modify the annotation format instead of ignoring all but one container per pod. Running checks against several containers in the same pod should be an option. What do you think of changing annotation keys like so?
```
com.datadoghq.sd/<full_image_name>/check_names
com.datadoghq.sd/<full_image_name>/init_configs
com.datadoghq.sd/<full_image_name>/instances
```

`AbstractConfigStore._get_kube_config` has both annotations and the identifier (which should already be `full_image_name` in this case) so we can easily do the matching [there](https://github.com/mikekap/dd-agent/blob/kube-sd-only-one/utils/service_discovery/abstract_config_store.py#L102).

This feature is not documented yet so it shouldn't be a problem.
What do you think?

mikekap:
We could definitely add support for monitoring all containers. It would probably be more intuitive to use the kubernetes container name rather than the image name though. One thing I'm not sure of - is the multi-container model more flexible? Currently, even though we only init checks for one of the containers, you can easily monitor the other containers in a pod by just changing the port (since all containers in a pod share the same ip).

On a higher level, the kubernetes model is a bit shoehorned into the docker model here. The "unit" of kubernetes is a pod. Ideally we'd want to monitor pods no matter how many containers are running or not running in those pods. That is, if some container fails to start, or docker is down, the monitoring should still apply. That's a bit outside the scope of this PR to fix though.

hkaj:
Using the kubernetes container name is fine since it's also quick to extract and guaranteed to exist. Go for it 👍 
The reason for the multi-container model is that to resolve template variables we use the `docker inspect` of the container (by looking through all ports used by the container and selecting the right one for `%%port%%` for instance). So tying the template to a container rather than a pod seems to make more sense (but I might be wrong).
The thing is service discovery is not targeting only kubernetes but container environments at large. So it is likely to stay container-centric and not pod-centric, at least for now. We have some other work in progress to make pod/service/deployment monitoring easier and more powerful though. Notably improvements to the kubernetes integration. So stay tuned ;)

mikekap:
Done. I haven't tested this outside of unit tests yet since I would need to migrate by existing usage.

utils/service_discovery/sd_docker_backend.py
```diff
@@ -223,6 +230,14 @@ def _get_additional_tags(self, container_inspect, *args):
             tags.append('pod_name:%s' % pod_metadata.get('name'))
         return tags
 
+    def _get_kube_container_name(self, c_id):
+        pods = self.kubeutil.retrieve_pods_list().get('items', [])
```
hkaj: not super excited about calling this for every container, can we maybe build the mapping `c_id -> c_name` once before entering this loop https://github.com/mikekap/dd-agent/blob/4790dd8df8417900a62fb776f0e10fa3491fb749/utils/service_discovery/sd_docker_backend.py#L258 to reduce api calls? It should be fine since it's coming from the kubelet but with nodes running hundreds of containers if there's a hiccups we'll feel it.

hkaj:
Looks like this scheme is not going to work:
```
The Pod "redis-django" is invalid:
* metadata.annotations: Invalid value: "com.datadoghq.sd/frontend/check_names": must match the regex ([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9] (e.g. 'MyName' or 'my.name' or '123-abc') with an optional DNS subdomain prefix and '/' (e.g. 'example.com/MyName'
```

Let's use dots instead of slashes?

mikekap:
Sounds good. I'm a bit swamped this week so I'll get around to making those changes on the weekend if you don't mind. Sorry for the delay

utils/service_discovery/abstract_config_store.py
```diff
@@ -26,7 +26,8 @@
 INIT_CONFIGS = 'init_configs'
 INSTANCES = 'instances'
 KUBE_ANNOTATIONS = 'kube_annotations'
-KUBE_ANNOTATION_PREFIX = 'com.datadoghq.sd/'
+KUBE_CONTAINER_NAME = 'kube_container_name'
+KUBE_ANNOTATION_PREFIX = 'sd.datadoghq.com'
```
mikekap: So I may have snuck this one in there. I changed the annotation to be `service-discovery.datadoghq.com/checks.XXXX`. I've realized that kube annotations don't follow the "reverse dns" scheme at all and I didn't want to deviate from that. Let me know if you'd prefer to change it back to what you were suggesting.

utils/service_discovery/sd_docker_backend.py
```diff
@@ -47,31 +91,42 @@ def __init__(self, agentConfig):
...
             conf_reload_set.update(set(checks))
 
         if conf_reload_set:
             self.reload_check_configs = conf_reload_set
 
-    def _get_checks_from_inspect(self, inspect):
+    def _get_checks_to_refresh(self, state, c_id):
         """Get the list of checks applied to a container from the identifier_to_checks cache in the config store.
         Use the DATADOG_ID label or the image."""
+        inspect = state.inspect_container(c_id)
```
hkaj: So we actually have another issue to take care of here, which is that if the container which the ID is from is already removed at this point, inspect will be None and we won't reload anything although there's a check that now fails on a missing container.

To be clear it's a pre-existing bug, not yours, but let's fix it here. In this situation we can't really decide which check to reload since the container isn't here anymore, so I think we should trigger a full reload.
Adding this block here does the job:
```python
        # if the container was removed we can't tell which check is concerned
        # so we have to reload everything
        if not inspect:
            self.reload_check_configs = True
            return
```

utils/service_discovery/abstract_config_store.py
```diff
@@ -98,11 +99,12 @@ def _populate_identifier_to_checks(self):
 
         return identifier_to_checks
 
-    def _get_kube_config(self, identifier, kube_annotations):
+    def _get_kube_config(self, identifier, kube_annotations, kube_container_name):
         try:
-            check_names = json.loads(kube_annotations[KUBE_ANNOTATION_PREFIX + CHECK_NAMES])
-            init_config_tpls = json.loads(kube_annotations[KUBE_ANNOTATION_PREFIX + INIT_CONFIGS])
-            instance_tpls = json.loads(kube_annotations[KUBE_ANNOTATION_PREFIX + INSTANCES])
+            prefix = '{}/{}.'.format(KUBE_ANNOTATION_PREFIX, kube_container_name)
```
mikekap: HK - I think you might have forgotten some of the context from above. The main point of doing this PR is so we instantiate the checks above once per pod rather than once per container (since pods can have a few containers and you can't annotate containers directly in kubernetes; this was previously creating N instances of the checks above where N=containers in a pod).

hkaj: welp, yes indeed.

hkaj:
🎉 thanks again @mikekap

mikekap:
Wow - thanks for getting this in @hkaj! I was just coming back from the break and getting ready to tackle merging this. Sorry you had to go through that.

hkaj:
No worries, it was an easy conflict. I'll finish polishing a documentation update for it and we should be good to go!

PR #2906, opened 2016-10-12
[CI][Kong] if cassandra isn't there, kong env is broken

masci:
### What does this PR do?

Takes care of having a working Cassandra

### Motivation

don't ask, don't tell

### Notes

Had to update Cassandra version, the old archive was giving a 404
I'm using rsync instead of curl to grab `utils-linux` b/c the curl shipped with Travis is quite old and doesn't support recent SSL implementations


hkaj:
👍

PR #2766, opened 2016-08-17
RFC: Kubernetes: Add cpu and memory requests, limits, and capacity info.

markine:
# An RFC: first cut of code, looking for green light for further cleanup. Also seeking any guidance on testing from authors of the kubernetes Check code.

### What does this PR do?

Add reporting of Kubernetes resource management info: CPU and memory requests, limits, and capacity.

### Motivation

Monitoring of cluster resources from a scheduling/allocation perspective is important. E.g. over-allocation of CPU resources to deployments may cause pod scheduling failures even if actual CPU usage in the cluster is low. We want to get ahead of these errors by monitoring request/limit/capacity data in DataDog.

### Testing Guidelines

An overview on [testing](https://github.com/DataDog/dd-agent/blob/master/tests/README.md) is available in our contribution guidelines.

PR #2908, opened 2016-10-12
Fix parsing of units in Kubernetes resource quantities

therc:
This fixes parsing of resource quantities in Kubernetes resources and limits. Right now suffixes like G, M, m, etc. get simply ignored, leading to wrong numbers (unless you manage to always end up with absolute bytes/CPU numbers without any units).

It's based on the second part of #2766, but with some optimizations (precomputing stuff once at the module level).

### DIFF
--- a/checks.d/kubernetes.py
+++ b/checks.d/kubernetes.py
@@ -50,6 +50,28 @@ FUNC_MAP = {
 
 EVENT_TYPE = 'kubernetes'
 
+# Suffixes per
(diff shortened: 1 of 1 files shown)


therc:
cc @markine

markine:
Any updates here? I'm attempting a rebase of #2766 against latest and will be using this patch.

masci:
@markine thanks for the heads up, merged

PR #2909, opened 2016-10-13
[CI][kong] compile kong in the install stage

masci:
### What does this PR do?

move `make && make install` instructions for kong to the install stage

### Motivation

kong tests are slow


hkaj:
good point 👍

PR #2910, opened 2016-10-13
[mesos_slave] Connect to master with same protocol as slave

mcclurmc:
if you have not yet done so.*

### What does this PR do?

Fixes a bug in the mesos-slave.py check. This check queries both the localhost and the master node. The bug is that even if you specify SSL for communicating with the localhost (slave), the master is always queried with plain HTTP. This is definitely a bug because it is impossible to configure a mesos cluster where master and slave are using different protocols.

### Motivation

I've got SSL enabled on my mesos cluster and this check just plain doesn't work with SSL enabled.

### Testing Guidelines

### Additional Notes

This work is a follow-on to Zendesk ticket: 62968

### DIFF
--- a/checks.d/mesos_slave.py
+++ b/checks.d/mesos_slave.py
@@ -143,12 +143,15 @@ class MesosSlave(AgentCheck):
 
     def _get_constant_attributes(self, url, timeout, master_port, verify):
         state_metrics = None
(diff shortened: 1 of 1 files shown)


masci:
Nice! Thanks @mcclurmc

Issue #2340, opened 2016-03-10
Riak CS integration 'legend' error


millerdev:
I'm getting an error after setting up the riakcs integration:

``` sh
$ dd-agent info
===================
Collector (v 5.7.1)
===================

  Status date: 2016-03-10 19:46:45 (15s ago)
  Pid: 26946
  Platform: Linux-3.13.0-79-generic-x86_64-with-debian-jessie-sid
  Python Version: 2.7.11
  Logs: <stderr>, /var/log/datadog/collector.log, syslog:/dev/log

  Clocks
  ======

    NTP offset: 2.5778 s
    System UTC time: 2016-03-10 19:47:01.797154

  Paths
  =====

    conf.d: /etc/dd-agent/conf.d
    checks.d: /opt/datadog-agent/agent/checks.d

  Hostnames
  =========

    ... snip ...

  Checks
  ======

    riakcs
    ------
      - instance #0 [ERROR]: "'legend'"
      - Collected 0 metrics, 0 events & 2 service checks

    ... snip ...

$ riak-cs version
2.1.0


```

Relevant source line: https://github.com/DataDog/dd-agent/blob/26e811c887a48ffaedf850e00600c4beea8e0e0a/checks.d/riakcs.py#L46
Comment from basho about removing 'legend': https://github.com/basho/riak_cs/issues/764#issuecomment-138485465

olivielpeau:
Thanks @millerdev for reporting this issue in details! This is definitely something that needs to be fixed.

Adding this to the "Triage" milestone until we assess the priority of this issue.

Feel free to open a PR if you have time to write a fix :)

PR #2920, opened 2016-10-17
[riakcs] support Riak CS 2.1+ stats format

millerdev:
### What does this PR do?

Fixes https://github.com/DataDog/dd-agent/issues/2340
### Motivation

Our Riak CS integration has been saying "no data received" for a long time. This adapts the existing collector to the [new stats format introduced in Riak CS 2.1](https://github.com/basho/riak_cs/wiki/%5BRFC%5D-Riak-CS-and-Stanchion-metrics-2.1).
### Testing Guidelines

There are no tests for the `riakcs` stats check. However, I dropped `riakcs.py` from this PR on one of our servers and verified that it works and that the stats can be queried in the web interface.
### Additional Notes

The new stats format includes many more stats (>1000), and the app link for _riakcs_ in our infrastructure list on the website now takes FOREVER to load, and it freezes the whole page while it's loading. Seems like that could be optimized, but it's outside the scope of this change.


checks.d/riakcs.py
```diff
@@ -107,8 +114,16 @@ def _get_stats(self, s3, aggregation_key):
...
+        data = json.loads(text)
+        if "legend" in data:
+            # riak cs before v2.1 had duplicate keys
+            data = json.JSONDecoder(object_pairs_hook=multidict).decode(text)
+        return data
+
+
+STATS_METHODS = {
+    "one": "count",
+    "total": "monotonic_count",
```
millerdev: For metrics ending in `_one` or `_total` I guessed they would be better tracked as counts than a as simple gauge. It wasn't entirely clear which method was the best for each of these. More info on various types of stats here: https://github.com/basho/riak_cs/wiki/%5BRFC%5D-Riak-CS-and-Stanchion-metrics-2.1

millerdev: As I'm reviewing these again and after reviewing the stats reported in my test, I think `total` maybe should use `gauge` rather than `monotonic_count`. Thoughts?

remh:
Thanks a lot @millerdev 
Could you add some tests ? Riak CS tests are actually mocked https://github.com/DataDog/dd-agent/blob/master/tests/checks/mock/test_riakcs.py given that setting up RiakCS in a test environment is not straight forward.

Could you update this test with the new format ?

Also i'm a bit worried about the number of new metrics it collects. Looks like it will start collecting thousands of metrics while it was just collecting in the hundreds. Any possibility to trim that down and to focus on the most relevant metrics ?

checks.d/riakcs.py
```diff
@@ -107,8 +123,191 @@ def _get_stats(self, s3, aggregation_key):
...
+#   - *_time_error_*
+#   - *_time_100
+#
+# Any of these excluded metrics in addition to many others (there are
+# over 1000 to choose from) can be added by specifying them in the
+# riakcs.yaml config file under the "metrics" key in the instance
+# config; the value should be a list of metric names.
+#
+# Helpful references:
+# - https://github.com/basho/riak_cs/wiki/Riak-cs-and-stanchion-metrics
```
millerdev: I chose to only collect a limited set of S3 API and memory metrics by default. I'm not sure how to identify the set that most people will want, so I just went with a set that seemed reasonable for my usage (although I haven't done anything with it yet, so this may evolve). Here are some reasons for each of the exclusions listed above:

- `bucket_acl_(get|put)` and `object_acl_(get|put)` - I'm not using ACLs.
- `bucket_policy_(get|put|delete)` - I'm not doing any automated bucket policy updates.
- `*_in_(one|total)` - guessing this is a similar metric to the `_out_` variants.
- `*_time_error_*` - it seems like the timing of errors is less important than the count of errors.
- `*_time_100` - had a hard time figuring out what this even means. Is it 99.9th percentile? Decided to exclude it since it was not in the old metrics format.

PR #2924, opened 2016-10-18
[dns_check] change to network check

sjenriquez:
### What does this PR do?

Changes the DNS check to a network type check that allows instances to run in separate threads.
### Motivation

Per customer request.
### Additional Notes

Will affect this open PR: https://github.com/DataDog/dd-agent/pull/2799


checks.d/dns_check.py
```diff
@@ -20,66 +20,96 @@
...
 
-class DNSCheck(AgentCheck):
+class DNSCheck(NetworkCheck):
     SERVICE_CHECK_NAME = 'dns.can_resolve'
     DEFAULT_TIMEOUT = 5
 
-    def __init__(self, name, init_config, agentConfig, instances=None):
-        AgentCheck.__init__(self, name, init_config, agentConfig, instances)
-        self.default_timeout = init_config.get('default_timeout', self.DEFAULT_TIMEOUT)
+    def __init__(self, name, init_config, agentConfig, instances):
```
remh: You can remove the constructor entirely if you are just calling the parent constructor

checks.d/dns_check.py
```diff
@@ -20,66 +20,96 @@
...
-            return
+        self.default_timeout = init_config.get('default_timeout', self.DEFAULT_TIMEOUT)
 
-        timeout = float(instance.get('timeout', self.default_timeout))
-        self.query_dns(instance, timeout)
+    def _load_conf(self, instance):
+        # Fetches the conf
+        try:
+            hostname = instance.get('hostname')
+        except Exception:
```
remh:
```python
instance.get('hostname')
```
cannot return an exception.
Better to do something like

```python
def _load_conf(self, instance):
    hostname = instance.get('hostname')
    if not hostname:
        raise BadConfException('A valid "hostname" must be specified')
    ...
```

checks.d/dns_check.py
```diff
@@ -20,66 +20,96 @@
...
         record_type = instance.get('record_type', 'A')
 
-        status = AgentCheck.CRITICAL
+        return hostname, timeout, nameserver, record_type, resolver
+
+    def _check(self, instance):
+        hostname, timeout, nameserver, record_type, resolver = self._load_conf(instance)
+
+        # Perform the DNS query, and report its duration as a gauge
+        start_time = end_time = 0.0
```
remh: You don't need to declare them, as you are declaring them a few lines later

checks.d/dns_check.py
```diff
@@ -20,66 +20,96 @@
...
-            raise
+            self.log.exception('DNS resolution of {0} has failed.'.format(hostname))
+            return Status.CRITICAL, 'DNS resolution of {0} has failed'.format(hostname)
+
         else:
+            tags = self._get_tags(instance)
             if end_time - start_time > 0:
-                self.gauge('dns.response_time', end_time - start_time, tags=tags)
-                self.service_check(self.SERVICE_CHECK_NAME, AgentCheck.OK, tags=self._get_tags(instance))
+                self.gauge('dns.response_time', end_time - start_time, tags)
```
remh: better to keep the keyword argument `tags=tags`

checks.d/dns_check.py
```diff
@@ -20,66 +20,96 @@
...
+            self.log.debug('Resolved hostname: {0}'.format(hostname))
+            return Status.UP, 'UP'
 
     def _get_tags(self, instance):
+        instance_name = instance.get('name')
```
remh: I think people upgrading from a previous version won't have a `name` so we should default to the `hostname`

checks.d/dns_check.py
```diff
@@ -20,66 +20,96 @@
...
+        tags = custom_tags + ['nameserver:{0}'.format(nameserver),
+                              'resolved_hostname:{0}'.format(hostname),
+                              'instance:{0}'.format(instance_name),
+                              'record_type:{0}'.format(record_type)]
         return tags
+
+    def report_as_service_check(self, sc_name, status, instance, msg=None):
+        tags = self._get_tags(instance)
+
+        # FIXME: 5.3, Setting skip_event True here to hide deprecated option in yaml.
```
remh: Outdated comment

checks.d/dns_check.py
sjenriquez: I did this to set the `default_timeout` when configured in `init_config`. `network.py` only passes `instance` to `_check()`, not `init_config`. 
https://github.com/DataDog/dd-agent/blob/master/checks/network_checks.py#L138

Is there a better way to set `default_timeout` when configured in the yaml file?

sjenriquez:
When a `name` isn't passed this Exception in `network.py` is raised:
https://github.com/DataDog/dd-agent/blob/master/checks/network_checks.py#L79-L80

olivielpeau:
Moving this PR to Triage since it's not a blocker for 5.10 but feel free to merge it whenever it's ready, I'll update the milestone accordingly

checks.d/dns_check.py
remh: Good point

remh:
@sjenriquez Then this will break compatibility for current users of the DNS check, if they upgrade their agent, the check will break. Can you fix that please ?

PR #2931, opened 2016-10-19
[kubernetes] add namespace tag to kubernetes.pods.running

hkaj:
_Note: Please remember to review the Datadog [Contribution Guidelines](https://github.com/DataDog/dd-agent/blob/master/CONTRIBUTING.md)
if you have not yet done so._
### What does this PR do?

Add the `kube_namespace` tag to `kubernetes.pods.running`.
### Motivation

User request, it makes aggregation easier and more useful.
### Testing Guidelines

Relevant tests have been update.
### Additional Notes

We should update the `kube_replication_controller` tag name (must be done carefully as it will break things for users relying on it). It used to be accurate but now pods can be created by daemon sets, jobs, deployments, and replica sets too.

### DIFF
--- a/checks.d/kubernetes.py
+++ b/checks.d/kubernetes.py
@@ -368,25 +368,25 @@ class Kubernetes(AgentCheck):
             "ReplicaSet",
         ]
 
-        controllers_map = defaultdict(lambda: {'count': 0, 'namespace': None})
+        # (create-by, namespace): count
+        controllers_map = defaultdict(int)
         for pod in pods['items']:
             try:
                 created_by = json.loads(pod['metadata']['annotations']['kubernetes.io/created-by'])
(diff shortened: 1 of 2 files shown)


checks.d/kubernetes.py
```diff
@@ -326,22 +326,25 @@ def _update_pods_metrics(self, instance, pods):
             "ReplicaSet",
         ]
 
-        controllers_map = defaultdict(int)
+        controllers_map = defaultdict(lambda: {'count': 0, 'namespace': None})
```
remh: This is a weird datastructure you use.

You should keep a `defaultdict(int)` (or a `collections.Counter`) where the keys would be a tuple `(namespace, controller)`

remh:
👍

PR #2934, opened 2016-10-19
[dev mode] fallback to /tmp to dump pstats

hkaj:
### What does this PR do?

If dumping pstats in the current dir fails, fall back to /tmp.

### Motivation

User noticed that they didn't get the dumps because the agent was trying to write to `/` as the default path is relative and they were probably running it with supervisor.

### Additional Notes

This was just a minor bug as the agent should not run in dev mode in production (it ruins its perf) and usually runs in the foreground from a directory where it has write access.

### DIFF
--- a/agent.py
+++ b/agent.py
@@ -225,8 +225,13 @@ class Agent(Daemon):
         self.collector = Collector(self._agentConfig, emitters, systemStats, hostname)
 
         # In developer mode, the number of runs to be included in a single collector profile
-        self.collector_profile_interval = self._agentConfig.get('collector_profile_interval',
-                                                                DEFAULT_COLLECTOR_PROFILE_INTERVAL)
+        try:
+            self.collector_profile_interval = int(
+                self._agentConfig.get('collector_profile_interval', DEFAULT_COLLECTOR_PROFILE_INTERVAL))
+        except ValueError:
+            log.warn('collector_profile_interval is invalid. '
+                     'Using default value instead (%s).' % DEFAULT_COLLECTOR_PROFILE_INTERVAL)
+            self.collector_profile_interval = DEFAULT_COLLECTOR_PROFILE_INTERVAL
 
         # Configure the watchdog.
         self.check_frequency = int(self._agentConfig['check_freq'])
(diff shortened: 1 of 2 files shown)


utils/profile.py
```diff
@@ -41,8 +42,12 @@ def disable_profiling(self):
...
         log.debug("Agent profiling is disabled")
         if self.DUMP_TO_FILE:
-            log.debug("Pstats dumps are enabled. Dumping pstats output to {0}".format(self.STATS_DUMP_FILE))
-            ps.dump_stats(self.STATS_DUMP_FILE)
+            try:
+                ps.dump_stats(self.STATS_DUMP_FILE)
+                log.debug("Pstats dumps are enabled. Dumping pstats output to {0}".format(self.STATS_DUMP_FILE))
+            except IOError:
+                log.debug('Dumping pstats output to {} failed, using /tmp instead.'.format(self.STATS_DUMP_FILE))
+                ps.dump_stats(urljoin('/tmp/', self.STATS_DUMP_FILE))
```
degemer: `os.path.join` !

hkaj:
thanks for the review @degemer 
I updated according to your comments and found&fixed another related bug while testing.

agent.py
```diff
@@ -225,8 +225,12 @@ def run(self, config=None):
         self.collector = Collector(self._agentConfig, emitters, systemStats, hostname)
 
         # In developer mode, the number of runs to be included in a single collector profile
-        self.collector_profile_interval = self._agentConfig.get('collector_profile_interval',
-                                                                DEFAULT_COLLECTOR_PROFILE_INTERVAL)
+        try:
+            self.collector_profile_interval = int(self._agentConfig.get('collector_profile_interval'))
```
degemer: The default is still needed in the `get`, otherwise it will always fail when not configured.

agent.py
```diff
@@ -225,8 +225,12 @@ def run(self, config=None):
...
+        except ValueError:
```
degemer: I think it's a TypeError: `TypeError: int() argument must be a string or a number, not 'NoneType'`.

hkaj:
doh! I added the default back. The exception is a `ValueError` if `collector_profile_interval` is provided but not a string we can turn into an int. (try with `int('foo')`)

PR #2935, opened 2016-10-19
Kubernetes: Add CPU and memory capacity reporting

markine:
### What does this PR do?

Add CPU and memory capacity reporting to the Kubernetes check.

This is a rebase of #2766 with a part of the code moved to KubeUtil.

### Motivation

Monitoring of cluster resources from a scheduling/allocation perspective is important. E.g. over-allocation of CPU resources to deployments may cause pod scheduling failures even if actual CPU usage in the cluster is low. We want to get ahead of these errors by monitoring capacity data in DataDog.

### DIFF
--- a/checks.d/kubernetes.py
+++ b/checks.d/kubernetes.py
@@ -346,6 +346,31 @@ class Kubernetes(AgentCheck):
                     self.log.debug("Container object for {}: {}".format(c_name, container))
 
         self._update_pods_metrics(instance, pods_list)
+        self._update_node(instance)
+
+    def _update_node(self, instance):
+        # These vars and fetch stuff should live in kubeuitl:
+        from utils.http import retrieve_json
+        from urlparse import urljoin
+
+        self.log.info('Hostname: %s' % self.kubeutil.host)
+        self.log.info('Metrics URL: %s' % self.kubeutil.metrics_url)
+        self.log.info('Pod list URL: %s' % self.kubeutil.pods_list_url)
+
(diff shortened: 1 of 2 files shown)


markine:
FYI I'll update the tests.

markine:
Update: tests are ready, please review. I am not familiar with your build system but I don't think the Travis CI failure is related to my change.

checks.d/kubernetes.py
```diff
@@ -346,6 +346,18 @@ def parse_quantity(s):
...
+        self._update_node(instance)
+
+    def _update_node(self, instance):
+        machine_info = self.kubeutil.retrieve_machine_info()
+        num_cores = machine_info.get('num_cores', 0)
+        memory_capacity = machine_info.get('memory_capacity', 0)
+
+        tags = instance.get('tags', [])
+        self.publish_gauge(self, NAMESPACE + '.cpu.capacity', float(num_cores), tags)
+        self.publish_gauge(self, NAMESPACE + '.memory.capacity', float(memory_capacity), tags)
```
masci: Could you add `node.` prefix to those metrics, so that we have `kubernetes.node.*`?

masci: @markine you're right, thanks for pointing out, I underestimated the number of metrics regarding nodes we already have. Hope we'll be able to refactor the naming scheme eventually, for now let's keep those as they are in your code!

markine:
Thank you @masci

PR #2937, opened 2016-10-19
[core][packaging] Adds SUSE to install script

gmmeyer:
This makes a few small changes that add SUSE to the install script. I use the same logic to detect that it's SLES here that I do everywhere else.

While OpenSUSE can use the RH build, SLES cannot, so I just use an else if where SLES comes after OpenSUSE.

### DIFF
--- a/packaging/datadog-agent/source/install_agent.sh
+++ b/packaging/datadog-agent/source/install_agent.sh
@@ -168,7 +168,7 @@ elif [ $OS = "SUSE" ]; then
   fi
 
   echo -e "\033[34m\n* Installing YUM Repository for Datadog\n\033[0m"
-  $sudo_cmd sh -c "echo -e '[datadog]\nname=datadog\nenabled=1\nbaseurl=https://yum.datadoghq.com/suse-rpm/x86_64\ntype=rpm-md\ngpgcheck=1\nrepo_gpgcheck=0\ngpgkey=https://yum.datadoghq.com/DATADOG_RPM_KEY.public' > /etc/zypp/repos.d/datadog.repo"
+  $sudo_cmd sh -c "echo -e '[datadog]\nname=datadog\nenabled=1\nbaseurl=https://yum.datadoghq.com/suse/rpm/x86_64\ntype=rpm-md\ngpgcheck=1\nrepo_gpgcheck=0\ngpgkey=https://yum.datadoghq.com/DATADOG_RPM_KEY.public' > /etc/zypp/repos.d/datadog.repo"
 
   echo -e "\033[34m\n* Refreshing repositories\n\033[0m"
   $sudo_cmd zypper --non-interactive refresh
--- a/packaging/datadog-agent/source/install_agent.sh
+++ b/packaging/datadog-agent/source/install_agent.sh
@@ -81,7 +81,7 @@ elif [ -f /etc/system-release -o "$DISTRIBUTION" == "Amazon" ]; then
 # Arista is based off of Fedora14/18 but do not have /etc/redhat-release
 elif [ -f /etc/Eos-release -o "$DISTRIBUTION" == "Arista" ]; then
     OS="RedHat"
-elif [ -f /etc/issue -o "$DISTRIBUTION" == "SUSE" ]; then
+elif [ "$DISTRIBUTION" == "SUSE" ]; then
     OS="SUSE"
 fi
 
--- a/packaging/datadog-agent/source/install_agent.sh
+++ b/packaging/datadog-agent/source/install_agent.sh
@@ -56,8 +56,8 @@ else
 fi
 
 if [ ! $apikey ]; then
-  printf "\033[31mAPI key not available in DD_API_KEY environment variable.\033[0m\n"
-  exit 1;
+    printf "\033[31mAPI key not available in DD_API_KEY environment variable.\033[0m\n"
+    exit 1;
 fi
 
 # OS/Distro Detection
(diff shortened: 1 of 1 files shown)


packaging/datadog-agent/source/install_agent.sh
```diff
@@ -56,13 +56,13 @@ else
 fi
 
 if [ ! $apikey ]; then
-    printf "\033[31mAPI key not available in DD_API_KEY environment variable.\033[0m\n"
-    exit 1;
+  printf "\033[31mAPI key not available in DD_API_KEY environment variable.\033[0m\n"
+  exit 1;
```
olivielpeau: why change the indentation here? :(

packaging/datadog-agent/source/install_agent.sh
```diff
@@ -81,6 +81,8 @@ elif [ -f /etc/system-release -o "$DISTRIBUTION" == "Amazon" ]; then
 # Arista is based off of Fedora14/18 but do not have /etc/redhat-release
 elif [ -f /etc/Eos-release -o "$DISTRIBUTION" == "Arista" ]; then
     OS="RedHat"
+elif [ -f /etc/issue -o "$DISTRIBUTION" == "SUSE" ]; then
```
olivielpeau: `/etc/issue` seems to be present on a lot of platforms (it's present on both debians and rhels at least from what I've checked).

It looks like `/etc/SuSE-release` would be better to detect the distribution, what do you think?

packaging/datadog-agent/source/install_agent.sh
```diff
@@ -158,6 +160,24 @@ If the cause is unclear, please contact Datadog support.
 "
     $sudo_cmd apt-get install -y --force-yes datadog-agent
     ERROR_MESSAGE=""
+elif [ $OS = "SUSE" ]; then
+  UNAME_M=$(uname -m)
+  if [ "$UNAME_M"  == "i686" -o "$UNAME_M"  == "i386" -o "$UNAME_M"  == "x86" ]; then
+    printf "The Datadog Agent installer is only available for 64 bit SUSE Enterprise machines."
```
olivielpeau: nit: it'd be nicer if you printed this error in bright red :)

```bash
printf "\033[31mThe Datadog Agent installer is only available for 64 bit SUSE Enterprise machines.\033[0m\n"
```

packaging/datadog-agent/source/install_agent.sh
```diff
@@ -158,6 +160,24 @@ If the cause is unclear, please contact Datadog support.
...
+    exit;
+  else
+      ARCHI="x86_64"
+  fi
+
+  echo -e "\033[34m\n* Installing YUM Repository for Datadog\n\033[0m"
+  $sudo_cmd sh -c "echo -e '[datadog]\nname=datadog\nenabled=1\nautorefresh=0\nbaseurl=https://yum.datad0g.com/suse-testing/x86_64\ntype=rpm-md\ngpgcheck=1\nrepo_gpgcheck=0\ngpgkey=https://yum.datadoghq.com/DATADOG_RPM_KEY.public' > /etc/zypp/repos.d/datadog.repo"
```
olivielpeau: just curious: why do you set `autorefresh=0` here?

olivielpeau: Also, and more importantly, we need to update the `baseurl` to the prod/stable repo before we merge right?

packaging/datadog-agent/source/install_agent.sh
gmmeyer: I think that was an accident, I'll fix it.

packaging/datadog-agent/source/install_agent.sh
gmmeyer: I had to check, because I thought I had a reason for not choosing this. `/etc/SuSE-release` is deprecated:

```
greg@linux-ely7:/etc> cat SuSE-release 
SUSE Linux Enterprise Server 12 (x86_64)
VERSION = 12
PATCHLEVEL = 1
# This file is deprecated and will be removed in a future service pack or release.
# Please check /etc/os-release for details about this release.
```

We can use /etc/os-release or /etc/issue.

packaging/datadog-agent/source/install_agent.sh
gmmeyer: good idea!

gmmeyer:
I addressed those notes, though there's still one open question.

packaging/datadog-agent/source/install_agent.sh
```diff
@@ -158,6 +160,24 @@ If the cause is unclear, please contact Datadog support.
...
+  UNAME_M=$(uname -m)
+  if [ "$UNAME_M"  == "i686" -o "$UNAME_M"  == "i386" -o "$UNAME_M"  == "x86" ]; then
+      printf "\033[31mThe Datadog Agent installer is only available for 64 bit SUSE Enterprise machines.\033[0m\n"
+      exit;
+  else
+      ARCHI="x86_64"
+  fi
+
+  echo -e "\033[34m\n* Installing YUM Repository for Datadog\n\033[0m"
+  $sudo_cmd sh -c "echo -e '[datadog]\nname=datadog\nenabled=1\nbaseurl=https://yum.datadoghq.com/suse-testing/x86_64\ntype=rpm-md\ngpgcheck=1\nrepo_gpgcheck=0\ngpgkey=https://yum.datadoghq.com/DATADOG_RPM_KEY.public' > /etc/zypp/repos.d/datadog.repo"
```
olivielpeau: `suse-testing/x86_64` -> `suse-rpm/x86_64` right? :)

gmmeyer: 👍

PR #2938, opened 2016-10-20
[core] display expected service check count

degemer:
### What does this PR do?

We add one service check per check, to see whether the check is broken
or not. The info command then shows a misleading number service checks.

Let's fix it and not count the additional service check.

### Motivation

Misleading service check count.

### Testing Guidelines
Run info.

### Additional Notes
Backward compatibility issue: unlikely, this only issue would be custom scripts depending on the info output.


yannmh:
I am all for it.

degemer:
Thanks for the review, merging.

PR #2940, opened 2016-10-20
[Mongodb] Check was sending password in the clear if the password inc…

derekwbrown:
### What does this PR do?

PR fixes two problems which resulted in passwords being sent in the clear.
Problem 1 was that passwords are URL decoded for supplying to mongo, but the replace string didn't account for that.

Problem 2 was that the regular expression for replacing passwords in a URI looks for username:password@host....  If the username had a "_" in it, the regex wouldn't match and the password would be left.

### Motivation

Customer report.

…ludes

a character that needs to be URL encoded (i.e. '/').  Account for
passwords that include URL encoded chars

Also allow "_" to be a username in the regular expression matching
username/password in a URI.  Otherwise, a username with a "_" will not
match, and the password will be sent in the clear

### DIFF
--- a/checks.d/mongo.py
+++ b/checks.d/mongo.py
@@ -628,7 +628,14 @@ class MongoDb(AgentCheck):
         db_name = parsed.get('database')
         additional_metrics = instance.get('additional_metrics', [])
 
-        clean_server_name = server.replace(password, "*" * 5) if password else server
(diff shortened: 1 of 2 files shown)


remh:
Test is failing on httpbin having issues. Merging. Thanks @derekwbrown

PR #2941, opened 2016-10-20
[collector][emitter] Ignore all non-unicode-decodable characters

olivielpeau:
### What does this PR do?

On Japanese and Korean versions of Windows, We've seen cases where
the json encoding would fail on a UnicodeDecodeError because of
non-utf8-decodable characters in strings, which causes the whole emitter to fail.

This fixes the issue by making another pass on the
whole payload to remove all the non-decodable characters and log the
affected strings.

### Testing Guidelines

Added a mock test on the function that removes non-decodable chars.

### DIFF
--- a/emitter.py
+++ b/emitter.py
@@ -29,34 +29,47 @@ control_chars = ''.join(map(unichr, range(0, 32) + range(127, 160)))
 control_char_re = re.compile('[%s]' % re.escape(control_chars))
 
 
-def remove_control_chars(s):
+def remove_control_chars(s, log):
     if isinstance(s, str):
         sanitized = control_char_re.sub('', s)
     elif isinstance(s, unicode):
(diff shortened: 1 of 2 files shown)


olivielpeau:
Basically, when `json.dumps` can't encode the payload to json (and only in that case), this removes the non-ascii characters that are in python `string`s and that can't be decoded using the `utf-8` encoding (instead of failing completely)

So events with non-roman characters should be fine as long as they're utf8-decodable.

@gmmeyer Let me know if that's clear enough for you, if so I'll merge this :)

gmmeyer:
Yes, merge this if you feel it's ready! I was just curious what the outcome would be.

olivielpeau:
Ok, thanks @gmmeyer! Merging

PR #2942, opened 2016-10-20
[win32] Handle tzname encodings

olivielpeau:
### What does this PR do?

Fix encodings of timezone metadata on Windows systems on which `time.tzname` returns non-`ascii` characters. I've been able to confirm on a Korean and on a Japanese system that `locale.getpreferredencoding` returns the correct encoding.

### Motivation
- Non-ascii characters make the emitter crash (at least until #2941)
- the timezone is useless when it's not decoded

### Testing

Added some tests that give example inputs/outputs
This has also been tested successfully on customers' machines.

### DIFF
--- a/checks/collector.py
+++ b/checks/collector.py
@@ -4,6 +4,7 @@
 
 # stdlib
 import collections
+import locale
 import logging
 import pprint
 import socket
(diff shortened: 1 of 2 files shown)


olivielpeau:
@gmmeyer thanks for the review! Unfortunately I don't have other examples of inputs/outputs for other localizations as I don't have environments or found customers running versions of Windows with other "non-ascii" localizations. Spinning up other windows environments would take a while and I'm not sure it'd be worth it. What do you think?

gmmeyer:
I think that's reasonable. Don't let my observation be blocking.

PR #2947, opened 2016-10-21
[status] Display api key status in the info page

remh:
### What does this PR do?

Display api key status in the info page

### Motivation

Feature request

### DIFF
--- a/checks/check_status.py
+++ b/checks/check_status.py
@@ -19,6 +19,7 @@ import time
 
 # 3p
 import ntplib
+import requests
 import yaml
 
 # project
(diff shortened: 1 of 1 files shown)


degemer:
Nice! Shouldn't we just stop the agent when starting if the it's a bad API key?
If we want to keep it in the status, could we let the forwarder do if from time to time and add it to the `ForwarderStatus.pickle`? (since https://github.com/DataDog/dd-agent/blob/0eb8392b0c644668809bd1412d204e96fedbde05/checks/check_status.py#L137)

remh:
No some people run the agent without a valid api key on purpose.

I don't think we should have the forwarder do it, why would it be better ? The `ForwarderStatus.pickle` was created to have the forwarder process dump information that only the forwarder has. In this case that's not the case, it's easy to retrieve the api key status so we might as well do it when displaying the info page.

PR #2948, opened 2016-10-21
[consul] Support multiple instances

olivielpeau:
### What does this PR do?

Support multiple instances of the `consul` check

### Motivation

Customer request: the Agent needs to support multiple instances of the check when an Agent monitors multiple containers running consul servers.

### Testing Guidelines

Updated the mock tests.

### Additional Notes

Also adds support of yaml-configured tags on each instance to
distinguish data from different instances running on the same Agent

### DIFF
--- a/checks.d/consul.py
+++ b/checks.d/consul.py
@@ -11,6 +11,7 @@ from urlparse import urljoin
 
 # project
 from checks import AgentCheck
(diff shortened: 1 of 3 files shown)


yannmh:
Looks good to me !

PR #2950, opened 2016-10-24
[core][subprocess output] Do not set close_fds=True

remh:
### What does this PR do?
Do not set close_fds=True

This was introduced by
https://github.com/DataDog/dd-agent/commit/24cf7823c254925ab061baf6aff712678a4c793f

But it’s not needed anymore and as a result the collector is looping
over thousand of file descriptors, trying to close all file descriptors between 3 and the max number of open files descriptors (most of the
time it cannot) and use a lot of CPU to do so.

See: https://github.com/python/cpython/blob/2.7/Lib/subprocess.py#L1185-L1197 for implementation

This commit change that
and revert to the default `close_fds=False`

Also I’m removing a bunch of options that are not used anywhere.

### Motivation
Thanks a lot to @ross for pointing that out.

### DIFF
--- a/checks.d/varnish.py
+++ b/checks.d/varnish.py
@@ -107,7 +107,8 @@ class Varnish(AgentCheck):
 
     def _get_version_info(self, varnishstat_path):
         # Get the varnish version from varnishstat
-        output, error, _ = get_subprocess_output([varnishstat_path, "-V"], self.log)
+        output, error, _ = get_subprocess_output([varnishstat_path, "-V"], self.log,
+            raise_on_empty_output=False)
 
         # Assumptions regarding varnish's version
         use_xml = True
--- a/utils/subprocess_output.py
+++ b/utils/subprocess_output.py
@@ -9,16 +9,13 @@ import logging
 import subprocess
 import tempfile
 
-# project
-from utils.platform import Platform
-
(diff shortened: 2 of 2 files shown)


ross:
Obviously not a big deal with `MAXFD` is small, but can result in lots of systems calls when it's really high. Thanks!

remh:
Investigating the failing varnish test failure

checks.d/varnish.py
```diff
@@ -107,7 +107,11 @@ def check(self, instance):
 
     def _get_version_info(self, varnishstat_path):
         # Get the varnish version from varnishstat
-        output, error, _ = get_subprocess_output([varnishstat_path, "-V"], self.log)
+        try:
+            output, error, _ = get_subprocess_output([varnishstat_path, "-V"], self.log)
```
olivielpeau: 🙀 
```
  File "/home/travis/build/DataDog/dd-agent/checks.d/varnish.py", line 120, in _get_version_info
    m1 = re.search(r"varnish-(\d+)", output, re.MULTILINE)
UnboundLocalError: local variable 'output' referenced before assignment
```

remh: 😅

PR #2955, opened 2016-10-25
[consul] Remove warning when no whitelist is defined

### LINKED ISSUE #2954: Consul Integration issue warning when whitelist not specified and < 50 services
The Consul integration will indicate that there is an integration issue when a whitelist has not been specified.

But in our current scenario we have less than 50 services registering with consul, and would not like to have to maintain or engineer a solution to update that whitelist every time a dynamic service comes online or goes offline from consul.

We would like the consul integration to only report a warning when there are more than 50 services discovered and a whitelist has not been specified, because at that point the integration starts to truncate the service discovery list. 

Thanks,

Jackson
Alianza Inc.

olivielpeau:
### What does this PR do?

As long as the number of polled services is lower than the limit, let's
just log that no whitelist has been defined. We should only add a
warning when there's an actual issue.

### Motivation

Fixes #2954, removes a spurious warning from the info page of the Agent and from the infra list of the webapp.

### Testing

None for this trivial change, but we have mock and integration tests on this check.

### DIFF
--- a/checks.d/consul.py
+++ b/checks.d/consul.py
@@ -204,7 +204,7 @@ class ConsulCheck(AgentCheck):
             services = [s for s in services if s in service_whitelist][:self.MAX_SERVICES]
         else:
             if len(services) <= self.MAX_SERVICES:
-                self.warning('Consul service whitelist not defined. Agent will poll for all %d services found' % len(services))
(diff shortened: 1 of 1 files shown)


gmmeyer:
👍

PR #2956, opened 2016-10-25
[Kafka] Document specifying multiple Kafka / ZK nodes

### LINKED ISSUE #2943: If kafka_consumer.py can accept multiple ZK/Kafka servers, document it
It's not clear from https://github.com/DataDog/dd-agent/blob/master/conf.d/kafka_consumer.yaml.example whether the Kafka and Zookeeper hosts can be a list or only a single host. In a production environment, a list is preferable for redundancy.

If this is possible, it should be documented.

jeffwidman:
### What does this PR do?

Documents how to specify multiple Kafka/ZK nodes. 

### Motivation

In a production environment, it's often useful to specify multiple Kafka / Zookeper nodes for a single check instance. This way you only generate a single check process, but if one host goes down, KafkaClient / KazooClient will try contacting the next host. Generating multiple check processes is a bad solution because then you're doing redundant reads against Kafka (not a big deal) and Zookeeper (can be a bottleneck).

### Testing Guidelines

Modify config to see if it works.

Fixes #2943

### DIFF
--- a/conf.d/kafka_consumer.yaml.example
+++ b/conf.d/kafka_consumer.yaml.example
@@ -11,3 +11,15 @@ instances:
   #   consumer_groups:
   #     my_consumer:
   #       my_topic: [0, 1, 4, 12]
+  
+  # Production example with redundant hosts:
+  # In a production environment, it's often useful to specify multiple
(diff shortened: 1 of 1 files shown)


remh:
Thank you!

PR #2957, opened 2016-10-25
[psutil] Upgrade to 4.4.1

olivielpeau:
### What does this PR do?

Upgrades psutil to the latest version (`4.4.1`), and related changes on `dd-agent` (see details in commit messages):
- [disk] Use `in_use` percentage provided by psutil
- [win32][system] Use `psutil` to sample `cpu.interrupt`: tested on my local windows VM: `psutil` and `WMI` send the same values for `cpu.interrupt`

### Motivation

The upgrade was requested by a customer so that we ship a version that includes this fix: https://github.com/giampaolo/psutil/issues/761
More generally it's good to keep an up-to-date version.

### Testing Guidelines

Went through the changelog of `psutil` and through the occurences of `psutil` in `dd-agent` to find things that could be affected. Not sure I've spotted everything but I'll closely monitor the deploy of the updated Agent on our staging infrastructure (especially for the `gunicorn` check that could potentially be affected).

### Additional Notes

Related PR on `omnibus-software`: https://github.com/DataDog/omnibus-software/pull/82
`psutil` changelog: https://github.com/giampaolo/psutil/blob/master/HISTORY.rst

### DIFF
--- a/checks/system/win32.py
+++ b/checks/system/win32.py
@@ -201,65 +201,20 @@ class Cpu(Check):
     def __init__(self, logger):
         Check.__init__(self, logger)
 
-        # Sampler(s)
-        self.wmi_sampler = WMISampler(
-            logger,
-            "Win32_PerfRawData_PerfOS_Processor",
-            ["Name", "PercentInterruptTime"]
-        )
-
         self.counter('system.cpu.user')
         self.counter('system.cpu.idle')
-        self.gauge('system.cpu.interrupt')
         self.counter('system.cpu.system')
+        self.counter('system.cpu.interrupt')
 
     def check(self, agentConfig):
-        try:
-            self.wmi_sampler.sample()
-        except TimeoutException:
-            self.logger.warning(
-                u"Timeout while querying Win32_PerfRawData_PerfOS_Processor WMI class."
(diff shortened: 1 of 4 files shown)


checks/system/win32.py
```diff
@@ -201,66 +201,21 @@ class Cpu(Check):
...
-            logger,
-            "Win32_PerfRawData_PerfOS_Processor",
-            ["Name", "PercentInterruptTime"]
-        )
-
         self.counter('system.cpu.user')
         self.counter('system.cpu.idle')
-        self.gauge('system.cpu.interrupt')
         self.counter('system.cpu.system')
+        self.counter('system.cpu.interrupt')
```
remh: why ? 
is that a breaking change ?

olivielpeau: Basically an old-style-check `counter`s is the equivalent of a `rate`s of a new-style check. It's still stored as a gauge in the backend, so there's no backwards-compatibility issue.

We compute a `rate` here because `psutil.cpu_times` returns for `interrupt` the same kind of value as for the other cpu values, i.e. the total time the cpu(s) spent in the interrupt state since startup, so to get a percentage from this we need to compute the rate divided by the number of cpus and multiply that by 100.

PR #2958, opened 2016-10-26
[elasticsearch] Added a couple of tweaks to work with Elasticsearch 5.x and release candidates

ptinsley:
_Note: Please remember to review the Datadog [Contribution Guidelines](https://github.com/DataDog/dd-agent/blob/master/CONTRIBUTING.md)
if you have not yet done so._
### What does this PR do?

There are two changes in this PR: 
- Allows the elasticsearch plugin to handle pre-release versions of elasticsearch. 5.0.0-rc1 as an example, the plugin expect a purely numeric result. As the status of RC vs GA should not affect the metrics / feature flags in the plugin, I simply truncate the -RCx or anything - and beyond.
- Allows the plugin to work with 5.x versions of elasticsearch, they have changed the behavior of the API endpoint for node statistics in 5.0.0 to return all stats by default and requiring you to filter down if you _don't_ want all the stats. As a result they have removed the flag to ask for all.

Request against 5.x with the all parameter defined:

```
curl http://X.X.X.X:9200/_nodes/_local/stats?all=true
{"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"request [/_nodes/_local/stats] contains unrecognized parameter: [all]"}],"type":"illegal_argument_exception","reason":"request [/_nodes/_local/stats] contains unrecognized parameter: [all]"},"status":400}
```

Removing the ?all=true from the url causes the request to complete successfully and return statistics.
### Motivation

We are using ES 5 and I wanted to try and tune performance with Datadog.
### Testing Guidelines

### Additional Notes

I'm not experienced in Python, sorry if I did something silly, feel free to rewrite the code as you see fit :)

As for the approach I took of setting a flag for ?all vs overriding the setting of the variable directly. It's currently set in 4 different places based on a couple of different sets of variables. It seemed like the right thing to just append that if by default and let a 5.x version check ask for it to be left off.

Also, sorry if others disagree with the truncation of the -'d portion of the version string, it just seemed like the simplest approach given that it shouldn't have any behavioral impact on interacting with ES which appears to be the entire reason the plugin grabs the version in the first place.

### DIFF
--- a/checks.d/elastic.py
+++ b/checks.d/elastic.py
@@ -586,7 +586,7 @@ class ESCheck(AgentCheck):
                 verify=verify,
                 cert=cert
             )
- 
+
             resp.raise_for_status()
         except Exception as e:
             if send_sc:
--- a/checks.d/elastic.py
+++ b/checks.d/elastic.py
@@ -397,6 +397,7 @@ class ESCheck(AgentCheck):
         # (URLs and metrics) accordingly
         version = self._get_es_version(config)
 
+
         health_url, nodes_url, stats_url, pshard_stats_url, pending_tasks_url, stats_metrics, \
             pshard_stats_metrics = self._define_params(version, config.cluster_stats)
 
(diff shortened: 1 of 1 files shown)


bai:
Confirming that without this patch Elasticsearch checks fail with `400 Bad Request`.

olivielpeau:
Hi @ptinsley, thanks for this PR!

@alexef has opened a pretty similar PR (#2980), there are small things I like and dislike about both of PRs actually, so I'll open a separate PR (and preserve your authorship ;) )

olivielpeau:
Resolved via #2989, thanks again @ptinsley!

Closing

PR #2967, opened 2016-10-27
[pgbouncer] SSL support. Fix service check messages and tags

remh:
### What does this PR do?

This fixes the service check messages and tags to support SSL which was introduced with #2885


masci:
LGTM 👍

PR #2973, opened 2016-10-28
[collector][system] Catch exception raised by system checks

olivielpeau:
### What does this PR do?

Catch exceptions raised by system checks and log them
### Motivation

There's no reason to let the Agent crash when system checks fail
### Additional Notes

We already do this for Windows system checks so it should
be safe to do it for unix checks.


remh:
Thanks @olivielpeau 
Most exceptions should be called at the `check()` level. Was there a case where it was not the case ? If so we should catch it at that level as well.

PR #2976, opened 2016-10-29
Send all configured tags with process checks.

### LINKED ISSUE #2974: Process checks do not pass configured tags
The [`process`](https://github.com/DataDog/dd-agent/blob/master/checks.d/process.py) check does not pass the [tags](https://github.com/DataDog/dd-agent/blob/master/checks.d/process.py#L301) configured on the `instance` block for the check. Right now it only sends the [process name](https://github.com/DataDog/dd-agent/blob/master/checks.d/process.py#L369).

Could the check include the same tags as the metrics?

jdavisp3:
Send the tags configured on the `instance` with the process checks. Given the tests around the old behavior I imagine there might be some reason why this isn't the case now, but just in case there isn't I think it would be useful if the checks had the same tags that were configured for the metrics.

This fixes #2974, FYI.

### DIFF
--- a/checks.d/process.py
+++ b/checks.d/process.py
@@ -374,7 +374,7 @@ class ProcessCheck(AgentCheck):
             if vals:
                 self.rate('system.processes.%s' % mname, sum(vals), tags=tags)
 
-        self._process_service_check(name, len(pids), instance.get('thresholds', None))
+        self._process_service_check(name, len(pids), instance.get('thresholds', None), tags)
 
     def _get_pid_set(self, pid):
         try:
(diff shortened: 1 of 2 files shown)


jdavisp3:
Oh, right, I missed that the two tags were actually different. I added the `process:name` tag back in f272356. Thanks @olivielpeau !

olivielpeau:
Thanks @jdavisp3! I've made a few cosmetic changes and pushed to your branch.

Merging!

jdavisp3:
Thanks very much!

PR #2982, opened 2016-11-01
[Win32] Terminate agent-manager.exe on uninstall/upgrade.

derekwbrown:
### What does this PR do?

Modifies the installer so that if agent-manager.exe is running, it is automatically terminated during uninstall and/or upgrade.  This prevents requiring a reboot after upgrade.

### Motivation

Multiple failed installs due to agent-manager.exe running in the background.

### Additional Notes

Modifies install script to terminate agent-manager during uninstall or
upgrade.  Otherwise, agent-manager is running in the background and can't
be upgraded, leading to an unstable install

### DIFF
--- a/win32/gui.py
+++ b/win32/gui.py
@@ -845,6 +845,10 @@ if __name__ == '__main__':
     if Platform.is_windows():
         # Let's kill any other running instance of our GUI/SystemTray before starting a new one.
         kill_old_process()
+        if len(sys.argv) > 1 and "-stop" in sys.argv:
+            # just return.  The kill_old_process() should have terminated the process,
+            # and now we're done.
+            sys.exit(0)
 
     app = QApplication([])
     if Platform.is_mac():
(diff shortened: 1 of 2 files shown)


win32/gui.py
```diff
@@ -845,6 +845,10 @@ def kill_old_process():
     if Platform.is_windows():
         # Let's kill any other running instance of our GUI/SystemTray before starting a new one.
         kill_old_process()
+        if len(sys.argv) > 1 and "-stop" in sys.argv:
+            # just return.  The kill_old_process() should have terminated the process,
+            # and now we're done.
+            sys.exit(0)
```
yannmh: I don't understand what this is for.

derekwbrown: I can take it out if need be. Originally, I thought I was going to need to "shell out" from the installer (in Windows parlance, a Custom Action) to stop the agent-manager if it's running.  Turns out I didn't need to.  However, it seems potentially useful to be able to stop it from the command line.  There'd be no harm removing.

yannmh: So
```
agent-manager.exe -stop
```
should stop the background process ?

derekwbrown: Yes.

yannmh:
Let's merge it 🚀

PR #2983, opened 2016-11-02
password cast to string

michalyao:
if you have not yet done so.*

### What does this PR do?

Password casted to string in mysql.py

### Motivation

Pure number password without  quotes causes an error in pymysql. 

### Additional Notes

If password only contains number, it must be surrounded by quotes, otherwise the pymysql will raise an error. So I cast it to a string to avoid misunderstanding. Or leave it alone to let the user to determine when to use a string? Just my personal opinion.


remh:
Thanks!

PR #2985, opened 2016-11-02
[JMXFetch] [service discovery] enable service discovery for jmx checks

hkaj:
# WIP


agent.py
```diff
@@ -82,6 +88,18 @@ def __init__(self, pidfile, autorestart, start_event=True, in_developer_mode=Fal
         self.reload_configs_flag = False
         self.sd_backend = None
 
+        if Platform.is_windows():
+            pipe_name = SD_PIPE_WIN_PATH.format(pipename=SD_PIPE_NAME)
+        else:
+            pipe_name = os.path.join("/tmp", "dd-service_discovery")
```
remh: You shouldn't use "/tmp"

PR #2999, opened 2016-11-04
[kubernetes state] Use a less ambiguous metric name and tags

remh:
### What does this PR do?

This renames a metric name to make it less ambiguous. Otherwise we could end with 
`kubernetes_state.nodes.unschedulable {status:available}` which is quite confusing


masci:
👍

PR #3000, opened 2016-11-04
[postgres] explicit that the drivers behave differently when using unix sockets

masci:
### What does this PR do?

Adds a line of comment

### Motivation

I was confused by how pg8000 behaves when using unix sockets


remh:
3000!

PR #3004, opened 2016-11-07
[core] hard-deprecate start/stop/restart/status commands

degemer:
### What does this PR do?

This commits still allows these options, but only in developer mode,
removing the risk for most people to misuse them.

It also removes the deprecation notice for the old command line
executables: they are not shipped anymore, and the main reason why they
were deprecated was the start/stop/restart/status commands. Since this
commit takes care of that, the notices are not needed anymore.

### Motivation

All of these are handled by Supervisord, which launches the processes in
the foreground. This is a complete legacy code, and some people still
use it wrongly.


miketheman:
@degemer How are people meant to use non-supervisord approaches, where all we want is the dogstatsd listener and relay? (see linked issue above).

This has basically broken the Heroku Buildpack approach.

degemer:
@miketheman Sorry to hear that. 😞 
As a temporary workaround, you can set `developer_mode: yes` in datadog.conf - https://github.com/DataDog/dd-agent/blob/master/datadog.conf.example#L54. It will allow you to use the `start` command (and doesn't affect dogstatsd, developer_mode is only used in the collector).

I don't know Heroku, would it be possible to do the same as the dogstatsd-only docker image and run supervisor with a custom configuration (starting only dogstatsd & the forwarder)? https://github.com/DataDog/docker-dd-agent/blob/master/dogstatsd/Dockerfile & https://github.com/DataDog/docker-dd-agent/blob/master/dogstatsd/supervisor.conf
Or is it just impossible to run supervisor?

irabinovitch:
@jyee could you help take a look at this as well?

miketheman:
I'd point out that removing a working interface in a non-6.x version seems really wrong, SemVer-wise.

PR #3005, opened 2016-11-08
[haproxy] unix socket stats urls

sj26:
### What does this PR do?

Support unix sockets for haproxy stats collection.

### Motivation

Running a separate stats socket with tight permissions instead of over a tcp port provides greater security.

### Testing Guidelines

Sorry, I have no idea how to get the tests actually working in a local environment. It has been a world of pain. I'm hoping travis does a better job at running them and providing some useful feedback.

I'm on macOS 10.12, using homebrew, ruby 2.3.1, python 2.7.12. `rake` aborts on rubocop errors. So does `rake ci:run`. `rake ci:run[haproxy]` tries to compile haproxy which can't find openssl, even if I supply correct CFLAGS and LDFLAGS to my homebrew'd openssl install as env. Subsequent `rake ci:run[haproxy]` runs don't try to recompile haproxy, they fail with `Errno::ENOENT No such file or directory - /Users/sj26/Projects/dd-agent/embedded/haproxy_1.5.11/haproxy`. And cleanup fails because it can't find an haproxy pid.

I haven't attempted to figure out mocks yet.

### Additional Notes

I realise this introduces a new dependency which requires updating your omnibus, but I'd like to get the code working first before jumping on that dragon.

### DIFF
--- a/checks.d/haproxy.py
+++ b/checks.d/haproxy.py
@@ -8,10 +8,7 @@ import copy
 import re
 import socket
 import time
-try:
-    import urlparse
-except LoadError:
-    import urllib.parse as urlparse
+import urlparse
 
 # 3rd party
 import requests
--- a/checks.d/haproxy.py
+++ b/checks.d/haproxy.py
@@ -6,12 +6,15 @@
 from collections import defaultdict
 import copy
 import re
+import socket
 import time
+try:
+    import urlparse
+except LoadError:
+    import urllib.parse as urlparse
 
 # 3rd party
 import requests
-import requests_unixsocket
-requests_unixsocket.monkeypatch()
 
 # project
 from checks import AgentCheck
@@ -99,8 +102,20 @@ class HAProxy(AgentCheck):
 
     def check(self, instance):
         url = instance.get('url')
-        username = instance.get('username')
-        password = instance.get('password')
+        self.log.debug('Processing HAProxy data for %s' % url)
+
+        parsed_url = urlparse.urlparse(url)
+
+        if parsed_url.scheme == 'unix':
+            data = self._fetch_socket_data(parsed_url.path)
+
+        else:
+            username = instance.get('username')
+            password = instance.get('password')
+            verify = not _is_affirmative(instance.get('disable_ssl_validation', False))
+
+            data = self._fetch_url_data(url, username, password, verify)
+
         collect_aggregates_only = _is_affirmative(
             instance.get('collect_aggregates_only', True)
         )
@@ -129,12 +144,6 @@ class HAProxy(AgentCheck):
 
         custom_tags = instance.get('tags', [])
 
-        verify = not _is_affirmative(instance.get('disable_ssl_validation', False))
-
-        self.log.debug('Processing HAProxy data for %s' % url)
-
-        data = self._fetch_data(url, username, password, verify)
-
         process_events = instance.get('status_check', self.init_config.get('status_check', False))
 
         self._process_data(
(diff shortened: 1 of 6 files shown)


masci:
Hi @sj26 and thanks for your code, we now that running the tests locally is not as easy as it should be, sorry for that!

Aside from the details that are making the CI fail, I'm a little bit concerned about the overall approach - if I understood correctly, by `monkey_patch`ing requests you're changing the behaviour of the check entirely, switching from TCP to unix sockets.

I agree moving to unix socket would improve security but we cannot break the compatibility with existing configurations. I'd make the check compatible with both approaches, defaulting to the old behaviour and making available the new one through the config file.

What do you think?

sj26:
The monkey patch shouldn't prevent requests from functioning as normal, or change any existing integrations, it just adds an extra adapter for the "http+unix" scheme. :-)

ci/haproxy.rb
```diff
@@ -44,8 +44,13 @@ def haproxy_rootdir
...
     task before_script: ['ci:common:before_script'] do
       %w(haproxy haproxy-open).each do |name|
+        # Older haproxy doesn't support ENV interpolation
+        config = File.read("#{ENV['TRAVIS_BUILD_DIR']}/ci/resources/haproxy/#{name}.cfg")
+        config.gsub!("$VOLATILE_DIR", ENV["VOLATILE_DIR"])
+        File.write("#{ENV["VOLATILE_DIR"]}/#{name}.cfg", config)
+
         pid = spawn("#{haproxy_rootdir}/haproxy", '-d', '-f',
-                    "#{ENV['TRAVIS_BUILD_DIR']}/ci/resources/haproxy/#{name}.cfg",
+                    "#{ENV["VOLATILE_DIR"]}/#{name}.cfg",
```
masci: change these to single quotes to make rubocop stop complaining

masci:
Nice, that was my impression looking at the CI but I discovered the existence of `requests_unixsocket` only this morning 😄

checks.d/haproxy.py
```diff
@@ -147,19 +158,38 @@ def check(self, instance):
...
+
+        response = ""
+        output = sock.recv(8192)
+        while output:
+            response += output.decode("ASCII")
+            output = sock.recv(8192)
 
-        r = requests.get(url, auth=auth, headers=headers(self.agentConfig), verify=verify, timeout=self.default_integration_http_timeout)
-        r.raise_for_status()
+        sock.close()
```
sj26: Most of this was borrowed from https://github.com/neurogeek/haproxyctl/blob/master/haproxy/conn.py#L51-L83

sj26:
@masci sorry for the delays.

Turns out I was completely wrong and the admin/stats sockets don't run HTTP at all but uses their own line-based protocol. I've updated the implementation to accept `unix:///var/run/haproxy.sock` style values and talk unix sockets directly instead of using http in those cases. It actually makes things easier, and we don't need another egg.

Take another look?

checks.d/haproxy.py
```diff
@@ -147,19 +158,38 @@ def check(self, instance):
...
+        ''' Hit a given stats socket and return the stats lines '''
+
+        self.log.debug("Fetching haproxy stats from socket: %s" % socket_path)
+
+        sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+        sock.connect(socket_path)
+        sock.send("show stat\r\n")
+
+        response = ""
+        output = sock.recv(8192)
```
masci: Can you define this number as a constant on the top of the module? Like:
`BUFSIZE=8192`

conf.d/haproxy.yaml.example
```diff
@@ -5,6 +5,10 @@ instances:
     # username: username
     # password: password
     #
+    # or, with a unix stats or admin socket:
+    #
+    # url: unix:///var/run/haproxy.sock
```
masci: Nitpick: could you add a leading `-` to the example so it's clear that should be another instance and not an attribute of the first one?
```
 instances:
   - url: http://localhost/admin?stats
      # username: username
      # password: password
   # or, with a unix stats or admin socket:
   # - url: unix:///var/run/haproxy.sock
```

checks.d/haproxy.py
```diff
@@ -6,7 +6,12 @@
 from collections import defaultdict
 import copy
 import re
+import socket
 import time
+try:
+    import urlparse
+except LoadError:
```
sj26: :+1:

checks.d/haproxy.py
sj26: :+1:

conf.d/haproxy.yaml.example
sj26: Hmm, that looks odd to me, but updated.

masci: Yeah at first I thought it'd be clearer but doesn't seem so...
Feel free to adjust it as you think it's better from the user perspective, or leave it as it is and we'll fix that later.

sj26:
Thanks!

PR #3010, opened 2016-11-09
[service_discovery][jmx] trying to pick-up JMX changes with SD.

truthbk:
### What does this PR do?
Enables service discovery for JMX integrations. 

Once a service discovery index config is detected, we parse the config from JSON (as stored in etcd/consul) to YAML and the relayed over to JMXfetch over a named pipe `dd-service_discovery` that lives in `/tmp` by default.

If multiple changes are detected, these are all submitted to JMXfetch by a single write operation to avoid race conditions. The pipe should take care of locking from the reading and writing ends, and by dropping all relevant configs in a single write operation we can guarantee all configs will be read on the JMXfetch in a single iteration and thus help avoid having multiple versions of the same config in the pipe at any given time.

### Motivation

Enabling Service Discovery for JMXfetch integrations.

### Testing Guidelines

### Additional Notes

Related PR: https://github.com/DataDog/jmxfetch/pull/113

### DIFF
--- a/agent.py
+++ b/agent.py
@@ -19,6 +19,8 @@ import os
 import signal
 import sys
 import time
+import supervisor.xmlrpc
+import xmlrpclib
 from copy import copy
 
 # For pickle & PID files, see issue 293
@@ -32,10 +34,12 @@ from config import (
     get_parsed_args,
     get_system_stats,
     load_check_directory,
-    load_check
+    load_check,
+    generate_jmx_configs
 )
 from daemon import AgentSupervisor, Daemon
 from emitter import http_emitter
+from utils.platform import Platform
 
 # utils
 from util import Watchdog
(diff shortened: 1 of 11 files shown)


agent.py
```diff
@@ -50,11 +54,21 @@
...
 PID_NAME = "dd-agent"
 PID_DIR = None
 WATCHDOG_MULTIPLIER = 10
 RESTART_INTERVAL = 4 * 24 * 60 * 60  # Defaults to 4 days
 
+JMX_SUPERVISOR_ENTRY = 'datadog-agent:jmxfetch'
+JMX_GRACE_SECS = 2
+SERVICE_DISCOVERY_PREFIX = 'SD-'
+SD_PIPE_NAME = "dd-service_discovery"
+SD_PIPE_UNIX_PATH = "/tmp"
```
degemer: Quick note: I think using `/opt/datadog-agent/run` is more appropriate and easier to adapt to Windows, unless you have a good reason to use `/tmp`. 😃

truthbk: Named pipes are only supported on windows (and I don't think all versions apply) in a special path, so it doesn't really apply to windows.

I used `/tmp` to avoid issues with the source installs.

truthbk:
TODO: I could probably add some tests for the `generate_jmx_configs` function.

agent.py
```diff
@@ -216,9 +248,17 @@ def run(self, config=None):
         if self._agentConfig.get('service_discovery'):
             self.sd_backend = get_sd_backend(self._agentConfig)
 
+        # Initialize Supervisor proxy (unix specific)
+        self.supervisor_proxy = self._get_supervisor_socket(self._agentConfig)
```
hkaj: do we want to init that if SD is not enabled?

agent.py
```diff
@@ -216,9 +248,17 @@ def run(self, config=None):
...
+
         # Load the checks.d checks
         self._checksd = load_check_directory(self._agentConfig, hostname)
 
+        # Load JMX configs if available
+        jmx_sd_configs = generate_jmx_configs(self._agentConfig, hostname)
+        if jmx_sd_configs:
+            self._submit_jmx_service_discovery(jmx_sd_configs)
```
hkaj: ditto, this will trigger even without SD, is that expected?

config.py
```diff
@@ -1,4 +1,3 @@
-# (C) Datadog, Inc. 2010-2016
```
truthbk: Not sure how I got rid of that! 😓

config.py
```diff
@@ -1098,6 +1103,35 @@ def load_check(agentConfig, hostname, checkname):
...
+                            'instances': sd_instances}
+
+            try:
+                yaml = config_to_yaml(check_config)
+                # generated["{}_{}".format(check_name, idx)] = yaml
+                generated["{}_{}".format(check_name, 0)] = yaml
+                log.debug("YAML generated: %s", yaml)
+            except Exception as e:
+                log.exception("Unable to generate YAML config for %s: %s", check_name, e)
+
```
truthbk: good catch!

hkaj:
LGTM

truthbk:
No changes to the logic, just added the test. We're green, so merging! Thanks @hkaj

PR #3013, opened 2016-11-10
[psutil] Only set `psutil.PROCFS_PATH` once in the collector

olivielpeau:
### What does this PR do?

Fixes the `procfs_path` customization in the `process` check.  Previously the `procfs_path` specified in `process.yaml` wouldn't be taken into account because `psutil.PROCFS_PATH` would get overwritten in the `__init__` method of other checks (in particular the `disk` check since it runs by default)

Also, since `PROCFS_PATH` is a global setting of the `psutil` lib it makes sense to set it once for the whole collector and only allow its customization from `datadog.conf`, so let's deprecate the setting in `process.yaml`.

### Additional Notes

🐮

### DIFF
--- a/checks.d/process.py
+++ b/checks.d/process.py
@@ -76,12 +76,16 @@ class ProcessCheck(AgentCheck):
             )
         )
 
+        self._conflicting_procfs = False
+        self._deprecated_init_procfs = False
         if Platform.is_linux():
             procfs_path = init_config.get('procfs_path')
-            if not procfs_path:
-                procfs_path = self.agentConfig.get('procfs_path', '/proc').rstrip('/')
-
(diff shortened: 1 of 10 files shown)


checks.d/process.py
```diff
@@ -305,6 +309,13 @@ def check(self, instance):
         pid = instance.get('pid')
         pid_file = instance.get('pid_file', None)
 
+        if self._conflicting_procfs:
+            self.warning('The `procfs_path` defined in `process.yaml` is different from the one defined in '
+                         '`datadog.conf`. This is currently not supported by the Agent')
```
truthbk: We should probably state here we're defaulting to the `datadog.conf` specified path.

PR #3018, opened 2016-11-10
[forwarder] stop flushing after 10s

degemer:
### What does this PR do?

[forwarder] stop flushing after 10s (so that new payloads can be sent).

### Motivation

Very long running flushes which delay the metrics.


transaction.py
```diff
@@ -197,6 +197,12 @@ def flush(self):
     def flush_next(self):
 
         if self._trs_to_flush is not None and len(self._trs_to_flush) > 0:
+            # Running for too long?
+            if datetime.utcnow() - self._flush_time >= timedelta(seconds=10):
```
olivielpeau: could you use a var named `_MAX_FLUSH_DURATION`or something similar?

PR #3028, opened 2016-11-15
Updating Kubernetes check to handle multiple namespaces fix #2838

### LINKED ISSUE #2838: [kubernetes] Events from all namespaces
from https://github.com/DataDog/dd-agent/pull/2551

Currently the Kubernetes events feature only gets events from the `default` namespace.  This means that datadog is unable to capture critical events from the `kube-system` namespace as well as any other user-defined namespaces.

I believe the dd-agent kubernetes configuration should support two options:
1. Capture events from all namespaces (defaults to `False`)
2. Capture events from list of namespaces (defaults to `default` only)

This will not change the functionality of new 5.9 deployments but will enable improved kubernetes support down the road for all users.

relevant code: https://github.com/DataDog/dd-agent/blob/master/checks.d/kubernetes.py#L362

hush-hush:
### What does this PR do?

This PR allows the agent to collect events from multiple Kubernetes namespaces by using a list of names and/or a regexp.

Two new configuration keys are added:
'namespaces': a list of names
'namespace_name_regexp': a regexp to match any namespace

### Additional Notes

The old configuration value 'namespace' is still handled by the agent: if present it is added to the 'namespaces' list.

### DIFF
--- a/conf.d/kubernetes.yaml.example
+++ b/conf.d/kubernetes.yaml.example
@@ -22,10 +22,18 @@ instances:
   # collect_events: false
   #
   #
-  # The namespace for which events should be collected.
-  # If not modified, the default namespace will be used.
+  # The namespaces for which events should be collected.
+  # If not modified, the 'default' namespace will be used.
   #
-  # namespace: default
+  # namespaces:
+  #  - default
+
+  # The regexp used to select namespaces for which events should be collected.
+  # The matched namespaces will be added to the "namespaces" list.
+  # If empty, regexp selection will be ignored.
+  #
+  # namespace_name_regexp:
+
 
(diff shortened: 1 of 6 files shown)


conf.d/kubernetes.yaml.example
```diff
@@ -22,10 +22,18 @@ instances:
...
+  # The namespaces for which events should be collected.
+  # If not modified, the 'default' namespace will be used.
   #
-  # namespace: default
+  # namespaces:
+  #  - default
+
+  # The regexp used to select namespaces for which events should be collected.
+  # The matched namespaces will be added to the "namespaces" list.
+  # If empty, regexp selection will be ignore.
```
hkaj: `s/ignore/ignored`

Issue #3034, opened 2016-11-16
[postgres] Allow the collection of postgresql.database_size metrics to be disabled


jstotz:
For a large database with a large number of relations the `pg_database_size` function can be very slow and expensive. Under heavy load it's possible for it to take longer than the check interval to execute causing queries to stack up and negatively affect system performance.

There should be a configuration flag to disable the collection of these metrics.

PR #3035, opened 2016-11-16
[postgres] Allow disabling postgresql.database_size metrics

jstotz:
### What does this PR do?

Makes the collection of PostgreSQL database size metrics optional, defaulting to enabled.

### Motivation

For large databases with many relations the calculation of this metric can be very slow and expensive. See #3034.

### Additional Notes

I'm not a Python programmer so feedback is very welcome!


jstotz:
@masci Done!

PR #3036, opened 2016-11-18
[install_script] Make `dd-agent` group of `datadog.conf`

olivielpeau:
Now that all our linux packages create a `dd-agent` group


masci:
👍

PR #3038, opened 2016-11-21
[service_discovery] Add a Zookeeper service discovery implementation.

manolama:
if you have not yet done so.*

### What does this PR do?

Adds support for using Apache Zookeeper as the backend for Service Discovery features when running DataDog agents in a cluster.

### Motivation

Zookeeper is widely deployed and used as the default for a DC/OS Mesos cluster. Instead of using resources for a Console or Etcd cluster just for DataDog configs, it makes sense to use the existing Zookeeper instance. And many other systems still use Zookeeper as a key value store.

### Testing Guidelines

NA

### Additional Notes

Works similar to the other stores, just has to iterate each time, unfortunately, to fetch all of the children. One way to bypass that would be to just change the "datadog/config_checks" mtime on each update but users would have to know to do that.

### DIFF
--- a/utils/service_discovery/zookeeper_config_store.py
+++ b/utils/service_discovery/zookeeper_config_store.py
@@ -1,0 +1,99 @@
+# (C) Datadog, Inc. 2016
+# All rights reserved
+# Licensed under Simplified BSD License (see LICENSE)
+
+# std
+import logging
+
+from kazoo.client import KazooClient, NoNodeError
+from utils.service_discovery.abstract_config_store import AbstractConfigStore, KeyNotFound
+
+DEFAULT_ZK_HOST = '127.0.0.1'
+DEFAULT_ZK_PORT = 2181
+DEFAULT_TIMEOUT = 5
+log = logging.getLogger(__name__)
+
+class ZookeeperStore(AbstractConfigStore):
+    """Implementation of a config store client for Zookeeper"""
+
+    def _extract_settings(self, config):
+        """Extract settings from a config object"""
+        settings = {
+            'host': config.get('sd_backend_host', DEFAULT_ZK_HOST),
+            'port': int(config.get('sd_backend_port', DEFAULT_ZK_PORT)),
+        }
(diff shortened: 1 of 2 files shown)


hkaj:
Hi @manolama 
Thanks a lot for contributing! That's a feature we've been wanting for a while but couldn't find time to write, so cheers :)
I'll review it shortly.

manolama:
Happy to help @hkaj. It looks like it works the same with the Etcd plugin performs but there may be cases I missed. And I need to learn how to UT in Python ;)

utils/service_discovery/zookeeper_config_store.py
```diff
@@ -0,0 +1,99 @@
...
+            raise KeyNotFound("The key %s was not found in Zookeeper" % path)
+
+    def recursive_list(self, path, results):
+        """Recursively walks the children from the given path and build a list of key/value tuples"""
+        try:
+            data, stat = self.client.get(path)
+
+            if data:
+                node_as_string = data.decode("utf-8")
+                if node_as_string[0] == '[':
```
hkaj: @manolama would an empty string work for you here instead of `'['`? This would allow us to do `if not node_as_string` (cleaner) and would avoid swallowing formatting errors (people forgetting that they need to use a list instead of a single dict, etc.). If that works for you I can add this change and merge your PR.

manolama: Hi @hkaj, yeah that'd work. The ZK CLI won't let you create an empty node but the API does. I could write a little CLI to include with DD that would make it easier to create and configure the nodes. In the mean time, please make that tweak and merge this if you don't mind. Thanks!

manolama:
Hi @hkaj, I tweaked the code as per your last comment. Lemme know if it's ok. thanks!

hkaj:
This looks great. CI passes and i tested it manually. I can't wait to see it in 5.11!
Thanks a lot @manolama!

PR #3040, opened 2016-11-21
[flare] ignore whitespace before api_key

sjenriquez:
### What does this PR do?

Fixes the case when there was whitespace before the `api_key` in datadog.conf causing a pattern match failure.

### Motivation

Support case.

### DIFF
--- a/utils/flare.py
+++ b/utils/flare.py
@@ -74,7 +74,7 @@ class Flare(object):
     ]
     MAIN_CREDENTIALS = [
         CredentialPattern(
-            re.compile('\s*api_key:( *\w+(\w{5}) ?,?)+$'),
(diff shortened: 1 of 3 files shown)


utils/flare.py
```diff
@@ -74,7 +74,7 @@ class Flare(object):
     ]
     MAIN_CREDENTIALS = [
         CredentialPattern(
-            re.compile('^api_key:( *\w+(\w{5}) ?,?)+$'),
+            re.compile('\s*api_key:( *\w+(\w{5}) ?,?)+$'),
```
degemer: Could you keep the `^` ? This regex could match other things (commented api_key for instance). I can't think of a case where it really matters, but I'd say let's avoid that altogether and restrict the possible cases.

sjenriquez:
@degemer Added a test and put the `^` back in per your feedback.

PR #3042, opened 2016-11-22
[Win32] Make basic system checks use psutil & win32pdh rather than WMI

derekwbrown:
### What does this PR do?

Makes the basic system checks use other means besides WMI.  The WMI checks are having anecdotal reports of slow checks and other problems.  This change removes use of the WMI check (though leaves the WMI checks available, since standalone WMI checks are supported).

### Motivation

Support tickets.

### Additional Notes

While there are reports of the WMI check failing in unpredictable ways, it's not been reported in-house.  Therefore, there's no conclusive proof this checkin is an improvement.


checks/libs/win/winpdh.py
```diff
@@ -0,0 +1,49 @@
+import time
+import sys
+import win32pdh
+import psutil
+
+class WinPDHCounter(object):
+    
+    def __init__(self, class_name, instance_name):
+        self._class_name = class_name
+        self._instance_name = instance_name
```
gmmeyer: Could we make this class stateless, make the functions into class methods and just mass these variables into those methods? If I'm not missing anything, it seems like you're only calling it once with each of these.

derekwbrown: I did it this way to match as closely as possible the previous calling convention.  I'd be happy to make that change if it's helpful.

checks/libs/win/winpdh.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def get_single_value(self):
+        hq = win32pdh.OpenQuery()
+        path = win32pdh.MakeCounterPath((None, self._class_name, None, None, 0, self._instance_name))
+        counter_handle = win32pdh.AddCounter(hq, path)
+        win32pdh.CollectQueryData(hq)
+        t, val = win32pdh.GetFormattedCounterValue(counter_handle, win32pdh.PDH_FMT_LARGE)
```
yannmh: `PDH_FMT_LARGE` vs `PDH_FMT_LONG ` (in `get_all_values`). Also https://www.cac.cornell.edu/wiki/index.php?title=Performance_Data_Helper_in_Python_with_win32pdh suggests to use `PDH_FMT_FLOAT`

checks/libs/win/winpdh.py
```diff
@@ -0,0 +1,49 @@
...
+    
+    def __init__(self, class_name, instance_name):
+        self._class_name = class_name
+        self._instance_name = instance_name
+
+    def get_single_value(self):
+        hq = win32pdh.OpenQuery()
+        path = win32pdh.MakeCounterPath((None, self._class_name, None, None, 0, self._instance_name))
+        counter_handle = win32pdh.AddCounter(hq, path)
+        win32pdh.CollectQueryData(hq)
```
yannmh: From what i understand `hq` is stateful

> Some performance counters are instantaneous. Others require you collect data for a few seconds. See references for details. A safe bet is to collect data twice before getting values.
 
(https://www.cac.cornell.edu/wiki/index.php?title=Performance_Data_Helper_in_Python_with_win32pdh)


Shall it be kept in the instance to persist over time?

checks/libs/win/winpdh.py
```diff
@@ -0,0 +1,49 @@
...
+    
+            win32pdh.CollectQueryData(hq)
+            try:
+                t, val = win32pdh.GetFormattedCounterValue(counter_handle, win32pdh.PDH_FMT_LONG)
+                ret[inst] = val
+            except Exception as e:
+                # exception usually means self type needs two data points to calculate. Wait
+                # a bit and try again
+                time.sleep(0.01)
+                win32pdh.CollectQueryData(hq)
```
yannmh: Same as above about persisting query objects. What do you think?

checks/libs/win/winpdh.py
```diff
@@ -0,0 +1,49 @@
...
+            counter_handle = win32pdh.AddCounter(hq, path)
+    
+            win32pdh.CollectQueryData(hq)
+            try:
+                t, val = win32pdh.GetFormattedCounterValue(counter_handle, win32pdh.PDH_FMT_LONG)
+                ret[inst] = val
+            except Exception as e:
+                # exception usually means self type needs two data points to calculate. Wait
+                # a bit and try again
+                time.sleep(0.01)
```
yannmh: Let's put this time in a CONSTANT at the class or module level.

checks/libs/win/winpdh.py
```diff
@@ -0,0 +1,49 @@
...
+                win32pdh.CollectQueryData(hq)
+                # if we get exception self time, just return it up
+                try:
+                    t, val = win32pdh.GetFormattedCounterValue(counter_handle, win32pdh.PDH_FMT_LONG)
+                    ret[inst] = val
+                except Exception as e:
+                    win32pdh.CloseQuery(hq)
+                    raise e
+            win32pdh.CloseQuery(hq)
```
yannmh: Python has a `finally` clause for that 🍬 

```python
try:
    try:
    except Exception:
        do_something()
except Exception:
    do_something()
finally:
    win32pdh.CloseQuery(hq)  # Where we close the query

```

checks/libs/win/winpdh.py
derekwbrown: see below.

checks/libs/win/winpdh.py
derekwbrown: So, this was taken directly from a sample on how to use winpdh.  I don't know if it's better to just to the brief sleep, or do the state as you mention above.

derekwbrown:
Older PR, got tripped up with git merging.  Will create new PR.

PR #3051, opened 2016-11-27
allow override of kubelet host with KUBERNETES_KUBELET_HOST env var

stonith:
if you have not yet done so.*

### What does this PR do?

dd-agent assumes that the kubelet API is available at the default gateway of the container. This PR allows overriding of the kubelet host if the KUBERNETES_KUBELET_HOST environment variable is found. As an example, the kubernetes downward API in 1.4+ can be used to specify the node name:

```
          - name: KUBERNETES_KUBELET_HOST
            valueFrom:
              fieldRef:
                fieldPath: spec.nodeName
```

### Motivation

There are implementations of kubernetes where the kubelet API is not available at the default gateway. An example of this would be when Calico is used: projectcalico/calico-containers#1099

We're testing Calico with kubernetes and a docker-dd-agent daemonset does not function properly. 

### Additional Notes

I can cut a PR to update docs in DataDog/docker-dd-agent if this gets merged.

- this addresses kubernetes environments where the kubelet api is
  not reachable at the default gateway
- addresses projectcalico/calico-containers#1099

### DIFF
--- a/conf.d/kubernetes.yaml.example
+++ b/conf.d/kubernetes.yaml.example
@@ -3,7 +3,8 @@ init_config:
 instances:
   # The kubernetes check retrieves metrics from cadvisor running under kubelet.
   # By default we will assume we're running under docker and will use the address
-  # of the default router to reach the cadvisor api.
+  # of the default router to reach the cadvisor api unless the environment variable
+  # KUBERNETES_KUBELET_HOST is found.
   #
   # To override, e.g. in the case of a standalone cadvisor instance, use the following:
   #
--- a/utils/kubernetes/kubeutil.py
+++ b/utils/kubernetes/kubeutil.py
@@ -57,14 +57,15 @@ class KubeUtil:
 
         self.method = instance.get('method', KubeUtil.DEFAULT_METHOD)
         self.host = instance.get("host") or self.docker_util.get_hostname()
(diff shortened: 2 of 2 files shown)


caseydavenport:
@stonith Just to clarify my understanding:

```
fieldPath: spec.nodeName
```

means that the node's name in Kubernetes needs to be resolvable by the pod (or a valid IP address for that node), correct?

stonith:
@caseydavenport in this example yes, the pod would have to be able to resolve the node's name.

caseydavenport:
@stonith cool, thanks! This looks good to me.

remh:
Thanks! This looks great to me. @hkaj can you review as well ?

hkaj:
Thanks @stonith. It looks good to me too, I'll probably merge it today.
As you mentioned we'll also need a PR in `docker-dd-agent` to add the option in here: https://github.com/DataDog/docker-dd-agent/blob/11.0.5101/entrypoint.sh#L95-L104

Mind taking care of it too @stonith ? 
Thanks again!

hkaj:
Just tested, all good.

stonith:
PR for the [README](DataDog/docker-dd-agent/pull/144). 

NOTE:
When trying to implement this with CNI networking I ran into kubernetes/kubernetes#31307 which unfortunately means hostPort daemonset with my above example doesn't currently work. A NodePort service should work though.

PR #3053, opened 2016-12-01
[sqlserver] remove warning message leaking password strings

masci:
### What does this PR do?

Remove a `log.warning` call outputting a string containing connection password in case of errors. The log operation was redundant anyways since the exception is correctly and safely propagated.

### DIFF
--- a/checks.d/sqlserver.py
+++ b/checks.d/sqlserver.py
@@ -329,7 +329,6 @@ class SQLServer(AgentCheck):
         except Exception as e:
             cx = "%s - %s" % (host, database)
             message = "Unable to connect to SQL Server for instance %s." % cx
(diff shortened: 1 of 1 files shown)


checks.d/sqlserver.py
```diff
@@ -329,7 +329,6 @@ def open_db_connections(self, instance):
         except Exception as e:
             cx = "%s - %s" % (host, database)
             message = "Unable to connect to SQL Server for instance %s." % cx
-            self.log.warning("%s Exception: %s", message, e)
```
truthbk: Was it the exception that contained the connection string, and leaked the password? If not we could leave out the exception body, and just print out the `cx`? I think this integration is multi-instance....

truthbk: disregard.

PR #3055, opened 2016-12-02
[vsphere] check revamp addressing several issues

truthbk:
### What does this PR do?

This PR aims to address several issues with the check. Mostly focusing on the following points:

- Identifying all relevant infrastructure elements - VMs, hosts, datastores, datacenters. etc. We had issues when due to the infrastructure hierarchy we weren't properly traversing the tree. By resorting to container views and allowing pyvmomi to do the traversal for us we should be guaranteed to view all infrastructure we're privileged enough to see.
- Improving vCenter/vSphere 6+ compatibility.
- Improving tagging.
- Fixing a bug where we were submitting vsphere "rates" as rates instead of gauges, effectively computing the rate of rates. Vsphere rates are precomputed and should be submitted as gauges. This has been broken for a long time and might affect some customers, but we weren't doing the right thing.
- Performance should have improved, however collecting all metrics is still CPU intensive.

### Motivation

A general need to improve our vsphere check to better address customer requirements.


checks.d/vsphere.py
```diff
@@ -860,16 +860,13 @@ def _collect_metrics_atomic(self, instance, mor):
...
-
-                record_metric(
+                # vsphere "rates" should be submitted as gauges (rate is
+                # precomputed).
+                self.gauge(
                     "vsphere.%s" % metric_name,
                     value,
                     hostname=mor['hostname'],
-                    tags=['instance:%s' % instance_name]
+                    tags=['instance:%s' % instance_name] + mor['tags']
```
yannmh: I don't think this is necessary. `more['tags']` should be host tags, not metric tags.

checks.d/vsphere.py
```diff
@@ -700,12 +707,20 @@ def _cache_morlist_process_atomic(self, instance, mor):
...
+        else:
+            interval = None
+
         available_metrics = perfManager.QueryAvailablePerfMetric(
-            mor['mor'], intervalId=REAL_TIME_INTERVAL)
+            mor['mor'], intervalId=interval)
 
-        mor['metrics'] = self._compute_needed_metrics(instance, available_metrics)
-        mor_name = str(mor['mor'])
+        if available_metrics:
```
truthbk: You're right, we probably don't need it, going to try with passing around the empty list.

checks.d/vsphere.py
```diff
@@ -692,17 +661,15 @@ def _cache_morlist_raw(self, instance):
...
-                "Skipping morlist collection now, RAW results "
-                "processing not over (latest refresh was {0}s ago)".format(
-                    time.time() - self.cache_times[i_key][MORLIST][LAST])
-            )
-            return
-        self.morlist_raw[i_key] = []
-
-        server_instance = self._get_server_instance(instance)
-        root_folder = server_instance.content.rootFolder
+        for resource_type in RESOURCE_TYPE_MAP:
```
truthbk: Well, I did that because now we pop() the elements per resource type in `_cache_morlist_process()`, before we knew when the list was empty, we were done. Now we may have processed all ComputeResources, or whatever, but not the VM's, so just checking for the length of the dictionary wouldn't be enough. Besides, we pop from the lists, not the keys.  `self.morlist_raw[i_key][resource_type]` dicts are created the first time around, but they then exist through out, I believe, so the `len(self.morlist_raw[i_key]) == len(RESOURCE_TYPE_MAP)` in subsequent iterations.

PR #3060, opened 2016-12-05
[Service Discovery] create a template cache to reduce calls to the KV store

hkaj:
if you have not yet done so.*

### What does this PR do?

**Create a cache for configuration templates**
- it keeps auto_configuration templates and templates coming from the KV store
- it invalides the KV templates when the agent detects an update in the config store
- to reload auto_conf or JMX templates we still need to restart the collector or send it a `SIGHUP`.

**Minor clean ups and test updates**

### Motivation

Using this cache avoids querying the config store at every container update.

### Testing Guidelines

Tests were updated accordingly. A custom docker image using this build is also available at the `TODO` tag.

### DIFF
--- a/tests/core/test_service_discovery.py
+++ b/tests/core/test_service_discovery.py
@@ -2,6 +2,7 @@
 import copy
 import mock
 import unittest
+from collections import defaultdict
 
 # 3p
 from nose.plugins.attrib import attr
@@ -11,7 +12,8 @@ from config import generate_jmx_configs
 from utils.service_discovery.config_stores import get_config_store
 from utils.service_discovery.consul_config_store import ConsulStore
 from utils.service_discovery.etcd_config_store import EtcdStore
-from utils.service_discovery.abstract_config_store import AbstractConfigStore, CONFIG_FROM_KUBE
+from utils.service_discovery.abstract_config_store import AbstractConfigStore, \
+    _TemplateCache, CONFIG_FROM_KUBE, CONFIG_FROM_TEMPLATE, CONFIG_FROM_AUTOCONF
 from utils.service_discovery.sd_backend import get_sd_backend
 from utils.service_discovery.sd_docker_backend import SDDockerBackend, _SDDockerBackendConfigFetchState
 
@@ -66,11 +68,11 @@ def client_read(path, **kwargs):
     if 'all' in kwargs:
         return {}
     else:
-        return TestServiceDiscovery.mock_tpls.get(image)[0][config_parts.index(config_part)]
+        return TestServiceDiscovery.mock_raw_templates.get(image)[0][config_parts.index(config_part)]
 
 
 def issue_read(identifier):
-    return TestServiceDiscovery.mock_tpls.get(identifier)
+    return TestServiceDiscovery.mock_raw_templates.get(identifier)
 
 @attr('unix')
 class TestServiceDiscovery(unittest.TestCase):
@@ -122,7 +124,7 @@ class TestServiceDiscovery(unittest.TestCase):
     }
 
     # raw templates coming straight from the config store
-    mock_tpls = {
+    mock_raw_templates = {
         # image_name: ('[check_name]', '[init_tpl]', '[instance_tpl]', expected_python_tpl_list)
         'image_0': (
             ('["check_0"]', '[{}]', '[{"host": "%%host%%"}]'),
@@ -509,12 +511,12 @@ class TestServiceDiscovery(unittest.TestCase):
     def test_get_auto_config(self):
         """Test _get_auto_config"""
         expected_tpl = {
-            'redis': ('redisdb', None, {"host": "%%host%%", "port": "%%port%%"}),
-            'consul': ('consul', None, {
-                "url": "http://%%host%%:%%port%%", "catalog_checks": True, "new_leader_checks": True
-            }),
-            'redis:v1': ('redisdb', None, {"host": "%%host%%", "port": "%%port%%"}),
-            'foobar': None
+            'redis': [('redisdb', None, {"host": "%%host%%", "port": "%%port%%"})],
+            'consul': [('consul', None, {
+                        "url": "http://%%host%%:%%port%%", "catalog_checks": True, "new_leader_checks": True
+                        })],
+            'redis:v1': [('redisdb', None, {"host": "%%host%%", "port": "%%port%%"})],
+            'foobar': []
         }
 
         config_store = get_config_store(self.auto_conf_agentConfig)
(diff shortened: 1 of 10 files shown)


utils/service_discovery/abstract_config_store.py
```diff
@@ -34,6 +34,125 @@ class KeyNotFound(Exception):
...
+        # miss
+        else:
+            tpls = self._issue_read(identifier)
+            if tpls:
+                templates[CONFIG_FROM_TEMPLATE] = tpls
+                self.kv_templates[identifier] = tpls
+            # if _issue_read doesn't return anything we still create a key in the cache
+            # so that subsequent reads don't trigger issue_read every time
+            else:
+                self.kv_templates[identifier] = None
```
truthbk: I know this is a NOOP but I noticed that we do include in the results `templates[CONFIG_FROM_TEMPLATE] = self.kv_templates[identifier]` even when `self.kv_templates[identifier] is None` if we have a hit, but we don't do that with a miss. 

Just saying because it occured to me we could refactor this as:

```
if identifier not in self.kv_templates:
    tpls = self._issue_read(identifier)
    if tpls:
        templates[CONFIG_FROM_TEMPLATE] = tpls
        self.kv_templates[identifier] = tpls
    else:
        self.kv_templates[identifier] = None

templates[CONFIG_FROM_TEMPLATE] = self.kv_templates[identifier]
```

because we always populate after a hit or fetch.

utils/service_discovery/abstract_config_store.py
```diff
@@ -34,6 +34,125 @@ class KeyNotFound(Exception):
...
+        """Retrieve auto_conf templates"""
+        raw_templates = get_auto_conf_images(full_tpl=True)
+        for image, tpls in raw_templates.iteritems():
+            for check_name, init_tpl, instance_tpl in zip(*tpls):
+                if image in self.auto_conf_templates:
+                    if check_name in self.auto_conf_templates[image][0]:
+                        log.warning("Conflicting templates in auto_conf for image %s and check %s. "
+                                "Please check your template files." % (image, check_name))
+                        continue
+                    self.auto_conf_templates[image][0].append(check_name)
```
truthbk: Oh I get it, we stick to lists for "compatibility" across the board when dealing with the configs. I think we could maybe refactor in the future, but it's a legit choice.

utils/service_discovery/abstract_config_store.py
```diff
@@ -152,11 +244,12 @@ def get_checks_to_refresh(self, identifier, **kwargs):
         if to_check:
             return to_check
         else:
-            # lastly fallback to auto_conf
-            return set(self.identifier_to_checks[self._get_image_ident(identifier)])
+            # lastly fallback to legacy name for auto-conf
+            return self.template_cache.get_check_names(self._get_image_ident(identifier))
```
truthbk: Question: can we have checks under the legacy name that should also be applied? So a case where `identifier` and the legacy `self._get_image_ident(identifier)` would both apply to an image?

tests/core/test_service_discovery.py
```diff
@@ -570,8 +572,8 @@ def test_get_config_id(self):
                     expected_ident)
                 clear_singletons(self.auto_conf_agentConfig)
 
-    @mock.patch.object(AbstractConfigStore, '_issue_read', side_effect=issue_read)
-    def test_read_config_from_store(self, issue_read):
+    @mock.patch.object(_TemplateCache, '_issue_read', side_effect=issue_read)
+    def test_read_config_from_store(self, args):
```
truthbk: Big 🙇 for all these awesome tests!

utils/checkfiles.py
```diff
@@ -100,5 +100,18 @@ def get_auto_conf_images(agentConfig):
         # extract the supported image list
         images = auto_conf.get('docker_images', [])
         for image in images:
-            auto_conf_images[image] = check_name
+            if full_tpl:
+                init_tpl = auto_conf.get('init_config') or {}
+                instance_tpl = auto_conf.get('instances', [])
+                if image not in auto_conf_images:
+                    auto_conf_images[image] = [[check_name], [init_tpl], [instance_tpl]]
```
hkaj: you mean like `auto_conf_images[(image, check_name)] = [init_tpl, instance_tpl]`?

Problem with that is if users have several instances of the same check for the same image (and a customer actually asked how to do that yesterday so it happens). In this case we still need to wrap init_tpl and instance_tpl in lists and end up with the same logic as we have now.

But maybe I didn't understand correctly your suggestion

utils/service_discovery/abstract_config_store.py
hkaj: man that snippet is on point! 👌

utils/service_discovery/abstract_config_store.py
hkaj: Good catch! That's a use case we didn't support before (if user-defined templates were found we skipped auto config) but I added logic for that [here](https://github.com/DataDog/dd-agent/blob/haissam/sd-cache-kv-tpls/utils/service_discovery/abstract_config_store.py#L122-L133) and forgot to update this part. Will do it now.

utils/checkfiles.py
truthbk: I think I originally meant `auto_conf_images[image] = {'checkname': {'init': init_tpl, 'instances: [intance_tpl_0, instance_tpl_1,...]}}`

But then I read on, and I saw it wasn't necessarily a good idea given the current structure/logic. I felt it would make the code a little more maintainable (perhaps). We can maybe just keep in mind for the future (it'd be a pretty big refactor), there's nothing wrong with the current logic.

utils/service_discovery/abstract_config_store.py
```diff
@@ -34,6 +35,141 @@ class KeyNotFound(Exception):
...
+                tpls = self._issue_read(identifier)
+            except NotImplementedError:
+                # expected when get_check_names is called in auto-conf mode
+                tpls = None
+            except Exception:
+                tpls = None
+                log.exception('Failed to retrieve a template for %s.' % identifier)
+            # create a key in the cache even if _issue_read doesn't return a tpl
+            # so that subsequent reads don't trigger issue_read
+            self.kv_templates[identifier] = tpls or None
```
truthbk: I'm not sure you need the `or None`? `tpls` will already be something if `self._issue_read(identifier)` returns either a list or `None`, or the `None` we've set it in the `except` blocks.

hkaj:
Thanks @truthbk ! I addressed your comments and squashed the commits. Will merge once the tests pass 🎉

PR #3066, opened 2016-12-08
[powerdns_recursor] add PowerDNS recursor version 4 support

jimmystewpot:
### What does this PR do?

This PR adds support for powerdns recursive server version 4 while still retaining support for version 3 configuration files. The idea behind that is it allows for users to update their agent and keep their configuration files in place without a breakage. i.e. without the new instance:version syntax.

The PR also adds an increased number of metrics which had previously been submitted in PR-2919. The additional metrics add more robust production monitoring.

### Motivation

We run a large number of production high throughput DNS servers. The additional metrics allow us to get better visibility of variations in their performance. We also have plans to migrate to version 4 of the recursive server which is now supported in this PR. Allowing us to run multiple versions in production concurrently.

Version 4 provides more in-depth DNSSEC and IPv4 vs IPv6 data via the statistics interface which is a major win for our organisation.

### Testing Guidelines

As per PR-2919 @olivielpeau  the tests have been updated to reflect both the V3 and V4 Gauge and Rate metrics. 

On my local test setup I have both Version 3 and Version 4 recursive name servers running which is why the Config values are configured as _v3 and _v4 and the tests are broken into two separate runs.

That means for these tests to pass successfully on your travis I assume that you will need to configure powerdns 3 and 4 to be configured as well.

### Additional Notes

The experimental web server is no longer experimental in version 4 of the recursive server.  As such the configuration file has been amended to reflect this change in language.


jimmystewpot:
The tests are failing on this PR because the test_check_v4 is expecting a Version 4 powerdns recursor to be available on port 8083 as defined on L156 of teh test_powerdns_recursor.py file. Once that is available the tests should complete successfully.

jimmystewpot:
Any news on this?

truthbk:
@jimmystewpot sorry for the slow time to respond! I'll review this in the next day or so. I will also try to set you up with a flavor for version 4 to run the CI on, that should probably do it. Thanks a ton for your PR 🙇

checks.d/powerdns_recursor.py
```diff
@@ -23,24 +25,141 @@ class PowerDNSRecursorCheck(AgentCheck):
...
+    GAUGE_METRICS_V4 = [
+        'cache-entries',
+        'concurrent-queries',
+        'failed-host-entries',
+        'negcache-entries',
+        'packetcache-entries',
+        'throttle-entries',
+        'fd-usage',
+    ]
+    RATE_METRICS_V4 = [
```
truthbk: It seems like we have many metrics which are common to V3 and V4... should we maybe create a base set of metrics common to both versions, and then add to the catalog depending on the version? The less duplicates we have the easier it'll be to maintain and less likely we might end up leaving something out. This consolidation is always something we could do in the future, but ideally that would be awesome.

truthbk:
@jimmystewpot I've added some feedback above. Overall this looks great, just a few things that might help us with testing and maintenance.

conf.d/powerdns_recursor.yaml.example
```diff
@@ -17,6 +23,8 @@ instances:
     port: 8082
   # Recursor web server api key.
     api_key: pdns_api_key
+  # Version 3 or 4
+    version: 4
```
jimmystewpot: Actually it can be completely absent and it will default to version 3. i.e. existing configurations would remain working after an upgrade of the agent.

I will update the configuration regardless to ensure that it is consistent with the default.

tests/checks/integration/test_powerdns_recursor.py
```diff
@@ -22,60 +24,196 @@ class TestPowerDNSRecursorCheck(AgentCheckTest):
...
+        for metric in self.RATE_METRICS_V3:
             self.assertMetric(self.METRIC_FORMAT.format(metric), tags=[])
 
         service_check_tags = ['recursor_host:127.0.0.1', 'recursor_port:8082']
         self.assertServiceCheckOK('powerdns.recursor.can_connect', tags=service_check_tags)
 
         self.coverage_report()
 
+
+    def test_check_v4(self):
```
jimmystewpot: I am unsure what you are referring to here. 

When I run version 4 and version 3 locally with the configuration defined in self.config_v3 and self.config_v4 (both versions running on different ports) the tests pass locally. Would you be able to elaborate further on what your expectations are?

```
# nosetests tests/checks/integration/test_powerdns_recursor.py
......
----------------------------------------------------------------------
Ran 6 tests in 3.164s

OK

# curl -H 'X-API-Key: pdns_api_key' -s http://127.0.0.1:8083/api/v1/servers/localhost | jq
{
  "config_url": "/api/v1/servers/localhost/config{/config_setting}",
  "daemon_type": "recursor",
  "id": "localhost",
  "type": "Server",
  "url": "/api/v1/servers/localhost",
  "version": "0.0.g9413017",
  "zones_url": "/api/v1/servers/localhost/zones{/zone}"
}

# curl -H 'X-API-Key: pdns_api_key' -s http://127.0.0.1:8082/servers/localhost | jq
{
  "type": "Server",
  "id": "localhost",
  "url": "/servers/localhost",
  "daemon_type": "recursor",
  "version": "3.7.3",
  "config_url": "/servers/localhost/config{/config_setting}",
  "zones_url": "/servers/localhost/zones{/zone}"
}
```

I am keen to ensure that we are on the same page so there is no confusion.

checks.d/powerdns_recursor.py
jimmystewpot: Adjusted as per your recommendation.

jimmystewpot:
The most recent commit has the following changes

- Update the configuration yaml example to be version 3 by default. As per your recommendation.
- Update the metrics lists to be baseline or version 3, and then have the v4 extended metrics. As per your recommendation.
- rolled up the metrics collection into a single function where the URL is the only change based on config.version == 4. To simplify all the things.

I have currently refrained from updating the check to automatically check the version of the software in use primarily because I have outstanding questions above in regards to the flavors that you talk about.

jimmystewpot:
Any update on this?

truthbk:
@jimmystewpot this totally flew under my radar during the holidays. Apologies for that. Happy New Year, by the way, I hope you had a great one.

Thanks a lot for your changes, this should now be backward compatible and easier to maintain after your latest update.

So the issue with the test as it runs in the CI environments (Travis mainly) is that it brings up a particular version of the PowerDNS recursor. You can see that here: 

- https://github.com/DataDog/dd-agent/blob/master/ci/powerdns_recursor.rb#L8
- https://github.com/DataDog/dd-agent/blob/master/.travis.yml#L74

As you can see in the travis YAML, we are currently not specifying a version, so it's defaulting to `3.7.3` and that seems like it will fail for the v4 test case. We should try to make an `4.x` version be brought up, and then have the right test be executed depending on the recursor version installed. Or just have the test (or the actual check) detect the version and run the appropriate code. I hope this makes sense, but let me know if you have any questions. 

I'd also be happy to help with setting up the CI if I find some spare cycles, if you need that. 

Thanks!

jimmystewpot:
@truthbk happy new year and thanks for the response, I will see if I can get the travis flavour stuff working. It now makes sense to me what you described in the previous comments. Having not used Travis previously i'll  probably take a few attempts.

truthbk:
@jimmystewpot I actually just realized we merged a powerDNS recursor check with the integration SDK here: https://github.com/DataDog/integrations-core/pull/28

This integration SDK will actually be the preferred way of receiving community contributions like yours in the near-future. So if you want to be one of our SDK pioneer contributors, you could actually open the PR against that repo. I am telling you this because we actually already have both flavors `3.x` and `4.x` enabled for the `PowerDNS` check. So it might be easier for you. 

It's really up to you, if you prefer to just move forward with this PR and figure out the flavor stuff, that's fine. If you want to give the SDK a shot that'd be awesome! One way or another we will have to port it to the SDK eventually anyhow. 

Thanks a ton! Sorry I didn't tell you about this sooner, I hadn't realized how it could've possibly saved you work.

jimmystewpot:
@truthbk Thanks for the tip, I will have a look at the SDK. It seems logical that I will update the SDK repo to report on the additional metrics. I will also cancel the ```brocade_vadc``` PR that I have and move that over to the new repo.

I had a crack at the travis configuration. I have no idea why it's failing at the moment as it runs end to end locally for both the 3.7.3 and 4.0.3 releases. However via the travis ci pipeline the 4.x release is failing. ```Exit code 127``` suggests "command is not found". Suggesting the task ```before_install``` is not completing correctly. However I do not see any of the steps from that task even attempting to be executed in the log. 

When these tests run locally the 4.x build takes ~5minutes to complete. As this is the first time that I have used rake or travis I would be really interested to know what the solution is.

truthbk:
@jimmystewpot just wanted to check back in, see if you made any progress with porting this PR to the integration SDK, and harness the relative ease with which we could enable tests with a v4 container there. 

We're going to be freezing the `dd-agent` as we segregate the checks from the agent core. So it's something we should be doing sooner rather than later. I'll be happy to help if you need assistance.

truthbk:
Superseding PR merged here https://github.com/DataDog/dd-agent/pull/3166, identical to this one with @jimmystewpot's work, plus a couple of minor tweaks to the CI infrastructure, configurations, etc.

PR #3069, opened 2016-12-09
[HTTP CHECK] Add an Option to mark check as DOWN if a content is found

oceanef:
### What does this PR do?

Adds a feature to the Http_check : allows users to look for a certain content in the URL and shows the check as DOWN if that content is found. 

User will need to set the `reverse_content_match` to `true` in order to do so. 

### Motivation

Several Customer Requests 

### Testing Guidelines

Added some test to the `tests/checks/integration/test_http_check.py` file to test the new feature independantly while checking there was no retro compatibility issue.

Ran the rake ci:run[system] and the behavior was expected

### Additional Notes

Option to Enable is reverse_content_match, defaults to False

### DIFF
--- a/checks.d/http_check.py
+++ b/checks.d/http_check.py
@@ -223,15 +223,15 @@ class HTTPCheck(NetworkCheck):
 
             instance_proxy = self.proxies.copy()
 
-            # disable proxy if necessary
-            # if skip_proxy:
-            #     instance_proxy.pop('http')
-            #     instance_proxy.pop('https')
-            # else:
-            #     for url in self.proxies['no'].replace(';', ',').split(","):
-            #         if url in parsed_uri.netloc:
-            #             instance_proxy.pop('http')
-            #             instance_proxy.pop('https')
+            disable proxy if necessary
+            if skip_proxy:
+                instance_proxy.pop('http')
+                instance_proxy.pop('https')
+            else:
+                for url in self.proxies['no'].replace(';', ',').split(","):
+                    if url in parsed_uri.netloc:
+                        instance_proxy.pop('http')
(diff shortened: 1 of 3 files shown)


checks.d/http_check.py
```diff
@@ -295,27 +313,21 @@ def _check(self, instance):
             if content_match:
                 # r.text is the response content decoded by `requests`, of type `unicode`
                 content = r.text if type(content_match) is unicode else r.content
-                if re.search(content_match, content, re.UNICODE):
-                    self.log.debug("%s is found in return content" % content_match)
-                    service_checks.append((
-                        self.SC_STATUS, Status.UP, "UP"
-                    ))
+                if not reverse_content_match and re.search(content_match, content, re.UNICODE):
```
masci: The code is correct but maybe something like this could be more readable:
```
if re.search(content_match, content, re.UNICODE):
    if reverse_content_match:
        # do smth
    else:
       # do smth else
else:
      if reverse_content_match:
          # do smth
      else:
          # do smth else
```
you could also move the `if reverse_content_match:` blocks within a separate function where you evaluate (and maybe send) the final status.

conf.d/http_check.yaml.example
```diff
@@ -29,6 +29,11 @@ instances:
     # content_match: 'In Stock'
     # content_match: '^(Bread|Apples|Very small rocks|Cider|Gravy|Cherries|Mud|Churches|Lead) float(s)? in water'
 
+    # The (optional) reverse_content_match parameter will allow the content_match
+    # to work the other way around. That means that the check will report
+    # as DOWN if the string is DOWN
```
masci: `# as DOWN if the string is found`

Issue #3045, opened 2016-11-23
More metrics from Kubernetes State API


therc:
5.10 introduced new metrics from the kubernetes-state-metrics API.
There are still some in the API that haven't made it to DataDog, e.g.:

kube_pod_container_status_ready{container="dd-agent",namespace="kube-system",pod="dd-agent-0fv0y"} 1
kube_pod_container_status_restarts{container="dd-agent",namespace="kube-system",pod="dd-agent-0fv0y"} 0
kube_pod_container_status_running{container="dd-agent",namespace="kube-system",pod="dd-agent-0fv0y"} 1
kube_pod_container_status_terminated{container="dd-agent",namespace="kube-system",pod="dd-agent-0fv0y"} 0
kube_pod_container_status_waiting{container="dd-agent",namespace="kube-system",pod="dd-agent-0fv0y"} 0

kube_pod_container_requested_cpu_cores{container="kubedns",namespace="kube-system",node="10.X.Y.Z",pod="kube-dns-v20-juul3"} 0.1
kube_pod_container_requested_memory_bytes{container="kubedns",namespace="kube-system",node="ip-10.X.Y.Z",pod="kube-dns-v20-juul3"} 1.048576e+08

At least the restarts and the requested resources (as opposed to limits) are very interesting to us.

I guess one question is how to map the pod and container names into DataDog: separate tags or combined into one?

schasse:
+1

ojongerius:
Is the plan to keep this as the meta issue for improvements?

ahawkins:
I think #3305 dupes this.

hkaj:
this should be fixed now, we're also working on supporting kube-state-metrics v0.5

PR #3070, opened 2016-12-09
[Kubernetes] Add state metrics for pods and containers

therc:
### What does this PR do?

It adds more kubernetes state metrics:
 * Add mem/cpu requests and limits
 * Add container restarts
 * Add pod phase: Pending|Running|Succeeded|Failed|Unknown

It fixes part of #3045

### Motivation

We'd like to know things such as which containers are restarting, which pods are in pending or failed states, container requests/limits, etc.

### Additional Notes

Tests might fail because the prometheus protobuf.bin blob lacks the
resource limits added in https://github.com/kubernetes/kube-state-metrics/pull/49

I can comment those out in the tests or in the code. Or maybe someone can add an updated protobuf dump.

### DIFF
--- a/utils/kubernetes/kube_state_processor.py
+++ b/utils/kubernetes/kube_state_processor.py
@@ -4,12 +4,21 @@
 
 NAMESPACE = 'kubernetes_state'
 
-
 class KubeStateProcessor:
     def __init__(self, kubernetes_check):
         self.kube_check = kubernetes_check
         self.log = self.kube_check.log
         self.gauge = self.kube_check.gauge
+        self.service_check = kubernetes_check.service_check
+        # Original camelcase keys have already been converted to lowercase.
+        self.pod_phase_to_status = {
+            'pending':   kubernetes_check.WARNING,
+            'running':   kubernetes_check.OK,
+            'succeeded': kubernetes_check.OK,
+            'failed':    kubernetes_check.CRITICAL,
+            # Rely on lookup default value
+            # 'unknown':   AgentCheck.UNKNOWN
+        }
 
     def process(self, message, **kwargs):
         """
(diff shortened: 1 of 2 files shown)


therc:
Any feedback, especially about the tests or the TODOs?
#3045 is tagged for milestone 5.11. I'm not sure what the deadline is for that, but I'd really like to get at least some form of this PR in.

utils/kubernetes/kube_state_processor.py
```diff
@@ -143,6 +143,72 @@ def kube_deployment_spec_replicas(self, message, **kwargs):
...
+            val = metric.gauge.value
+            tags = ['{}:{}'.format(label.name, label.value) for label in metric.label]
+            self.gauge(metric_name, val, tags)
+
+    # TODO: implement kube_pod_info
+    # TODO: implement kube_pod_status_ready
+    # TODO: implement kube_pod_status_scheduled
+
+    # Labels attached: namespace, pod, phase=Pending|Running|Succeeded|Failed|Unknown
+    def kube_pod_status_phase(self, message, **kwargs):
```
hkaj: The phase is probably more interesting as a service check than as a set of metrics split on a phase tag with a 0/1 value. Could you use a service check with the pod and namespace tags in the `kubernetes_state.py` check instead please?

therc: Do you mean like kube_node_status_ready/kube_nod_status_out_of_disk? They live in kube_state_processor.py already, not in kubernetes_state.py.

therc:
I squashed and rebased the changes. Sorry for the commit spam.
Only failure is from rabbitmq testing; I don't think it's related at all.

hkaj:
thanks @therc !

therc:
Thanks, @hkaj! Now the question is: when is 5.11 going to be released? :)

PR #3072, opened 2016-12-09
adds proper pagefile memory on windows

gmmeyer:
### What does this PR do?

This PR fixes the naming of pagefile memory on windows, it also fixes where we're pulling it from in psutil. Previously, we were reporting the physical memory, rather than the pagefile.

### DIFF
--- a/checks/system/win32.py
+++ b/checks/system/win32.py
@@ -114,10 +114,10 @@ class Memory(Check):
         self.gauge('system.mem.usable')
         self.gauge('system.mem.pct_usable')
         #  details about the usage of the pagefile.
-        self.gauge('system.mem.page_total')
-        self.gauge('system.mem.page_used')
-        self.gauge('system.mem.page_free')
-        self.gauge('system.mem.page_pct_free')
+        self.gauge('system.mem.pagefile_total')
+        self.gauge('system.mem.pagefile_used')
+        self.gauge('system.mem.pagefile_free')
+        self.gauge('system.mem.pagefile_pct_free')
 
     def check(self, agentConfig):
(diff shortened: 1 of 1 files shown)


checks/system/win32.py
```diff
@@ -114,10 +114,10 @@ def __init__(self, logger):
         self.gauge('system.mem.usable')
         self.gauge('system.mem.pct_usable')
         #  details about the usage of the pagefile.
-        self.gauge('system.mem.page_total')
-        self.gauge('system.mem.page_used')
-        self.gauge('system.mem.page_free')
-        self.gauge('system.mem.page_pct_free')
+        self.gauge('system.mem.pagefile_total')
```
degemer: Why not `system.pagefile.total` ? (same for the others)
I know this has already been discussed, but why not `system.swap` ? Or maybe just an alias (two metric names for the same metric).

gmmeyer: Sure, we can do that!

gmmeyer: The reason I chose this was because theres a metric: `system.mem.paged`

gmmeyer:
@degemer 51 orgs use this metric
@remh should we change the name? should we report both names back until we bump to a new minor (5.11 or 5.12)?

remh:
Given it's actually a bug fix, it's better to fix it entirely and to reach out to the users using this metric.

Also as we are creating a new metric, we might as well send it in bytes directly and add the proper metrics metadata for it.

gmmeyer:
@remh okay I remove the b2mb, I'll put in a related change to our metadata

PR #3073, opened 2016-12-09
[aggregator] if sample rate is bad, fix it but still parse tags.

truthbk:
### What does this PR do?

We seem to have a bug where we would not parse the tags if a bad rate was reported, we wouldn't even attempt to parse them. This fix will simply correct the bad rate like we were doing earlier without raising an assert that would stop us from grabbing the tags.

### Motivation

Behavior was identified by customer.

### DIFF
--- a/aggregator.py
+++ b/aggregator.py
@@ -492,13 +492,14 @@ class Aggregator(object):
                     # Parse the sample rate
                     if m[0] == '@':
                         sample_rate = float(m[1:])
-                        assert 0 <= sample_rate <= 1
+                        # in case it's in a bad state
+                        sample_rate = 1 if sample_rate < 0 or sample_rate > 1 else sample_rate
                     elif m[0] == '#':
                         tags = tuple(sorted(m[1:].split(',')))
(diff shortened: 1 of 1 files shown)


aggregator.py
```diff
@@ -492,13 +492,14 @@ def parse_metric_packet(self, packet):
...
                     elif m[0] == '#':
                         tags = tuple(sorted(m[1:].split(',')))
-            except (IndexError, AssertionError):
+            except IndexError:
                 log.warning(u'Incorrect metric metadata: metric_name:%s, metadata:%s',
                             name, u' '.join(value_and_metadata[2:]))
-                sample_rate = 1  # In case it's in a bad state
+
+            # in case it's in a bad state
+            sample_rate = 1 if sample_rate < 0 or sample_rate > 1 else sample_rate
```
degemer: What do you think of moving it where the `assert` was ? I think it's easier to read, since it's the only case it's used.

truthbk:
Failure is unrelated. Merging.

PR #3076, opened 2016-12-12
Exclude shutdown-script and google-container-manifest metadata.

fotinakis:
### What does this PR do?

Adds `shutdown-script`, `configure-sh`, and `google-container-manifest` to the list of excluded GCE metadata keys.

### Motivation

The `google-container-manifest` metadata likely contains sensitive info, and is a kubernetes config file so it doesn't make any sense to use as a datadog tag in any case. It's similar to what can be stored in the `user-data` key (which is already excluded) but is specifically used for configuring the older container-vm images on GCE. The `configure-sh` metadata is a big kubernetes bash script that should also be excluded.

Excluding `shutdown-script` is equivalent to the current exclusion for `startup-script`.

### DIFF
--- a/utils/cloud_metadata.py
+++ b/utils/cloud_metadata.py
@@ -19,8 +19,8 @@ class GCE(object):
     TIMEOUT = 0.3 # second
     SOURCE_TYPE_NAME = 'google cloud platform'
     metadata = None
-    EXCLUDED_ATTRIBUTES = ["kube-env", "startup-script", "sshKeys", "user-data",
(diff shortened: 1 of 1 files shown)


fotinakis:
Hey datadog friends — can anyone take a look at this?

remh:
Seems more than reasonable @fotinakis . Thanks!

fotinakis:
Hey @remh — do you know when this will make it into an official release? I just pulled the latest docker-dd-agent image (because of the recent APM updates) and it seems to be the old version of this file.

PR #3084, opened 2016-12-14
Update on tag_by_filesystem section of disk.yaml.default

ls339:
### What does this PR do?

Makes a small change in `disk.yaml.default`.

### Motivation

A client is using the disk integration's `tag_by_filesystem` option and expecting tags like `#filesystem:nfs` as the example states in disk.yaml.default. 

Tags are being generated without the `filesystem:` prefix.

### Testing Guidelines

### Additional Notes


hkaj:
Thanks @ls339

Issue #3085, opened 2016-12-15
Collector should only run one gohai at a time


directionless:
Hi there! I'm running datadog-agent version 5.9.1-1, and running into an odd issue. I don't think it's a strictly a datadog bug, but I think there are some things the collector should do to mitigate it.

Basically, what we see happening is a build up of gohai processes, each chewing up a cpu core, with kernel errors saying `rcu_sched self-detected stall on CPU`. We spent a little debugging this, and we don't see anything really wrong with gohai.

But we did notice that it was running at higher priority. This is because, in our environment, sshd runs at high priority, so daemons started from within an ssh session (and apt calling /etc/init.d/datadog-agent is one) inherit that priority. This higher priority causes some contention and eventually stalls.

One change I think datadog should make, is to only spawn one gohai process at a time. If it doesn't come back within the polling period, it seems counter productive to spawn another from the collector. This seems like a pretty simple patch, though I'm not sure how y'all would like it implemented. 

I don't know that there's a reasonable way for the agent to run gohai at lower priority. And some sites might want it running at a higher priority, so I think that's something for me to shake out in our infrastructure.

PR #3086, opened 2016-12-15
Add in support for /etc/default/datadog-agent and set nicelevel from it

directionless:
### What does this PR do?

This adds support for `/etc/default/datadog-agent` into the debian and centos init scripts.
It then uses the `NICELEVEL` if set.

In debian, start-stop-daemon has a `--nicelevel` argument.

In centos, daemon does not, but seems to honor the NICELEVEL environmental variable

### Motivation

In my environment, we run sshd at a high priority to aid in access during host issues. This higher priority is inherited by apt, and by the init script. Thus accidentally causing datadog-agent to run at high priority

We're finding that when datadog-agent runs at a high priority, gohai ends up causing some CPU locks.  I referenced this in https://github.com/DataDog/dd-agent/issues/3085 This is workaround.

### Additional Notes

I don't know how to add this to the systemd configs. Though they won't have the same inheritance issue I'm running into.

### DIFF
--- a/packaging/centos/datadog-agent.init
+++ b/packaging/centos/datadog-agent.init
@@ -42,6 +42,7 @@ DD_OPT_PROC_REGEX="dogstatsd|jmxfetch|go-metro"
 # Source function library.
 . /etc/rc.d/init.d/functions
 
+[ -r /etc/default/datadog-agent ] && . /etc/default/datadog-agent
 
 check_status() {
     # run checks to determine if the service is running or use generic status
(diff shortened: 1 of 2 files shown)


davexunit:
This patch would be extremely useful to me.  Any chance of it getting merged soon?

packaging/centos/datadog-agent.init
```diff
@@ -42,6 +42,7 @@ DD_OPT_PROC_REGEX="dogstatsd|jmxfetch|go-metro"
 # Source function library.
 . /etc/rc.d/init.d/functions
 
+[ -r /etc/default/datadog-agent ] && . /etc/default/datadog-agent
```
olivielpeau: on centos I think the convention is to source `/etc/sysconfig/datadog-agent` instead of `/etc/default/datadog-agent`

directionless: Okay! I don't know the centos conventions. I'll push an update

directionless:
systemd does have more options. But it's easy for end users to adjust the service files. 

One idiom I've seen is to set `EnvironmentFile` to that config file. Eg: https://github.com/facebook/osquery/blob/master/tools/deployment/osqueryd.service I'm not sure if that makes sense here.

PR #3093, opened 2016-12-22
Skip dbname's we can't access

pbrisbin:
The datadog user may only have access to the databases which the customer
wishes to monitor. Without this patch, a failure to invoke dbstats on _any_
database would raise an exception and prevent monitoring for _all_ databases.
With this patch, databases that cannot be accessed are ignored with a warning.

### DIFF
--- a/checks.d/tokumx.py
+++ b/checks.d/tokumx.py
@@ -12,6 +12,7 @@ import bson
 from pymongo import (
     MongoClient,
     ReadPreference,
+    errors,
     uri_parser,
     version as py_version,
 )
@@ -424,7 +425,11 @@ class TokuMX(AgentCheck):
                 db_tags = list(tags)
                 db_tags.append('db:%s' % dbname)
                 db = conn[dbname]
-                stats = db.command('dbstats')
+                try:
+                    stats = db.command('dbstats')
+                except errors.OperationFailure:
+                    self.log.warning("Cannot access dbstats on database %s" % dbname)
+                    continue
                 for m, v in stats.items():
                     if m in ['db', 'ok']:
                         continue


checks.d/tokumx.py
```diff
@@ -424,7 +424,11 @@ def collect_metrics(self, instance, server, conn, db, tags):
                 db_tags = list(tags)
                 db_tags.append('db:%s' % dbname)
                 db = conn[dbname]
-                stats = db.command('dbstats')
+                try:
+                    stats = db.command('dbstats')
+                except:
```
remh: Is there a more specific exception that can be caught ? 
It's usually bad practice to have bare `except:` statements:
https://docs.python.org/2/howto/doanddont.html#except

pbrisbin: Thanks, totally agree. Here's the error message without this `except`:

```
OperationFailure: command SON([('dbstats', 1)]) on namespace *.$cmd failed: unauthorized
```

I assume this means `except OperationFailure:` is what I want?

pbrisbin: Hmm, or maybe I need to use a pattern like [this](https://github.com/DataDog/dd-agent/blob/master/checks.d/tokumx.py#L380)?

remh: Yes you should import the `OperationFailure` class (see http://api.mongodb.com/python/current/api/pymongo/errors.html ) and then `except OperationFailure`

pbrisbin:
@remh thanks for the review. Updated!

pbrisbin:
FYI: i just applied this patch on a newly provisioned server (after changing the `catch` to be more specific) and it worked as expected.

pbrisbin:
Ping @remh -- is there anything left here preventing Merge?

remh:
Thanks! Looks great to me.

PR #3094, opened 2016-12-22
[process] Emit a warning when no process was found

### LINKED ISSUE #3088: Bug: info command prints OK for process checks that don’t find a process
This is unintuitive and unhelpful.

remh:
### What does this PR do?

This emits a warning when no process is found in the process check

### Motivation

#3088

### DIFF
--- a/checks.d/process.py
+++ b/checks.d/process.py
@@ -356,6 +356,9 @@ class ProcessCheck(AgentCheck):
         self.log.debug('ProcessCheck: process %s analysed', name)
         self.gauge('system.processes.number', len(pids), tags=tags)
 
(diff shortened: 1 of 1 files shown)


aviflax:
Where does the warning go? Will this just go to a log file, or will it change the output of `info`?

remh:
This will show up in the logs, in the info page and as well in the infrastructure overview page.

aviflax:
But `OK` will still appear in the output? If so, that seems misleading to me.

remh:
no.
That's how it would look like:

image

aviflax:
Excellent! Thank you!

Issue #3080, opened 2016-12-13
[core] network_check active threads count causing check aborts


ross:
There's a [check on network_checks.py:120](https://github.com/DataDog/dd-agent/blob/6df4554465338b22441f5bba04913538fe7ef1a2/checks/network_checks.py#L120-L121) that looks to see if there are too many active threads.

```Python
        if threading.activeCount() > 5 * self.pool_size + 5: # On Windows the agent runs on multiple threads so we need to have an offset of 5 in case the pool_size is 1
            raise Exception("Thread number (%s) is exploding. Skipping this check" % threading.activeCount())
```

This works fine as long as you only have one network_check running on a box or otherwise have network checks configured with similar values for `pool_size`. This breaks when you have one or more large pools alongside smaller ones. This is easy situation to get into if you have a large number of instances in one network_check as it sets the pool size to the number of instances.

All of the big pools instance check just fine, but the smaller network_checks never run. We have a custom snmpwalk based check with a VERY large number of instances that runs alongside a small set. As soon as we turned on the large set we lost the small one.

**Additional environment details (Operating System, Cloud provider, etc):**

**Steps to reproduce the issue:**
1. Configure a `NetworkCheck` derived check with a large number of instances each of which will be active for a bit of time
2. Configure another `NetworkCheck` derived check with a small number of instances, e.g. 1

**Describe the results you received:**

```
2016-12-13 12:33:25 PST | ERROR | dd.collector | checks.snmpwalk(__init__.py:788) | Check 'snmpwalk' instance #0 failed
Traceback (most recent call last):
  File "/opt/datadog-agent/agent/checks/__init__.py", line 771, in run
    self.check(copy.deepcopy(instance))
  File "/opt/datadog-agent/agent/checks/network_checks.py", line 122, in check
    raise Exception("Thread number (%s) is exploding. Skipping this check" % threading.activeCount())
Exception: Thread number (32) is exploding. Skipping this check
```

**Describe the results you expected:**

The active thread count should be looked at globally before kicking off the whole check process and take into account the sum of all the pool_sizes rather than compare the global active_threads to a per-instance pool_size.

ross:
[...]

PR #3095, opened 2016-12-22
[network checks] 

remh:
### What does this PR do?

Use a global counter for the pool size.
This way our assertion on the max number of threads will be accurate.

### Motivation

See #3080 for details

### DIFF
--- a/checks/network_checks.py
+++ b/checks/network_checks.py
@@ -18,6 +18,8 @@ DEFAULT_SIZE_POOL = 6
 MAX_LOOP_ITERATIONS = 1000
 FAILURE = "FAILURE"
 
+_global_current_pool_size = 0
+
 
 class Status:
     DOWN = "DOWN"
@@ -95,6 +97,10 @@ class NetworkCheck(AgentCheck):
         default_size = min(self.instance_count(), DEFAULT_SIZE_POOL)
         self.pool_size = int(self.init_config.get('threads_count', default_size))
 
+        # To keep track on the total number of threads we should have running
+        global _global_current_pool_size
+        _global_current_pool_size += self.pool_size
+
         self.pool = Pool(self.pool_size)
 
         self.resultsq = Queue()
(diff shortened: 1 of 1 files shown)


checks/network_checks.py
```diff
@@ -117,7 +128,7 @@ def restart_pool(self):
     def check(self, instance):
         if not self.pool_started:
             self.start_pool()
-        if threading.activeCount() > 5 * self.pool_size + 5: # On Windows the agent runs on multiple threads so we need to have an offset of 5 in case the pool_size is 1
+        if threading.activeCount() > 5 * _global_current_pool_size + 1:
```
ross: Does this need a `global _global_current_pool_size`? It appears in the other 2 places this is used.

checks/network_checks.py
```diff
@@ -18,6 +18,8 @@
 MAX_LOOP_ITERATIONS = 1000
 FAILURE = "FAILURE"
 
+_global_current_pool_size = 0
```
ross: Could this just be a property on the `NetworkCheck` class? Should save the need for the `globals` and tie it to a logically connected class.

checks/network_checks.py
remh: It doesn't, in python the need for the global keyword is not needed when you read the value of a variable.

checks/network_checks.py
```diff
@@ -117,7 +125,7 @@ def restart_pool(self):
     def check(self, instance):
         if not self.pool_started:
             self.start_pool()
-        if threading.activeCount() > 5 * self.pool_size + 5: # On Windows the agent runs on multiple threads so we need to have an offset of 5 in case the pool_size is 1
+        if threading.activeCount() > 5 * NetworkCheck._global_current_pool_size + 1:
```
olivielpeau: why change the offset from `5` to `1`?

remh: The reason was that the windows agent is not using threads to run its different components (collector dogstatsd forwarder) anymore but multiprocessing, which was the reason we had this offset in the first place. However Windows checks use wmi which is itself multithreaded so you're right we should keep this offset. Fixing that.

PR #3103, opened 2016-12-29
[sdk][tests][common] load actual SDK checks (classes/modules) not dd-agent's 

truthbk:
### What does this PR do?

From the SDK, whenever `load_class` or `load_check` were invoked we were actually loading from the sources in the `dd-agent` repo. The aim of this PR is to ensure that when running from SDK-land (both locally or on the CI environments: Travis, CircleCI) we can guarantee the "common" test logic (shared by `dd-agent` and `integrations-core`, `integrations-extra`) actually loads the corresponding modules/classes. 

### Motivation

We need consistent testing, and preferably having a single source for all testing facilities (at least as long as we have commonalities between `dd-agent` and the SDK repos).  We still have checks/logic that store state (docker, kubernetes, etc) and will remain in `dd-agent` for some time, these checks rely largely on the same testing code. Similarly, `integrations-core` and `integrations-extra` both will use this code, so it's definitely something we probably want to do: keep the testing code in `dd-agent`.

### Testing Guidelines

`dd-agent` remains unaffected, the SDK repos now load the right logic as reflected in https://github.com/DataDog/integrations-core/pull/84.

### Additional Notes

There's one caveat to the way this PR actually handles the importing of modules from the SDK, we actually have to "mangle" the SDK module names a little bit to avoid naming conflicts (one great example would be our `rabbitmq` check, the SDK module is named `rabbitmq` as is the official module we used within our check). I have decided to prepend an underscore '_' to SDK imported modules (so the SDK `rabbitmq` will be imported as `_rabbitmq`), this should be transparent to the user but may play a role when patching/mocking as you can see in the [Openstack SDK PR](https://github.com/DataDog/integrations-core/pull/84). If we have any feedback here, I'd love to hear it.

### DIFF
--- a/config.py
+++ b/config.py
@@ -745,7 +745,10 @@ def get_sdk_integrations_path(osname=None):
         if os.environ.get('TRAVIS'):
             path = os.environ['TRAVIS_BUILD_DIR']
         elif os.environ.get('CIRCLECI'):
-            path = os.environ['CIRCLE_PROJECT_REPONAME']
+            path = os.path.join(
+                os.environ['HOME'],
+                os.environ['CIRCLE_PROJECT_REPONAME']
+            )
         elif os.environ.get('APPVEYOR'):
             path = os.environ['APPVEYOR_BUILD_FOLDER']
         else:
--- a/config.py
+++ b/config.py
@@ -740,14 +740,14 @@ def get_checksd_path(osname=None):
 def get_sdk_integrations_path(osname=None):
     if not osname:
         osname = get_os()
-    if osname in ['windows']:
-        raise PathNotFound()
 
     if os.environ.get('INTEGRATIONS_DIR'):
         if os.environ.get('TRAVIS'):
             path = os.environ['TRAVIS_BUILD_DIR']
         elif os.environ.get('CIRCLECI'):
             path = os.environ['CIRCLE_PROJECT_REPONAME']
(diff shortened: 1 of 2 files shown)


config.py
```diff
@@ -740,11 +740,21 @@ def get_checksd_path(osname=None):
...
-    cur_path = os.path.dirname(os.path.realpath(__file__))
-    path = os.path.join(cur_path, '..', SDK_INTEGRATIONS_DIR)
+    if os.environ.get('INTEGRATIONS_DIR'):
+        if os.environ.get('TRAVIS'):
+            path = os.environ['TRAVIS_BUILD_DIR']
+        elif os.environ.get('CIRCLECI'):
+            path = os.environ['CIRCLE_PROJECT_REPONAME']
+        else:
+            cur_path = os.environ['INTEGRATIONS_DIR']
+            path = os.path.join(cur_path, '..') # might need tweaking in the future.
```
gmmeyer: Why would this be `..`? Wouldn't the `INTEGRATIONS_DIR` environment variable be correct?

truthbk: I think you are indeed correct... I'll double check just in case. 👍

truthbk: This was actually correct... we can maybe fix it in the future (very possibly) but INTEGRATIONS_DIR is actually pointing to `SDK_DIR/embedded`, hence the `..`

PR #3104, opened 2017-01-03
pkg: add trace-agent to supervisor

talwai:
In a future release, the `datadog-agent` package will bundle the [datadog-trace-agent](https://github.com/DataDog/datadog-trace-agent)

This change will add a process definition for the trace-agent to the bundled supervisor config. It will also add this proc to the `datadog-agent` process group so that existing install methods will start it up along with existing processes.

Should be merged after https://github.com/DataDog/dd-agent-omnibus/pull/107


gmmeyer:
@talwai you missed this one: https://github.com/DataDog/dd-agent/blob/master/packaging/suse/datadog-agent.init

talwai:
thanks @gmmeyer , should be fixed

PR #3109, opened 2017-01-06
[php_fpm] Include custom tags in service check

jippi:
### What does this PR do?

Include custom tags in service check

### Motivation

Inside a Docker scheduler, using the `ping_url` is bad, since each restart of the process will assign a new unique port to the process, thus marking the old service as gone, even though its in fact running somewhere else successfully.

Including own tags to the service, a cluster alert can be created to mitigate this issue :)

### DIFF
--- a/checks.d/php_fpm.py
+++ b/checks.d/php_fpm.py
@@ -103,7 +103,7 @@ class PHPFPMCheck(AgentCheck):
         if ping_reply is None:
             ping_reply = 'pong'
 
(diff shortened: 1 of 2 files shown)


olivielpeau:
Hi @jippi, thanks for contributing!

The change looks good to me, could you also update the integration test here? https://github.com/DataDog/dd-agent/blob/master/tests/checks/integration/test_php_fpm.py

You should only need to add the configured `tags` to the service check assertions (`assertServiceCheck`)

PR #3111, opened 2017-01-06
[Spark]  Fix check to enumerate job IDs

derekwbrown:
### What does this PR do?

Spark integration, at least against Spark 1.5, was using the wrong value of 'id'.  It appears there are two values, the generic job id name from 8080/json:

 "activeapps" : [ {
    "starttime" : 1483727736675,
    "id" : "app-20170106103536-0024",
    "name" : "PythonPi",
    "user" : "db",
    "memoryperslave" : 1024,
    "submitdate" : "Fri Jan 06 10:35:36 PST 2017",
    "state" : "RUNNING",
    "duration" : 5357
  } ],

However, when you chase down individual job URL, in this case 4040/api/v1/applications/, you get
[ {
  "id" : "PythonPi",
  "name" : "PythonPi",
  "attempts" : [ {
    "startTime" : "2017-01-06T18:35:34.758GMT",
    "endTime" : "1969-12-31T23:59:59.999GMT",
    "sparkUser" : "",
    "completed" : false
  } ]
} ]

When creating the URLs to retrieve the individual job statistics, the api appears to want the id from the 2nd, not the first.  The code was using the id from the first, and failing.

### Motivation

Customer report.

### DIFF
--- a/checks.d/spark.py
+++ b/checks.d/spark.py
@@ -82,6 +82,7 @@ from bs4 import BeautifulSoup
 
 # Project
 from checks import AgentCheck
+from config import _is_affirmative
 
 # Identifier for cluster master address in `spark.yaml`
 MASTER_ADDRESS = 'spark_url'
@@ -93,6 +94,9 @@ SPARK_CLUSTER_MODE = 'spark_cluster_mode'
 SPARK_YARN_MODE = 'spark_yarn_mode'
 SPARK_STANDALONE_MODE = 'spark_standalone_mode'
 SPARK_MESOS_MODE = 'spark_mesos_mode'
+
+# option enabling compatibility mode for Spark ver < 2
+SPARK_PRE_20_MODE = 'spark_pre_20_mode'
 
 # Service Checks
 SPARK_STANDALONE_SERVICE_CHECK = 'spark.standalone_master.can_connect'
@@ -263,7 +267,9 @@ class SparkCheck(AgentCheck):
             cluster_mode = SPARK_YARN_MODE
 
         if cluster_mode == SPARK_STANDALONE_MODE:
-            return self._standalone_init(master_address)
+            # check for PRE-20
+            pre20 = _is_affirmative(instance.get(SPARK_PRE_20_MODE, False))
+            return self._standalone_init(master_address, pre20)
 
         elif cluster_mode == SPARK_MESOS_MODE:
             running_apps = self._mesos_init(master_address)
@@ -278,7 +284,7 @@ class SparkCheck(AgentCheck):
             raise Exception('Invalid setting for %s. Received %s.' % (SPARK_CLUSTER_MODE,
                 cluster_mode))
 
-    def _standalone_init(self, spark_master_address):
+    def _standalone_init(self, spark_master_address, pre_20_mode):
         '''
         Return a dictionary of {app_id: (app_name, tracking_url)} for the running Spark applications
         '''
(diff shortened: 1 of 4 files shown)


derekwbrown:
I have tested on Spark 2.0.  This change is not necessary for Spark 2.0, but it does still work (just one additional http request to the Spark API).  So, we could 
a) leave as is
b) when we get the 404 fall back to this method
c) have a configuration option for pre 2.0 spark
The fix appears to only be necessary for older (pre 2.0) versions of Spark.

derekwbrown:
Added the conditional, so that the new code is only used when configured (on older versions of Spark).  Did the following tests
1) Tried w/o config on Spark 1.5.2, verified bug reproduced.
2) Tried w config on Spark 1.5.2, verified bug fixed
3) Tried w/o config on Spark 2.0, verified works as expected
4) Tried w config on Spark 2.0, verified still works as expected (albeit with extra http request)

checks.d/spark.py
```diff
@@ -198,6 +201,7 @@ def __init__(self, name, init_config, agentConfig, instances=None):
         AgentCheck.__init__(self, name, init_config, agentConfig, instances)
         self.previous_jobs = {}
         self.previous_stages = {}
+        self._spark_pre_20_mode = False
```
truthbk: I don't think we need this as an attribute, it will change from instance to instance and it's not too expensive to look up in the instance dict (something we'd have to do anyway if you made the `self._spark_pre_20_mode` into a dict). We can probably just lookup the value in `check()` for the relevant code paths.

checks.d/spark.py
```diff
@@ -263,6 +267,11 @@ def _get_running_apps(self, instance, tags):
             cluster_mode = SPARK_YARN_MODE
 
         if cluster_mode == SPARK_STANDALONE_MODE:
+            # check for PRE-20
+            pre20 = instance.get(SPARK_PRE_20_MODE)
```
truthbk: I'd use `_is_affirmative()` here (you'll have to import it)...
```
pre20 = _is_affirmative(instance.get(SPARK_PRE_MODE, False))
```
By doing this we achieve two things, if the key doesn't exist (ie. it hasn't been set in the YAML) we'll default to false, and we can also handle stuff like "yes", "no", etc if it is set. You won't need to do the validation like you do in the next couple of lines.

checks.d/spark.py
```diff
@@ -263,6 +267,11 @@ def _get_running_apps(self, instance, tags):
...
+            if pre20 is not None:
+                if pre20 is True:
+                    self._spark_pre_20_mode = True
             return self._standalone_init(master_address)
```
truthbk: We can maybe pass the `pre20` argument to `_standalone_init()` as an optional argument. Prototype would change to:
```
def _standalone_init(self, spark_master_address, pre20=False):
```

checks.d/spark.py
```diff
@@ -624,6 +648,7 @@ def _rest_request(self, address, object_path, service_name, *args, **kwargs):
             url = urljoin(url, '?' + query)
 
         try:
+            self.log.debug('Spark check URL: %s' % url)
```
truthbk: Do we really need this line? I can see it being of value when debugging support cases, but it'll be quite verbose, no? Your call.

truthbk:
Also, since you have the payload formats, etc... it could be worth it to add a test for this use-case here: https://github.com/DataDog/dd-agent/blob/master/tests/checks/mock/test_spark.py

checks.d/spark.py
derekwbrown: It will be quite verbose, but that's why it's debug.  It's the first thing we added (and then sent the customer a patch) when the card came in, so it seems worthwhile since the default will be not to log (info level)

tests/checks/mock/test_spark.py
```diff
@@ -206,6 +212,69 @@ def raise_for_status(self):
             body = f.read()
             return MockStandaloneResponse(body, 200)
 
+def standalone_requests_pre20_get_mock(*args, **kwargs):
```
truthbk: Ideally we could maybe refactor some code out of `standalone_requests_pre20_get_mock()` and `standalone_requests_get_mock()` so we'd have less duplicated code?

maybe something like:

```
class MockStandaloneResponse:
    text = ''
 
    def __init__(self, json_data, status_code):
         self.json_data = json_data
         self.status_code = status_code
         self.text = json_data

    def json(self):
        return json.loads(self.json_data)

    def raise_for_status(self):
        return True

def standalone_request_get_mock(*args, **kwargs):
    if args[0] == STANDALONE_APP_URL:
        with open(Fixtures.file('spark_standalone_apps'), 'r') as f:
        body = f.read()
        return MockStandaloneResponse(body, 200)
    # REMAINING logic

def standalone_request_pre20_get_mock(*args, **kwargs):
    if args[0] == STANDALONE_SPARK_APP_URL:
        with open(Fixtures.file('spark_apps_pre20'), 'r') as f:
            body = f.read()
            return MockStandaloneResponse(body, 200)
    else:
        return standalone_request_get_mock(args, kwargs)
```

PR #3114, opened 2017-01-09
[core] [integrations] Removes checks and config

gmmeyer:
I apologize for the size of this Pull Request. It's gigantic. It's so large that Github starts getting glitchy when I try to view it here.

Here's what it does, in general terms:

First, this removes nearly everything. It removes almost all checks except hdfs, mesos and jenkins because they are deprecated, and kubernetes, docker and agent metrics because they are tied too closely to the agent itself. It removes all config files except the ones for those checks and the same for the autoconf files.

Second, a bunch of existing tests do things like loading existing checks or existing yaml files. 

Unfortunately, they specifically check for the check they're loading, checking to see whether or not the "disk" check is running. The way I think is best to fix this, because it adds the least complexity, is to add them to fixtures, move them to the proper place when the test runs, and then remove them afterwards.

This is also the solution that is least confusing, as these checks are no longer properly part of the codebase of the agent except as fixtures for the tests.

Third, this updates the readme and contributions files.

Fourth, this removes the relevant rakefiles and the commands from Travis.

I think that this about covers everything, however, if I have forgotten anything, let me know and I will add it back in here.


gmmeyer:
@degemer It's also used by service discovery. I didn't think of modifying the path, do you think that would be less disruptive?

I can remove them, yes.

gmmeyer:
I think that this still adds less complexity than that option, but am open to the discussion!

CONTRIBUTING.md
```diff
@@ -28,6 +28,11 @@ Your Pull Request **must** always pass the Travis/Appveyor tests before being me
 
 _If you are adding a dependency (python module, library, ...), please check the [corresponding section](#add-dependencies)._
 
+## [Integrations](https://github.com/DataDog/integrations-core)
+
+All integrations, except for the kubernetes and docker ones, have been moved to the [Integration SDK](https://github.com/DataDog/integrations-core). Please look there to contribute to them.
```
irabinovitch: Are we planning to move docker and k8s as well?

README.md
```diff
@@ -27,6 +23,10 @@ If you do **not** agree with the licensing terms and wish to contribute code non
 please email us at <info@datadoghq.com> before submitting your
 pull request.
 
+# [Integration SDK](https://github.com/DataDog/integrations-core)
+
+All integrations, except for the kubernetes and docker ones, have been moved to the [Integration SDK](https://github.com/DataDog/integrations-core). Please look there to see and contribute to any Integrations.
```
irabinovitch: All integrations, except for the kubernetes and docker, have been moved to the [Integration SDK](https://github.com/DataDog/integrations-core). Please look there to submit related issues, PRs, or review the latest changes.

CONTRIBUTING.md
gmmeyer: Oh I need to update that as well, we have the deprecated hdfs, jenkins and mesos checks in here too, along with the agent_metrics check.

We are but not before the release. Currently they are too tied into the core agent to remove.

irabinovitch: Lets put that in their individual README pages too?

gmmeyer: Individual readme pages?

tests/checks/fixtures/checks/disk.py
```diff
@@ -43,6 +43,9 @@ def check(self, instance):
         # Windows and Mac will always have psutil
         # (we have packaged for both of them)
         if self._psutil():
+            if Platform.is_linux():
```
truthbk: What was this fixture?

conf.d/auto_conf/apache.yaml
```diff
@@ -1,7 +0,0 @@
-docker_images:
```
truthbk: we have to add these to integrations-core....

truthbk: https://github.com/DataDog/integrations-core/pull/216

tests/checks/fixtures/checks/disk.py
truthbk: Got it, for service discovery.

PR #3116, opened 2017-01-09
[docker/hostname] improve get_hostname for containers

hkaj:
if you have not yet done so.*

### What does this PR do?

When the agent is containerized it makes sense to use
docker info's Name as the hostname even if it doesn't resolve.
Because otherwise the agent falls back to the container short ID
which is not really helpful.

This PR drops the resolvability check for the name.

### Motivation

Fixing https://github.com/DataDog/docker-dd-agent/issues/139

### Testing Guidelines

### DIFF
--- a/utils/dockerutil.py
+++ b/utils/dockerutil.py
@@ -177,25 +177,29 @@ class DockerUtil:
 
         return None
 
-    def get_hostname(self, use_default_gw=True):
+    def get_hostname(self, use_default_gw=True, should_resolve=False):
         '''
         Return the `Name` param from `docker info` to use as the hostname
         Falls back to the default route.
         '''
+        # return or raise
+        is_resolvable = lambda host: socket.gethostbyname(host)
 
         if self.hostname is not None:
             # Use cache
-            return self.hostname
+            try:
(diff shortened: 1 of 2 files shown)


degemer:
We first need to make sure that the issue this code was fixing - https://github.com/DataDog/dd-agent/pull/2753 - cannot happen anymore.

hkaj:
I believe this was fixing an issue where the agent couldn't reach the local kubelet due to the hostname not being routable. #3142 introduces new methods to reach kubelet so it should be fine.

utils/dockerutil.py
```diff
@@ -156,25 +156,26 @@ def get_gateway(cls, proc_prefix=""):
...
         Return the `Name` param from `docker info` to use as the hostname
         Falls back to the default route.
         '''
+        # return or raise
+        is_resolvable = lambda host: socket.gethostbyname(host)
 
         if self.hostname is not None:
             # Use cache
-            return self.hostname
+            if not should_resolve or is_resolvable(self.hostname):
```
degemer: This line can raise an uncaught error if the hostname is not resolvable.

PR #3117, opened 2017-01-10
Add udp metric InCsumErrors for checksum errors

diver-in-sky:
### What does this PR do?

Adds one more udp related metric.

### Motivation

InCsumErrors is not available. Here is good explanation of different fields from `/proc/net/snmp`: https://blog.packagecloud.io/eng/2016/06/22/monitoring-tuning-linux-networking-stack-receiving-data/#procnetsnmp

> `InCsumErrors`: Incremented when a UDP checksum failure is detected. Note that in all cases I could find, `InCsumErrors` is incrememnted at the same time as `InErrors`. Thus, `InErrors` - `InCsumErros` should yield the count of memory related errors.


remh:
Looks good to me.
Thanks !

PR #3118, opened 2017-01-10
[tcp_check] Set instance tags in response_time metric

mberner-ch:
Before this commit, adding instance tags in the tcp_check configuration
only applied the tags to the (deprecated) service_check but not the
response_time metric sent to datadog.
With this improvement, the custom instance tags are also applied to the
response_time metric, as I would expect.


degemer:
Hey @mberner-ch, thanks for the contribution! I took the liberty of slightly editing your changes (mostly adding a test). Going to merge it now, it will be released with 5.12!

mberner-ch:
Excellent, cheers! :)

iancward:
Would it be possible to update the tcp_check example to indicate custom tag support?

PR #3122, opened 2017-01-11
[packaging] fixes source install for sdk

gmmeyer:
This fixes the source install after https://github.com/DataDog/dd-agent/pull/3114 is merged.

I tested the new logic on dash, bash and zsh.

### DIFF
--- a/packaging/datadog-agent/source/setup_agent.sh
+++ b/packaging/datadog-agent/source/setup_agent.sh
@@ -389,25 +389,31 @@ tar -xz -C "$DD_HOME/agent" --strip-components 1 -f "$DD_HOME/agent.tar.gz"
 rm -f "$DD_HOME/agent.tar.gz"
 print_done
 
-print_console "* Downloading integrations from GitHub"
-mkdir -p "$DD_HOME/integrations"
-$DOWNLOADER "$DD_HOME/integrations.tar.gz" "https://api.github.com/repos/DataDog/integrations-core/tarball"
-print_done
+AGENT_VERSION_ARRAY=(${AGENT_VERSION//./ })
+AGENT_MAJOR_VERSION=${AGENT_VERSION_ARRAY[0]}
+AGENT_MINOR_VERSION=${AGENT_VERSION_ARRAY[1]}
+# Only install the integrations from the integrations-core if it's version 5.11 or above.
+if [[ (("$AGENT_MAJOR_VERSION" = "5")) && (("$AGENT_MINOR_VERSION" > "11")) ]]; then
+  print_console "* Downloading integrations from GitHub"
+  mkdir -p "$DD_HOME/integrations"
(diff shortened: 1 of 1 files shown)


packaging/datadog-agent/source/setup_agent.sh
```diff
@@ -389,6 +389,26 @@ tar -xz -C "$DD_HOME/agent" --strip-components 1 -f "$DD_HOME/agent.tar.gz"
 rm -f "$DD_HOME/agent.tar.gz"
 print_done
 
+print_console "* Downloading integrations from GitHub (~5 MB)"
+mkdir -p "$DD_HOME/3rd-party"
+$DOWNLOADER "$DD_HOME/integrations.tar.gz" "https://api.github.com/repos/DataDog/integrations-core/tarball"
```
gmmeyer: when we end up versioning the integrations core with the version number of the agent, we can download it like this:

`https://api.github.com/repos/DataDog/integrations-core/tarball/{version}`

see more here: https://developer.github.com/v3/repos/contents/#get-archive-link

`GET /repos/:owner/:repo/:archive_format/:ref`

gmmeyer:
The alternative to this is to create an agent source install tarball that we can put up on s3. I favor this approach.

packaging/datadog-agent/source/setup_agent.sh
```diff
@@ -389,6 +389,26 @@ tar -xz -C "$DD_HOME/agent" --strip-components 1 -f "$DD_HOME/agent.tar.gz"
 rm -f "$DD_HOME/agent.tar.gz"
 print_done
 
+print_console "* Downloading integrations from GitHub"
+mkdir -p "$DD_HOME/integrations"
+$DOWNLOADER "$DD_HOME/integrations.tar.gz" "https://api.github.com/repos/DataDog/integrations-core/tarball"
```
olivielpeau: let's pull a pinned version of the integrations-core repo

packaging/datadog-agent/source/setup_agent.sh
```diff
@@ -389,6 +389,26 @@ tar -xz -C "$DD_HOME/agent" --strip-components 1 -f "$DD_HOME/agent.tar.gz"
...
+print_console "* Uncompressing tarball"
+tar -xz -C "$DD_HOME/integrations" --strip-components 1 -f "$DD_HOME/integrations.tar.gz"
+rm -f "$DD_HOME/integrations.tar.gz"
+print_done
+
+print_console "* Trying to install integration requirements"
+for INT_DIR in "$DD_HOME/integrations/*"; do
+  if [ -d $INT_DIR ]; then
+    if [-f "$INT_DIR/requirements.txt" ]; then
+      "$DD_HOME/agent/utils/pip-allow-failures.sh" "$INT_DIR/requirements.txt"
```
olivielpeau: let's concatenate all the requirement files (agent + integrations) in one constraint file and use that constraint with every pip call so that we're sure we're not installing conflicting dependencies.

We can even use the output of `pip freeze` instead of the agent's `requirements.txt` file to generate that constraint file

PR #3123, opened 2017-01-11
[install_agent.sh] Fixed openSUSE autoinstallation failures

ScaffaCS:
Automatic repo add + install on openSUSE was failing. Fixed in lines 85 through 88 for both current and future releases.

if you have not yet done so.*

### What does this PR do?

Adds in openSUSE support, which was broken on openSUSE Leap 42.2

### Motivation

Pending DataDog deployment on OpenStack system which uses mix of SUSE / openSUSE.

### Testing Guidelines

### DIFF
--- a/packaging/datadog-agent/source/install_agent.sh
+++ b/packaging/datadog-agent/source/install_agent.sh
@@ -81,9 +81,7 @@ elif [ -f /etc/system-release -o "$DISTRIBUTION" == "Amazon" ]; then
 # Arista is based off of Fedora14/18 but do not have /etc/redhat-release
 elif [ -f /etc/Eos-release -o "$DISTRIBUTION" == "Arista" ]; then
     OS="RedHat"
-# openSUSE and SUSE use /etc/SuSE-release (legacy), in the future it will be /etc/os-release
-elif [ -f /etc/os-release -o "$DISTRIBUTION" == "SUSE" -o "$DISTRIBUTION" == "openSUSE" ]; then
-    OS="SUSE"
+# openSUSE and SUSE use /etc/SuSE-release
 elif [ -f /etc/SuSE-release -o "$DISTRIBUTION" == "SUSE" -o "$DISTRIBUTION" == "openSUSE" ]; then
     OS="SUSE"
 fi
--- a/packaging/datadog-agent/source/install_agent.sh
+++ b/packaging/datadog-agent/source/install_agent.sh
@@ -81,7 +81,10 @@ elif [ -f /etc/system-release -o "$DISTRIBUTION" == "Amazon" ]; then
 # Arista is based off of Fedora14/18 but do not have /etc/redhat-release
 elif [ -f /etc/Eos-release -o "$DISTRIBUTION" == "Arista" ]; then
     OS="RedHat"
-elif [ "$DISTRIBUTION" == "SUSE" ]; then
+# openSUSE and SUSE use /etc/SuSE-release (legacy), in the future it will be /etc/os-release
+elif [ -f /etc/os-release -o "$DISTRIBUTION" == "SUSE" -o "$DISTRIBUTION" == "openSUSE" ]; then
+    OS="SUSE"
+elif [ -f /etc/SuSE-release -o "$DISTRIBUTION" == "SUSE" -o "$DISTRIBUTION" == "openSUSE" ]; then
     OS="SUSE"
 fi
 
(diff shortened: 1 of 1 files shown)


gmmeyer:
Hey @ScaffaCS! 

Thanks for the contribution!

Honestly this PR looks fine, the change makes sense. However, I need to test this to ensure it doesn't break anything for anyone else.

Can you tell me what error you're getting when you try to install it now? I'm going to attempt to replicate this, but I want to know what I'm replicating.

Thanks again! It's really appreciated! 😃

ScaffaCS:
Hey @gmmeyer 

Installing on a fresh OpenSUSE LEAP 42.2 box with this command:

<pre>linux-1o6l:/home/craig # DD_API_KEY=XXXXXXX bash -c "$(curl -L https://raw.githubusercontent.com/DataDog/dd-agent/master/packaging/datadog-agent/source/install_agent.sh)"
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  9822  100  9822    0     0  98977      0 --:--:-- --:--:-- --:--:-- 99212

* Installing YUM sources for Datadog

sh: /etc/yum.repos.d/datadog.repo: No such file or directory</pre>

The issue is that openSUSE / SUSE use the directory;

<pre>/etc/zypp/repos.d/<pre>

The logic to use this location is already there for distributions detected as "SUSE" on line 172.

All we need to do is correct the OS detection to flag openSUSE as "SUSE" rather than RedHat.

Thanks,

Craig

packaging/datadog-agent/source/install_agent.sh
```diff
@@ -73,15 +73,18 @@ Please use the 1-step script available at https://app.datadoghq.com/account/sett
...
     OS="RedHat"
 # Some newer distros like Amazon may not have a redhat-release file
 elif [ -f /etc/system-release -o "$DISTRIBUTION" == "Amazon" ]; then
     OS="RedHat"
 # Arista is based off of Fedora14/18 but do not have /etc/redhat-release
 elif [ -f /etc/Eos-release -o "$DISTRIBUTION" == "Arista" ]; then
     OS="RedHat"
-elif [ "$DISTRIBUTION" == "SUSE" ]; then
+# openSUSE and SUSE use /etc/SuSE-release (legacy), in the future it will be /etc/os-release
+elif [ -f /etc/os-release -o "$DISTRIBUTION" == "SUSE" -o "$DISTRIBUTION" == "openSUSE" ]; then
```
gmmeyer: Is os-release SUSE specific? If not, can you change this to something else because this could trigger on this incorrectly.

ScaffaCS: @gmmeyer 

It looks like os-release is part of systemd, so yes it's very likely this will be present on other Linux distributions.

Source: [freedesktop.org:osrelease](https://www.freedesktop.org/software/systemd/man/os-release.html)

<pre>NAME=

    A string identifying the operating system, without a version component, and suitable for presentation to the user. If not set, defaults to "NAME=Linux". Example: "NAME=Fedora" or "NAME="Debian GNU/Linux"". </pre>

gmmeyer: Yea I just confirmed that myself as well, unfortunately.

ScaffaCS:
What do you think of this:

- Remove /etc/os-release references. 
- Leave in the /etc/SuSE-release reference as a way of detecting SUSE / openSUSE.
- Make another pull request to work on adding in the /etc/os-release logic.

**/etc/os-release** seems to be a "moving forward" file. I've found it in Ubuntu 16.04 / CentOS-7 / SUSE 12sp1 / openSUSE LEAP.

However, all these distributions still contain the /etc/lsb-release or /etc/redhat-release or /etc/debian-release and such. So including /etc/os-release could cause problems, but leaving it out will not.

gmmeyer:
I think that makes sense to me

ScaffaCS:
Thanks @gmmeyer 

I'll take a look at the systemd stuff once I'm back from the states.

gmmeyer:
Wait, what systemd stuff?

ScaffaCS:
The /etc/os-release references which I put in, but removed as it would've
caused problems.

gmmeyer:
Oh, yes! Sorry, you're right. Those are annoyingly non specific.

PR #3124, opened 2017-01-12
[mongo] add metric with number of databases

mpidlisnyi:
We want to see number of mongodb databases because our databases can grow dynamically. I was thinking about this code. The best solution which I see it's usage of GAUGE link to AgentCheck.gauge function with static metric's name.
Please correct me if I am wrong
Would like to see this metric in the oficial repository
Thanks

### DIFF
--- a/checks.d/mongo.py
+++ b/checks.d/mongo.py
@@ -829,7 +829,7 @@ class MongoDb(AgentCheck):
             pass
 
         dbnames = cli.database_names()
-        GAUGE(self, 'mongodb.dbs', len(dbnames), tags=tags)
+        self.gauge('mongodb.dbs', len(dbnames), tags=tags)
 
         for db_n in dbnames:
             db_aux = cli[db_n]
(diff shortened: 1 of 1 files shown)


gmmeyer:
Hey @mpidlisnyi! Thanks for your pull request. I made a small change to it, I wanted to bring it in line with our other checks, but I think it looks pretty good overall. I'm going to merge it just after the tests run! Thank you very much for your contribution! It's very much appreciated.

checks.d/mongo.py
```diff
@@ -829,7 +829,7 @@ def total_seconds(td):
             pass
 
         dbnames = cli.database_names()
-        GAUGE(self, 'mongodb.dbs', len(dbnames), tags=tags)
+        self.gauge('mongodb.dbs', len(dbnames), tags=tags)
```
truthbk: One question, I see the tags list is extended in the big `try...except` block above. I'm not sure if all of those `tags` would apply to this particular metric. My knowledge of the mongo check is a bit limited, we should make sure of that before merging.

gmmeyer: I think you might be right that line might be problematic. However, it affects a lot of the check, so it should be separate from this change.

PR #3125, opened 2017-01-12
[win32][gui] Fix old process detection and termination

olivielpeau:
### What does this PR do?

Ignore "access denied" errors when accessing the old process by PID.

Also, switch to psutil for simplicity. I think we didn't use psutil
before because it still had some bugs on Windows, but it looks like
these bugs have been fixed since then.

### Motivation

The agent manager would crash on startup if the stored PID was
referring to a process that the user couldn't access or terminate (usually
when the process is a system process).

Example stack trace in `agent-manager.exe.log`:
```
  File "gui.py", line 847, in <module>
  File "gui.py", line 822, in kill_old_process
pywintypes.error: (5, 'OpenProcess', 'Access is denied.')
```

### Testing Guidelines

Tested manually on a custom build of the agent, by playing a little bit with the
PID stored in the PID file.

### DIFF
--- a/win32/gui.py
+++ b/win32/gui.py
@@ -62,6 +62,7 @@ spyderlib.baseconfig.IMG_PATH = [""]
 from spyderlib.widgets.sourcecode.codeeditor import CodeEditor
 
 # 3rd Party others
+import psutil  # psutil is always present on both windows and OS X installs
 import tornado.template as template
 import yaml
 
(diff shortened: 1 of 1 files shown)


olivielpeau:
Hmm you're right the comment doesn't reflect what's actually done by the function, I'll update it

olivielpeau:
Updated the comment, going to merge

PR #3127, opened 2017-01-13
[marathon] catch can't connect instead of failing on nodata found

supernovae:
### What does this PR do?

If you use marathon.can_connect to check for marathon service availability and health, it won't fail if the check can't connect (if marathon is down) 

### Motivation

The check should raise a critical event if marathon can't connect. Otherwise if marathon is down you may only get an alert if you have enabled "Alert on no data" which can be a high timeframe as well.

### Additional Notes

I'm not a python dev by anymeans, but perhaps other "can_connect" alerts should be updated to send critical alert on "Can't connect" rather than rely on no data found.


truthbk:
Thanks a lot @supernovae! Definitely a case we hadn't considered! 🙇

PR #3133, opened 2017-01-18
Added burrow agent check

lypnol:
### What does this PR do?

Adds https://github.com/linkedin/Burrow (kafka 09 consumer monitoring tool) to AgentCheck

### Motivation

Get lag metric for kafka09 clusters

### DIFF
--- a/checks.d/burrow.py
+++ b/checks.d/burrow.py
@@ -1,0 +1,86 @@
+# (C) Datadog, Inc. 2010-2017
+# All rights reserved
+# Licensed under Simplified BSD License (see LICENSE)
+
+# stdlib
+import urlparse
+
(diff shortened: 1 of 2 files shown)


gmmeyer:
We're moving all the checks out of this repo into https://github.com/datadog/integrations-core and https://github.com/DataDog/integrations-extras. As such we're not accepting any PRs for new checks into this repo. 

Can you move it to the appropriate repo and add some integration tests? If you clone either repo and type `rake generate:skeleton[burrow]` it'll get you a good bit of the way there, then just spin up the infra you need and write a small test (you can look in integrations core if you need some assistance in how we do this).

Thank you! 😃

PR #3142, opened 2017-01-23
[kubernetes] add support for kubelet authN/authZ

hkaj:
if you have not yet done so.*

### What does this PR do?

Make the agent able to authenticate against kubelet in case the `read-only` port is disabled and we have to use the HTTPS API.
The flow is as follow:
- try connecting to the read-only port without TLS
- fall back to HTTPS port, with auth if necessary/configured

The default behavior in case the read-only port is disabled is to enable cert validation for the kubelet. If the user's PKI setup is standalone and the cert can't be verified, all kubernetes-related features will fail (the kubernetes check, parts of service discovery, part of the docker check tagging). In this case, passing a certificate to use for verification is supported, and recommended.

The agent also sends automatically its service account bearer token, so kubelet's authZ strategy can rely on that. It is also possible to specify a client cert/key pair that the agent will use to authenticate against kubelet.

It's possible to explicitly disable certificate validation, however this should be used with caution as it may create the risk of leaking the agent's service account bearer token.

### Motivation

Kubernetes 1.5(.1) introduced kubelet authentication, let's be good citizens and support it.
Also, fix https://github.com/DataDog/docker-dd-agent/issues/145

### Testing Guidelines

kubeutil tests have been updated. Integration tests are still, uh... manual, but I'll add some doc about how to proceed in the wiki.

### Additional Notes

Clean up some code too.


utils/kubernetes/kubeutil.py
```diff
@@ -140,33 +209,33 @@ def retrieve_metrics(self):
         """
         return retrieve_json(self.metrics_url)
 
-    def filter_pods_list(self, pods_list, host_ip):
+    def retrieve_kubelet_url(self, url, verbose=True, timeout=10):
```
remh: This is not an explicit name for what it does. This sounds like it's returning a url as a string.

utils/http.py
```diff
@@ -8,7 +8,7 @@
 DEFAULT_TIMEOUT = 10
 
 
-def retrieve_json(url, timeout=DEFAULT_TIMEOUT):
-    r = requests.get(url, timeout=timeout)
+def retrieve_json(url, timeout=DEFAULT_TIMEOUT, verify=True):
+    r = requests.get(url, timeout=timeout, verify=verify)
```
remh: 👍

utils/kubernetes/kubeutil.py
```diff
@@ -56,27 +60,107 @@ def __init__(self, instance=None):
...
-        self.kubelet_api_url = '%s://%s:%d' % (self.method, self.kubelet_host, self.kubelet_port)
-        self.cadvisor_url = '%s://%s:%d' % (self.method, self.kubelet_host, self.cadvisor_port)
+        # apiserver
         self.kubernetes_api_url = 'https://%s/api/v1' % (os.environ.get('KUBERNETES_SERVICE_HOST') or self.DEFAULT_MASTER_NAME)
 
+        # kubelet
+        self.ssl_settings = self._init_ssl_settings(instance)
+        self.kubelet_api_url = self._locate_kubelet(instance)
+        if not self.kubelet_api_url:
+            log.exception("Kubernetes check exiting, cannot run without access to kubelet.")
```
remh: Should we raise an exception instead ? You shouldn't use `log.exception` if there is no exception btw.

utils/kubernetes/kubeutil.py
```diff
@@ -56,27 +60,107 @@ def __init__(self, instance=None):
...
+
+            https_url = 'https://%s:%s' % (host, port)
+            test_url = urljoin(https_url, KubeUtil.KUBELET_HEALTH_PATH)
+            self.retrieve_kubelet_url(test_url)
+
+            return https_url
+        except Exception as ex:
+            log.error("Kubelet unreachable.\nPlease configure the no-auth endpoint or configure authentication: %s" % str(ex))
+
+    def get_node_hostname(self, host):
```
remh: can you add a docstring to this function to detail what it does ?

utils/kubernetes/kubeutil.py
```diff
@@ -56,27 +60,107 @@ def __init__(self, instance=None):
...
+        try:
+            port = instance.get('kubelet_port', KubeUtil.DEFAULT_HTTPS_KUBELET_PORT)
+
+            https_url = 'https://%s:%s' % (host, port)
+            test_url = urljoin(https_url, KubeUtil.KUBELET_HEALTH_PATH)
+            self.retrieve_kubelet_url(test_url)
+
+            return https_url
+        except Exception as ex:
+            log.error("Kubelet unreachable.\nPlease configure the no-auth endpoint or configure authentication: %s" % str(ex))
```
remh: i think we should just raise the error here ?

utils/kubernetes/kubeutil.py
```diff
@@ -56,27 +60,107 @@ def __init__(self, instance=None):
...
+        if not host:
+            # if no hostname was provided, use the docker hostname if ssl
+            # validation is not required, the kubernetes hostname otherwise.
+            docker_hostname = self.docker_util.get_hostname()
+            if self.ssl_settings['verify']:
+                try:
+                    k8s_hostname = self.get_node_hostname(docker_hostname)
+                    host = k8s_hostname or docker_hostname
+                except Exception as ex:
+                    log.error(str(ex))
```
remh: What are the scenarios where an exception can be raised ?

utils/kubernetes/kubeutil.py
```diff
@@ -56,27 +60,107 @@ def __init__(self, instance=None):
...
+            self.retrieve_kubelet_url(test_url)
+
+            return https_url
+        except Exception as ex:
+            log.error("Kubelet unreachable.\nPlease configure the no-auth endpoint or configure authentication: %s" % str(ex))
+
+    def get_node_hostname(self, host):
+        node_filter = 'labelSelector=kubernetes.io/hostname'
+        node = self.retrieve_json_auth(
+            self.kubernetes_api_url + '/nodes?%s%%3D%s' % (node_filter, host),
```
remh: wanna use url encode instead ?

utils/kubernetes/kubeutil.py
hkaj: Ha apparently i couldn't decide between logging something or raising something. Thanks.

utils/kubernetes/kubeutil.py
hkaj: `get_node_hostname` can raise if reaching the apiserver (connecting, validating its cert...) fails.

Right now it can also happen if the host filtering doesn't work and we can't pick which hostname to use for the local kubelet, but I'll actually make that a `log.error` since we can try and fallback to the docker hostname and fail later if cert validation doesn't like that.

utils/kubernetes/kubeutil.py
hkaj: 👍 if we can't reach the kubelet there's no point running the check

hkaj: actually the exception is raised in __init__ anyway, i'll concat here.

hkaj:
@remh thanks for the review! Updated according to your comments. I also stopped sending the bearer token when cert auth is explicitly configured by the user to avoid confusing the backend with multiple identities.

hkaj:
@remh I fixed the conflicts coming from the work on apiserver auth, ready for another round :)

remh:
This looks good to me @hkaj but you'll need a rebase with the integration SDK

hkaj:
i kinda figured, thanks

PR #3144, opened 2017-01-23
[linux_proc_extras] backport linux_proc_extras changes from the sdk

gmmeyer:
I made a few changes to linux_proc_extras when I moved it to the sdk. It makes it easier to test. Since this is still the source of truth, we have to backport it here.

### DIFF
--- a/checks.d/linux_proc_extras.py
+++ b/checks.d/linux_proc_extras.py
@@ -1,4 +1,5 @@
 # (C) Cory Watson <cory@stripe.com> 2016
+# (C) Datadog, Inc. 2016
 # All rights reserved
 # Licensed under Simplified BSD License (see LICENSE)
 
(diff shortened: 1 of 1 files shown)


checks.d/linux_proc_extras.py
```diff
@@ -24,47 +25,66 @@
 }
 
 class MoreUnixCheck(AgentCheck):
+    tags = []
+    proc_path_map = {
```
degemer: Is it used?

checks.d/linux_proc_extras.py
```diff
@@ -24,47 +25,66 @@
 }
 
 class MoreUnixCheck(AgentCheck):
+    tags = []
```
degemer: Same, it's set anyway at every check, unless I'm missing something?

checks.d/linux_proc_extras.py
```diff
@@ -24,47 +25,66 @@
...
+        with open(self.proc_path_map['stat_info'], 'r') as stat_info:
             lines = [line.strip() for line in stat_info.readlines()]
-
             for line in lines:
                 if line.startswith('ctxt'):
                     ctxt_count = float(line.split(' ')[1])
-                    self.monotonic_count('system.linux.context_switches', ctxt_count, tags=tags)
+                    self.monotonic_count('system.linux.context_switches', ctxt_count, tags=self.tags)
                 elif line.startswith('processes'):
+                    self.log.info(line)
```
degemer: Do we really want this at INFO level? Side note: if we're only reading the log, without knowing the check, this line doesn't make sense. Could we prefix it with what it is/means?

checks.d/linux_proc_extras.py
```diff
@@ -24,47 +25,66 @@
...
                     process_count = int(line.split(' ')[1])
-                    self.monotonic_count('system.linux.processes_created', process_count, tags=tags)
+                    self.log.info(process_count)
```
degemer: Same

checks.d/linux_proc_extras.py
gmmeyer: It's used in a bunch of places.

checks.d/linux_proc_extras.py
gmmeyer: This I can get rid of

checks.d/linux_proc_extras.py
gmmeyer: Yup I'll remove it thanks!

checks.d/linux_proc_extras.py
gmmeyer: Yup I'll remove it thanks!

checks.d/linux_proc_extras.py
gmmeyer: It's also static, the tags thing was wrong as it's not static, but this makes sense to include here

gmmeyer: Actually it's redefined, so no I don't think it is. You're right.

PR #3145, opened 2017-01-24
[kubernetes] support cert auth against apiserver

hkaj:
if you have not yet done so.*

### What does this PR do?

Authentication against the apiserver was limited to bearer token. This PR adds support for X509 cert auth.

### Motivation

#3142 adds the feature for authenticating against kubelet, apiserver deserves some love too.
Also, it was requested by existing users ( 👋 @netvisao )

### Testing Guidelines

Tests were updated accordingly.

### DIFF
--- a/conf.d/kubernetes.yaml.example
+++ b/conf.d/kubernetes.yaml.example
@@ -13,6 +13,14 @@ instances:
   # method: http
  - port: 4194
 
+  # Authentication against the apiserver
+  # By default the agent authenticates against the apiserver with its service account bearer token.
+  # X509 certificate authentication can also be enabled by providing here the paths to
+  # a client cert/key pair.
+  # The recommended way to expose these files to the agent is by using Kubernetes Secrets.
+  # apiserver_client_crt: /path/to/client.crt
+  # apiserver_client_key: /path/to/client.key
+  #
   # collect_events controls whether the agent should fetch events from the kubernetes API and
   # ingest them in Datadog. To avoid duplicates, only one agent at a time across the entire
   # cluster should have this feature enabled. To enable the feature, set the parameter to `true`.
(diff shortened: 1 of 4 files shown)


utils/kubernetes/kubeutil.py
```diff
@@ -77,6 +78,23 @@ def __init__(self, instance=None):
         # default value is 0 but TTL for k8s events is one hour anyways
         self.last_event_collection_ts = 0
 
+    def _init_tls_settings(self, instance):
+        """
+        Initialize TLS settings for connection to apiserver and kubelet.
```
hkaj: actually not for kubelet yet, but it's introduced in another existing PR

utils/kubernetes/kubeutil.py
```diff
@@ -159,7 +177,7 @@ def filter_pods_list(self, pods_list, host_ip):
         pods_list['items'] = filtered_pods
         return pods_list
 
-    def retrieve_json_auth(self, url, auth_token, timeout=10):
+    def retrieve_json_auth(self, url, timeout=10):
         """
         Kubernetes API requires authentication using a token available in
         every pod.
```
masci: I'd put a docstring here about certificates options and the fact that the token is ignored in that case

conf.d/kubernetes.yaml.example
```diff
@@ -13,6 +13,14 @@ instances:
   # method: http
  - port: 4194
 
+  # Authentication against the apiserver
+  # By default the agent authenticates against the apiserver with its service account bearer token.
+  # X509 certificate authentication can also be enabled by providing here the paths to
+  # a client cert/key pair.
```
masci: Should we advertise that this overrides the token?

PR #3150, opened 2017-01-26
[Service Discovery][jmx] is_jmx flag in auto_conf

sjenriquez:
### What does this PR do?

- Enables configuration of JMX yaml files with any name when they have the `is_jmx: true` flag  in the `init_config` section. This mirrors the same functionality for JMX yaml files in `/conf.d`.
- Adds validation for file names in the `/auto_conf` directory to only configure files with `.yaml` extension. This mirrors the same functionality as in `/conf.d`. For example, `foo.yaml.example` will be ignored.

### Motivation

Support case that found the `is_jmx` limitation.

### DIFF
--- a/agent.py
+++ b/agent.py
@@ -59,7 +59,6 @@ from utils.service_discovery.config_stores import get_config_store
 from utils.service_discovery.sd_backend import get_sd_backend
 
 # Constants
-from jmxfetch import JMX_CHECKS
 PID_NAME = "dd-agent"
 PID_DIR = None
 WATCHDOG_MULTIPLIER = 10
@@ -137,8 +136,7 @@ class Agent(Daemon):
                 jmx_sd_configs = generate_jmx_configs(self._agentConfig, hostname)
         else:
             new_checksd = copy(self._checksd)
-            auto_conf_is_jmx_checks = get_jmx_checks()
-            all_jmx_checks = JMX_CHECKS + auto_conf_is_jmx_checks
+            all_jmx_checks = get_jmx_checks(auto_conf=True)
             jmx_checks = [check for check in checks_to_reload if check in all_jmx_checks]
             py_checks = set(checks_to_reload) - set(jmx_checks)
             self.refresh_specific_checks(hostname, new_checksd, py_checks)
--- a/agent.py
+++ b/agent.py
@@ -39,11 +39,12 @@ from config import (
     generate_jmx_configs,
     _is_affirmative,
     SD_PIPE_NAME
-
 )
 from daemon import AgentSupervisor, Daemon
 from emitter import http_emitter
 from utils.platform import Platform
+from jmxfetch import get_jmx_checks
+
 
 # utils
 from util import Watchdog
@@ -136,8 +137,9 @@ class Agent(Daemon):
                 jmx_sd_configs = generate_jmx_configs(self._agentConfig, hostname)
         else:
             new_checksd = copy(self._checksd)
-
-            jmx_checks = [check for check in checks_to_reload if check in JMX_CHECKS]
+            auto_conf_is_jmx_checks = get_jmx_checks()
+            all_jmx_checks = JMX_CHECKS + auto_conf_is_jmx_checks
+            jmx_checks = [check for check in checks_to_reload if check in all_jmx_checks]
             py_checks = set(checks_to_reload) - set(jmx_checks)
             self.refresh_specific_checks(hostname, new_checksd, py_checks)
             if self._jmx_service_discovery_enabled:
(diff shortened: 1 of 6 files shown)


agent.py
```diff
@@ -136,8 +137,9 @@ def reload_configs(self, checks_to_reload=set()):
                 jmx_sd_configs = generate_jmx_configs(self._agentConfig, hostname)
         else:
             new_checksd = copy(self._checksd)
-
-            jmx_checks = [check for check in checks_to_reload if check in JMX_CHECKS]
+            auto_conf_is_jmx_checks = get_jmx_checks()
+            all_jmx_checks = JMX_CHECKS + auto_conf_is_jmx_checks
```
truthbk: I noticed that practically everywhere where we're calling `get_jmx_checks()` we then do:
```
all_jmx_checks = get_jmx_checks()
all_jmx_checks += JMX_CHECKS 
```

Maybe we should just do that in the actual function instead when called with the auto_conf paramenter set to `True`? Let me know if you think this doesn't make sense.

Also, I'm not sure if it might be a problem having duplicates (possibly not, but it would still be cleaner not to have them), so we could maybe use a set instead?

config.py
```diff
@@ -1129,16 +1129,22 @@ def load_check(agentConfig, hostname, checkname):
...
 def generate_jmx_configs(agentConfig, hostname, checknames=None):
     """Similar logic to load_check_directory for JMX checks"""
-    from jmxfetch import JMX_CHECKS
+    from jmxfetch import JMX_CHECKS, get_jmx_checks
+
+    jmx_checks = JMX_CHECKS
+    auto_conf_jmx_checks = get_jmx_checks()
+    for check in auto_conf_jmx_checks:
+        if check not in JMX_CHECKS:
+            jmx_checks.append(check)
```
truthbk: This is the kind of thing we could maybe not need to do if we rely on sets, or make sure the list doesn't contain any duplicates when returning from `get_jmx_checks()`

datadog.conf.example
```diff
@@ -103,7 +103,7 @@ gce_updated_hostname: yes
 
 # ========================================================================== #
 # Service Discovery                                                          #
-# See https://github.com/DataDog/dd-agent/wiki/Service-Discovery for details #
+# See http://docs.datadoghq.com/guides/servicediscovery/ for details #
```
truthbk: 🍰

jmxfetch.py
```diff
@@ -460,6 +451,40 @@ def _get_path_to_jmxfetch(self):
         return os.path.realpath(os.path.join(os.path.abspath(__file__), "..", "..",
                                 "jmxfetch", JMX_FETCH_JAR_NAME))
 
+def get_jmx_checks(auto_conf=True, confd_path=None):
```
sjenriquez: Not a bad idea to maintain consistency with the fact that service discovery came after the use of `confd_path`.

utils/checkfiles.py
```diff
@@ -90,28 +90,33 @@ def get_auto_conf_images(full_tpl=False):
...
-                    auto_conf_images[image] = [[check_name], [init_tpl], [instance_tpl]]
-                else:
-                    for idx, item in enumerate([check_name, init_tpl, instance_tpl]):
-                        auto_conf_images[image][idx].append(item)
-            else:
-                if image in auto_conf_images:
-                    auto_conf_images[image].append(check_name)
+        # Ignore files that do not end in .yaml
+        extension = yaml_file.split('.')[-1]
+        if extension != 'yaml':
```
sjenriquez: `.yml` is parsed successfully, however, [we look for the file](https://github.com/DataDog/dd-agent/pull/3150/files#diff-305e19f343add5fc856b48d568ee8d97R63) by adding `.yaml` on to `check_name`.

 This assumes all config files end in `.yaml`. The same is true for files in `/conf.d`, just tested with `docker_daemon.yml` and it was not picked up.

agent.py
sjenriquez: Had the same thought! Adding to `get_jmx_checks()` and changing to a set, we don't want duplicates.

config.py
sjenriquez: When `get_jmx_checks()` is [called from jmxfetch.py](https://github.com/DataDog/dd-agent/pull/3150/files#diff-64a477c26f008c29252aad6c1499c099R202) it expects a dict in return for each returned check. A set of dicts is troublesome, it may be best to keep as a list and add logic to avoid duplicates?

Will push a refactor shortly.

jmxfetch.py
```diff
@@ -460,6 +451,40 @@ def _get_path_to_jmxfetch(self):
...
+    if auto_conf:
+        path = confd_path + '/auto_conf'
+    else:
+        path = confd_path
+
+    for conf in glob.glob(os.path.join(path, '*.yaml')):
+        filename = os.path.basename(conf)
+        check_name = filename.split('.')[0]
+        if os.path.exists(conf):
+            f = open(conf)
```
sjenriquez: 👍

truthbk:
rabbitmq flake, yet again... tests are good (otherwise)! 👍

PR #3151, opened 2017-01-26
[dev] Added TLS auth for the builtin http_check.py

ipfaffy:
### What does this PR do?

This PR adds the ability to make a request with TLS client authentication if you present client-side certs.

### Motivation

I use this in my org and thought it might be useful for others.


irabinovitch:
@ipfaffy Thanks for creating a PR for this. This should be a very useful addition. Over the past few months we have been moving our integrations out of this repo and over to: DataDog/integrations-core/

Any chance you could close this PR and open it there?

The benefit of doing this is that starting in Datadog Agent 5.12 you'll be able to update/install checks independently of agent releases, so we'll be able to release enhances such as yours on a more frequent cadence.

Thank you for being a member of the Datadog community.

ipfaffy:
Sure, closing to move to DataDog/integrations-core.

PR #3157, opened 2017-01-30
[redis] Bump python lib bugfix version

olivielpeau:
So that it's consistent with the version in integrations-core


olivielpeau:
CI is failing because of the flaky `rabbitmq` test, going ahead and merging

PR #3158, opened 2017-01-30
[network] dont combine connection states

gmmeyer:
This supercedes #2856. It fixes the failing tests and puts the changes behind a flag.


conf.d/network.yaml.default
```diff
@@ -13,3 +13,6 @@ instances:
     # Optionally completely ignore any network interface
     # matching the given regex:
     # excluded_interface_re: my-network-interface.*
+
+    # Do not combine connection states
```
olivielpeau: Could you add more details on what the option does? And I think it's worth explicitly mentioning what the default value is.

gmmeyer: That's a good note, thank you.

gmmeyer:
@olivielpeau I can add some tests, however we'll have to add them in integrations core, the changes to the file structure forced me to change the tests for this one pretty substantially :(

gmmeyer:
I'll put it in the issues there

PR #3160, opened 2017-01-31
[kubernetes] make apiserver cacert configurable

hkaj:
### What does this PR do?

To verify the identity of the Kubernetes API server we use the CA cert passed in the agent pod service account by default. This PR makes it possible to pass a custom ca cert instead.

### Motivation

User request

### Testing Guidelines

Test updated accordingly


masci:
LGTM! 👍

PR #3164, opened 2017-02-01
[trace-agent] adding output of 'trace-agent -info'

ufoot:
### What does this PR do?

When calling `datadog-agent info`, this patch makes a subcall to `trace-agent -info` to show trace agent (APM) status.

### Motivation

It's easier to monitor the agent status this way, provides a quick way to know what's happening.

### Testing Guidelines

- Check that  `datadog-agent status` display relevant info.
- Check the return code of that command (should be 0 on  success, obviously)

### Additional Notes

- Wondering wether Debian support is enough or if all platforms should be done...
- What happens when the agent is not running, should we report an error (return code 1) or return 0 as APM is not part of the base, core install

*IMPORTANT* -> must **NOT** ship before: https://github.com/DataDog/datadog-trace-agent/pull/214

### DIFF
--- a/packaging/datadog-agent/source/agent
+++ b/packaging/datadog-agent/source/agent
@@ -121,10 +121,6 @@ case $action in
         RETURN_VALUE=$(($RETURN_VALUE || $?))
         python agent/ddagent.py info
         RETURN_VALUE=$(($RETURN_VALUE || $?))
-        if [ -x ./bin/trace-agent ]; then
-            ./bin/trace-agent -info
-            RETURN_VALUE=$(($RETURN_VALUE || $?))
-        fi
         exit $RETURN_VALUE
         ;;
 
--- a/packaging/datadog-agent/source/agent
+++ b/packaging/datadog-agent/source/agent
@@ -121,6 +121,10 @@ case $action in
         RETURN_VALUE=$(($RETURN_VALUE || $?))
         python agent/ddagent.py info
         RETURN_VALUE=$(($RETURN_VALUE || $?))
+        if [ -x ./bin/trace-agent ]; then
+            ./bin/trace-agent -info
+            RETURN_VALUE=$(($RETURN_VALUE || $?))
+        fi
         exit $RETURN_VALUE
         ;;
 
--- a/packaging/debian/datadog-agent.init
+++ b/packaging/debian/datadog-agent.init
@@ -19,6 +19,7 @@ DOGSTATSDPATH="/opt/datadog-agent/agent/dogstatsd.py"
 KILL_PATH="/opt/datadog-agent/embedded/bin/kill"
 AGENTUSER="dd-agent"
 FORWARDERPATH="/opt/datadog-agent/agent/ddagent.py"
+TRACEAGENTPATH="/opt/datadog-agent/bin/trace-agent"
 NAME="datadog-agent"
 DESC="Datadog Agent"
 SUPERVISOR_PIDFILE="/opt/datadog-agent/run/datadog-supervisord.pid"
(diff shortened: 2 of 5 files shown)


packaging/datadog-agent/source/agent
```diff
@@ -121,6 +121,10 @@ case $action in
         RETURN_VALUE=$(($RETURN_VALUE || $?))
         python agent/ddagent.py info
         RETURN_VALUE=$(($RETURN_VALUE || $?))
+        if [ -x ./bin/trace-agent ]; then
```
talwai: can skip this file, we don't ship with the source install

packaging/osx/datadog-agent
```diff
@@ -111,6 +112,10 @@ case $1 in
         RETURN_VALUE=$(($RETURN_VALUE || $?))
         $FORWARDERPATH info
         RETURN_VALUE=$(($RETURN_VALUE || $?))
+        if [ -x $TRACEAGENTPATH ]; then
```
talwai: can skip this file, we don't ship with osx

packaging/suse/datadog-agent.init
```diff
@@ -417,7 +418,12 @@ case "$1" in
         DOGSTATSD_RETURN=$?
         su $AGENTUSER -c "$FORWARDERPATH info"
         FORWARDER_RETURN=$?
-        exit $(($COLLECTOR_RETURN+$DOGSTATSD_RETURN+$FORWARDER_RETURN))
+        TRACEAGENT_RETURN=0
+        if [ -x $TRACEAGENTPATH ];
+           su $AGENTUSER -c "$TRACEAGENTPATH -info"
+           TRACEAGENT_RETURN=$?
+        fi
+        exit $(($COLLECTOR_RETURN+$DOGSTATSD_RETURN+$FORWARDER_RETURN+$TRACEAGENT_RETURN))
```
talwai: as mentioned in chat, i'd prefer if we don't include $TRACEAGENT_RETURN in the exit code. we mark it as an optional process, disabled by default. since `info` is sometimes used as a healthcheck - i think we should consider the agent "healthy" even when `trace-agent -info` exits non-zero. @olivielpeau any thoughts on this?

talwai:
missed a file: https://github.com/DataDog/dd-agent/blob/master/packaging/centos/datadog-agent.init

packaging/suse/datadog-agent.init
truthbk: The way I see it, the exit code should reflect trace agent's status if it's enabled and discard it/ignore it otherwise.

truthbk:
I'm not convinced the trace agent status should be ignored if it's indeed enabled. I'd like to hear @olivielpeau input on this one.

packaging/suse/datadog-agent.init
ufoot: Yes @truthbk really makes sense -> @talwai do we have a way to do that beyond parsing the conf for `apm_enable=true` & reading the env var `DD_APM_ENABLED` ?

talwai: > do we have a way to do that beyond parsing the conf for apm_enable=true & reading the env var DD_APM_ENABLED ?

this is the only way. i agree with @truthbk 's logic in principle if this is not too messy to implement

packaging/debian/datadog-agent.init
```diff
@@ -176,6 +177,11 @@ case "$1" in
         DOGSTATSD_RETURN=$?
         su $AGENTUSER -c "$FORWARDERPATH info"
         FORWARDER_RETURN=$?
+        TRACEAGENT_RETURN=0
+        if [ -x $TRACEAGENTPATH ];
```
olivielpeau: applies to the 2 other init files as well: I don't think we need to check that the trace agent binary exists and is executable, these init files are only used by the packaged Linux agent and it'll always ship the trace agent going forward, with correct exec permissions. Unless I'm missing something.

olivielpeau:
Also if the plan is to ship this for the `5.11.0` agent release please set the related milestone on the PR :)

PR #3172, opened 2017-02-07
[docker] fix image tag extraction

hkaj:
if you have not yet done so.*

### What does this PR do?

Fix `image_tag_extractor` in the case where `RepoTags` is empty.
Also add support for image name extraction from `RepoDigests`.

### Motivation

Working on an other PR I realized we were getting a warning:
```Failed to count Docker images. Exception: 'NoneType' object is not iterable```

### Additional Notes

Fix `test_tags_options`.

### DIFF
--- a/utils/dockerutil.py
+++ b/utils/dockerutil.py
@@ -382,7 +382,7 @@ class DockerUtil:
                 # the split will be like [repo_url, repo_port/image_name, image_tag]. Let's avoid that
                 split = [':'.join(split[:-1]), split[-1]]
             return [split[key]]
-        if "RepoTags" in entity:
+        if entity.get('RepoTags'):
             splits = [el.split(":") for el in entity["RepoTags"]]
             tags = set()
             for split in splits:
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: utils/dockerutil.py [hunk 1]
```diff
@@ -392,6 +392,14 @@ def image_tag_extractor(cls, entity, key):
                     tags.add(split[key])
             if len(tags) > 0:
                 return list(tags)
+        elif entity.get('RepoDigests'):
+            # the human-readable tag is not mentioned in RepoDigests
+            if key > 0:
```
hush-hush: so after that point if we didn't return, key is lower than 0 and we use it as an index. Seems wrong, no ?

utils/dockerutil.py
```diff
@@ -382,7 +382,7 @@ def image_tag_extractor(cls, entity, key):
                 # the split will be like [repo_url, repo_port/image_name, image_tag]. Let's avoid that
                 split = [':'.join(split[:-1]), split[-1]]
             return [split[key]]
-        if "RepoTags" in entity:
+        if entity.get('RepoTags'):
```
hush-hush: why get instead of `in` ?

hkaj: because it's always there but can be set to `{}` which is useless

utils/dockerutil.py
hkaj: key is lower or equal to 0 (it's `== 0` actually). I'll update to `!=` for clarity

PR #3174, opened 2017-02-08
[utils/hostname] better EC2 detection

degemer:
If the hostname is None, we should try to get the EC2 instance-id.
This was reported by a customer, and they proposed this fix.
The hostname can be None when `/bin/hostname -f` returns an error, which
can happen at AMI creation when the hosts file is not ready.`


degemer:
It's breaking backward compatibility on Windows for now.

degemer:
I tested this specific case, and we handle it "properly", no new host is created, the old one is used.

PR #3176, opened 2017-02-08
[service_disco] add replica_set, daemon_set & namespace tags for k8s

hkaj:
if you have not yet done so.*

### What does this PR do?

When service discovery is enabled and we're running in kubernetes, find the namespace, replica set and daemon set (when available) and add them as tags to all metrics.

### Motivation

Someone asked it to me, but I can't remember who :thinking: 

### Testing Guidelines

### Additional Notes

The replica set tag is open to discussion. Of course it's useful as is, but a later version will also add the deployment as a tag, and in the case of deployments, replica set is not stable. The deployment name is stable but it churns through replica sets at each update. So when we have the deployment thing working we'll want to remove the replica set tag when there's a deployment on top of the RS (and report the RS name as container metadata, but that's another story). So to avoid breaking stuff it might be better to leave it out to begin with.

### DIFF
--- a/utils/service_discovery/sd_docker_backend.py
+++ b/utils/service_discovery/sd_docker_backend.py
@@ -242,18 +242,37 @@ class SDDockerBackend(AbstractSDBackend):
                 log.warning("Failed to fetch pod metadata for container %s."
                             " Kubernetes tags may be missing." % c_id[:12])
                 return []
-            # get labels
+
+            # get pod labels
             kube_labels = pod_metadata.get('labels', {})
             for label, value in kube_labels.iteritems():
                 tags.append('%s:%s' % (label, value))
 
-            # get replication controller
-            created_by = json.loads(pod_metadata.get('annotations', {}).get('kubernetes.io/created-by', '{}'))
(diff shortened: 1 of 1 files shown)


hkaj:
requesting your review @remh mostly to have your opinion on what I wrote in `Additional Notes`

utils/service_discovery/sd_docker_backend.py
```diff
@@ -242,18 +242,33 @@ def get_tags(self, state, c_id):
...
+            tags.append('kube_namespace:%s' % namespace)
+
+            # get created-by
+            created_by = json.loads(pod_metadata.get('annotations', {}).get('kubernetes.io/created-by', '{}'))
+            creator_kind = created_by.get('reference', {}).get('kind')
+            creator_name = created_by.get('reference', {}).get('name')
+
+            # add creator tags
+            if creator_kind == 'ReplicationController':
+                tags.append('kube_replication_controller:%s' % creator_name)
```
hush-hush: `creator_name` may be `None`, we should check it if we don't want to create tag like `kube_replication_controller:None`

PR #3180, opened 2017-02-08
[core] Submitting metrics to the api endpoint

remh:
### What does this PR do?

This PR splits the collector payload into a `legacy` payload and a `metrics` payload.

It does so by extracting the metrics that are in the "new" format from the current payload, transforming them, a putting them into a new payload to be posted to the `api/v1/series` endpoint.

### Motivation

This will help up phasing out a backend service used by the old payload.

### Testing Guidelines

 - Added a unit test on the split function
 - Tested that this was a no op on my agent (i.e. all metrics keep flowing seamlessly)
 - This will need to be tested at higher scale as well during QA.


tests/core/test_emitter.py
```diff
@@ -2,17 +2,48 @@
...
+                attributes['hostname'] = s['host']
+            if s.get('tags'):
+                attributes['tags'] = s['tags']
+            if s.get('device'):
+                attributes['device_name'] = s['device']
+
+            formatted_sample = [s['metric'], s['points'][0][0], s['points'][0][1], attributes]
+            legacy_payload_split['metrics'].append(formatted_sample)
+
+        self.assertEqual(legacy_payload, legacy_payload_split)
```
yannmh: This will always be `True`, `split_payload` modifies the original payload by reference, in other words
```python
# &legacy_payload_split == &legacy_payload
legacy_payload_split, metrics_payload = split_payload(legacy_payload)
```

emitter.py
```diff
@@ -124,13 +115,57 @@ def http_emitter(message, log, agentConfig, endpoint):
...
+            if ts[3].get('hostname'):
+                sample['host'] = ts[3]['hostname']
+            if ts[3].get('tags'):
+                sample['tags'] = ts[3]['tags']
+            if ts[3].get('device_name'):
+                sample['device'] = ts[3]['device_name']
+
+        metrics_payload["series"].append(sample)
+
+    return legacy_payload, metrics_payload
```
yannmh: Naming nitpick: `legacy_payloads` refers to both the original input, and the modified/stripped output. Shall we use two different names?

emitter.py
```diff
@@ -124,13 +115,57 @@ def http_emitter(message, log, agentConfig, endpoint):
...
+        if len(ts) >= 4:
+            if ts[3].get('type'):
+                sample['type'] = ts[3]['type']
+            if ts[3].get('hostname'):
+                sample['host'] = ts[3]['hostname']
+            if ts[3].get('tags'):
+                sample['tags'] = ts[3]['tags']
+            if ts[3].get('device_name'):
+                sample['device'] = ts[3]['device_name']
+
```
yannmh: We are not handling any exception in this block. 
Considering that metrics are strongly formatted,  I believe it is "safe" to accept it and avoid a `try ... except ... continue` overhead. 

@olivielpeau any thoughts?

emitter.py
```diff
@@ -124,13 +115,57 @@ def http_emitter(message, log, agentConfig, endpoint):
...
         log.exception("Unable to post payload.")
-        try:
-            log.error("Received status code: {0}".format(r.status_code))
-        except Exception:
-            pass
 
 
-def post_headers(agentConfig, payload):
+def split_payload(legacy_payload):
+    metrics = list(legacy_payload['metrics'])
```
yannmh: Nitpick: the method performs reads only on `metrics`, thus, it is not required to make a copy of it.

emitter.py
```diff
@@ -124,13 +115,57 @@ def http_emitter(message, log, agentConfig, endpoint):
...
+
+        if len(ts) >= 4:
+            if ts[3].get('type'):
+                sample['type'] = ts[3]['type']
+            if ts[3].get('hostname'):
+                sample['host'] = ts[3]['hostname']
+            if ts[3].get('tags'):
+                sample['tags'] = ts[3]['tags']
+            if ts[3].get('device_name'):
+                sample['device'] = ts[3]['device_name']
```
yannmh: We could change those keys directly in the `MetricAggregator` to speed up.
```
sample.update(ts[3])
```

emitter.py
```diff
@@ -124,13 +115,57 @@ def http_emitter(message, log, agentConfig, endpoint):
...
+    metrics = list(legacy_payload['metrics'])
+    del legacy_payload['metrics']
+
+    metrics_payload = {"series": []}
+
+    # See https://github.com/DataDog/dd-agent/blob/5.11.1/checks/__init__.py#L905-L926 for format
+    for ts in metrics:
+        sample = {
+            "metric": ts[0],
+            "points": [[ts[1], ts[2]]]
```
olivielpeau: nit: could be a `tuple` instead of a `list`

emitter.py
olivielpeau: should `device_name` be changed to `device`? Dogstatsd uses `device_name` in its API payload (https://github.com/DataDog/dd-agent/blob/5.11.3/aggregator.py#L991)

emitter.py
olivielpeau: not handling exceptions here looks safe to me, the format is well-defined and the code is safe

tests/core/test_emitter.py
olivielpeau: yes, and so that things are clearer maybe `split_payload` should either return a `legacy_payload_split ` and not mutate the `legacy_payload` passed as argument _or_ not return a `legacy_payload_split` and mutate the `legacy_payload` (which probably makes more sense here)

emitter.py
remh: @yann i didn't want to change the formatter as i thought it was used by the legacy checks. Turns out it's not the case. I'll update this.

@olivielpeau what do you mean ? legacy payload processing will use `device_name`, new payload expects `device` hence my change.

emitter.py
remh: it doesn't matter and it's still a legacy payload

emitter.py
remh: indeed.

emitter.py
remh: Yes this code is safe.

emitter.py
remh: Actually i want to keep this logic kinda isolated as it's a hack. Better to keep it in a single place to keep the code cleaner. Agent 6 will get rid of all of this anyway.

tests/core/test_emitter.py
remh: good catch :man_facepalming:

remh: i'll just pass a copy of the payload to the test for 2 reasons.

1. Better to be explicit, if the function is mutating the dict, having it being returned is better than silently mutating it.
2. We could make a copy of the dict in the split payload function but this would come with a memory cost. 

So i'll just make a copy of the payload in the test function.
Good catch

remh:
Addressed some of your comments and fixed a few things. Let me know if there is other stuff but i think we're good now.

PR #3185, opened 2017-02-13
Update to kube-state-metrics 0.4 data

therc:
### What does this PR do?

It adds new metrics exported by kube-state-metrics 0.4+, see https://github.com/kubernetes/kube-state-metrics/blob/master/CHANGELOG.md

Rearrange methods to match order in kube-state-metrics' README
Add/fix comments
Add new daemonset metrics
Add rolling update metric for deployments
Add limits for containers
Add namespace resource quotas

As far as I can tell, kube-state-metrics hasn't moved to integrations-core

cc @hkaj

### DIFF
--- a/utils/kubernetes/kube_state_processor.py
+++ b/utils/kubernetes/kube_state_processor.py
@@ -72,6 +72,53 @@ class KubeStateProcessor:
                 return label.value
         return None
 
+    # All these methods try to follow the same order as https://github.com/kubernetes/kube-state-metrics/blob/master/README.md
+
+    # Labels attached: node, kernel_version, os_image, container_runtime_version, kubelet_version, kubeproxy_version
+    def kube_node_info(self, message, **kwargs):
+        """ Node information and versions. """
+        metric_name = NAMESPACE + '.node.info'
+        for metric in message.metric:
+            val = metric.gauge.value
+            tags = ['{}:{}'.format(label.name, label.value) for label in metric.label]
+            self.gauge(metric_name, val, tags)
+
+    def kube_node_spec_unschedulable(self, message, **kwargs):
+        """ Whether a node can schedule new pods. """
+        metric_name = NAMESPACE + '.node.status'
+        statuses = ('schedulable', 'unschedulable')
+        for metric in message.metric:
+            tags = ['{}:{}'.format(label.name, label.value) for label in metric.label]
+            status = statuses[int(metric.gauge.value)]  # value can be 0 or 1
+            tags.append('status:{}'.format(status))
+            self.gauge(metric_name, 1, tags)  # metric value is always one, value is on the tags
+
+    def kube_node_status_ready(self, message, **kwargs):
+        """ The ready status of a cluster node. """
+        service_check_name = NAMESPACE + '.node.ready'
+        for metric in message.metric:
+            name, val = self._eval_metric_condition(metric)
+            tags = ['node:{}'.format(self._extract_label_value("node", metric.label))]
+            if name == 'true' and val:
+                self.service_check(service_check_name, self.kube_check.OK, tags=tags)
+            elif name == 'false' and val:
+                self.service_check(service_check_name, self.kube_check.CRITICAL, tags=tags)
+            else:
+                self.service_check(service_check_name, self.kube_check.UNKNOWN, tags=tags)
+
+    def kube_node_status_out_of_disk(self, message, **kwargs):
+        """ Whether the node is out of disk space. """
+        service_check_name = NAMESPACE + '.node.out_of_disk'
+        for metric in message.metric:
+            name, val = self._eval_metric_condition(metric)
+            tags = ['node:{}'.format(self._extract_label_value("node", metric.label))]
+            if name == 'true' and val:
+                self.service_check(service_check_name, self.kube_check.CRITICAL, tags=tags)
+            elif name == 'false' and val:
+                self.service_check(service_check_name, self.kube_check.OK, tags=tags)
+            else:
+                self.service_check(service_check_name, self.kube_check.UNKNOWN, tags=tags)
+
     def kube_node_status_capacity_cpu_cores(self, message, **kwargs):
         """ The total CPU resources of the node. """
         metric_name = NAMESPACE + '.node.cpu_capacity'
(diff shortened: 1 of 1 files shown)


utils/kubernetes/kube_state_processor.py
```diff
@@ -228,38 +286,70 @@ def kube_pod_status_phase(self, message, **kwargs):
             status = self.pod_phase_to_status.get(phase, self.kube_check.UNKNOWN)
             self.service_check(check_basename + phase, status, tags=tags)
 
-    def kube_node_status_ready(self, message, **kwargs):
-        """ The ready status of a cluster node. """
-        service_check_name = NAMESPACE + '.node.ready'
+    # TODO: implement kube_pod_status_ready
```
ojongerius: Maybe refer to https://github.com/DataDog/dd-agent/issues/3045 if you planned to keep that open, or open one or more issues for this so you or others can pick it up.

utils/kubernetes/kube_state_processor.py
```diff
@@ -228,38 +286,70 @@ def kube_pod_status_phase(self, message, **kwargs):
...
         for metric in message.metric:
+            val = metric.gauge.value
             tags = ['{}:{}'.format(label.name, label.value) for label in metric.label]
-            status = statuses[int(metric.gauge.value)]  # value can be 0 or 1
-            tags.append('status:{}'.format(status))
-            self.gauge(metric_name, 1, tags)  # metric value is always one, value is on the tags
+            self.gauge(metric_name, val, tags)
+
+    # Labels attached: namespace, pod, container, node
+    def kube_pod_container_limits_cpu_cores(self, message, **kwargs):
```
ojongerius: I see quite a lot of duplication, how about a simple helper function, something like:

```python
def save(self, metric_name, metric):
    for metric in message.metric:
        val = metric.gauge.value
        tags = ['{}:{}'.format(label.name, label.value) for label in metric.label]
        self.gauge(metric_name, val, tags)
```

And we could change most of the functions to something like the following:

```python
def kube_pod_container_limits_cpu_cores(self, message, **kwargs):
      """ CPU cores limit for a container in a pod. """
      save(self, NAMESPACE + '.container.cpu_limit', message)
```

Generating code, something I've done for Go and configs but never for Python, could possibly be another way
to make this work a little less tedious and and error prone. But out of scope for a small inline comment in a code review :wink:

ojongerius:
Thanks for doing this work @therc!

FWIW Travis errors look, at first glance, unrelated to the changes.

utils/kubernetes/kube_state_processor.py
```diff
@@ -72,6 +72,53 @@ def _extract_label_value(self, name, labels):
                 return label.value
         return None
 
+    # All these methods try to follow the same order as github.com/kubernetes/kube-state-metrics/README.md
```
therc: Done

utils/kubernetes/kube_state_processor.py
```diff
@@ -72,6 +72,53 @@ def _extract_label_value(self, name, labels):
...
+        service_check_name = NAMESPACE + '.node.ready'
+        for metric in message.metric:
+            name, val = self._eval_metric_condition(metric)
+            tags = ['node:{}'.format(self._extract_label_value("node", metric.label))]
+            if name == 'true' and val:
+                self.service_check(service_check_name, self.kube_check.OK, tags=tags)
+            elif name == 'false' and val:
+                self.service_check(service_check_name, self.kube_check.CRITICAL, tags=tags)
+            elif name == 'unknown' and val:
+                self.service_check(service_check_name, self.kube_check.UNKNOWN, tags=tags)
```
therc: I didn't write the original code, I'm just moving it around (check the left side in the GitHub files view).

I agree with you that it makes more sense that way, but perhaps there's a reason why things are done that way.

utils/kubernetes/kube_state_processor.py
```diff
@@ -72,6 +72,53 @@ def _extract_label_value(self, name, labels):
...
+        """ Whether the node is out of disk space. """
+        service_check_name = NAMESPACE + '.node.out_of_disk'
+        for metric in message.metric:
+            name, val = self._eval_metric_condition(metric)
+            tags = ['node:{}'.format(self._extract_label_value("node", metric.label))]
+            if name == 'true' and val:
+                self.service_check(service_check_name, self.kube_check.CRITICAL, tags=tags)
+            elif name == 'false' and val:
+                self.service_check(service_check_name, self.kube_check.OK, tags=tags)
+            elif name == 'unknown' and val:
```
therc: Ditto.

utils/kubernetes/kube_state_processor.py
```diff
@@ -120,6 +167,43 @@ def kube_node_status_allocatable_pods(self, message, **kwargs):
             tags = ['node:{}'.format(self._extract_label_value("node", metric.label))]
             self.gauge(metric_name, val, tags)
 
+    # Labels attached: daemonset, namespace
+    def kube_daemonset_status_current_number_scheduled(self, message, **kwargs):
+        """ Number of nodes running at least 1 daemon pod that are supposed to run one. """
```
therc: It might be easier to parse, but is it English?

The number of nodes running at least one daemon pod and that are supposed to?
The number of nodes running at least one daemon pod which are supposed to?
The number of nodes that are running at least one daemon pod and are supposed to?

I think the last one might be the best.

utils/kubernetes/kube_state_processor.py
therc: Yeah, I am not sure why it wasn't done originally. I'm just repeating the style that was already there.

utils/kubernetes/kube_state_processor.py
ojongerius: Agreeing the last one is a good compromise.

utils/kubernetes/kube_state_processor.py
ojongerius: Ah ok. I guess you are just trying to get something done. Fair enough 😄

utils/kubernetes/kube_state_processor.py
ojongerius: I'm curious why that reason would be would be, maybe @hkaj can shed some light on it. If it's a repetition of an exisiting pattern, maybe out of scope of your work to fix it.

utils/kubernetes/kube_state_processor.py
```diff
@@ -72,6 +72,53 @@ def _extract_label_value(self, name, labels):
                 return label.value
         return None
 
+    # All these methods try to follow the same order as github.com/kubernetes/kube-state-metrics/README.md
+
+    # Labels attached: node, kernel_version, os_image, container_runtime_version, kubelet_version, kubeproxy_version
+    def kube_node_info(self, message, **kwargs):
+        """ Node information and versions. """
+        metric_name = NAMESPACE + '.node.info'
```
hkaj: FYI here's how we could report them:
```
- `container_runtime_version` --> metadata_payload['host-tags']
- `kernel_version` --> metadata_payload['host-tags']
- `kubelet_version` --> metadata_payload['host-tags']
- `kubeproxy_version` --> metadata_payload['host-tags']
- `node` --> metadata_payload['meta']['host_alias']
- `os_image` --> metadata_payload['host-tags']
```

utils/kubernetes/kube_state_processor.py
```diff
@@ -120,6 +167,43 @@ def kube_node_status_allocatable_pods(self, message, **kwargs):
...
+    def kube_daemonset_status_desired_number_scheduled(self, message, **kwargs):
+        """ Number of nodes that should be running a daemon pod. """
+        metric_name = NAMESPACE + '.daemonset.desired_scheduled'
+        for metric in message.metric:
+            val = metric.gauge.value
+            tags = ['{}:{}'.format(label.name, label.value) for label in metric.label]
+            self.gauge(metric_name, val, tags)
+
+    # Labels attached: daemonset, namespace
+    def kube_daemonset_metadata_generation(self, message, **kwargs):
```
hkaj: I don't think this one is interesting to collect, do you have a use for it in mind?

utils/kubernetes/kube_state_processor.py
```diff
@@ -152,63 +239,34 @@ def kube_deployment_spec_replicas(self, message, **kwargs):
...
-    # TODO: implement kube_pod_container_status_running
-    # TODO: implement kube_pod_container_status_terminated
-    # TODO: implement kube_pod_container_status_waiting
-
-    # Labels attached: namespace, pod, container
-    def kube_pod_container_status_restarts(self, message, **kwargs):
-        """ Number of desired pods for a deployment. """
-        metric_name = NAMESPACE + '.container.restarts'
+    # Labels attached: deployment, namespace
+    def kube_deployment_metadata_generation(self, message, **kwargs):
```
hkaj: same as the other `metadata_generation` one, do you have a use case where it could be useful? Otherwise let's not collect it.

jkurz:
This PR can be closed correct @hkaj? These metrics are currently in master and it has been refactored to implement them in a much cleaner way.

hkaj:
@jkurz correct, and it will be even more simplified soon with the [Prometheus check](https://github.com/DataDog/dd-agent/blob/bc6f2a6f897a6d66bf25ee483aaa8746d66c8b62/checks/prometheus_check.py)

Thanks nonetheless @therc !

PR #3190, opened 2017-02-13
[kubernetes] reduce log message from error to debug

pmahoney:
A PodSpec with containers with no explicit [ResourceRequirements](https://kubernetes.io/docs/resources-reference/v1.5/#resourcerequirements-v1) results in error logs from the collector:

    Unable to retrieve container requests for $container: 'requests'
    Unable to retrieve container requests for $container: 'requests'
    Unable to retrieve container requests for $container: 'requests'
    ...

### What does this PR do?

Reduce a noisy log from ERROR to DEBUG level.

### Motivation

This is a normal config; no reason to fill the log.


hkaj:
Thanks @pmahoney !

PR #3196, opened 2017-02-15
[mysql] use `information_schema` when < 5.6.0

yannmh:
The `performance_schema.threads` table is not available on MySQL <
`5.6.0`. As a result, retrieving the slaves' statuses triggers the
following error:

```
ProgrammingError: (1146, u"Table 'performance_schema.threads' doesn't
exist")
```

This is a regression introduced by
https://github.com/DataDog/dd-agent/pull/2603

### DIFF
--- a/checks.d/mysql.py
+++ b/checks.d/mysql.py
@@ -513,8 +513,8 @@ class MySql(AgentCheck):
             metrics.update(GALERA_VARS)
 
         performance_schema_enabled = self._get_variable_enabled(results, 'performance_schema')
-        if _is_affirmative(options.get('extra_performance_metrics', False)) and \
-                self._version_compatible(db, host, "5.6.0") and \
+        above_560 = self._version_compatible(db, host, "5.6.0")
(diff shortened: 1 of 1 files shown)


checks.d/mysql.py
```diff
@@ -863,12 +863,17 @@ def _get_replica_stats(self, db):
             self.warning("Privileges error getting replication status (must grant REPLICATION CLIENT): %s" % str(e))
             return {}
 
-    def _get_slave_status(self, db, nonblocking=True):
+    def _get_slave_status(self, db, above_560):
```
truthbk: should we keep the non-blocking optional parameter?

yannmh: Not sure what's the benefits in doing that. `_get_slave_status` is only called here and once.

truthbk: I'm not sure if that was added on a customer request, not specifically, but asking to make the check more performant. But you are correct, it's only holding the mutex once per check run.

truthbk: Your call.

truthbk: 👍

PR #3197, opened 2017-02-15
[core] add a DNS cache

truthbk:
### What does this PR do?
This PR adds a DNS cache to the agent, if disabled the agent just skips any local DNS resolution. Otherwise we've implemented a very simple cache that should allow us to reduce the number of DNS queries made.

### Motivation

A customer was complaining about too many DNS queries. We suggested some local caching like `dnsmasq`, apparently that was not an option for them. Unfortunately there was no way to do this using `requests`, which was the preferred approach.

### Testing Guidelines

### DIFF
--- a/ddagent.py
+++ b/ddagent.py
@@ -22,6 +22,7 @@ from Queue import Full, Queue
 from socket import error as socket_error, gaierror
 import sys
 import threading
+from urlparse import urlparse
 import zlib
 
 # For pickle & PID files, see issue 293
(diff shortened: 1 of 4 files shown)


ddagent.py
```diff
@@ -209,6 +212,8 @@ def __sizeof__(self):
 
     def get_url(self, endpoint, api_key):
         endpoint_base_url = get_url_endpoint(endpoint)
+        if self._application.agent_dns_caching:
+            endpoint_base_url = self._application.get_from_dns_cache(endpoint_base_url)
```
truthbk: This could mess up the SSL/TLS verification, which is domain name based.

truthbk:
Should we document the option in `datadog.conf`?

PR #3200, opened 2017-02-16
[openstack] support regexp for exclusion

### LINKED ISSUE #3198: openstack plugin does not accept wildcards for exclude_server_ids
This is a critical bug because at this moment it is impossible to filterout the servers from monitoring. 
We are interested only about monitoring the open stack project resource usages and not the running servers themselves because we are using them for continuous-integration testing, meaning that a server lives on average for only ~30 minutes and we really don't care about its status.

We do create/destroy probably more than 100 VMs per day which means that the datadog account would be spammed with monitoring machines that have a very short livespan and current config does not allow us to exclude them.

DataDog is recording GUIDs as hostnames making really impossible to filter them out. This bug is critical because without it we would not even be able to consider going to a paid plan.

If we would be able to add "*" to the the exclude_server_ids we would be able to solve this problem.

yannmh:
Support regular expressions for Network IDs exclusion.

Fix #3198

### DIFF
--- a/checks.d/openstack.py
+++ b/checks.d/openstack.py
@@ -5,6 +5,7 @@
 # stdlib
 from datetime import datetime, timedelta
 from urlparse import urljoin
+import re
 
 # project
 from checks import AgentCheck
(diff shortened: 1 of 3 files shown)


truthbk:
Please merge into the SDK repo.

yannmh:
Moved to https://github.com/DataDog/integrations-core/pull/181.

PR #3201, opened 2017-02-16
adds json to gemfile

gmmeyer:
httparty took json out of their gem, this will break very old versions of ruby (pre 1.9)

https://github.com/jnunemaker/httparty/commit/37dfb42382e4365fb79d0323cd531198121cfb81


gmmeyer:
nevermind there are substantial other problems with our setup on 1.8

PR #3202, opened 2017-02-16
[dockerutil] break singleton instantiation loop.

truthbk:
### What does this PR do?

There was an infinite instantiation loop for DockerUtil(). Since it implements the singleton pattern and `fetch_swarm_state()` is in the constructor, if `fetch_swarm_state()` calls DockerUtil() again, it will call the constructor again (and so forth) and the class instance will never make it to the Singleton class map.

### Motivation

Found the issue while testing another PR branched off `master` that included the `swarm` changes. 

### Testing Guidelines

### Additional Notes

### DIFF
--- a/utils/dockerutil.py
+++ b/utils/dockerutil.py
@@ -135,7 +135,7 @@ class DockerUtil:
     def fetch_swarm_state(self):
         self.swarm_node_state = None
         try:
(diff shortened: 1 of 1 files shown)


hkaj:
oops, thanks! LGTM

PR #3203, opened 2017-02-16
[core][sdk] ensure manifest validation when running checks.

truthbk:
### What does this PR do?

Let's validate the SDK checks to ensure we can detect checks with the wrong manifests.

### Motivation

We might need this to warn/default to core checks if the core agent is updated from under an SDK installed check. Otherwise checks might break and we wouldn't have anything meaningful to debug in that aspect.

### Testing Guidelines

### DIFF
--- a/config.py
+++ b/config.py
@@ -904,7 +904,7 @@ def get_checks_places(osname, agentConfig):
 
     try:
         if Platform.is_windows():
-            places.append(get_windows_sdk_check)
+            places.append(lambda name: (get_windows_sdk_check(name), None))
         else:
             sdk_integrations = get_sdk_integrations_path(osname)
             places.append(lambda name: (os.path.join(sdk_integrations, name, 'check.py'),
--- a/config.py
+++ b/config.py
@@ -22,6 +22,9 @@ import sys
 import traceback
 from urlparse import urlparse
 
+# 3p
+import simplejson as json
+
 # project
 from util import check_yaml, config_to_yaml
 from utils.platform import Platform, get_os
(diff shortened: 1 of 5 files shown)


config.py
```diff
@@ -962,19 +973,52 @@ def _update_python_path(check_config):
...
             continue
 
         check_is_valid, check_class, load_failure = get_valid_check_class(check_name, check_path)
         if not check_is_valid:
             continue
 
+        if manifest_path:
+            validated = validate_sdk_check(manifest_path)
+            if not validated:
+                log.warn("The SDK check (%s) was designed for a different agent core "
```
truthbk: This has been implemented as a best-effort. If the check runs, it runs. Alternatively we could just `continue` and fallback to the `core` bundled check...

truthbk:
Windows appveyor issue fixed here: https://github.com/DataDog/dd-agent/pull/3230

config.py
```diff
@@ -890,18 +900,19 @@ def get_checks_places(osname, agentConfig):
         log.error(e.args[0])
         sys.exit(3)
 
-    places = [lambda name: os.path.join(agentConfig['additional_checksd'], '%s.py' % name)]
+    places = [lambda name: (os.path.join(agentConfig['additional_checksd'], '%s.py' % name), None)]
 
     try:
         if Platform.is_windows():
-            places.append(get_windows_sdk_check)
+            places.append(lambda name: (get_windows_sdk_check(name), None))
```
gmmeyer: You should add this to the get_windows_sdk_check function. It should be relatively easy to return the tuple there as well.

config.py
```diff
@@ -965,12 +976,39 @@ def _update_python_path(check_config):
...
+
+            for minfield in MANIFEST_VALIDATION['min']:
+                min_version = manifest.get(minfield)
+                if not min_version:
+                    continue
+
+                min_validated = False if min_version > get_version() else True
+                break
+    except IOError:
+        log.warn("Manifest file (%s) not present " % manifest_path)
```
gmmeyer: I think it's unnecessary to get two warnings in a row. Can you add this to the warning below instead of here? 

This might be valuable as a debug log line, but I don't think you need it as a warning.

Is an IOError the only error you can get here? I think you will also want to catch json parsing errors.

config.py
truthbk: Good catch.

PR #3205, opened 2017-02-17
[docker] bogus pid

truthbk:
### What does this PR do?

This PR attempts to address the case where we either have a bogus container pid reported, or no pid was collected, in which case the container must be skipped. 

### Motivation

Issue has been affecting some customers.

### Testing Guidelines

Further testing must be conducted to reproduce this reliably, I have been unable to so far. But then again I haven't had the time to dedicate a solid half-day to reproducing.

### DIFF
--- a/checks.d/docker_daemon.py
+++ b/checks.d/docker_daemon.py
@@ -134,6 +134,10 @@ def compile_filter_rules(rules):
     return patterns, tag_names
 
 
+class BogusPIDException(Exception):
+    pass
+
+
 class DockerDaemon(AgentCheck):
     """Collect metrics and events from Docker API and cgroups."""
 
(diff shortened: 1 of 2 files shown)


checks.d/docker_daemon.py
```diff
@@ -575,11 +575,14 @@ def _report_performance_metrics(self, containers_by_id):
...
-                continue
-            self._report_net_metrics(container, tags)
+            try:
+                self._report_cgroup_metrics(container, tags)
+                if "_proc_root" not in container:
+                    containers_without_proc_root.append(DockerUtil.container_name_extractor(container)[0])
+                    continue
+                self._report_net_metrics(container, tags)
+            except BogusPIDException:
+                self.log.exception('Unable to report cgroup metrics.')
```
hkaj: This is still going to fill the logs with tracebacks. Maybe `log.error` is enough? With the container id or name for reference.

checks.d/docker_daemon.py
```diff
@@ -592,6 +595,9 @@ def _report_performance_metrics(self, containers_by_id):
 
     def _report_cgroup_metrics(self, container, tags):
         cgroup_stat_file_failures = 0
+        if not container.get('_pid'):
+            raise BogusPIDException('Cannot report on bogus pid(0)')
```
hkaj: does that cover `0`-ed PIDs?

truthbk: it does

checks.d/docker_daemon.py
truthbk: sounds good, I'll update that.

truthbk:
closing this one, opened here:

PR #3214, opened 2017-02-23
[travis][ci] launch core and extras if CI run successful

truthbk:
### What does this PR do?

This PR will run the SDK repos CI if we have a successful test run (maybe it should run always regardless).

### Motivation

During the SDK transition we have moved the checks to the [core](https://github.com/DataDog/integrations-core) and [extras](https://github.com/DataDog/integrations-extras) repositories. These are somewhat coupled to `dd-agent`, so we must ensure we launch the tests in their respective CI's after changes are made here.

### Testing Guidelines

### Additional Notes


.travis.yml
```diff
@@ -43,9 +37,15 @@ script:
...
   - mkdir -p $INTEGRATIONS_DIR
   - ls -al $INTEGRATIONS_DIR
-  - 'rake ci:run'
+  - 'rake ci:run[default]'
+  - 'rake ci:run[core_integration]'
+  - 'rake ci:run[checks_mock]'
+  - 'rake ci:run[system]'
   - ls -al $INTEGRATIONS_DIR
 
+after_success:
```
truthbk: should we do this after every run (failed or not)? This should not run with PR's so it's not something that will clobber our CI in the SDK repos (and CI activity in the `dd-agent` should come way down anyway, so it shouldn't be too bad).

degemer:
I like the idea behind the PR, however I think we should to do it another way.
I see 3 possible cases:
- I open a PR and I know it's potentially breaking core/extra, then I'll check by myself.
- I open a PR that doesn't break core/extra, then nothing to do.
- (the case this PR was made for) I open a PR not knowing that it's going to break core/extra. CI is ✅ , and I won't check core/extra CI. CI is 🔴  on core/extra master, but we don't check that very often. A new PR is created on core/extra, CI is 🔴 , we have a look and we determine (hopefully easily?) it's my previous PR. 

Pros:
- we can check easily when core/extra broke and track it down to dd-agent (assuming tests are not too flaky).
Cons:
- it slows the dd-agent CI (by ~2min)
- it slows the entire CI (by triggering a lot more integrations-core builds); as you said the volume of dd-agent changes should go way down, so it might not be a real issue.

I think that being really careful about changes in `utils/` could be enough (and catching the rest with usual CI on core/extra), and that we don't have a lot to win by triggering a CI that doesn't fail the CI on PRs.

Let me know what you think, I don't really have a solution. Maybe use the new Travis cron feature daily - https://travis-ci.org/DataDog/integrations-core/settings ?

PR #3215, opened 2017-02-23
[integrations] [core] removes final checks and tests

gmmeyer:
### What does this PR do?

Removes the last checks from the agent.

### Motivation

Now that the other checks have been moved out, this will remove them from the agent and remove their associated config and tests.

### Testing Guidelines

This removes a bunch of tests, making testing easier and quicker!

### Additional Notes

This needs to be gone over with some care, to ensure I didn't remove anything important, but I do not believe I did.


tests/checks/common.py
```diff
@@ -33,6 +33,10 @@
 AUTO_CONFD_PATH = os.path.join(os.path.dirname(os.path.realpath(__file__)), '..', '..', 'conf.d', 'auto_conf')
 
 def copy_checks():
+    if not os.path.exists(CHECKSD_PATH):
```
truthbk: I saw the convo the other day regarding this. We should maybe add a card or a repo issue or something to not forget tweaking this and avoid having to copy/remove the checks.

gmmeyer: I think it's reasonable to do something else now that the folders no longer exist, though when it did it made sense to keep doing this.

We should at least look into it, it's not a bad idea. Let's make a card.

PR #3220, opened 2017-02-24
[jmxfetch] prepare source script for post jmxfetch repo bundle.

truthbk:
if you have not yet done so.*

### What does this PR do?

Removes jmxfetch from the `dd-agent` repo. It fixes the source install to pull the jmxfetch jarfile from the relevant location.

### Motivation

We're probably going to stop bundling jmxfetch in the `dd-agent` repository, it doesn't make much sense to keep a binary artifact here and we already have the jmxfetch repository with the sources. 

### Testing Guidelines

### Additional Notes


olivielpeau:
Quick note: we should probably also pull/update jmxfetch with a rake command for the dev environment and the CI

packaging/datadog-agent/source/setup_agent.sh
```diff
@@ -389,18 +403,21 @@ tar -xz -C "$DD_HOME/agent" --strip-components 1 -f "$DD_HOME/agent.tar.gz"
 rm -f "$DD_HOME/agent.tar.gz"
 print_done
 
+# get the version from the actual config file in the branch
+AGENT_VERSION=$(cat $DD_HOME/agent/config.py | grep AGENT_VERSION | cut -f2 -d'=')
```
olivielpeau: `grep AGENT_VERSION` returns 2 lines here, we should probably use `grep "AGENT_VERSION ="` instead (it's a bit brittle, if for some reason we use `AGENT_VERSION=` in `config.py` in the future then the grep would fail, so it might be worth adding a comment in `config.py`...)

We could probably do the same thing for `JMX_VERSION` to be on the safe side.

Another solution: use `python`  to import `config.py` and print out these 2 versions. Probably safer.

olivielpeau: One more thing: let's use a different variable name than `AGENT_VERSION` to avoid overwriting the existing var :)

packaging/datadog-agent/source/setup_agent.sh
```diff
@@ -389,18 +403,21 @@ tar -xz -C "$DD_HOME/agent" --strip-components 1 -f "$DD_HOME/agent.tar.gz"
...
-if [ "$AGENT_MAJOR_VERSION" -eq "5" -a "$AGENT_MINOR_VERSION" -gt "11" ]; then
+if check_version $PRE_SDK_RELEASE $AGENT_VERSION;
+then
   print_console "* Downloading integrations from GitHub"
   mkdir -p "$DD_HOME/integrations"
   mkdir -p "$DD_HOME/agent/checks.d"
   mkdir -p "$DD_HOME/agent/conf.d/auto_conf"
 
-  $DOWNLOADER "$DD_HOME/integrations.tar.gz" "https://api.github.com/repos/DataDog/integrations-core/tarball/$AGENT_VERSION"
+  $DOWNLOADER "$DD_HOME/integrations.tar.gz" "https://api.github.com/repos/DataDog/integrations-core/tarball/$INTEGRATIONS_VERSION"
```
olivielpeau: 🍰

packaging/datadog-agent/source/setup_agent.sh
```diff
@@ -436,6 +453,15 @@ $DOWNLOADER "$DD_HOME/requirements-opt.txt" "$BASE_GITHUB_URL/requirements-opt.t
 "$DD_HOME/agent/utils/pip-allow-failures.sh" "$DD_HOME/requirements-opt.txt"
 print_done
 
+if check_version $LAST_JMXFETCH_BUNDLE_RELEASE $AGENT_VERSION;
+then
+    print_console "* Trying to install JMXFetch jarfile from $JMXFETCH_URL"
+    mkdir -p "$DD_HOME/libs"
```
olivielpeau: I _think_ the correct path is `$DD_HOME/agent/checks/libs`

packaging/datadog-agent/source/setup_agent.sh
```diff
@@ -436,6 +453,15 @@ $DOWNLOADER "$DD_HOME/requirements-opt.txt" "$BASE_GITHUB_URL/requirements-opt.t
...
+    JMX_VERSION=$(cat $DD_HOME/config.py | grep JMX_VERSION | cut -f2 -d'=')
+    $DOWNLOADER "$DD_HOME/libs/" "$JMXFETCH_URL/jmxfetch-$(JMX_VERSION)-jar-with-dependencies.jar"
```
olivielpeau: `$(JMX_VERSION)` -> `$JMX_VERSION` no?

packaging/datadog-agent/source/setup_agent.sh
```diff
@@ -214,6 +217,16 @@ and we'll do our very best to help you solve your problem."
     fi
 }
 
+# Allows us to compare versions
+# Returns 0 if the second arg version is > than the first.
+check_version()
+{
+    local version=$1 check=$2
+    local winner=$(echo -e "$version\n$check" | sort -t '.' -k1,1 -k2,2 -k3,3 | head -n 1)
```
truthbk: forgot the `-n` switch. It's working now....

```
vagrant@vagrant:~$ ./version.sh 5.11.1 5.10.99; echo $?
1
vagrant@vagrant:~$ ./version.sh 5.9.99 5.10.0; echo $?
0
vagrant@vagrant:~$ ./version.sh 5.9.99 5.10.0; echo $?
0
```

packaging/datadog-agent/source/setup_agent.sh
truthbk: Wrong braces, meant to have `{}` for clarity.

packaging/datadog-agent/source/setup_agent.sh
truthbk: That's correct.

packaging/datadog-agent/source/setup_agent.sh
truthbk: I can do that, but that was the whole point: overwriting the version... with what's provided in `config.py`.

Say you want to try your feature branch: `AGENT_VERSION="oli/foo"`, as it stands, it'll pull the right tarball, and then grab the version from config.py. If you set `5.13.2` it'll match, so you'll overwrite it with the same version. If you set `5.13.x`, that'll work too, and will set the version for the remainder of the script to the version in config.py.

olivielpeau: hmm I understand the idea but to me `AGENT_VERSION` should always be the version that the user requested no?

The only places where we need a "resolved" agent version (i.e. the one in `config.py`) instead of the one the user requested is when we compare the version with known versions to determine if/how to install jmxfetch/integrations.

Further, if I'm using `AGENT_VERSION="oli/foo"`, I'd expect `INTEGRATIONS_VERSION` to default to `"oli/foo"` too, not to the tag from `config.py`.

I don't really mind either way, this is really a nitpick, feel free to ignore

packaging/datadog-agent/source/setup_agent.sh
```diff
@@ -389,18 +413,22 @@ tar -xz -C "$DD_HOME/agent" --strip-components 1 -f "$DD_HOME/agent.tar.gz"
 rm -f "$DD_HOME/agent.tar.gz"
 print_done
 
+# get the version from the actual config file in the branch
+AGENT_VERSION=$(get_from_py_config AGENT_VERSION)
+INTEGRATIONS_VERSION=${INTEGRATIONS_VERSION:-$AGENT_VERSION}
```
olivielpeau: could you document `INTEGRATIONS_VERSION` in this comment section? https://github.com/DataDog/dd-agent/blob/89cb32acb09419df38dfa16339f9986d322696ee/packaging/datadog-agent/source/setup_agent.sh#L23-L45 (and document which default version is set when it's not specified :) )

PR #3221, opened 2017-02-27
[kubernetes] - Do not check for kubernetes auth token secret if we have TLS certs configured

chen-anders:
### What does this PR do?

This PR makes it so we don't check for the kubernetes auth_token if TLS certs are already provided for authentication to the kube-apiserver (when the dd-agent is running on the host). The pre-existing dd-agent should work, but would log noisy "Token not found at /var/run/secrets/kubernetes.io/serviceaccount/token" errors to the logs. 

### Motivation

What inspired you to submit this pull request? Less noisy logs :)

### DIFF
--- a/utils/kubernetes/kubeutil.py
+++ b/utils/kubernetes/kubeutil.py
@@ -94,9 +94,10 @@ class KubeUtil:
         if apiserver_cacert and os.path.exists(apiserver_cacert):
             tls_settings['apiserver_cacert'] = apiserver_cacert
 
-        token = self.get_auth_token()
-        if token:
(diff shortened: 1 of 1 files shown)


xvello:
Hello @chen-anders,

This commit https://github.com/DataDog/dd-agent/pull/3401/commits/8711e643b2eeeb17575742ee8fcfec7d8dcc4ed6 from https://github.com/DataDog/dd-agent/pull/3401 should fix your use case and remove the log.

Can you please confirm you are happy with this and we can close your PR once this is merged into master?


Regards

chen-anders:
Works for me. Thanks @xvello

PR #3230, opened 2017-02-28
[info+flare] include SDK-relevant information if available

truthbk:
### What does this PR do?

This PR adds certain SDK-related facilities/improvements to the `flare` and `CheckStatus`, and as a result to the `info` page.

It also fixes several issues encountered with the testing infrastructure including build issues on `travis` and `appveyor` (also local).

I have also snuck in a fix to the SNMP flare. In particular, the `community_string` should be considered a credential and as such should be redacted.

### Motivation

With the advent of the SDK we need to be able to track what integration versions users have installed. This will allow them and us to understand what code actually got loaded, and what versions of the checks they're running at a glance.

### Testing Guidelines

### Additional Notes

### DIFF
--- a/checks/__init__.py
+++ b/checks/__init__.py
@@ -10,6 +10,7 @@ The Check class is being deprecated so don't write new checks with it.
 # stdlib
 from collections import defaultdict
 import copy
+import inspect
 import logging
 import numbers
 import os
@@ -29,11 +30,13 @@ import yaml
 
 # project
 from checks import check_status
+from config import AGENT_VERSION
 from util import get_next_id, yLoader
 from utils.hostname import get_hostname
 from utils.proxy import get_proxy
 from utils.platform import Platform
 from utils.profile import pretty_statistics
+from utils.sdk import load_manifest
 if Platform.is_windows():
     from utils.debug import run_check  # noqa - windows debug purpose
 
@@ -43,6 +46,9 @@ log = logging.getLogger(__name__)
 DEFAULT_PSUTIL_METHODS = ['memory_info', 'io_counters']
 
 AGENT_METRICS_CHECK_NAME = 'agent_metrics'
+
+AGENT_CHECK_FRAME = 2
+NETWORK_CHECK_FRAME = 3
 
 
 # Konstants
@@ -350,11 +356,14 @@ class AgentCheck(object):
         self.service_checks = []
         self.instances = instances or []
         self.warnings = []
+        self.check_version = None
         self.library_versions = None
         self.last_collection_time = defaultdict(int)
         self._instance_metadata = []
         self.svc_metadata = []
         self.historate_dict = {}
+
+        self.set_check_version(manifest=self._get_check_manifest())
 
         # Set proxy settings
         self.proxy_settings = get_proxy(self.agentConfig)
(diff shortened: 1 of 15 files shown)


truthbk:
Note: we might need to tweak a couple of things on windows (in particular, an assumption is being made for the location of manifest which I believe does not hold true for windows).

checks/__init__.py
```diff
@@ -375,6 +384,36 @@ def __init__(self, name, init_config, agentConfig, instances=None):
...
+        Do not use outside of the constructor if called without args, it assumes the following call chain:
+            CheckClass.__init__() -> AgentCheck.__init__() -> AgentCheck._get_check_manifest()
+        """
+        from checks.network_checks import NetworkCheck
+        if not manifest_path:
+            frames = inspect.stack()
+            frame_idx = NETWORK_CHECK_FRAME if isinstance(self, NetworkCheck) else AGENT_CHECK_FRAME
+            caller_frame = frames[frame_idx]  # frame_idx _should_ be the check __init__ frame
+            module_path = os.path.dirname(caller_frame[1])  # idx 1 contains the filename.
+            manifest_path = os.path.join(module_path, 'manifest.json')
```
truthbk: will this hold for `windows`? I think so, but I'd like someone to confirm.

utils/flare.py
```diff
@@ -499,6 +525,15 @@ def _get_path_agent_exec(self):
             )
         return agent_exec
 
+    # Find SDK integrations path
+    def _get_sdk_integrations_path(self):
```
truthbk: We'll need this to play well on `windows`. See #3231.

checks/__init__.py
gmmeyer: I'm not sure, that's a good question, you should test it on windows

gmmeyer: Is this tested on windows?

utils/sdk.py
```diff
@@ -0,0 +1,27 @@
...
+# 3p
+import simplejson as json
+
+# project
+
+
+log = logging.getLogger(__name__)
+
+
+def load_manifest(path):
```
gmmeyer: This should catch json parsing errors, too.

config.py
```diff
@@ -195,6 +196,11 @@ def _confd_path(directory):
 
 
 def _checksd_path(directory):
+    path_override = os.environ.get('CHECKSD_OVERRIDE')
```
gmmeyer: 🍰 this is a good solution, thank you.

checks/__init__.py
```diff
@@ -375,6 +384,36 @@ def __init__(self, name, init_config, agentConfig, instances=None):
...
+        return load_manifest(manifest_path)
+
+    def set_check_version(self, manifest=None):
+        version = AGENT_VERSION
+
+        if manifest:
+            try:
+                version = "{core}:{sdk}".format(core=AGENT_VERSION,
+                                            sdk=manifest['version'])
+            except Exception:
```
gmmeyer: Can you specify this a bit more than Exception?

checks/__init__.py
```diff
@@ -375,6 +384,36 @@ def __init__(self, name, init_config, agentConfig, instances=None):
...
+                self.log.info('Invalid manifest - unable to extract check version.')
+                version = 'Unknown'
```
gmmeyer: Why set it to unknown? Won't you usually be able to set it to `AGENT_VERSION`? 

I don't know how many errors are possible, but that would usually work, wouldn't it?

checks/__init__.py
truthbk: Yup, it worked 👍

checks/__init__.py
truthbk: Well, because `AGENT_VERSION` represents the core agent shipped checks. That wouldn't be the case here, it would be an SDK check with an unknown version.

checks/__init__.py
```diff
@@ -350,12 +356,15 @@ def __init__(self, name, init_config, agentConfig, instances=None):
...
         self.instances = instances or []
         self.warnings = []
+        self.check_version = None
         self.library_versions = None
         self.last_collection_time = defaultdict(int)
         self._instance_metadata = []
         self.svc_metadata = []
         self.historate_dict = {}
 
+        self.set_check_version(manifest=self._get_check_manifest())
```
truthbk: I don't think this would fail, if `load_manifest()` fails to be able to load the manifest, it will return `None`. I don't think the `inspect` calls or `os.path` calls will error out.

checks/__init__.py
truthbk: I can set this to {core}:unknown

utils/sdk.py
truthbk: It does, `except Exception` was intentionally not very specific.... Geared towards `IOErrors` and `JSONDecodeError`.

utils/windows_configuration.py
```diff
@@ -32,17 +32,49 @@ def get_registry_conf(agentConfig):
 
     return registry_conf
 
+def get_sdk_integration_path(hkey, reg_path):
```
truthbk: Mainly, for flexibility... you might call this as hkey being the key to `_winreg.HKEY_LOCAL_MACHINE` and have `reg_path` be `foo.bar`, or you could have hkey be the key to `_winreg.HKEY_LOCAL_MACHINE.foo` and reg_path=`bar`. I didn't see a reason to be more specific then I needed to.

gmmeyer: makes sense

utils/sdk.py
gmmeyer: okay that's good

checks/__init__.py
gmmeyer: You're probably right! Thats fine

checks/__init__.py
```diff
@@ -375,6 +384,36 @@ def __init__(self, name, init_config, agentConfig, instances=None):
...
+    def _get_check_manifest(self, manifest_path=None):
+        """
+        Attempts to collect the manifest for SDK checks.
+
+        Do not use outside of the constructor if called without args, it assumes the following call chain:
+            CheckClass.__init__() -> AgentCheck.__init__() -> AgentCheck._get_check_manifest()
+        """
+        from checks.network_checks import NetworkCheck
+        if not manifest_path:
+            frames = inspect.stack()
```
truthbk: Added some comments, let me know what you think.

utils/sdk.py
truthbk: I made the exception handling a little more specific.

PR #3231, opened 2017-03-01
[core] adds logic for windows sdk check loading

gmmeyer:
On Windows, the integrations installer will allow them to place an integration anywhere they want, and it will write that location in the registry.

This forces our sdk loader to act a differently than otherwise, it grabs the key from the registry, then it loads the check from it.

### DIFF
--- a/config.py
+++ b/config.py
@@ -33,7 +33,7 @@ from utils.subprocess_output import (
     get_subprocess_output,
     SubprocessOutputEmptyError,
 )
-from utils.windows_configuration import get_registry_conf
+from utils.windows_configuration import get_registry_conf, get_windows_sdk_check
 
 
 # CONSTANTS
@@ -893,8 +893,11 @@ def get_checks_places(osname, agentConfig):
     places = [lambda name: os.path.join(agentConfig['additional_checksd'], '%s.py' % name)]
 
     try:
-        sdk_integrations = get_sdk_integrations_path(osname)
-        places.append(lambda name: os.path.join(sdk_integrations, name, 'check.py'))
+        if Platform.is_windows():
+            places.append(get_windows_sdk_check)
+        else:
+            sdk_integrations = get_sdk_integrations_path(osname)
+            places.append(lambda name: os.path.join(sdk_integrations, name, 'check.py'))
     except PathNotFound:
         log.debug('No sdk integrations path found')
 
@@ -968,6 +971,9 @@ def load_check_from_places(check_config, check_name, checks_places, agentConfig)
     load_success, load_failure = {}, {}
     for check_path_builder in checks_places:
         check_path = check_path_builder(check_name)
+        # The windows SDK function will return None, so we should also skip 
+        if not check_path:
+            continue
         if not os.path.exists(check_path):
             continue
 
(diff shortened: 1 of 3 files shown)


config.py
```diff
@@ -968,6 +971,10 @@ def load_check_from_places(check_config, check_name, checks_places, agentConfig)
     load_success, load_failure = {}, {}
     for check_path_builder in checks_places:
         check_path = check_path_builder(check_name)
+        # The windows SDK function will return None,
+        # so the loop should also continue if there is no path.
+        if not check_path:
+            continue
```
truthbk: maybe:
```
if not (check_path and os.path.exists(check_path)):
    continue
```
in a single line?

utils/windows_configuration.py
```diff
@@ -31,6 +31,15 @@ def get_registry_conf(agentConfig):
 
     return registry_conf
 
+def get_windows_sdk_check(name):
+    sdk_reg_path = WINDOWS_REG_PATH + "\\Integrations\\" + name
+    try:
+        with _winreg.OpenKey(_winreg.HKEY_LOCAL_MACHINE,sdk_reg_path) as reg_key:
+            directory = _winreg.QueryValueEx(reg_key, "InstallDir")
+            return os.path.join(directory, 'check.py')
+    except WindowsError:
```
truthbk: we might want to log something if this happens?

utils/windows_configuration.py
```diff
@@ -31,6 +31,15 @@ def get_registry_conf(agentConfig):
 
     return registry_conf
 
+def get_windows_sdk_check(name):
+    sdk_reg_path = "SOFTWARE\\Datadog\\Datadog Agent\\Integrations\\{}".format(name)
```
truthbk: Can we put `SOFTWARE\\Datadog\\Datadog Agent` in a constant?

utils/windows_configuration.py
```diff
@@ -31,6 +31,15 @@ def get_registry_conf(agentConfig):
...
+    try:
+        with _winreg.OpenKey(_winreg.HKEY_LOCAL_MACHINE,sdk_reg_path) as reg_key:
+            directory = _winreg.QueryValueEx(reg_key, "InstallPath")[0]
+            return os.path.join(directory, 'Integrations', name, 'check.py')
+    except WindowsError as e:
```
truthbk: Will this only happen if the key doesn't exist?

utils/windows_configuration.py
gmmeyer: Yes, it is in one already actually, I copied it while debugging

config.py
gmmeyer: that makes more sense

utils/windows_configuration.py
gmmeyer: I know that's one time it does, I don't know about only. @derekwbrown?

derekwbrown: that's the only time I would expect; others would be lack of proper permissions (which shouldn't happen because we're running as system), things like that.  If the OpenKey() fails, there's not much we can do but move on, anyway.  IMO this is fine.

utils/windows_configuration.py
```diff
@@ -31,6 +31,15 @@ def get_registry_conf(agentConfig):
 
     return registry_conf
 
+def get_windows_sdk_check(name):
+    sdk_reg_path = "SOFTWARE\\Datadog\\Datadog Agent\\Integrations\\{}".format(name)
+    try:
+        with _winreg.OpenKey(_winreg.HKEY_LOCAL_MACHINE,sdk_reg_path) as reg_key:
```
degemer: Nit: space after `,`

PR #3237, opened 2017-03-06
add metrics from kube-state-metrics v0.4.0 to kube_state_processor.py

sophaskins:
### What does this PR do?

Adds support for new `kube-state-metrics` metrics from version 0.4.0. I've uncommented the following metrics:

- kube_pod_container_limits_cpu_cores
- kube_pod_container_limits_memory_bytes

and written the following new ones:

- kube_daemonset_status_current_number_scheduled
- kube_daemonset_status_desired_number_scheduled
- kube_daemonset_status_number_misscheduled
- kube_deployment_spec_paused
- kube_deployment_spec_strategy_rollingupdate_max_unavailable
- kube_deployment_status_replicas
- kube_node_status_phase
- kube_pod_container_status_ready
- kube_pod_container_status_running
- kube_pod_container_status_terminated
- kube_pod_container_status_waiting
- kube_pod_status_ready
- kube_pod_status_scheduled
- kube_replicaset_spec_replicas
- kube_replicaset_status_fully_labeled_replicas
- kube_replicaset_status_ready_replicas
- kube_replicaset_status_replicas

### Motivation

I wanted to use the pod-state metrics in my own Kubernetes installation, but they weren't present in the Datadog integration.

### Testing Guidelines

This file doesn't have an extant test - if we want one, I think refactoring the file is probably in order.

### Additional Notes

I'd be totally willing to re-work this as a PR to de-duplicate a ton of the code in this file. I think it's gotten to the point where there's only a handful of different method bodies. I *haven't* implemented the `-info` and `-generation` metrics because I don't think they make sense in DataDog in quite the same way they do in Prometheus.

If there is a reasonable way to move this file to https://github.com/DataDog/integrations-core, I'd also be happy to do a pass on that work. Right now, it looks like integrations are implicitly required to be only one Python file (check.py), which makes this one a bit tricky.

### DIFF
--- a/utils/kubernetes/kube_state_processor.py
+++ b/utils/kubernetes/kube_state_processor.py
@@ -245,7 +245,7 @@ class KubeStateProcessor:
             tags = ['{}:{}'.format(label.name, label.value) for label in metric.label]
             if val:
                 self.service_check(service_check_name, self.kube_check.OK, tags=tags)
-            else
+            else:
                 self.service_check(service_check_name, self.kube_check.CRITICAL, tags=tags)
 
     # note that these three prometheus metrics share a service_check_name
--- a/utils/kubernetes/kube_state_processor.py
+++ b/utils/kubernetes/kube_state_processor.py
@@ -120,6 +120,14 @@ class KubeStateProcessor:
             tags = ['node:{}'.format(self._extract_label_value("node", metric.label))]
             self.gauge(metric_name, val, tags)
 
+    def kube_deployment_status_replicas(self, message, **kwargs):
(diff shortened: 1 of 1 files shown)


ojongerius:
Nice work! Probably overlaps with https://github.com/DataDog/dd-agent/pull/3185, but looks like you have ticked off some of the TODOs there. Care to rebase on top of @therc's work?

jkurz:
So this PR can be closed correct? @hkaj container restarts is already in master.

sophaskins:
Yes! What's in master looks great!

PR #3238, opened 2017-03-06
Proposing a way to customize aggregations and percentils for histograms

wk8:
### What does this PR do?

Currently, it's only possible to do this at a system-wide level, using
the following agent configuration options:

```
histogram_aggregates: max, median, avg, count
histogram_percentiles: 0.95
```

This patch makes this behaviour customizable on a metric basis, based
on a list of regexes and their corresponding configs; e.g. the following
agent configuration:
```
histogram_aggregates: ^my_prefix: max ; max, median, avg, count
histogram_percentiles: ^my_prefix: 0.95, 0.90 ; 0.95
```
would run only the `max` aggregation but report both the 90th and 95th
percentiles on any prefix starting with `my_prefix`; but any other metric
would have the same aggregations as above (ie `max, median, avg, count`),
and report only the 95th percentile.

The first matching regex in the configuration will be applied. If none
match, defaults to the config with no regex, or else to the hard-coded
agent default.

In particular, that makes this patch entirely backward compatible.

### Motivation

This would come in handy to be able to limit aggregations for some
metrics but not all of them on any given box; in particular to avoid
pushing too many custom metrics with respect to plan limits.

For example, we use Datadog to do some high-level profiling in some
of our codebases; we'd be perfectly happy having only the `median`
aggregation for these metrics, which would allow us to lower our
number of custom metrics; but we do not want to stop pushing
other aggregations for other histogram metrics on the same boxes.

### Tests

Updated exisiting tests, and added a minimal test on the new feature, but
will add more if/when the Datadog dev team validates the approach. In particular,
tests should be added at least for:
 - several matching regexes: the 1st matching one should prevail
 - no defined default: the hard-coded default should be applied
 - invalid regex in config gets ignored with a warning

### Additional Notes

Related to customer ticket 82067.

### DIFF
--- a/aggregator.py
+++ b/aggregator.py
@@ -232,16 +232,27 @@ class Histogram(Metric):
         self.name = name
         self.count = 0
         self.samples = []
-        self.aggregates = extra_config['aggregates'] if\
-            extra_config is not None and extra_config.get('aggregates') is not None\
-            else DEFAULT_HISTOGRAM_AGGREGATES
-        self.percentiles = extra_config['percentiles'] if\
-            extra_config is not None and extra_config.get('percentiles') is not None\
-            else DEFAULT_HISTOGRAM_PERCENTILES
+        self.aggregates = self._extract_relevant_config(extra_config, 'aggregates',
+                                                        DEFAULT_HISTOGRAM_AGGREGATES)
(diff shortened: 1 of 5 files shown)


wk8:
Stale, closing

PR #3240, opened 2017-03-07
[emitter][payload split] make sure to have a hostname

remh:
In some scenarios: only for windows system metrics using the old check
interface, hostname was not passed. In that case we should use the
internalHostname one.

### DIFF
--- a/emitter.py
+++ b/emitter.py
@@ -128,10 +128,15 @@ def split_payload(legacy_payload):
         }
 
         if len(ts) >= 4:
-            if ts[3].get('type'):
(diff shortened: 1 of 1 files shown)


emitter.py
```diff
@@ -128,10 +128,15 @@ def split_payload(legacy_payload):
         }
 
         if len(ts) >= 4:
-            if ts[3].get('type'):
```
degemer: Why remove the type?

remh: it's moved not removed

PR #3246, opened 2017-03-08
[core] change agent.py rights to 755

degemer:
It was changed by mistake, and makes local dev harder, and breaks Docker installs.
Also modifying Rakefile, not sure why it had 755.


degemer:
Unrelated failure on Appveyor - https://github.com/DataDog/dd-agent/pull/3203#issuecomment-285102140

PR #3251, opened 2017-03-10
[core][sdk][info] fix version detection for WMI Checks

truthbk:
### What does this PR do?

Fixes a small bug where the check version was not properly being reported in the info page for WMI checks.

It also addresses another small issue, when we hit an issue when validating the YAML, we actually fail before attempting to load the check, so the check version is unknown at that point. We were reporting the check version as `None` in that case, this will now report it as `unknown` (as we have not yet loaded the check). Is `unknown` better than `None`? I went for it because it would be more consistent with other error messages when the version cannot be determined (ie. a bad manifest, etc).

### Motivation

Found the bug while testing the RC.

### Testing Guidelines

### DIFF
--- a/checks/__init__.py
+++ b/checks/__init__.py
@@ -49,6 +49,7 @@ AGENT_METRICS_CHECK_NAME = 'agent_metrics'
 
 AGENT_CHECK_FRAME = 2
 NETWORK_CHECK_FRAME = 3
+WMI_CHECK_FRAME = 3
 
 
 # Konstants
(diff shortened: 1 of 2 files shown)


checks/__init__.py
```diff
@@ -392,17 +393,26 @@ def _get_check_manifest(self, manifest_path=None):
             CheckClass.__init__() -> AgentCheck.__init__() -> AgentCheck._get_check_manifest()
         """
         from checks.network_checks import NetworkCheck
+        try:
+            from checks.wmi_check import WinWMICheck
```
truthbk: `pywintypes` unavailable on linux, so this will fail. No problem.

gmmeyer: We could also do if Platform.is_windows(), but generally we do the try/except import pattern, so I think it's better to follow the pattern we already have established

truthbk: Yeah, since we have to set WInWMICheck to `None` otherwise this seemed to be a pattern that worked well.

PR #3253, opened 2017-03-10
[core][sdk][version_manifest] Fix version comparison

olivielpeau:
### What does this PR do?

Fixes the version comparison in `validate_sdk_check` (added in #3251)

### Testing

Added a test case covering what this fixes.

### Additional Notes

We have to use tuples of integers if we want the comparison
to work as expected

### DIFF
--- a/config.py
+++ b/config.py
@@ -125,6 +125,11 @@ def get_version():
     return AGENT_VERSION
 
 
+def _version_string_to_tuple(version_string):
+    '''Return a (X, Y, Z) version tuple from an 'X.Y.Z' version string'''
+    return tuple(int(version_elem) for version_elem in version_string.split('.'))
+
+
 # Return url endpoint, here because needs access to version number
 def get_url_endpoint(default_url, endpoint_type='app'):
     parsed_url = urlparse(default_url)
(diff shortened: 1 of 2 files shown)


tests/core/test_config.py
```diff
@@ -324,11 +324,20 @@ def testConfigCustomOverDefault(self, *args):
...
+
+class TestManifestValidation(unittest.TestCase):
     def testManifestValidateOK(self, *args):
         manifest_path = '{}/manifest.json'.format(FIXTURE_PATH)
-        validate = validate_sdk_check(manifest_path)
-        self.assertEquals(True, validate)
+        with mock.patch('config.get_version', return_value='5.9.1'):
+            validate = validate_sdk_check(manifest_path)
+            self.assertEquals(True, validate)
+        with mock.patch('config.get_version', return_value='5.12.0'):
```
truthbk: Is the second test to illustrate the tuples are doing their thing versus lexicographical comparison? If so, I think just the second test would suffice.

config.py
```diff
@@ -125,6 +125,20 @@ def get_version():
...
 
 
+def _version_string_to_tuple(version_string):
+    '''Return a (X, Y, Z) version tuple from an 'X.Y.Z' version string'''
+    version_list = []
+    for elem in version_string.split('.'):
+        try:
+            elem_int = int(elem)
+        except ValueError:
+            log.warning("Unable to parse element '%s' of version string '%s'", elem, version_string)
```
truthbk: So your changes have actually highlighted something interesting. I actually think this should error out if we can't parse the version_string to ints. We know (or can assume with a very high degree of certainty) for a fact the `AGENT_VERSION` will be well formed, so that would be an indication of a bad string in the manifest. I would maybe favor raising an exception here and making sure `validate_sdk_check` returns `False`. The check will still run, as we provide a best-effort approach, and could be something valuable to log.

PR #3258, opened 2017-03-14
[core][sdk][win] add paths to sys.path if bundled with libs.

truthbk:
### What does this PR do?

This PR attempts to add to the relevant locations to `sys.path` before the checks are loaded in the collector solving potential library dependency issues in integrations-extras. It does not affect `core` integrations as these already bundle their deps with the `dd-agent` core package. It does not affect `unix` systems either because in those cases we **_do_** enfore the target location for the packages so we can drop the preferable `pth` files instead.  

### Motivation

On windows, integrations can be installed to an arbitrary location, meaning we don't know at build time what the `pth` files we'd drop in `site-packages` should look like. 

### Testing Guidelines

### DIFF
--- a/agent.py
+++ b/agent.py
@@ -62,6 +62,7 @@ from utils.profile import AgentProfiler
 from utils.service_discovery.config_stores import get_config_store
 from utils.service_discovery.sd_backend import get_sd_backend
 from utils.watchdog import Watchdog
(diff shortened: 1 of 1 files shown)


truthbk:
Note: that this should be a `NOP` until we actually have an integrations_extra package that actually bundles some libs. Nothing in the horizon right now.

PR #3260, opened 2017-03-14
[config][core] explicitly set manifest paths. Correct typo.

truthbk:
### What does this PR do?

Set manifest attributes, and check version explicitly after a successful check load. Stack introspection was not deemed reliable.

Typo in config.py. It's `dirname` not `basedir`!

### Motivation

Upon further review stack introspection to determine the check's nature was not the preferred approach. Instead set options manually. Collector was crashing on check failures due to typo.

### Testing Guidelines


truthbk:
@olivielpeau thanks for the input was too obfuscated on the former approach.

PR #3262, opened 2017-03-15
[travis][ci] Add a proxy test

zippolyte:
### What does this PR do?

Add a proxy test to the `core_integration` flavor.


tests/core/test_proxy.py
```diff
@@ -44,3 +63,50 @@ def test_no_proxy(self):
...
+        self.docker_client = DockerUtil().client
+
+        for line in self.docker_client.pull(CONTAINER_TO_RUN, stream=True):
+            log.info(line)
+
+        self.container = self.docker_client.create_container(CONTAINER_TO_RUN, detach=True, name=CONTAINER_NAME,
+            ports=[3128], host_config=self.docker_client.create_host_config(port_bindings={3128: 3128}))
+        log.info("Starting container: {0}".format(CONTAINER_TO_RUN))
+        self.docker_client.start(CONTAINER_NAME)
+        sleep(1) # Give time for the container to properly start, otherwise we get 'Proxy CONNECT aborted'
```
truthbk: This is sometimes prone to race conditions. Could we maybe inspect the logs for the `test-squid` container to determine if the process is up or not?

ddagent.py
```diff
@@ -261,7 +262,11 @@ def flush(self):
                 tornado.httpclient.AsyncHTTPClient.configure("tornado.curl_httpclient.CurlAsyncHTTPClient")
         else:
             log.debug("Using SimpleHTTPClient")
-        http = tornado.httpclient.AsyncHTTPClient()
+
+        if self._use_blocking_http_client:
```
truthbk: Although this is perfectly valid, you might be able to avoid this (which isn't really something we'd likely use) by using the tornado test classes which have some facilities to test the async stuff: http://www.tornadoweb.org/en/stable/testing.html

.travis.yml
```diff
@@ -28,9 +28,15 @@ env:
     - DD_CASHER_DIR=/tmp/casher
   matrix:
     - TRAVIS_FLAVOR=default
-    - TRAVIS_FLAVOR=core_integration
     - TRAVIS_FLAVOR=checks_mock
 
+matrix:
+  include:
```
truthbk: This only has an impact of about 40 seconds, and now that we've streamlined the CI after the SDK split, that shouldn't be a problem 👍

tests/core/test_proxy.py
```diff
@@ -44,3 +63,50 @@ def test_no_proxy(self):
         env.pop("https_proxy", None)
         env.pop("HTTP_PROXY", None)
         env.pop("HTTPS_PROXY", None)
+
+class TestProxy(TestCase):
+    @attr(requires='test')
```
truthbk: I don't think the test is running. I _think_ this should be `@attr(requires='core_integration')`. It doesn't look like `test_no_proxy` is running either actually.

tests/core/test_proxy.py
```diff
@@ -44,3 +63,50 @@ def test_no_proxy(self):
...
+        access_log = self.docker_client.exec_start(
+            self.docker_client.exec_create(CONTAINER_NAME, 'cat /var/log/squid/access.log')['Id'])
+        self.assertTrue("CONNECT" in access_log) # There should be an entry in the proxy access log
+        self.assertEquals(len(trManager._endpoints_errors), 1) # There should be an error since we gave a bogus api_key
+
+    def setUp(self):
+        self.docker_client = DockerUtil().client
+
+        for line in self.docker_client.pull(CONTAINER_TO_RUN, stream=True):
+            log.info(line)
```
truthbk: Not convinced we need to log this, if we fail pulling we'll see it.

tests/core/test_proxy.py
zippolyte: Indeed I forgot to replace that after my testing was done. For the `test_no_proxy` I thought it ran as part of the `default` flavor. Should I add it to the `core_integration` flavor then ?

.travis.yml
zippolyte: Yes but actually there are some tests in the `default` flavor that require IPv6 support, which can be only done if the tests are run in the container based environment (`sudo: false`) so I have to separate them.

tests/core/test_proxy.py
```diff
@@ -44,3 +65,63 @@ def test_no_proxy(self):
         env.pop("https_proxy", None)
         env.pop("HTTP_PROXY", None)
         env.pop("HTTPS_PROXY", None)
+
+
+class CustomAgentTransaction(AgentTransaction):
+
+    def on_response(self, response):
+        super(CustomAgentTransaction, self).on_response(response)
+        self._test.stop()
```
truthbk: Nice, elegant approach. Can you set `self._test` in a constructor, in case the class is reused or refactored to somewhere else in the future?

truthbk: Also, if you could add a comment as to what `test` should be, maybe pointing here (that would be great): http://www.tornadoweb.org/en/stable/testing.html

zippolyte: Actually that's what I did at first, but then when the `on_response` method is called, the attribute `_test` got deleted somehow and I got `'CustomAgentTransaction' object has no attribute '_test'`. I didn't find what was wrong so I changed my approach.

tests/core/test_proxy.py
```diff
@@ -44,3 +65,63 @@ def test_no_proxy(self):
...
+                "password": None
+            }
+        }
+
+        app = Application()
+        app.skip_ssl_validation = True
+        app._agentConfig = config
+
+        trManager = TransactionManager(MAX_WAIT_FOR_REPLAY, MAX_QUEUE_SIZE, THROTTLING_DELAY)
+        trManager._flush_without_ioloop = True  # Use blocking API to emulate tornado ioloop
```
zippolyte: Well if I don't set this to True, the transaction is never flushed. I think it is because the IOLoop is started only when I call `wait()`, which I do after creating the transaction.

tests/core/test_proxy.py
truthbk: Weird! OK, then can we at least check if the attribute is present before calling it?

PR #3263, opened 2017-03-15
Remove requirements for kafka_consumer.py

jeffwidman:
Now that `kafka_consumer.py` is in integrations-core repo, don't think it's requirements need to be present here anymore.


jeffwidman:
I appreciate that you guys got this work done in #3298 but it does make it feel like I'm wasting my time to submit PR's if this wasn't reviewed for three weeks and then the change is done in your own internal PR. cc @gmmeyer @remh

PR #3264, opened 2017-03-15
[install script] only refresh datadog repo on SUSE

degemer:
As otherwise if another repository is configured and failing it will
fail the install for nothing.


degemer:
Tested on a fresh suse 12.

PR #3265, opened 2017-03-16
Add ability to get docker IP address from Rancher labels for Rancher 1.3

zippolyte:


utils/service_discovery/sd_docker_backend.py
```diff
@@ -171,6 +171,14 @@ def _get_host_address(self, state, c_id, tpl_var):
             if pod_ip:
                 return pod_ip
 
+        # try to get the rancher IP address
+        log.debug("No IP address was found in container %s (%s) "
```
hkaj: Could you add `if Platform.is_rancher()` before this new code? Otherwise this rancher-specific log line will show up in non Rancher environments.

You'll need to create an `is_rancher` function in `platform.py` first. One way you could detect rancher is by looking for their agent container(s) (read more about that [here](https://docs.rancher.com/rancher/v1.6/en/faqs/agents/)).

hkaj:
Nice one @zippolyte 
As mentioned in the card description (`and add rancher labels as tags in SD and the docker check while we're parsing docker labels.`) could you also collect interesting labels as tags? Like the rancher service name, etc.
It would also be interesting to add these tags in the docker check (integration-core).

This way users will be able to graph their metrics based on rancher resources.

utils/service_discovery/sd_docker_backend.py
```diff
@@ -290,6 +301,12 @@ def get_tags(self, state, c_id):
             if swarm_svc:
                 tags.append('swarm_service:%s' % swarm_svc)
 
+        if Platform.is_rancher():
+            c_inspect = state.inspect_container(c_id)
+            service_name = c_inspect.get('Config', {}).get('Labels', {}).get('io.rancher.container.name')
```
hkaj: maybe make `io.rancher.container.name` a constant? It could live here or in dockerutil.

utils/service_discovery/sd_docker_backend.py
```diff
@@ -171,6 +171,15 @@ def _get_host_address(self, state, c_id, tpl_var):
             if pod_ip:
                 return pod_ip
 
+        if Platform.is_rancher():
+            # try to get the rancher IP address
+            log.debug("No IP address was found in container %s (%s) "
+                "trying with the Rancher label" % (c_id[:12], c_img))
+
+            ip_addr = c_inspect.get('Config', {}).get('Labels', {}).get('io.rancher.container.ip')
```
hkaj: maybe make `io.rancher.container.ip` a constant?

hkaj:
cool, i like it better too :)

PR #3281, opened 2017-03-23
Fix the 5.12.0 Source Install

gmmeyer:
### What does this PR do?

Fixes the source install, which failed when we removed all the checks from the agent. It was partially integrated, but did not entirely work.

### DIFF
--- a/packaging/datadog-agent/source/setup_agent.sh
+++ b/packaging/datadog-agent/source/setup_agent.sh
@@ -397,6 +397,9 @@ VERSION
 if [ "$AGENT_MAJOR_VERSION" -eq "5" -a "$AGENT_MINOR_VERSION" -gt "11" ]; then
   print_console "* Downloading integrations from GitHub"
   mkdir -p "$DD_HOME/integrations"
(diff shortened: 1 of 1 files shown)


packaging/datadog-agent/source/setup_agent.sh
```diff
@@ -405,15 +408,32 @@ if [ "$AGENT_MAJOR_VERSION" -eq "5" -a "$AGENT_MINOR_VERSION" -gt "11" ]; then
   rm -f "$DD_HOME/integrations.tar.gz"
   print_done
 
-  print_console "* Trying to install integration requirements"
-  for INT_DIR in "$DD_HOME/integrations/*"; do
+  print_console "* Setting up integrations"
+  INTEGRATIONS=$(ls $DD_HOME/integrations/)
```
truthbk: I think what you want want is `INTEGRATIONS=$(ls -d $DD_HOME/integrations/*)`

gmmeyer:
I think we should remove it, for some reason it seemed to cause a bug when I did. I will investigate but better to get this out than spend time on that.

PR #3286, opened 2017-03-24
remove all checks.d optionals requirements

degemer:
Only keep the core agent optional requirements.

### DIFF
--- a/requirements-opt.txt
+++ b/requirements-opt.txt
@@ -1,8 +1,8 @@
 ######################## WARNING ##########################
 # This file currently determines the python deps installed
-# for the dev env, the source install and the Win build.
+# for the dev env and the source install.
 # It is NOT used for the DEB/RPM packages and the OS X
-# build. For these please update:
+# and Windows builds. For these please update:
 # https://github.com/DataDog/omnibus-software
 ###########################################################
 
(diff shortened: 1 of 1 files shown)


requirements-opt.txt
```diff
@@ -14,42 +14,8 @@
...
-
-# checks.d/pgbouncer.py
-# Require libpq
-psycopg2==2.6.2
-
-# checks.d/win32_event_log.py
-# checks.d/wmi.py
-# It's a pure python module, it doesn't require anything to install,
-# but needs the pywin32 extension to work
-wmi==1.4.9
```
truthbk: Don't wee need this for the windows system checks? https://github.com/DataDog/dd-agent/blob/master/checks/system/win32.py#L16

gmmeyer: I believe this is right.

degemer: I'm not sure, I checked and it seems we always use our internal implementation. We don't have any 'import wmi' or 'from wmi' in our code. cc @yannmh

yannmh: 👍

truthbk: Yeah, just for posterity, the internal checks.libs.wmi.WMISampler relies on:
```
import pythoncom
from win32com.client import Dispatch
```

PR #3289, opened 2017-03-28
[docker] Support Nomad for SD and tags

xvello:
### What does this PR do?

Detects the Nomad orchestrator and:
- adds job/group/task names as tags
- tags caching and cache invalidation

Auto-detection works via the NOMAD_ALLOC_ID envvar, so the dd-agent must run in a container started via Nomad (we need to write an example specfile)

## What's to do?
- docker_deamon intergration PR is https://github.com/DataDog/integrations-core/pull/305

### DIFF
--- a/utils/orchestrator/nomadutil.py
+++ b/utils/orchestrator/nomadutil.py
@@ -1,0 +1,106 @@
+# (C) Datadog, Inc. 2017
+# All rights reserved
+# Licensed under Simplified BSD License (see LICENSE)
+
+# stdlib
+import logging
+from utils.dockerutil import DockerUtil
+from utils.singleton import Singleton
+
+
+log = logging.getLogger(__name__)
+
+NOMAD_TASK_NAME = 'NOMAD_TASK_NAME'
+NOMAD_JOB_NAME = 'NOMAD_JOB_NAME'
+NOMAD_ALLOC_NAME = 'NOMAD_ALLOC_NAME'
+
+
+class NomadUtil:
+    __metaclass__ = Singleton
+
+    def __init__(self, instance=None):
+        self.docker_util = DockerUtil()
+        if instance is None:
+            log.debug("New NomadUtil instance")
+            self._container_tags_cache = {}
+            self._enabled = None
+
+    def init_platform(self, agentConfig):
+        """
+        Enable/disable Nomad support depending on the docker_orchestrator config file entry
+        :param agentConfig: dict from config.get_config()
+        """
+        self._enabled = agentConfig.get('docker_orchestrator', None) == 'nomad'
+
+    def is_enabled(self):
+        """
+        Allows user classes to check whether nomad is to be activated.
+        init_platform has to be called once first
+        :return: True/False
+        """
+        if self._enabled is None:
+            log.warning("Calling nomad.is_nomad before init_platform, replying False")
+            return False
+        else:
+            return self._enabled
+
+    def extract_container_tags(self, co):
+        """
+        Queries docker inspect to get nomad tags in the container's environment vars.
+        As this is expensive, it is cached in the self._nomad_tags_cache dict.
+        The cache invalidation goes through invalidate_nomad_cache, called by docker_daemon
+
+        :param co: container dict returned by docker-py
+        :return: tags as list<string>, cached
+        """
+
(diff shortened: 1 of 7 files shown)


xvello:
When a container is deleted, the docker_daemon reload triggers a reload of NomadUtil and although it's a singleton, it gets re-instanciated and the cache is emptied. Any idea on how to avoid that?

utils/service_discovery/sd_docker_backend.py
```diff
@@ -320,6 +324,11 @@ def get_tags(self, state, c_id):
             if container_name:
                 tags.append('rancher_container:%s' % container_name)
 
+        elif Platform.is_nomad():
+            nomad_tags = self.nomadutil.extract_container_tags(state.inspect_container(c_id))
+            if nomad_tags:
+                tags.extend(list(nomad_tags))
```
hkaj: we don't need the `list()` here do we?

utils/service_discovery/sd_docker_backend.py
```diff
@@ -136,7 +140,7 @@ def _get_checks_to_refresh(self, state, c_id):
             return
 
         identifier = inspect.get('Config', {}).get('Labels', {}).get(DATADOG_ID) or \
-            inspect.get('Config', {}).get('Image')
+            DockerUtil().image_name_extractor(inspect)
```
hkaj: nitpick because it's not too expensive, but you should create a `self.dockerutil` in `__init__` and use `self.dockerutil` when you need it. That would avoid the existing instance lookup at every call.

utils/orchestrator/nomadutil.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def is_enabled(self):
+        """
+        Allows user classes to check whether nomad is to be activated.
+        init_platform has to be called once first
+        :return: True/False
+        """
+        if self._enabled is None:
+            log.warning("Calling nomad.is_nomad before init_platform, replying False")
+            return False
```
xvello: Because I specified "docker_orchestrator:nomad" in datadog.conf and not docker_daemon.yaml, which leads to this hacky way to get agentConfig.
We should probably just use a nomad_tags option in docker_daemon.yaml and move to a global orchestrator option in agent6

jvrplmlmn:
@xvello any updates on this?

I can help with code reviews or testing if needed 😃

xvello:
¡Hola @jvrplmlmn, thanks for your interest!

This is scheduled for 5.14, due in a few weeks. As I worked on other features in the meantime, this PR will need some conflict resolution and refactoring before being merged in master, hopefully next week. If you want to help with testing, ping me after merging and I'll build a beta docker image.

Could you tell me more (PM or here) about your Nomad setup (version, number of nodes...)?

utils/orchestrator/nomadutil.py
```diff
@@ -0,0 +1,82 @@
...
+            for var in envvars:
+                if var.startswith(NOMAD_TASK_NAME):
+                    tags.append('nomad_task:%s' % var[len(NOMAD_TASK_NAME) + 1:])
+                elif var.startswith(NOMAD_JOB_NAME):
+                    tags.append('nomad_job:%s' % var[len(NOMAD_JOB_NAME) + 1:])
+                elif var.startswith(NOMAD_ALLOC_NAME):
+                    try:
+                        start = var.index('.', len(NOMAD_ALLOC_NAME)) + 1
+                        end = var.index('[')
+                        tags.append('nomad_group:%s' % var[start:end])
```
hkaj: Could we add
```python
if end <= start:
    raise ValueError
```
before this line? It shouldn't be necessary for now but the format could change and break the tag easily.

utils/orchestrator/nomadutil.py
```diff
@@ -0,0 +1,82 @@
...
+    def invalidate_cache(self, events):
+        """
+        Allows cache invalidation when containers dies
+        :param events from self.get_events
+        """
+        try:
+            for ev in events:
+                if ev.get('status') == 'die' and ev.get('id') in self._container_tags_cache:
+                    del self._container_tags_cache[ev.get('id')]
+                    log.debug(self._container_tags_cache)
```
hkaj: this will spam the logs let's remove it

PR #3297, opened 2017-04-04
[kubernetes_state] refactor gauge submission, fix container.restarts

hkaj:
if you have not yet done so.*

### What does this PR do?

- removes some of the duplicated code for gauges
- fixes the `container.restarts` metric type
- introduces a type check for prometheus metrics

### Motivation

container.restarts is always `0` because we use the gauge value instead of counter value.

### Testing Guidelines

See https://github.com/DataDog/integrations-core/pull/317

### DIFF
--- a/utils/kubernetes/kube_state_processor.py
+++ b/utils/kubernetes/kube_state_processor.py
@@ -4,6 +4,11 @@
 
 NAMESPACE = 'kubernetes_state'
 
+# message.type is the index in this array
+# see: https://github.com/prometheus/client_model/blob/model-0.0.2/metrics.proto#L24-L28
+METRIC_TYPES = ['counter', 'gauge']
+
+
 class KubeStateProcessor:
     def __init__(self, kubernetes_check):
         self.kube_check = kubernetes_check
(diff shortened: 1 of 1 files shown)


utils/kubernetes/kube_state_processor.py
```diff
@@ -20,13 +25,38 @@ def __init__(self, kubernetes_check):
...
-        invoke it. Log some info if method was not found.
+        Handle a message according to the following flow:
+            - search self.metric_to_gauge for a prometheus.metric <--> datadog.metric mapping
+            - call check method with the same name as the metric
+            - log some info if none of the above worked
         """
         try:
-            getattr(self, message.name)(message, **kwargs)
+            if message.name in self.metric_to_gauge:
+                self._submit_gauges(self.metric_to_gauge[message.name], message)
```
xvello: We loose the label argument if passed to process(). Should we keep it in _submit_gauges or just remove the corresponding code branch?

utils/kubernetes/kube_state_processor.py
```diff
@@ -20,13 +25,38 @@ def __init__(self, kubernetes_check):
...
+            'kube_node_status_capacity_memory_bytes': NAMESPACE + '.node.memory_capacity',
+            'kube_node_status_capacity_pods': NAMESPACE + '.node.pods_capacity',
+            'kube_node_status_allocatable_cpu_cores': NAMESPACE + '.node.cpu_allocatable',
+            'kube_node_status_allocatable_memory_bytes': NAMESPACE + '.node.memory_allocatable',
+            'kube_node_status_allocatable_pods': NAMESPACE + '.node.pods_allocatable',
+            'kube_deployment_status_replicas_available': NAMESPACE + '.deployment.replicas_available',
+            'kube_deployment_status_replicas_unavailable': NAMESPACE + '.deployment.replicas_unavailable',
+            'kube_deployment_status_replicas_updated': NAMESPACE + '.deployment.replicas_updated',
+            'kube_deployment_spec_replicas': NAMESPACE + '.deployment.replicas_desired',
+            'kube_pod_container_resource_requests_cpu_cores': NAMESPACE + '.container.cpu_requested',
```
xvello: The CI check fail on .container.cpu_requested, looks like the valid k8s metric name is :  kube_pod_container_requested_cpu_cores

hkaj: That's actually a mistake in their doc, see: https://github.com/kubernetes/kube-state-metrics/blob/1751d86b25463c60234a37753024c7122d39cc26/pod.go#L85

hkaj: I think what makes the CI fail is that we need to merge the `protobuf.bin` fixture, otherwise the metric doesn't exist. Tests pass locally when I copy the files.

utils/kubernetes/kube_state_processor.py
hkaj: i feel like it could be useful someday, but yeah let's kill it for now since it's not used.

PR #3300, opened 2017-04-05
[k8s] KubeUtil + SD service and creator tags

xvello:
### What does this PR do?

Add deployment and service k8s tags to service discovery tags. 

Logic is in separate kube_event_retriever and pod_service_mapper classes

#### Deployment

The deployment name is inferred from the replicaset name, as this is hardcoded in the k8s source. Matching selectors would be costly, especially if we need to support matchExpressions set selectors

#### Services

The pod->service is implemented by matching every pod to every service's label selectors. The services are kept in memory cache and cache is checked every time service discovery is triggered. If the pod satisfies matches several services, multiple kube_service tags are added.

The cache is invalidated on kubernetes events pushed by the kubernetes check, mirroring how docker_daemon/dockerutils work. These events allow to trigger SD updates on service events.

It is then dependant on https://github.com/DataDog/integrations-core/pull/319

#### Event handling

KubeUtil now has a process_events() method that kubernetes.py calls on every check. New features can hook in this method to:
  - invalidate caches
  - add container_ids to reload SD for

### DIFF
--- a/utils/kubernetes/kubeutil.py
+++ b/utils/kubernetes/kubeutil.py
@@ -366,6 +366,7 @@ class KubeUtil:
         try:
             if self._services_cache_last_resourceversion == -1:
                 # Retrieving latest service event number with check_services_cache_freshness dry run
+                self._services_cache = {}   # check___freshness skips if None
                 self.check_services_cache_freshness()
             reply = self.retrieve_json_auth(self.kubernetes_api_url + '/services')
             self._services_cache = {}
@@ -386,7 +387,11 @@ class KubeUtil:
 
         We use the event's resourceVersion, as using the service's version wouldn't catch deletion
         """
-        log.debug("Testing service cache freshness, current latest: %d", self._services_cache_last_resourceversion)
+
+        # Don't check if cache is already empty
+        if self._services_cache is None:
+            return
+
         lastestVersion = None
         flush = False
         try:
--- a/utils/kubernetes/kubeutil.py
+++ b/utils/kubernetes/kubeutil.py
@@ -8,6 +8,8 @@ import logging
 import os
 from urlparse import urljoin
 from urllib import urlencode
+import time
+import calendar
 
 # project
 from util import check_yaml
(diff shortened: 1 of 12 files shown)


utils/service_discovery/sd_docker_backend.py
```diff
@@ -111,6 +111,9 @@ def _make_fetch_state(self):
     def update_checks(self, changed_containers):
         state = self._make_fetch_state()
 
+        if Platform.is_k8s():
+            self.kubeutil.check_services_cache_freshness()
```
hkaj: Maybe we should also add a call in `get_configs`. That's the second entrypoint of sd_docker_backend. What do you think?

hkaj:
another point: it's adding a lot of logic to kubeutil, I think it's at a point where tests are badly needed. Could you add some for the logic you add? We'll extend the coverage as we go.

utils/service_discovery/sd_docker_backend.py
xvello: added it, and added some code to check__freshness to skip duplicate checks if it were to happen

jonchiu:
This is slightly tangential, but is there a reason why there's no `kube_` prefix for docker SD tags?

https://github.com/DataDog/dd-agent/blob/cbc91ed7fad78f96866534b89c3b0b78bea16b0e/utils/service_discovery/sd_docker_backend.py#L287

xvello:
@jonchiu that is an oversight from us. Unfortunately, this is hard to change now as this would break some dashboards and alerts, so it will be fixed in the next major version of the agent.

utils/kubernetes/pod_service_mapper.py
```diff
@@ -0,0 +1,200 @@
...
+        We use the event's resourceVersion, as using the service's version wouldn't catch deletion
+        """
+
+        # Don't check if cache is already invalidated
+        if self._service_cache_invalidated:
+            return
+
+        lastestVersion = None
+        invalidate = False
+        try:
```
xvello: Needs to be refactored to KubeEventRetriever, I'm on it

utils/kubernetes/kubeutil.py
```diff
@@ -337,3 +364,118 @@ def get_auth_token(cls):
...
+        Match the pods labels with services' label selectors to determine the list
+        of services that point to that pod. Returns an array of service names.
+
+        Pass refresh=True if you want to bypass the cached cid->services mapping (after a service change)
+        """
+        s = self._service_mapper.match_services_for_pod(pod_metadata, refresh, names=True)
+        #log.warning("Matches for %s: %s" % (pod_metadata.get('name'), str(s)))
+        return s
+
+    def _process_events_get_cid_to_update(self, events):
```
hkaj: looks like we can get rid of this one?

utils/kubernetes/kubeutil.py
```diff
@@ -230,6 +243,20 @@ def retrieve_metrics(self):
...
 
+    def get_deployment_for_replicaset(self, rs_name):
+        """
+        Get the deployment name for a given replicaset name
+        For now, the rs name's first part always is the deployment's name, see
+        https://github.com/kubernetes/kubernetes/blob/release-1.6/pkg/controller/deployment/sync.go#L299
+        But it might change in a future k8s version. The other way to match RS and deployments is comparing
+        label selectors, which would be very costly, especially matchExpressions matching.
+        """
+        end = rs_name.rfind("-")
```
xvello: As the code uses rfind, it will match on the last dash or the name. The second part (spechash) can't include a dash.

utils/kubernetes/kubeutil.py
xvello: indeed, the code was moved to process_events

utils/kubernetes/kubeutil.py
xvello: added a check that the second part is only numbers
if we want to be 100% accurate, we can query /apis/extensions/v1beta1/replicasets

hkaj: cool, let's start with the `isdigit` thing and we'll change if needed 👍

utils/kubernetes/pod_service_mapper.py
```diff
@@ -0,0 +1,187 @@
...
+                selector = service.get('spec', {}).get('selector', {})
+                if uid == '' or len(selector) == 0:
+                    continue
+                self._service_cache_names[uid] = name
+                self._service_cache_selectors[uid] = selector
+            self._service_cache_invalidated = False
+        except Exception as e:
+            log.exception('Unable to read service list from apiserver: %s', e)
+            self._service_cache_selectors = defaultdict(dict)
+            self._service_cache_names = {}
```
xvello: That exception catch would be a very last resort (as we are only doing safe dict lookups). In the case it happens, I think it's better to have no service tagging than only half of them tagged (easier to troubleshoot).

hkaj: 👍

PR #3301, opened 2017-04-10
[centos initscript] Allow a mechanism to disable datadog-agent via sysconfig

generica:
### What does this PR do?

Adds a check to a non-packaged file to be able to disable datadog-agent

### Motivation

We currently overwrite the initscript to enable/disable datadog for different roles/environments. This is problematic, and means we need to constantly be in sync. Also, there is a period of time during a yum upgrade where the original initscript is run, then run again once we overwrite it.

A better solution would be an upstream change to check a sysconfig file to see if datadog-agent is disabled for some reason. In all other scenarios, the initscript should work as before.

### Testing Guidelines

### Additional Notes

Happy to adjust to suit!


olivielpeau:
Postponing to the `5.15` release. There are a few questions we should answer first:
1. should we include the `/etc/sysconfig/datadog-agent` file to the rpm package?
2. if we support this in the rpm pkg, we should probably also support it in the deb pkg (the equivalent would be `/etc/default/datadog-agent`). Question 1 applies too.
3. if we are to include this in agent 5.x, we need to make sure the next major version of the agent support this too.

generica:
@olivielpeau I'd be happy to add changes to do this, and package a sample file for centos/debian, however I can't find the spec files/debian build files in this repo.

olivielpeau:
@generica Thanks! We build the agent using omnibus (see https://github.com/DataDog/dd-agent-omnibus), which generates the spec files from templates, so it might be slightly complex to change them.

After putting some thought into this with @masci, we don't think that it's necessary to ship the `sysconfig`/`default` files in the package as long as we document the feature correctly, so the answer to question `1` would be no :)

If you could add the same feature to the debian init script it'd be great.

Do you know how this change would work on systemd-based distributions? I'm not sure how systemd's compatibility layer would take into account this change in a sysvinit init file.

generica:
I've attached a commit for the debian equivalent. I'm a little bit unsure about the best way to do this in systemd land, but can have a look around. Cheers.

olivielpeau:
Looks good, thanks @generica! Merging, should be part of the next minor release (`5.17.0`).

On systemd-based systems one way of having similar functionality would be to use a "drop-in" directory, e.g. `/etc/systemd/system/datadog-agent.service.d`, put a `.conf` file in that directory, and set overrides of `[Unit]` directives in that file with whatever works best for the administrators.

PR #3304, opened 2017-04-12
Add kafka to the list of jmx checks

masci:
### What does this PR do?

When SD uses `auto_conf` to init JMX-fetch related checks, it verifies that the check name is listed in the list `JMX_CHECKS`. Kafka was missing, making impossible to use SD with JMX an the autoconf config storage.

### DIFF
--- a/jmxfetch.py
+++ b/jmxfetch.py
@@ -58,6 +58,7 @@ JMX_CHECKS = [
     'jmx',
     'solr',
     'tomcat',
+    'kafka',
 ]
 JMX_COLLECT_COMMAND = 'collect'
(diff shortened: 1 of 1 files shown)


olivielpeau:
Just a question: adding `is_jmx: yes` to the `init_config` section of the kafka auto conf yaml should work no?

That's currently the way we detect that a regular `kafka.yaml` file is a jmxfetch yaml config, and I don't see a reason why this detection would stop working with auto conf.

That said, even if this detection works, I don't see any reason not to merge this PR, `kafka.yaml` always refers to a jmxfetch config AFAIK.

PR #3310, opened 2017-04-19
[kubernetes] catch errors from kubeutil init

hkaj:
if you have not yet done so.*

### What does this PR do?

Kubeutil raises an exception if it can't reach the kubelet. We should catch it.

### Motivation

Ugly collector logs

### Testing Guidelines

Run the agent on a non-k8s node, see that the exception is now caught.

### Additional Notes

sister PR: https://github.com/DataDog/integrations-core/pull/345

### DIFF
--- a/utils/service_discovery/sd_docker_backend.py
+++ b/utils/service_discovery/sd_docker_backend.py
@@ -89,8 +89,12 @@ class SDDockerBackend(AbstractSDBackend):
 
         self.docker_client = DockerUtil(config_store=self.config_store).client
         if Platform.is_k8s():
-            self.kubeutil = KubeUtil()
-
+            try:
+                self.kubeutil = KubeUtil()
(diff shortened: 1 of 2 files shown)


utils/service_discovery/sd_docker_backend.py
```diff
@@ -89,8 +89,11 @@ def __init__(self, agentConfig):
 
         self.docker_client = DockerUtil(config_store=self.config_store).client
         if Platform.is_k8s():
-            self.kubeutil = KubeUtil()
```
masci: nit: should we assign `self.kubeutil = None` in case the exception is raised, to avoid attribute errors later?

masci: Actually it could be an issue, we would have an error later accessing `self.kubeutil`

PR #3315, opened 2017-04-21
[packaging] Remove agent/checks from PYTHONPATH

olivielpeau:
### What does this PR do?

Remove `/opt/datadog-agent/agent/checks` from the `PYTHONPATH` set up by supervisor.

### Motivation

Seems to have been there forever, but I don't think we need it in the
python path for any (good) reason. From what I've checked we always
import with `import checks.abc` instead of `import abc`.

Having it in the python path pollutes the module root namespace. For
instance it prevents importing the `datadog` module (i.e. module from
`datadogpy`) correctly (its name collides with `checks/datadog.py`).

### Testing Guidelines

Haven't tested this in depth but I applied the changes on an installed agent and haven't seen any import error.

### Additional Notes

The only use case that this change would break is if custom checks
are doing direct imports (e.g. `import wmi_check` to import
`checks/wmi_check.py`). The fix is trivial so it doesn't sound
blocking to me.

### DIFF
--- a/packaging/osx/supervisor.conf
+++ b/packaging/osx/supervisor.conf
@@ -17,7 +17,7 @@ logfile_maxbytes = 50MB
 nodaemon = false
 pidfile = /opt/datadog-agent/run/datadog-supervisord.pid
 logfile_backups = 10
-environment=PYTHONPATH=/opt/datadog-agent/agent:/opt/datadog-agent/agent/checks,LANG=POSIX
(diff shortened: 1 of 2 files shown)


pallan:
As I reported this to support, the context is that I was trying to use the python client in a custom_emitter. The namespace collision generated errors since `checks/datadog.py` does include the `statsd` module.  Modifying the supervisor.conf as done here fixes the error and the python client is properly loaded and usable.

👍

Issue #3254, opened 2017-03-13
[prometheus] extract prometheus-format processing to its own module


hkaj:
Some of the logic in kube_state_processor (https://github.com/DataDog/dd-agent/blob/5.11.3/utils/kubernetes/kube_state_processor.py) can be generalized to all checks using the prometheus format.

This issue covers several points:
- we should move the k8s-specific part to integration-core. This would allow us to add new metrics/tags without releasing a new version of the core agent
- the generic methods (`process`, `init` and label-related ones) should be kept, generalized (remove kubernetes mentions) and extended with more helpers ( see @ojongerius note for example https://github.com/DataDog/dd-agent/pull/3185/files#r103859112)

aerostitch:
I guess that we'll need to rewrite this one once #3317 is merged or should we move that part to integration-core after the merge?

hkaj:
https://github.com/DataDog/integrations-core/pull/492 implements that

Issue #3119, opened 2017-01-10
Add support for agent integration to pull Prometheus text-based metrics via HTTP


aantono:
With Prometheus monitoring system gaining popularity, more and more services/libraries offer exposing/consuming metrics using Prometheus [text4](https://prometheus.io/docs/instrumenting/exposition_formats/#format-version-0.0.4) format.  This can typically be consumed over HTTP configurable endpoint (either `/metrics` or `/prometheus`).
There is already provided Python library for parsing said format, available at https://github.com/prometheus/client_python#parserd

Having an agent check integration for consuming metrics from services/applications that expose it in Prometheus format would be great increase for adoption and transition for users who already use Prometheus or consume Prometheus-based metrics in the text format. (Looking at the agent code, it appears that there is already internal support for Prometheus-based Protobuf encoded metrics, which is what Kubernetes uses to expose it's statistics- https://github.com/DataDog/dd-agent/tree/master/utils/prometheus, https://github.com/DataDog/dd-agent/blob/master/checks.d/kubernetes_state.py#L37)

aerostitch:
#3317 should fit that need.

hkaj:
totally does, closing. Thanks guys :)

PR #3317, opened 2017-04-25
Add support for pulling metrics from prometheus

aerostitch:
### What does this PR do?

Adds the necessary helpers to be able to easily build the prometheus integration.
You provide a dictionary which has the prometheus metric as key and the name of the corresponding dd metric as value.
Only the metrics listed in this dictionary will be pushed to datadog.

Note that the prometheus' histogram metrics have a specific behavior with the notion of buckets, so for each bucket I add them with a tag which is the upper_bound. Of course, this part is optional and can be disabled.

### Motivation

I need to pull data that has been pushed to prometheus by 3rd party applications and don't want to reinvent the wheel.
Plus, it seems I'm not the only one who wish to see it happening (#3119, #3254)

### Additional Notes

I need to add more tests (planning to add those tomorrow) but I wanted a 1st feedback on the proposed modifications.

Once that would be integrated, the integration `check.py` could be as easy as:
```python
# (C) Datadog, Inc. 2010-2017
# All rights reserved
# Licensed under Simplified BSD License (see LICENSE)
from checks import CheckException
from checks.prometheus_check import PrometheusCheck

EVENT_TYPE = SOURCE_TYPE_NAME = 'kubedns'

class KubeDNSCheck(PrometheusCheck):
    """
    Collect kube dns metrics from Prometheus
    """
        super(PrometheusDNSCheck, self).__init__(name, init_config, agentConfig, instances)
        self.client = PrometheusCheck(self)
        self.client.NAMESPACE='kubedns'

        self.client.metrics_mapper = {
            'skydns_skydns_dns_response_size_bytes': 'dns.response_size.bytes',
            'skydns_skydns_dns_request_duration_seconds': 'dns.request_duration.seconds',
            'skydns_skydns_dns_request_count_total': 'dns.request_count.total',
            'skydns_skydns_dns_error_count_total': 'dns.error_count.total',
            'skydns_skydns_dns_cachemiss_count_total': 'dns.cachemiss_count.total',
        }


    def check(self, instance):
        endpoint = 'http://localhost:10055/metrics'
        if endpoint is None:
            raise CheckException("Unable to find prometheus_endpoint in config file.")

        send_buckets = instance.get('send_histograms_buckets')
        if send_buckets is None:
            raise CheckException("Unable to find send_histograms_buckets in config file.")

        self.client.process(endpoint, send_histograms_buckets=send_buckets, instance=instance)
```

Of course, after that we could rewrite the kubernetes-state-metrics integration to use this.

@hkaj I see that you've been working on several PR around prometheus. What do you think about this idea?

Thanks
Joseph


hkaj:
Hi @aerostitch 
Thanks for the contrib! I actually have a branch with some work regarding this feature but didn't have time to complete it yet https://github.com/DataDog/dd-agent/commit/9be9af8bd49157c2985b29359ebf9c0707906383

Looks like your PR implements a similar approach, I'll review it shortly.
Thanks again!

hkaj:
Had a first pass at it. It's a cool feature and will definitely help writing similar checks quicker.
Here's some more feedback, not tied to a particular code line:
- I think a mother class would be better fitted for this, what do you think?
- defining the metric names in the conf file is not ideal for several reasons, the main one being that if we write official checks with this prometheus helper, its metrics will be officially supported (i.e. not counted as custom metrics) but if users don't like the name and change them in the yaml, they will become custom metrics without the user realizing it. So I'd rather have `metrics_mapper` be a hardcoded class variable as in https://github.com/DataDog/dd-agent/commit/9be9af8bd49157c2985b29359ebf9c0707906383#diff-b97e21cc8c9598269fd1833ccdd51ae3R39 so metric names are fixed for every check.

aerostitch:
Thanks a lot for the review! :)
Sorry for the mother class I thought that I should have used it only in the integration/core project.
I'll fix that and change the other elements you commented. Hopefully I'll get the updated version this morning (San Francisco time! :) )

utils/prometheus/processor.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+class PrometheusProcessor:
+    def __init__(self, conf):
+        # metrics_mapper is a dictionnary where the keys are the metrics to capture and the values are the corresponding metrics names to have in datadog.
+        self.metrics_mapper = {}
+        # send_histograms_buckets is used to specify if yes or no you want to send the buckets as tagged values when dealing with histograms.
+        self.send_histograms_buckets = True
+        # endpoint is the metrics endpoint to use to poll metrics from Prometheus
+        self.endpoint = 'http://localhost:10055/metrics'
```
aerostitch: Sure, I was thinking about overwriting it like that in the check method:
```        endpoint = instance.get('prometheus_endpoint')
        if endpoint is None:
            raise CheckException("Unable to find prometheus_endpoint in config file.")
        self.processor.endpoint = endpoint```
But having it as a parameter should do it. :)

File: utils/prometheus/processor.py [hunk 2]
```diff
@@ -0,0 +1,95 @@
...
+        # Conf from the agent
+        self.log = conf.log
+        self.gauge = conf.gauge
+
+    def process(self, **kwargs):
+        """
+        Polls the data from prometheus and pushes them as gauges
+        """
+        data = self.processor.poll()
```
aerostitch: Good catch! :) Thanks

aerostitch:
I'm also wondering if I shouldn't move the check to the `checks/` directory (if it's a check per say, it should probably not be in the `utils` directory where you find all the helpers). What do you think?

utils/prometheus/processor.py
aerostitch: Note for self: done

utils/prometheus/processor.py
```diff
@@ -0,0 +1,95 @@
...
+    def __init__(self, conf):
+        # metrics_mapper is a dictionnary where the keys are the metrics to capture and the values are the corresponding metrics names to have in datadog.
+        self.metrics_mapper = {}
+        # send_histograms_buckets is used to specify if yes or no you want to send the buckets as tagged values when dealing with histograms.
+        self.send_histograms_buckets = True
+        # endpoint is the metrics endpoint to use to poll metrics from Prometheus
+        self.endpoint = 'http://localhost:10055/metrics'
+        # Conf from the agent
+        self.log = conf.log
+        self.gauge = conf.gauge
```
aerostitch: Note for self: done

utils/prometheus/processor.py
```diff
@@ -0,0 +1,95 @@
...
+            else:
+                getattr(self, message.name)(message, **kwargs)
+        except AttributeError as err:
+            self.log.debug("Unable to handle metric: {} - error: {}".format(message.name, err))
+
+    def poll(self):
+        """
+        Polls the metrics from the prometheus endpoint
+        """
+        headers = {
```
aerostitch: Note for self: support of custom headers and the 2 formats done in the query call. Need to validate that `parse_metric_family` is compatible with the output provided in the text format (probably not, so moving that function inside the class and modifying it to accept the different will probably be in the TODO).

utils/prometheus/processor.py
```diff
@@ -0,0 +1,95 @@
...
+            'accept': 'application/vnd.google.protobuf; proto=io.prometheus.client.MetricFamily; encoding=delimited',
+            'accept-encoding': 'gzip',
+        }
+        req = requests.get(self.endpoint, headers=headers)
+        req.raise_for_status()
+        return req.content
+
+    def _submit_gauge(self, metric_name, val, metric, custom_tags=[]):
```
aerostitch: Not sure what you want to do there. If we want a filter to not use certains labels or if you want to overwrite certain labels.
If we want to use a filter based on the label, we could just do:
```python
    def _submit_gauge(self, metric_name, val, metrics, label_filters=None, custom_tags=[]):
        """
        Submit a metric as a gauge, additional tags provided will be added to
        the ones from the label provided via the metrics object.
        If the `label_filters` list is provided, the metrics labels available in
        the label_filters will not be assigned to the metric.
        """
        tags = custom_tags
        for label in metrics.label:
            if label_filters is None or label not in label_filters:
                tags.append('{}:{}'.format(label.name, label.value))
        self.gauge('{}.{}'.format(self.NAMESPACE, metric_name), val, tags)
```
Right?

aerostitch: Oh wait, ok, the comment explanation just arrived to my brain, took a long time, sorry!
Will do.

utils/prometheus/processor.py
```diff
@@ -0,0 +1,95 @@
...
+        """
+        prometheus_tags = ['{}:{}'.format(label.name, label.value) for label in metric.label]
+        tags = prometheus_tags + custom_tags
+        self.gauge(metric_name, val, tags)
+
+    def _submit_metric(self, metric_name, message):
+        """
+        For each metric in the message, report it as a gauge with all labels as tags
+        except if a labels dict is passed, in which case keys are label names we'll extract
+        and corresponding values are tag names we'll use (eg: {'node': 'node'}).
```
aerostitch: Note for self: done.

aerostitch:
Updated the PR according to the review. Let me know if it's fine with you now.
Will add the tests this afternoon (I need to check the case of parsing the metrics pulled using text format too).

aerostitch:
Sorry I've been sidetracked a lot yesterday and today. Will push the tests asap.

FYI I tested the parse_metric_family function with the text format and of course it failed! ;)
Not sure if treating the text format should be part of this PR or should be another feature. Not sure what it brings as prometheus always comes with the protobuf format support AFAIK and with the abstraction provided by the helpers, I don't think the format we use to query the prometheus endpoint really matters to the end user.
I'll do a 1st pass without integrating this and try to integrate them right after I got something well tested with the protobuf format.

aerostitch:
@hkaj I rewrote the class to be a motherclass based on AgentCheck.
Did the changes you asked me (except the decoding of the text format that will come probably tomorrow), added tests.
I also imported parse_metric_family into the class because I'm going to change it to support both text and protobuf format input. We can deprecate the other one I guess (it would make more sense to me anyway).
Can you review the new version of the implementation when you have some time please?
Thanks
Joseph

hkaj:
Thanks for the update @aerostitch !
We can keep text format for later. Just left a few more notes.

checks/prometheus_check.py
```diff
@@ -0,0 +1,232 @@
...
+# its metrics will count as cutom metrics and WILL impact billing.
+#
+# Minimal config for checks based on this class include:
+#   - implementing the check method
+#   - overriding self.NAMESPACE
+#   - overriding self.metric_to_gauge
+#     AND/OR
+#   - create method named after the prometheus metric they will handle (see self.prometheus_metric_name)
+#
+# Check class example:
```
aerostitch: Sure, no problem, that was in the pipe. Removing the example from the comment.

checks/prometheus_check.py
```diff
@@ -0,0 +1,232 @@
...
+        Submit a metric as a gauge, additional tags provided will be added to
+        the ones from the label provided via the metrics object.
+
+        If the `labels_mapper` dictionnary is provided, the metrics labels names
+        in the `labels_mapper` will use the corresponding value as tag name.
+        """
+        _tags = custom_tags
+        if _tags is None:
+            _tags = []
+        for label in metric.label:
```
aerostitch: Sure, no problem. Changing that.

aerostitch: Basically it would end up like:
``` _tags = custom_tags
        if _tags is None:
            _tags = []
        for label in metric.label:
            if exclude_labels is not None and label.name not in exclude_labels:
                tag_name = label.name
                if labels_mapper is not None and label.name in labels_mapper:
                    tag_name = labels_mapper[label.name]
                _tags.append('{}:{}'.format(tag_name, label.value))
        self.gauge('{}.{}'.format(self.NAMESPACE, metric_name), val, _tags)
```
Makes sense?

aerostitch:
Hi @hkaj I did the requested changes and almost finished to add the support for the parsing of the text format (currently the summary quantiles and histogram buckets values are represented as labels, but I should be able to fix that tomorrow and add the test).
In the mean time, if you have the time to have a look and see stuffs you don't like, please let me know.
Thanks

checks/prometheus_check.py
```diff
@@ -0,0 +1,293 @@
...
+                    yield "{} {}:{}".format(obj_map[_m], _m, _obj)
+        else:
+            raise self.UnknownFormatError('Unsupported content-type provided: {}'.format(content_type))
+
+    def _extract_labels_from_string(self,labels):
+        """
+        Extracts the labels from a string that looks like:
+        {label_name_1="value 1", label_name_2="value 2"}
+        """
+        lbl_pattern = re.compile(r'(\w+)="(.*?)"')
```
hkaj: this should be compiled in `__init__` and stored in `self`

checks/prometheus_check.py
hkaj: Yes that's what we'll need eventually. I would have kept it for later but since you already did it... ;)

Could you define labels_mapper and exclude_labels the same way as metric_mapper (hardcoded in the class)? We'll loose some flexibility by doing so, but it's a tradeoff that will tag names predictable which is required for official integrations (this way we can provide better default dashboards).

aerostitch:
Hi @hkaj!

Thanks a lot for the review again.
Sorry for the late update I've been slammed at work on Friday.
I've pushed an update on the PR which take in account your remarks and finish the mother class including the support of both protobuf and text format for pulling metrics from prometheus and the corresponding tests.

Let me know if you have remarks on this version.

Thanks a lot! :)
Joseph

checks/prometheus_check.py
```diff
@@ -0,0 +1,339 @@
...
+# It must be noted that if the check implementing this class is not officially
+# supported
+# its metrics will count as cutom metrics and WILL impact billing.
+#
+# Minimal config for checks based on this class include:
+#   - implementing the check method
+#   - overriding self.NAMESPACE
+#   - overriding self.metric_to_gauge
+#     AND/OR
+#   - create method named after the prometheus metric they will handle (see self.prometheus_metric_name)
```
hkaj: we also removed that example, either re-introduce it as it was written in https://github.com/DataDog/dd-agent/commit/9be9af8bd49157c2985b29359ebf9c0707906383#diff-b97e21cc8c9598269fd1833ccdd51ae3R154 or as a commented-out method -- it could be something like 
```python
def prometheus_metric_name(self, message, **kwargs):
    if message.type < len(self.METRIC_TYPES):
        for metric in message.metric:
            val = getattr(metric, self.METRIC_TYPES[message.type]).value
            tags = ['{}:{}'.format(label.name, label.value) for label in metric.label]
            self.gauge("datadog.metric.name", val, tags)
```
Or just remove the mention to it.

checks/prometheus_check.py
```diff
@@ -0,0 +1,339 @@
...
+# to collect metrics, events and service checks exposed via Prometheus.
+#
+# It must be noted that if the check implementing this class is not officially
+# supported
+# its metrics will count as cutom metrics and WILL impact billing.
+#
+# Minimal config for checks based on this class include:
+#   - implementing the check method
+#   - overriding self.NAMESPACE
+#   - overriding self.metric_to_gauge
```
aerostitch: Fixed.

aerostitch:
Good catch @hkaj , thanks! :)
I fixed the issues seen in the latest remarks.
Thanks! :)

hkaj:
Awesome! Thank you for your hard work on this PR. It is now ready to go and scheduled for 5.14. I'll now write a couple checks using it so we have examples available for users when it goes out.

discordianfish:
Can someone clarify whether this can be used on it's own? Is it possible to scrape arbitrary prometheus metrics endpoints as a datadog customer? Or is it necessary to add a custom check to the agent?

aerostitch:
@discordianfish You need to create a custom check to push arbitrary metrics in any case but it's really easy (just a few lines).
You can have a look at https://github.com/DataDog/integrations-core/blob/master/kube_dns/check.py as an example.

discordianfish:
@aerostitch Thanks, so basically like documented here? https://docs.datadoghq.com/guides/agent_checks/ - Haven't used custom agent checks so far.

aerostitch:
Yes. You can copy the script I mentioned earlier to the location used for custom checks, change the class name, the metrics and namespace (and give a value to the parameters) and you should be all set.

aerostitch:
@hkaj I've had several people I know hitting me directly asking the same question. Do you want me to document it somewhere?

hkaj:
Thanks for the ping @aerostitch 
I'll queue some work on documenting this. Will tag you for review if that's okay with you :)

MOZGIII:
Any update on the docs? I was looking for the docs everywhere and cound't find anythng except for this MR. Maybe just a hint on how to simply add prometheus metrics to the system?

aerostitch:
@MOZGIII did you check my 3 previous responses? That should give you an example and a good start on how to use them. I don't believe that @hkaj had the time yet to put together the documentation mentionned earlier but with my 3 previous responses you should be able to easily have a working integration that pulls metrics out of prometheus.

hkaj:
Hey @MOZGIII Sorry this is still in the backlog. If this comment thread doesn't help you feel free to contact our support and we'll help you set this up.

aerostitch:
In the mean time I wrote a quick how-to on creating a custom check that polls the metrics out of prometheus (that's just an explanation of the check and a copy/paste of some explanations that exist in the PrometheusCheck class mainly): https://github.com/aerostitch/aerostitch.github.io/blob/master/misc/datadog_pulling_metrics_from_prometheus.md
@MOZGIII @discordianfish let me know if that doc answers all your questions.
@hkaj let me know when you're all good with your doc, I'll put mine down.

discordianfish:
Yeah it was quite straight forward. Thanks!
What I'm wondering in general: Why not support pulling metrics from prometheus without having to create these 'boilerplate-only' checks?

I've created https://github.com/latency-at/datadog-agent-prometheus as a generic exporter but it would be easy to support this directly in dd-agent instead.

aerostitch:
Yes, that's what I wanted to do in the 1st place but it cannot be done like that. Datadog needs the full list of metrics that you're going to send if you want the metrics not to be considered as custom metrics (the pricing differs, which is the main blocker if I recall).
That's the reason why such a generic check could only be a custom check and not an official supported one.

MOZGIII:
@aerostitch, hohestly I didn't really read through all the messages, but I've checked out the kube dns check and got some info from there. Seems like you refer to that as well.
@discordianfish I'll check out your solution, cause I guess I'd write something like that for my needs. Looks like it already does what I wanted initially.
I still have to figure out what's the best way to put all this into a docker container without config volumes to just build and run.

MOZGIII:
@aerostitch that's interesting, so, that means datadogs pricing will be higher if I just put prometheus metrics into it?

aerostitch:
You pay the plan that includes a certain number of custom metrics allowed. The checks that come from integration-core like kube-dns that are fully supported by datadog and are not considered as custom metrics hence are not part of this bucket so you can use them without requiring a plan upgrade (that's the "pricing" difference I was referencing). See: https://help.datadoghq.com/hc/en-us/articles/204271775-What-is-a-custom-metric-and-what-is-the-limit-on-the-number-of-custom-metrics-I-can-have-

discordianfish:
Yeah as I understood it, it's a custom metric either way. Correct?
Or can I add a check similar to kube-dns etc and get them not to count into my custom metrics?

MOZGIII:
I see, s oif I want to just capture everything with prometheus I'll have to somehow convert the metrics captured to the datadogs compatible format? Like, something generic, i.e. host memory usage.

discordianfish:
@MOZGIII  That's what this here + my bridge does. But I would expect this to count towards the custom metrics, so you'd really need to reduce the number of metrics you keep..

aerostitch:
@discordianfish If you don't want your metrics to be custom metrics, you write something similar to kube-dns and have it integrated inside integration-core (I wrote the prometheus integration to be able to integrate kube-dns metrics as non-custom metrics). As long as it's not there, that's correct, it'll be a custom metric.

@MOZGIII I'm not entirely sure that not having metrics through the normal agent polls and getting all through a custom script, even if they have the same name in the end, I'm pretty sure they'd still be custom metrics. But I'm not part of Datadog, just a contributor so not 100% sure.

discordianfish:
Got it! So I probably should try to get as much stuff upstreamed as possible. :)

MOZGIII:
Found relevant code, but to busy to read it it through: https://github.com/DataDog/dd-agent/blob/master/checks/collector.py#L261 Jost mentioning it here for later.
Maybe you'll find this interesting too.

hkaj:
@MOZGIII @discordianfish @aerostitch We're also working on a check that can be configured through yaml to collect metrics automatically without writing boilerplate checks. Something similar to jmxfetch (but less hairy).

Using it with your own apps will result it custom metrics, but upstreaming  to `integrations-core` a check that has it configured for a specific program will make these metrics "standard" and included in your subscription.

For example if you write a config for this check that monitors a micro service of yours it's custom, but if you write one to collect metrics about influxdb and upstream it, we'll add it to the supported integrations and it will be included in everyone's subscription.

Actually I think this check will be similar to your implementation @discordianfish :)
The limitation here is that not all prometheus endpoints are straight-forward to parse and convert to our format, so some checks will still need to be written manually. We also need to support use cases like label/metrics renaming/blacklisting that are often requested by users. kube-dns is a good example of a check we'll be able to port to this new logic. OTOH I doubt kube-state-metrics will be supported as its collection is not as streamlined.

hkaj:
@aerostitch thanks a lot for the documentation! I pointed our doc team to it, i'm sure that will help them.

MOZGIII:
@hkaj I'm really looking forward to the usecase with fully-automated autodiscovery of prometheus metrics, is it something you're planing to work on?

hkaj:
Not myself, but you can expect to see PRs about it in a few weeks from @mfpierre 😉

PR #3318, opened 2017-04-25
Add warnings of breaking change coming with agent 6

zippolyte:
### What does this PR do?

Add warnings about a breaking change coming with agent 6 regarding the config of jmx checks

### DIFF
--- a/jmxfetch.py
+++ b/jmxfetch.py
@@ -439,7 +439,7 @@ class JMXFetch(object):
             if custom_jar_paths:
                 if isinstance(custom_jar_paths, basestring):
                     log.warn('Using a string when having only one custom jar will be deprecated in future versions' +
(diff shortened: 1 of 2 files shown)


jmxfetch.py
```diff
@@ -438,6 +438,8 @@ def _is_jmx_check(check_config, check_name, checks_list):
 
             if custom_jar_paths:
                 if isinstance(custom_jar_paths, basestring):
+                    log.warn('Using a string when having only one custom jar will be deprecated in future versions' +
+                        ' of the agent. Only the list syntax will be supported. %s' % LINK_TO_DOC)
```
olivielpeau: could you use this syntax: `log.warn("foo bar %s", LINK_TO_DOC)` instead? (internally, it only formats the string if the log entry is actually logged, we should try to prefer this syntax as much as possible)

PR #3319, opened 2017-04-26
[kubernetes_state] Record pod phase

ojongerius:
### What does this PR do?
Record the fact that we a pod is in a certain phase.

### Motivation
We like to monitor the amount of pending pods, the service check does not give the amount of control we like.

### DIFF
--- a/utils/kubernetes/kube_state_processor.py
+++ b/utils/kubernetes/kube_state_processor.py
@@ -131,7 +131,6 @@ class KubeStateProcessor:
         """ Phase a pod is in. """
         check_basename = NAMESPACE + '.pod.phase.'
         for metric in message.metric:
-            # The gauge value is always 1, no point in fetching it.
             phase = ''
             tags = []
             for label in metric.label:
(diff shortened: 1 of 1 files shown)


xvello:
Sister PR for documentation: https://github.com/DataDog/integrations-core/pull/395

hkaj:
Hi @ojongerius
Unfortunately we'll have to revert this PR at least for now because it creates a lot of new contexts (combination of metrics and tags) for very little added value. Can you tell us more about your use case? If it's only to display the value you can use the check status widget in screenboards to display it (and the equivalent in monitors to alert on it). If you need more details than that I would be interested in knowing what exactly.
Sorry about the inconvenience.

therc:
The original motivation says "to monitor the amount of pending pods", but IIUC this PR monitors the identity of the pending pods, hence the explosion in context count. The phase of each individual pod is a noisy metric, would it be possible to aggregate them by deployment name, namespace and phase before reporting them?
In other words, knowing that webserver-xrrqt and webserver-pgkmn are pending is not very useful (how do you even create a monitor for that?), while knowing that there are two pending deployment=webserver pods is.

PR #3320, opened 2017-04-26
adds bosh_settings to excluded tags, as it frequently includes passwords

gmmeyer:
bosh_settings frequently includes passwords. This will blacklist it.


gmmeyer:
I looked at the logging output and can confirm that it comes from the cloud_metadata (linked in private as I don't wanna post a password here).

PR #3326, opened 2017-05-02
[docker] fix image name when using sha256 for specs

xvello:
### What does this PR do?

Some orchestrators init containers with the image sha256 instead of the
image name, and docker inspect returns this. If it's the case, make
a docker image inspect call to get the image name and cache the result

No cache invalidation is coded as image deletion don't trigger events and obsolete image count should be significant on a typical host

### Motivation

Compatibility with k8s 1.6 & Nomad

### Testing Guidelines

Added unit test for the three cases we support

### DIFF
--- a/utils/dockerutil.py
+++ b/utils/dockerutil.py
@@ -481,20 +481,20 @@ class DockerUtil:
                     else:
                         image_spec = self.client.inspect_image(image)
                         try:
-                            name = image_spec.get('RepoTags')[0]
+                            name = image_spec['RepoTags'][0]
                             self._image_sha_to_name_mapping[image] = name
                             return name
-                        except Exception:
-                            pass
+                        except (LookupError, TypeError) as e:
+                            log.debug("Failed finding image name in RepoTag, trying RepoDigests: %s", e)
                         try:
-                            name = image_spec.get('RepoDigests')[0]
+                            name = image_spec['RepoDigests'][0]
                             name = name.split('@')[0]   # Last resort, we get the name with no tag
                             self._image_sha_to_name_mapping[image] = name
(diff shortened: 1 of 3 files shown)


utils/dockerutil.py
```diff
@@ -459,6 +463,42 @@ def image_tag_extractor(cls, entity, key):
...
+                try:
+                    if image in self._image_sha_to_name_mapping:
+                        return self._image_sha_to_name_mapping[image]
+                    else:
+                        image_spec = self.client.inspect_image(image)
+                        try:
+                            name = image_spec.get('RepoTags')[0]
+                            self._image_sha_to_name_mapping[image] = name
+                            return name
+                        except Exception:
```
hkaj: This exception handling is risky. Either we expect a precise error and we should only catch that, or we catch everything with `Exception` but log something. Otherwise we open the door to sneaky bugs.

utils/dockerutil.py
```diff
@@ -459,6 +463,42 @@ def image_tag_extractor(cls, entity, key):
...
+                            pass
+                        try:
+                            name = image_spec.get('RepoDigests')[0]
+                            name = name.split('@')[0]   # Last resort, we get the name with no tag
+                            self._image_sha_to_name_mapping[image] = name
+                            return name
+                        except Exception:
```
hkaj: diitto

utils/dockerutil.py
```diff
@@ -459,6 +463,42 @@ def image_tag_extractor(cls, entity, key):
...
+                            pass
+                except Exception:
```
hkaj: ditto

PR #3331, opened 2017-05-09
[utils][proxy] adding a utility function to deal with proxy configs+skipping

truthbk:
### What does this PR do?

Adds a couple of helper utility functions to avoid having code duplication in several places (dd-agent, checks that wish to disable the proxy config, etc).

### Motivation

While adding the `skip_proxy` feature to rabbitMQ: https://github.com/DataDog/integrations-core/pull/407, it made plently of sense to add the common code to a util package.

### Additional Notes

Merge before the sister PR.

Also sneaks in some fixes to Travis. CI was broken.

### DIFF
--- a/checks/__init__.py
+++ b/checks/__init__.py
@@ -29,12 +29,13 @@ import yaml
 
 # project
 from checks import check_status
-from config import AGENT_VERSION
+from config import AGENT_VERSION, _is_affirmative
 from util import get_next_id, yLoader
 from utils.hostname import get_hostname
 from utils.proxy import get_proxy
 from utils.platform import Platform
 from utils.profile import pretty_statistics
+from utils.proxy import get_no_proxy_from_env, config_proxy_skip
 if Platform.is_windows():
     from utils.debug import run_check  # noqa - windows debug purpose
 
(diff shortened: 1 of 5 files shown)


checks/__init__.py
```diff
@@ -390,6 +391,12 @@ def set_check_version(self, manifest=None):
 
         self.check_version = version
 
+    def get_instance_proxy(self, instance, uri):
+        proxies = self.proxies.copy()
+        proxies['no'] = get_no_proxy_from_env()
+
+        return config_proxy_skip(proxies, uri, _is_affirmative(instance.get('skip_proxy', False)))
```
truthbk: We could add the code in `utils` here, but since they're functions that operate over the proxy dicts we pass on to `requests` they have some utility out of the `AgentCheck()` scope.

utils/proxy.py
```diff
@@ -70,3 +74,23 @@ def get_proxy(agentConfig):
                   "Proxy is probably not set", str(e))
 
     return None
+
+
+def config_proxy_skip(proxies, uri, skip_proxy=False):
```
olivielpeau: Could you add a comment explaining what this function does?

Also, a unit test of this function in `tests/core/test_proxy.py` would be perfect :)

Issue #3340, opened 2017-05-12
Kubernetes Service discovery don't allow multiple instances or checks on the same pod annotation


aerostitch:
So I'm tracking down an issue where, when I add a 2nd instance to an annotation for a given check, only the 1st one is taken.
Basically you define the checks in an array with the following annotation: `service-discovery.datadoghq.com/rds-metrics.check_names`
Then you define the instances of your checks in another array in the following annotation: `service-discovery.datadoghq.com/rds-metrics.instances` which would contain a hash for each instance you want.

That's fine if you want 1 instance, 1 check. The problem gets tricky when you define a 2nd instance because in `abstract_config_store.py`, the `get_check_tpls` will do a:
```python
return [(source, vs)
          for vs in zip(check_names, init_config_tpls, instance_tpls)]
```
Which, is a good idea if you want to define several checks. Check 1 is related to Instance 1, check 2 to instance 2.

So I tell myself, sure, I want multiple instance in a single mysql check, so I'll do an array of instances which would look like (as we've already established that an array of hash don't solve my problem, we need an array of arrays of hash):
```
service-discovery.datadoghq.com/rds-metrics.instances: '[[{"pass": "pwd1", "port": 3306, "server": "rds_instance_1","tags": ["dbinstanceidentifier:rds_instance_1"], "user": "user1"},
{"pass": "pwd3", "port": 3306, "server": "rds_instance_3","tags": ["dbinstanceidentifier:rds_instance_3"], "user": "user3"},
{"pass": "pwd2", "port": 3306, "server": "rds_instance_2","tags": ["dbinstanceidentifier:rds_instance_2"], "user": "user2"}]]'
```

But when you do that, the collector ends up not collecting for this pod and throwing the following error in the log:
```
2017-05-12 02:50:32 UTC | ERROR | dd.collector | utils.service_discovery.sd_docker_backend(sd_docker_backend.py:376) | Building config for container 0a6c7d126ff7 based on image gcr.io/google_containers/pause-amd64:3.0 using service discovery failed, leaving it alone.
Traceback (most recent call last):
  File "/opt/datadog-agent/agent/utils/service_discovery/sd_docker_backend.py", line 361, in get_configs
    check_configs = self._get_check_configs(state, cid, identifier) or []
  File "/opt/datadog-agent/agent/utils/service_discovery/sd_docker_backend.py", line 394, in _get_check_configs
    config_templates = self._get_config_templates(identifier, **platform_kwargs)
[...]

PR #3341, opened 2017-05-12
Fix multi-instance service discovery

aerostitch:
Fixes: #3340

As explained in  #3340, having multiple instances for the same check in the annotations don't work well.
This brings the few adjustments to make it work.

The doc will need to be updated to say that if you want several instances of the same check in the annotations, you'll need to have an array for then inside the annotations array.

### DIFF
--- a/tests/core/test_service_discovery.py
+++ b/tests/core/test_service_discovery.py
@@ -116,13 +116,18 @@ class TestServiceDiscovery(unittest.TestCase):
         # image_name: ([(source, (check_name, init_tpl, instance_tpl, variables))], (expected_config_template))
         'image_0': (
             [('template', ('check_0', {}, {'host': '%%host%%'}, ['host']))],
-            ('template', ('check_0', {}, {'host': '127.0.0.1', 'tags': [u'docker_image:nginx', u'image_name:nginx']}))),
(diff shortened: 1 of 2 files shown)


aerostitch:
@hkaj Do you think we could setup this one for 5.14?
Thanks! :)

aerostitch:
Note: I've got this running in one of my environments. Works fine with a multi-instance declared via an annotation.

aerostitch:
Note: this will also require a fix for the page https://docs.datadoghq.com/guides/autodiscovery/
The current examples should be adjusted to be proper arrays.

aerostitch:
Hi @hkaj do you think we can plan that one for 5.14?
Thanks

aerostitch:
Makes sense @hkaj for the requested change. I update the PR.
I don't really think it's a feature, more a bug no?
If 5.14 is frozen, shouldn't we have that in 5.14.1?

hkaj:
Thanks, LGTM. Will merge once 5.14.x is branched off. It's more of a feature because it was explicitly designed for a single instance and that was the only configuration that was supported. Moved it to 5.14.1.

aerostitch:
Thanks a lot! :)

PR #3344, opened 2017-05-15
[Prometheus] Automatically add instance tags as custom tags and pass …

aerostitch:
…along the send_buckets_histograms parameter

### What does this PR do?

Automatically pulls the "tags" entry from the instance and push them to the underlying functions as "custom tags" so that they are sent with the gauge. This way the logic doesn't have to be added to each child class.
Also noticed a bug where we didn't pass the send_histograms_buckets from process to process_metrics which is inconvenient.

### Motivation

Wanted to have the same logic as other checks but built-in (I needed to add custom tags in some integrations).

### Additional Notes

Added unit tests for the bug and for the custom tags capture.

@hkaj FYI, as you wanted to write integrations based on this class, you might want to use this to avoid repeating code for adding the custom tags! ;)
The bug could be quite annoying, so would be great if we can have it scheduled for 5.14.
Thanks! :)

Note: I've packaged this and the kube_dns integration and am rolling that in my environments, so I'll push more PR if I notice other bugs.

### DIFF
--- a/checks/prometheus_check.py
+++ b/checks/prometheus_check.py
@@ -225,12 +225,18 @@ class PrometheusCheck(AgentCheck):
         """
         Polls the data from prometheus and pushes them as gauges
         `endpoint` is the metrics endpoint to use to poll metrics from Prometheus
+
+        Note that if the instance has a 'tags' attribute, it will be pushed
+        automatically as additionnal custom tags and added to the metrics
         """
         content_type, data = self.poll(endpoint)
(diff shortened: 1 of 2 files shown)


hkaj:
Hey @aerostitch 
Good catch for the bug! For the tags I completely agree, I realized we'd need it while reviewing your kube_dns PR.

aerostitch:
Thank you! :)

PR #3345, opened 2017-05-17
Support custom kubernetes label prefixes

### LINKED ISSUE #3321: [kubernetes] Support custom label prefixes
Currently, the `extract_kube_labels` function in `kubeutil.py` hard codes a `kube_` prefix on all pod labels. It would be great if this prefix were configurable (including allowing it to be blank) to give more control over how these labels appear in the Datadog UI.

The use case is that we're also exporting stats from pods (using their labels as tags), and we want the tags to match up in our Datadog dashboards; currently, because of the `kube_` prefix, the stats exported for the Docker Daemon and other integrations don't match up with the ones exported from the app itself.

travisjeffery:
Adds support for custom kubernetes label prefixes. Fixes #3321. 

The motivation is that we (Confluent) would like to remove the "kube_" prefix.

Thanks

### DIFF
--- a/utils/kubernetes/kubeutil.py
+++ b/utils/kubernetes/kubeutil.py
@@ -38,6 +38,7 @@ class KubeUtil:
     DEFAULT_HTTPS_KUBELET_PORT = 10250
     DEFAULT_MASTER_PORT = 8080
     DEFAULT_MASTER_NAME = 'kubernetes'  # DNS name to reach the master from a pod.
+    DEFAULT_LABEL_PREFIX = 'kube_'
     CA_CRT_PATH = '/var/run/secrets/kubernetes.io/serviceaccount/ca.crt'
     AUTH_TOKEN_PATH = '/var/run/secrets/kubernetes.io/serviceaccount/token'
 
@@ -80,6 +81,7 @@ class KubeUtil:
         self.kubelet_host = self.kubelet_api_url.split(':')[1].lstrip('/')
         self.pods_list_url = urljoin(self.kubelet_api_url, KubeUtil.PODS_LIST_PATH)
         self.kube_health_url = urljoin(self.kubelet_api_url, KubeUtil.KUBELET_HEALTH_PATH)
+        self.kube_label_prefix = os.environ.get('KUBERNETES_LABEL_PREFIX') or self.DEFAULT_LABEL_PREFIX
 
         # cadvisor
         self.cadvisor_port = instance.get('port', KubeUtil.DEFAULT_CADVISOR_PORT)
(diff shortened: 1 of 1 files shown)


utils/kubernetes/kubeutil.py
```diff
@@ -80,6 +81,7 @@ def __init__(self, instance=None):
         self.kubelet_host = self.kubelet_api_url.split(':')[1].lstrip('/')
         self.pods_list_url = urljoin(self.kubelet_api_url, KubeUtil.PODS_LIST_PATH)
         self.kube_health_url = urljoin(self.kubelet_api_url, KubeUtil.KUBELET_HEALTH_PATH)
+        self.kube_label_perfix = os.environ.get('KUBERNETES_LABEL_PREFIX') or self.DEFAULT_LABEL_PREFIX
```
jonchiu: typo here `self.kube_label_perfix`

utils/kubernetes/kubeutil.py
```diff
@@ -194,6 +196,7 @@ def extract_kube_labels(self, pods_list, excluded_keys=None):
         excluded_keys = excluded_keys or []
         kube_labels = defaultdict(list)
         pod_items = pods_list.get("items") or []
+        prefix = self.kube_label_perfix
```
jonchiu: same typo `self.kube_label_perfix`

utils/kubernetes/kubeutil.py
travisjeffery: 👍

utils/kubernetes/kubeutil.py
```diff
@@ -80,6 +81,7 @@ def __init__(self, instance=None):
         self.kubelet_host = self.kubelet_api_url.split(':')[1].lstrip('/')
         self.pods_list_url = urljoin(self.kubelet_api_url, KubeUtil.PODS_LIST_PATH)
         self.kube_health_url = urljoin(self.kubelet_api_url, KubeUtil.KUBELET_HEALTH_PATH)
+        self.kube_label_prefix = os.environ.get('KUBERNETES_LABEL_PREFIX') or self.DEFAULT_LABEL_PREFIX
```
xvello: For consistence, could you please use a configuration option in kubernetes.yaml instead of an environment variable? I suggest `label_to_tag_prefix`. If you need to configure it though an envvar (directly using our image), we could patch the entrypoint.

xvello:
Hi!

Sorry for the miscommunication, we are using the [kubernetes check's config](https://github.com/DataDog/integrations-core/blob/master/kubernetes/conf.yaml.example) to give settings to KubeUtil, this is what you get via the `instance` object. That's something we'll do differently in the next major version of the agent.

The documentation would be a PR on the integrations-core repo then. As I plan on using the same option for the label tags' kubernetes check (with a different default of `""` to keep backwards compatibility), I can take care of that, but a PR for the doc would be appreciated.

Regards

travisjeffery:
@xvello do you have a code example/link where the type of config that you want is being used?

xvello:
Don't worry about the conf option, I'll take care of it in https://github.com/DataDog/integrations-core/pull/434.
Thanks for your contribution!

PR #3346, opened 2017-05-19
Respect `ssl_skip_validation` when checking API key

### LINKED ISSUE #3303: sudo service datadog-agent info throwing CertificateError despite agent reporting just fine
We run all our Datadog agents behind an internal proxy. This appears to be working normally as they are reporting data just fine. 

However, when I run:

    sudo service datadog-agent info

I get the following exception:

    2017-04-11 21:06:29,353 | ERROR | dd.forwarder | checks.check_status(check_status.py:140) | Unable to validate API Key
      File "/opt/datadog-agent/agent/checks/check_status.py", line 130, in validate_api_key
        params={'api_key': config.get('api_key')}, proxies=request_proxy, timeout=3)
      File "/opt/datadog-agent/embedded/lib/python2.7/site-packages/requests/api.py", line 70, in get
        return request('get', url, params=params, **kwargs)
      File "/opt/datadog-agent/embedded/lib/python2.7/site-packages/requests/api.py", line 56, in request
        return session.request(method=method, url=url, **kwargs)
      File "/opt/datadog-agent/embedded/lib/python2.7/site-packages/requests/sessions.py", line 475, in request
        resp = self.send(prep, **send_kwargs)
      File "/opt/datadog-agent/embedded/lib/python2.7/site-packages/requests/sessions.py", line 596, in send
        r = adapter.send(request, **kwargs)
      File "/opt/datadog-agent/embedded/lib/python2.7/site-packages/requests/adapters.py", line 423, in send
        timeout=timeout
      File "/opt/datadog-agent/embedded/lib/python2.7/site-packages/requests/packages/urllib3/connectionpool.py", line 595, in urlopen
        chunked=chunked)
      File "/opt/datadog-agent/embedded/lib/python2.7/site-packages/requests/packages/urllib3/connectionpool.py", line 352, in _make_request
        self._validate_conn(conn)
      File "/opt/datadog-agent/embedded/lib/python2.7/site-packages/requests/packages/urllib3/connectionpool.py", line 831, in _validate_conn
        conn.connect()
      File "/opt/datadog-agent/embedded/lib/python2.7/site-packages/requests/packages/urllib3/connection.py", line 305, in connect
        _match_hostname(cert, self.assert_hostname or hostname)
      File "/opt/datadog-agent/embedded/lib/python2.7/site-packages/requests/packages/urllib3/connection.py", line 313, in _match_hostname
        match_hostname(cert, asserted_hostname)
      File "/opt/datadog-agent/embedded/lib/python2.7/site-packages/backports/ssl_match_hostname/__init__.py", line 147, in match_hostname
        % (hostname, ', '.join(map(repr, dnsnames))))
    CertificateError: hostname 'our.proxy.net' doesn't match either of '*.agent.datadoghq.com', 'agent.datadoghq.com'

I just wasted half an hour trying to figure out if this was a valid problem or just a cosmetic error. It appears, as best I can tell, to be only cosmetic. 

Rather than throwing an exception, could you instead check the value for `skip_ssl_validation` and if set to `yes` then swallow the exception? 

Additionally, if we are clearly running a proxy in the `dd_url` field than make the unable-to-validate-api-key message a little more detailed so it's easier to understand why:

    Unable to validate API key. You appear to be running a proxy because the value of
    dd_url (set in /etc/dd-agent/datadog.conf) doesn't match [expected value].

This error is appearing on agent version 5.12.

jeffwidman:
Tell the underlying `requests` library to only validate the SSL key when `ssl_skip_validation` is `False`.

Tested and working with `skip_ssl_validation` omitted/`True`/`False`.

Fix #3303

### DIFF
--- a/checks/check_status.py
+++ b/checks/check_status.py
@@ -127,7 +127,8 @@ def validate_api_key(config):
         if proxy:
             request_proxy = {'https': "http://{user}:{password}@{host}:{port}".format(**proxy)}
         r = requests.get("%s/api/v1/validate" % config['dd_url'].rstrip('/'),
-            params={'api_key': config.get('api_key')}, proxies=request_proxy, timeout=3)
+            params={'api_key': config.get('api_key')}, proxies=request_proxy,
+            timeout=3, verify=(not config.get('skip_ssl_validation', False)))
 
         if r.status_code == 403:
             return "API Key is invalid"


checks/check_status.py
```diff
@@ -127,7 +127,8 @@ def validate_api_key(config):
         if proxy:
             request_proxy = {'https': "http://{user}:{password}@{host}:{port}".format(**proxy)}
         r = requests.get("%s/api/v1/validate" % config['dd_url'].rstrip('/'),
-            params={'api_key': config.get('api_key')}, proxies=request_proxy, timeout=3)
+            params={'api_key': config.get('api_key')}, proxies=request_proxy, timeout=3,
+                         verify=(not config['skip_ssl_validation']))
```
truthbk: We have a small flake8 issue here: `E126 continuation line over-indented for hanging indent`

truthbk: Also even if all is apparently well, we're not 100% comfortable with saying "API Key is valid" when we cannot guarantee the connection hasn't been compromised. So if we could tweak the message [here line 144](https://github.com/DataDog/dd-agent/pull/3346/files#diff-74e3374e663661c0fc395cfe46d645d4R144) to say something along these lines (when SSL/TLS validation is disabled): `"API key appears valid. However, SSL validation has been disabled - please beware."` , I think would maybe be preferable.

jeffwidman: Yeah, those are both reasonable. Will fixup probably next week after Pycon. 

Also, I noticed you're using `config.get('api_key')` when I would have expected `config['api_key']` because the check doesn't really make sense without an API key. Is that on purpose?  

If so, should I be using `config.get('skip_ssl_validation', False)` as well? I haven't actually played with the agent yet, so not sure if it is already set to `False` if not explicitly set by the user.

jeffwidman: Also, I just tested this, and when `requests` is passed `verify=False`, it already throws a warning:

    /opt/datadog-agent/embedded/lib/python2.7/site-packages/requests/packages/urllib3/connectionpool.py:838: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/security.html InsecureRequestWarning)

Do you still want the extra warning? It seems superfluous since requests is now throwing this warning.

jeffwidman:
Just tested and confirmed working with `skip_ssl_validation` omitted/`True`/`False`. Ideally there would be automated tests for running the agent under these conditions, but not sure how to set those up.

jeffwidman-ns:
@truthbk when you get a chance can you address my questions? Or just merge this PR since we know we'll need this code regardless and then add your additional changes as a followup commit.

jeffwidman:
@truthbk when you get a chance can you address my questions? Or just merge this PR since we know we'll need this code regardless and then add your additional changes as a followup commit.

jeffwidman:
Awesome, thanks for pulling this in!

PR #3350, opened 2017-05-26
[SD] better docker tag extraction logic + tests

xvello:
- move SD docker tags extraction to dockerutil.extract_container_tags with error handling
- only inspect the container once in a shared c_inspect var
- update sd unit test fixtures
- extend test_dockerutil to test new method

### DIFF
--- a/utils/dockerutil.py
+++ b/utils/dockerutil.py
@@ -441,6 +441,25 @@ class DockerUtil:
 
         raise MountException("Cannot find Docker cgroup directory. Be sure your system is supported.")
 
+    def extract_container_tags(self, co):
+        """
+        Retrives docker_image, image_name and image_tag tags as a list for a
+        container. If the container or image is invalid, will gracefully
+        return an empty list
+        """
+        tags = []
+        docker_image = self.image_name_extractor(co)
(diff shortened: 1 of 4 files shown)


utils/dockerutil.py
```diff
@@ -441,6 +441,25 @@ def find_cgroup_filename_pattern(cls, mountpoints, container_id):
...
+        return an empty list
+        """
+        tags = []
+        docker_image = self.image_name_extractor(co)
+        image_name = self.image_tag_extractor(co, 0)
+        image_tag = self.image_tag_extractor(co, 1)
+
+        if docker_image:
+            tags.append('docker_image:%s' % docker_image)
+        if image_name and len(image_name) > 0:
```
masci: check on the string length should be redundant since `bool("")` is `False`

utils/dockerutil.py
```diff
@@ -441,6 +441,25 @@ def find_cgroup_filename_pattern(cls, mountpoints, container_id):
...
+            tags.append('image_name:%s' % image_name[0])
+        if image_tag and len(image_tag) > 0:
```
masci: same as above

PR #3365, opened 2017-06-01
Add container_name to additional tags

subuk:
Hello!

Sometimes, we have two docker containers with same image and same service but different container names on one physical host and we need to distinguish metrics from this two containers in datadog. Currently it is impossible, because metrics are not tagged by container name. This PR adds ability to use container name variable in auto_check templates, so you can add such tags.

We have applied this patch on our servers and it works.

Configuration example:

```
docker_images:
  - someimage
init_config:
instances:
  - nginx_status_url: 'http://%%host%%:8080/nginx_status/'
    tags:
      - 'container_name:%%containername%%'
```

### Tests
Looks like this functionality has already covered by tests.

### DIFF
--- a/utils/service_discovery/sd_docker_backend.py
+++ b/utils/service_discovery/sd_docker_backend.py
@@ -338,8 +338,19 @@ class SDDockerBackend(AbstractSDBackend):
 
         return tags
 
+    def _get_container_name(self, state, c_id):
+        container_inspect = state.inspect_container(c_id)
+        name = container_inspect.get('Name', '')
+        if name.startswith('/'):
+            name = name[1:]
+        return name
+
     def _get_additional_tags(self, state, c_id, *args):
         tags = []
+        container_name = self._get_container_name(state, c_id)
+        if container_name:
+            tags.append('container_name:%s' % container_name)
+
         if Platform.is_k8s():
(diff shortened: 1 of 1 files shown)


hkaj:
Hi @subuk 
Thanks for contributing, we'll review this shortly.

utils/service_discovery/sd_docker_backend.py
```diff
@@ -106,6 +106,7 @@ def __init__(self, agentConfig):
         self.VAR_MAPPING = {
             'host': self._get_host_address,
             'port': self._get_port,
+            'containername': self._get_container_name,
```
hkaj: Could you make it part of `%%tags%%` instead please? in `_get_additional_tags` you can call `_get_container_name` and add the `container_name:` prefix.

subuk: Yeah, sure

subuk:
Fixed. I have also noticed that configuration should looks like this to get additional tags work:
```
instances:
  - nginx_status_url: http://%%host%%:8080/nginx_status/
    tags: '%%tags%%'
```

subuk:
thanks @hkaj !

hkaj:
will be merged today as part of #3421 
Thanks again @subuk :)

PR #3375, opened 2017-06-06
Mesos tagging and generic container tagger

xvello:
### What does this PR do?

Introduces a generic container tagger mechanism to support all orchestrators through a common interface instead of expanding `SDDockerBackend.get_tags` and `DockerDaemon._get_tags` once again.

This introduces three classes:

- `Orchestrator.BaseUtil`: provides the common interface and the caching and container inspect logic
- `Orchestrator.MesosUtil`: inherits BaseUtil and just implements `_get_cacheable_tags` and `is_detected`
- `Orchestrator.Tagger`: triggers orchestrator detection and proxies calls to relevant BaseUtil classes

ECSUtil and NomadUtil will be moved to BaseUtil if this PR is approved.

### Test

unit tests in the test_orchestrator.py file

### DIFF
--- a/utils/orchestrator/baseutil.py
+++ b/utils/orchestrator/baseutil.py
@@ -25,10 +25,10 @@ class BaseUtil:
     __metaclass__ = Singleton
 
     def __init__(self):
-        # Whether your get___tags methods need the inspect result
-        self.needs_inspect = False
-        # Whether your methods need the env portion (not in partial inspect)
-        self.needs_env = False
+        # Whether your get___tags methods need the Config section inspect data
+        self.needs_inspect_config = False
+        # Whether your get___tags methods need the Labels section inspect data
+        self.needs_inspect_labels = False
 
         self.log = logging.getLogger(__name__)
         self.docker_util = DockerUtil()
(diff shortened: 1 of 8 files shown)


utils/orchestrator/baseutil.py
```diff
@@ -0,0 +1,85 @@
...
+                self.log.warning("Invalid container dict, returning empty tags")
+                return []
+
+        if cid in self._container_tags_cache:
+            return self._container_tags_cache[cid]
+        else:
+            if (self.needs_inspect or self.needs_env) and co is None:
+                co = self.docker_util.inspect_container(cid)
+            if self.needs_env and 'Env' not in co.get('Config', {}):
+                co = self.docker_util.inspect_container(cid)
```
hkaj: this feels a bit clunky. Maybe `if (self.needs_inspect or self.needs_env) and (co is None or 'Env' not in co.get('Config, {}))` and remove the second if?

xvello: The idea was to be able to accept the partial inspects (without env/labels) that the docker_daemon currently gets (needs_inspect = true / needs_env = false), but none of the ecs/rancher/nomad/mesos utils would work with a partial inspect anyway.

As the result is cached, I'll go with removing needs_env and move needs_inspect to check whether co != None and has the Env section

PR #3383, opened 2017-06-12
[docker] add docker/mesos version host tags

xvello:
### What does this PR do?

- add orchestrator.MetadataCollector.get_host_tags() and method in Mesosutil
- add DockerUtil.get_host_tags() for docker_version
- detect mesos and dcos agents via http/https tests
- retrieve docker_version, docker_swarm, mesos_version and dcos_version host tags
- add new collect_orchestrator_tags conf option (default True)

### Motivation

Allows easier troubleshooting for version-specific orchestrator issues

### Testing Guidelines

Unit tests in the CI

### DIFF
--- a/utils/dockerutil.py
+++ b/utils/dockerutil.py
@@ -239,8 +239,12 @@ class DockerUtil:
             tags.append('docker_version:%s' % version['Version'])
         else:
             log.debug("Could not determine Docker version")
+
         if self.is_swarm():
             tags.append('docker_swarm:active')
+        else:
+            tags.append('docker_swarm:inactive')
+
         return tags
 
     @property
--- a/utils/dockerutil.py
+++ b/utils/dockerutil.py
@@ -232,6 +232,17 @@ class DockerUtil:
 
         return self._default_gateway
 
+    def get_host_tags(self):
+        tags = []
+        version = self.client.version()
+        if version and 'Version' in version:
+            tags.append('docker_version:%s' % version['Version'])
+        else:
+            log.debug("Could not determine Docker version")
+        if self.is_swarm():
+            tags.append('docker_swarm:active')
+        return tags
+
     @property
     def client(self):
         return Client(**self.settings)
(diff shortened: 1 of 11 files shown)


utils/dockerutil.py
```diff
@@ -232,6 +232,17 @@ def get_hostname(self, use_default_gw=True, should_resolve=False):
...
 
+    def get_host_tags(self):
+        tags = []
+        version = self.client.version()
+        if version and 'Version' in version:
+            tags.append('docker_version:%s' % version['Version'])
+        else:
+            log.debug("Could not determine Docker version")
+        if self.is_swarm():
+            tags.append('docker_swarm:active')
```
hkaj: actually, if that's all we have maybe it's useless for now

utils/orchestrator/mesosutil.py
```diff
@@ -3,18 +3,33 @@
...
+MESOS_AGENT_HTTP_PORT = 5051
+DCOS_AGENT_HTTP_PORT = 61001
+DOCS_AGENT_HTTPS_PORT = 61002
+
+
+def MESOS_AGENT_VALIDATION(r):
+    return "version" in r.json()
+
+
+def DCOS_AGENT_VALIDATION(r):
```
hkaj: these 2 should be staticmethods

utils/dockerutil.py
xvello: Maybe not for the user, but for support and usage statistics I think that's relevant

utils/orchestrator/mesosutil.py
xvello: indeed, thanks

utils/dockerutil.py
hkaj: then let's add `inactive` as well? This way we could differentiate between users for whom we didn't collect it (error, old agent version) and those for whom we're sure it's inactive.

xvello: 👍

PR #3384, opened 2017-06-13
[jmxfetch][auto-discovery] fix windows handling of pipe.

truthbk:
### What does this PR do?

This PR attempts to fix the currently broken implementation of windows JMXfetch auto discovery based on named pipes. We must rely on `pywin32` to create, open the pipe. We can then use io.open() to use the file descriptor and get a regular file handle.

### Motivation

Ongoing work on auto-discovery made this issue surface.

### Testing Guidelines

Additional testing is necessary.

### DIFF
--- a/agent.py
+++ b/agent.py
@@ -16,6 +16,7 @@ initialize_logging('collector')
 # stdlib
 from copy import copy
 import logging
+import io
 import os
 import signal
 import sys
@@ -64,6 +65,12 @@ from utils.service_discovery.sd_backend import get_sd_backend
 from utils.watchdog import Watchdog
 from utils.windows_configuration import get_sdk_integration_paths
 
+# conditional win imports
+if Platform.is_windows():
+    import win32pipe
+    import win32file
+    import msvcrt
+
 # Constants
 PID_NAME = "dd-agent"
 PID_DIR = None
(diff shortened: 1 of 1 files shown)


agent.py
```diff
@@ -261,18 +267,39 @@ def run(self, config=None):
             pipe_path = get_jmx_pipe_path()
             if Platform.is_windows():
                 pipe_name = pipe_path.format(pipename=SD_PIPE_NAME)
+                exists = win32pipe.WaitNamedPipe(pipe_name,
+                                                 win32pipe.NMPWAIT_USE_DEFAULT_WAIT)
+                if not exists:
+                    pipe = win32pipe.CreateNamedPipe(pipe_name, win32pipe.PIPE_ACCESS_DUPLEX,
+                                            win32pipe.PIPE_TYPE_MESSAGE | win32pipe.PIPE_WAIT,
```
truthbk: not convinced this should be `PIPE_WAIT`... we're only writing from our end, so this should be fine either way..

olivielpeau:
Agent 6 (for all platforms including Windows) handles the communication between the Agent and JMXFetch with a network socket, which works well. So we can consider this PR deprecated and should encourage users to upgrade to Agent 6 for a fully-working JMXFetch + autodiscovery experience on Windows.

So I'll go ahead and close this :)

PR #3385, opened 2017-06-13
[checks] Deprecate `AgentCheck.set`

olivielpeau:
### What does this PR do?

Adds a `warning` message when the `set` method is used.

### Motivation

`AgentCheck.set` will be removed in Agent 6.0. Any custom check that currently uses it should be modified.

### Additional Notes

2 core checks currently use `AgentCheck.set`:
- `zk`: see https://github.com/DataDog/integrations-core/pull/473
- `docker`: check that's been deprecated since agent [`5.5.0`](https://github.com/DataDog/dd-agent/blob/master/CHANGELOG.md#550--09-17-2015), so we should probably remove it from the `integrations-core` altogether when Agent 6.0 is released

### DIFF
--- a/checks/__init__.py
+++ b/checks/__init__.py
@@ -563,6 +563,8 @@ class AgentCheck(object):
         :param hostname: (optional) A hostname for this metric. Defaults to the current hostname.
         :param device_name: (optional) The device name for this metric
         """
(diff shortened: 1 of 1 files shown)


masci:
`docker` check will be removed from the repo with 5.15

PR #3390, opened 2017-06-15
Update Mac restart command

nmuesch:
if you have not yet done so.*

### What does this PR do?

This PR modified the restart command of the Mac OSX packaged Agent. The restart command now checks to see if the supervisor process is still running, and waits up to 10 seconds before calling the start command.

### Motivation

Running the `stop` then `start` commands on Mac leads to this error quite often: `Error: Another program is already listening on a port that one of our HTTP servers is configured to use.  Shut this program down first before starting supervisord.`

This seems to be due to the supervisor process taking too long to fully exit and free up ports, so we wait up to 10 seconds for this to be done. 

Thanks @cmatteri For writing this restart

### Testing Guidelines

### Additional Notes

### DIFF
--- a/packaging/osx/datadog-agent
+++ b/packaging/osx/datadog-agent
@@ -89,6 +89,7 @@ case $1 in
         else
             echo "$DESC was not running"
         fi
+        sleep 5
         "$0" start
         ;;
 
--- a/packaging/osx/datadog-agent
+++ b/packaging/osx/datadog-agent
@@ -90,11 +90,14 @@ case $1 in
         else
             echo "$DESC was not running"
         fi
+        # This section checks to see if supervisor is still running before we attempt a start. We will wait 
+        # up to 10 seconds as this is how long Supervisor waits for processes to stop before force exiting using SIGKILL
+        # http://supervisord.org/configuration.html#program-x-section-values
         checks=0
         pid=$(cat $SUPERVISOR_PIDFILE 2>/dev/null)
         if [ -n $pid ]; then
-            while ps aux $pid | grep $SUPERVISOR_PROCESS >/dev/null && [ $checks -lt 10 ]; do
-                  let "checks = checks + 1"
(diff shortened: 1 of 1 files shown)


packaging/osx/datadog-agent
```diff
@@ -89,6 +90,14 @@ case $1 in
         else
             echo "$DESC was not running"
         fi
+        checks=0
```
olivielpeau: Could you add a comment explaining what you're doing in this block?

Also, it makes sense to wait 10 seconds because by default that's the time supervisor will wait for the child processes to stop gracefully before it force-kills them with a `SIGKILL` signal (see `stopwaitsecs` option in http://supervisord.org/configuration.html#program-x-section-values). Let's explain that in the comment :)

packaging/osx/datadog-agent
```diff
@@ -89,6 +90,14 @@ case $1 in
...
+        pid=$(cat $SUPERVISOR_PIDFILE 2>/dev/null)
+        if [ -n $pid ]; then
+            while ps aux $pid | grep $SUPERVISOR_PROCESS >/dev/null && [ $checks -lt 10 ]; do
+                  let "checks = checks + 1"
```
olivielpeau: could you fix the indentation here?

packaging/osx/datadog-agent
```diff
@@ -89,6 +90,14 @@ case $1 in
         else
             echo "$DESC was not running"
         fi
+        checks=0
+        pid=$(cat $SUPERVISOR_PIDFILE 2>/dev/null)
+        if [ -n $pid ]; then
+            while ps aux $pid | grep $SUPERVISOR_PROCESS >/dev/null && [ $checks -lt 10 ]; do
```
olivielpeau: (nitpick: we could test `[ $checks -lt 10 ]` first, and then do the `ps aux .....`)

nmuesch:
So, I made most of the requested changes but left the logic in the restart function.

PR #3394, opened 2017-06-21
[PrometheusCheck] add ignore_metrics array

xvello:
### What does this PR do?

Allows checks to define a `self.ignore_metrics` array that PrometheusCheck will ignore. This is used in the revamped kubernetes_state check from https://github.com/DataDog/integrations-core/pull/492

### Motivation

- Reduce log spam
- Improve performance
- While defining the array, check authors can comment why some metrics are left out on purpose (not relevant / high card / ...) and not just forgotten


xvello:
We'll need to eventually remove kube_state_processor.py, should that be a separate PR?

PR #3395, opened 2017-06-21
[core] Adds fraction of file handles in use over system limit

nmuesch:
### What does this PR do?

Adds a metric for the percent of allocated and used file handlers over the system's max. Uses `/proc/sys/fs/file-nr` as noted here - https://www.kernel.org/doc/Documentation/sysctl/fs.txt

This metric will be displayed as a fraction between 0 and 1, and is named: `system.fs.file_handles_in_use` to keep syntactically similar to `system.disk_in_use` which has similar units. 

This would look like this in app: https://cl.ly/3u2E2S3l2j0K

### Motivation

Users would like to be able to be alerted if the number of handlers being used was approaching the limit available. 

### Testing Guidelines

### Additional Notes

### DIFF
--- a/checks/collector.py
+++ b/checks/collector.py
@@ -206,7 +206,7 @@ class Collector(object):
             'processes': u.Processes(log),
             'cpu': u.Cpu(log),
             'system': u.System(log),
-            'fh': u.File_Handlers(log)
+            'file_handles': u.FileHandles(log)
         }
 
         # Win32 System `Checks
@@ -302,7 +302,7 @@ class Collector(object):
             # Unix system checks
             sys_checks = self._unix_system_checks
 
-            for check_name in ['load', 'system', 'cpu', 'fh']:
+            for check_name in ['load', 'system', 'cpu', 'file_handles']:
                 try:
                     result_check = sys_checks[check_name].check(self.agentConfig)
                     if result_check:
--- a/checks/collector.py
+++ b/checks/collector.py
@@ -205,7 +205,8 @@ class Collector(object):
             'memory': u.Memory(log),
             'processes': u.Processes(log),
             'cpu': u.Cpu(log),
-            'system': u.System(log)
+            'system': u.System(log),
+            'fh': u.File_Handlers(log)
         }
 
         # Win32 System `Checks
@@ -301,7 +302,7 @@ class Collector(object):
             # Unix system checks
             sys_checks = self._unix_system_checks
 
-            for check_name in ['load', 'system', 'cpu']:
+            for check_name in ['load', 'system', 'cpu', 'fh']:
                 try:
                     result_check = sys_checks[check_name].check(self.agentConfig)
                     if result_check:
--- a/checks/system/unix.py
+++ b/checks/system/unix.py
@@ -808,7 +808,7 @@ def main():
         print(mem.check(config))
         print("--- System ---")
         print(system.check(config))
-        print("--- File Handlers ---")
+        print("--- File Handles ---")
         print(fh.check(config))
         print("\n\n\n")
         # print("--- Processes ---")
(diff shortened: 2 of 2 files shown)


checks/system/unix.py
```diff
@@ -221,6 +221,28 @@ def check(self, agentConfig):
...
 
 
+class File_Handlers(Check):
+
+    def check(self, agentConfig):
+
+        if not Platform.is_linux():
+            return False
+
+        with open('/proc/sys/fs/file-nr', 'r') as file_handlers:
```
masci: let's put this within a try/except block, returning `False` in case of failure

checks/system/unix.py
```diff
@@ -221,6 +221,31 @@ def check(self, agentConfig):
             return False
 
 
+class File_Handlers(Check):
```
olivielpeau: nit: could you name this class `FileHandles` instead? Just to be consistent with the standard case of python classes, and also name it `handles` instead of `handlers`

checks/system/unix.py
```diff
@@ -221,6 +221,31 @@ def check(self, agentConfig):
...
+        handler_metrics = handler_contents.split()
+
+        # https://www.kernel.org/doc/Documentation/sysctl/fs.txt
+        allocated_fh = float(handler_metrics[0])
+        allocated_unused_fh = float(handler_metrics[1])
+        max_fh = float(handler_metrics[2])
+
+        fh_in_use = (allocated_fh - allocated_unused_fh) / max_fh
+
+        return {'system.file_handles_in_use': float(fh_in_use)}
```
olivielpeau: I _think_ we could name it `system.fs.file_handles.in_use` instead, what do you think? (we already have a few `system.fs.` metrics, I think it makes to put this new metric in that metric namespace too since it's collected from `/proc/sys/fs/` too)

checks/collector.py
```diff
@@ -205,7 +205,8 @@ def __init__(self, agentConfig, emitters, systemStats, hostname):
             'memory': u.Memory(log),
             'processes': u.Processes(log),
             'cpu': u.Cpu(log),
-            'system': u.System(log)
+            'system': u.System(log),
+            'fh': u.File_Handlers(log)
```
olivielpeau: just for the sake of having a clear log message if an error occurs during the check run, could we name the key `file_handles` instead?

checks/system/unix.py
```diff
@@ -221,6 +221,31 @@ def check(self, agentConfig):
...
+    def check(self, agentConfig):
+
+        if not Platform.is_linux():
+            return False
+
+        try:
+            with open('/proc/sys/fs/file-nr', 'r') as file_handlers:
+                handler_contents = file_handlers.read()
+        except Exception:
+            self.logger.exception("Cannot extract system file handler stats")
```
olivielpeau: `handlers` -> `handles`

could make sense to rename the variables too

checks/system/unix.py
```diff
@@ -221,6 +221,31 @@ def check(self, agentConfig):
...
 
+class File_Handlers(Check):
+
+    def check(self, agentConfig):
+
+        if not Platform.is_linux():
+            return False
+
+        try:
+            with open('/proc/sys/fs/file-nr', 'r') as file_handlers:
```
olivielpeau: on some systems the parent dir of the `sys/` directory is not `/proc`, could you use the value of `procfs_path` in the agent config if it's there? (unless there's a good reason to always use `/proc`)

See https://github.com/DataDog/dd-agent/blob/5.14.1/checks/system/unix.py#L228 for a similar example

nmuesch: Ahh, I didn't know that, thanks for the catch!

PR #3402, opened 2017-06-29
Add hostname detection for azure cloud provider

hush-hush:
### What does this PR do?

Detect an Azure host using the metadata instance service and use the VM ID as hostname.

### DIFF
--- a/utils/cloud_metadata.py
+++ b/utils/cloud_metadata.py
@@ -16,6 +16,36 @@ from utils.proxy import get_proxy
 
 log = logging.getLogger(__name__)
 
+class Azure(object):
+    URL = "http://169.254.169.254/metadata/instance?api-version=2017-04-02"
+    TIMEOUT = 0.3 # second
+
+    @staticmethod
+    def _get_metadata(agentConfig):
+        if not agentConfig.get('collect_instance_metadata', True):
+            log.info("Instance metadata collection is disabled: not collecting Azure metadata.")
+            return {}
+
+        try:
+            r = requests.get(
+                Azure.URL,
+                timeout=Azure.TIMEOUT,
+                headers={'Metadata': 'true'}
+            )
+            r.raise_for_status()
+            return r.json()
+        except Exception as e:
+            log.debug("Collecting Azure Metadata failed %s", str(e))
+            return None
+
+    @staticmethod
+    def get_host_aliases(agentConfig):
+        try:
+            host_metadata = Azure._get_metadata(agentConfig)
+            return [host_metadata['compute']['vmId']]
+        except Exception:
+            return []
+
 class GCE(object):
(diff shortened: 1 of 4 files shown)


utils/cloud_metadata.py
```diff
@@ -14,6 +14,36 @@
...
 log = logging.getLogger(__name__)
 
+class Azure(object):
+    URL = "http://169.254.169.254/metadata/instance?api-version=2017-04-02"
+    TIMEOUT = 0.3 # second
+
+    @staticmethod
+    def _get_metadata(agentConfig):
+        if not agentConfig.get('collect_instance_metadata', True):
+            log.info("Instance metadata collection is disabled. Not collecting it.")
```
olivielpeau: nit: could you make the message here more specific, `Not collecting Azure metadata` for example :)

utils/cloud_metadata.py
```diff
@@ -14,6 +14,36 @@
...
+            return {}
+
+        try:
+            r = requests.get(
+                Azure.URL,
+                timeout=Azure.TIMEOUT,
+                headers={'Metadata': 'true '}
```
olivielpeau: Is the trailing space in `'true '` intentional?

hush-hush:
@olivielpeau We could push it behind a configuration option (and making it the default behavior for the next major version of the agent) ?

olivielpeau:
@hush-hush: After discussing this with the various backend teams, we should send the vm id as a host alias (see https://github.com/DataDog/dd-agent/blob/5.14.1/checks/collector.py#L757-L759 for an example of how we add host aliases from GCP)

utils/hostname.py
```diff
@@ -72,6 +72,12 @@ def get_hostname(config=None):
         if is_valid_hostname(gce_hostname):
             return gce_hostname
 
+    # Try to get Azure VM ID
+    azure_hostname = Azure.get_hostname(config)
+    if azure_hostname is not None:
+        if is_valid_hostname(azure_hostname):
+            return azure_hostname
```
olivielpeau: Actually we don't want to change the canonical hostname that's used by the agent (as this could cause hosts to change names in the app, and would require a complex migration process on our part and our users' part).

Instead of using the azure VM ID as the hostname, we should only add it to the host aliases in the metadata payload here: https://github.com/DataDog/dd-agent/blob/5.14.1/checks/collector.py#L757-L759

Let me know if you agree/disagree or have any questions :)

hush-hush: Good, idea. I update the PR

utils/hostname.py
```diff
@@ -8,7 +8,7 @@
 import socket
 
 # project
-from utils.cloud_metadata import EC2, GCE
+from utils.cloud_metadata import EC2, GCE, Azure
```
olivielpeau: you should remove the import from here and add it to `checks/collector.py` (CI is failing on this)

Issue #157, opened 2012-08-17
If hostname is set in datadog.conf, don't call socket.gethostname for aliasing


alq666:
On debian, socket.gethostname returns only the first part of the fqdn, which means that hosts like:

s0.domainA
s0.domainB

are incorrectly considered to be aliases.

PR #3406, opened 2017-06-29
Get real host aliases for Cloud Foundry

gmmeyer:
### What does this PR do?

I added a few ways to get the proper host alias for cloud foundry, including some undocumented options in the config file and environment variables. I can use or remove any that people think is a good idea. But, basically it will detect if it's cloud foundry and then grab the proper host alias from a few places.

### Motivation

I want the firehose nozzle metrics to be associated with their respective host, this will make that easier to do.

### Testing Guidelines

### Additional Notes

### DIFF
--- a/checks/collector.py
+++ b/checks/collector.py
@@ -751,13 +751,13 @@ class Collector(object):
             except Exception:
                 pass
 
-        if self.agentConfig.get('use_socket_hostname_as_alias'):
-            try:
-                if not metadata["host_aliases"]:
-                    metadata["host_aliases"] = []
-                metadata["host_aliases"] += CloudFoundry.get_host_aliases
-            except Exception:
-                pass
+        if not metadata["host_aliases"]:
+            metadata["host_aliases"] = []
+
+        try:
+            metadata["host_aliases"] += CloudFoundry.get_host_aliases
+        except Exception:
+            pass
 
         try:
             metadata["socket-fqdn"] = socket.getfqdn()
@@ -770,8 +770,6 @@ class Collector(object):
         # Add cloud provider aliases
         host_aliases = GCE.get_host_aliases(self.agentConfig)
         if host_aliases:
-            if not metadata["host_aliases"]:
-                metadata["host_aliases"] = []
             metadata['host_aliases'] += host_aliases
 
         return metadata
--- a/checks/collector.py
+++ b/checks/collector.py
@@ -751,7 +751,7 @@ class Collector(object):
             except Exception:
                 pass
 
-        if not metadata["host_aliases"]:
+        if not metadata.get("host_aliases"):
             metadata["host_aliases"] = []
 
         try:
--- a/checks/collector.py
+++ b/checks/collector.py
@@ -745,11 +745,17 @@ class Collector(object):
 
         if self.agentConfig.get('hostname'):
             metadata['agent-hostname'] = self.agentConfig.get('hostname')
+        else:
+            try:
+                metadata["socket-hostname"] = socket.gethostname()
+            except Exception:
+                pass
 
-        try:
-            metadata["socket-hostname"] = socket.gethostname()
-        except Exception:
-            pass
+        if self.agentConfig.get('danger_always_add_socket_hostname'):
+            try:
+                metadata["socket-hostname"] = socket.gethostname()
+            except Exception:
+                pass
 
         try:
             metadata["socket-fqdn"] = socket.getfqdn()
--- a/checks/collector.py
+++ b/checks/collector.py
@@ -745,11 +745,12 @@ class Collector(object):
 
         if self.agentConfig.get('hostname'):
             metadata['agent-hostname'] = self.agentConfig.get('hostname')
-        else:
-            try:
-                metadata["socket-hostname"] = socket.gethostname()
-            except Exception:
-                pass
+
+        try:
+            metadata["socket-hostname"] = socket.gethostname()
+        except Exception:
+            pass
+
         try:
             metadata["socket-fqdn"] = socket.getfqdn()
         except Exception:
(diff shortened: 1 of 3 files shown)


olivielpeau:
I do understand the use case but I'm a bit concerned about the host aliasing nightmare that could happen if a user inadvertently sets the same host alias on 2 different agents...

Could you try setting the same host alias on 2 agents and check:
1. what happens in the app
2. that the host aliasing logic in the app recovers nicely once the common host alias is removed from the agent

gmmeyer:
This is a very reasonable objection. I've updated the collector with different hostname logic, this will also get the same thing. I'm going to see if I can get it another way as well, that will make sure that it'll work fine.

olivielpeau:
@gmmeyer I'm not sure I understand how your new change solves your issue, could you elaborate on that? :)

gmmeyer:
Oh, the hostname I looking for is available as the socket hostname

olivielpeau:
@gmmeyer ok I see thanks, hmm this change could have undesirable side effects (host aliases changing, and potentially being identical for different hosts) for people who have set a hostname in `datadog.conf` (depends on how the aliasing is done on the backend, and what different environments return for `socket.gethostname()`).

See https://github.com/DataDog/dd-agent/issues/157 for instance (found the issue while `git blame`-ing the current logic)

I'm not a fan of having even more config flags, but maybe we could add one here to force sending `socket-hostname` even when a `hostname` is set in `datadog.conf`.

gmmeyer:
ah fair enough

gmmeyer:
let me think about how to put this together

utils/cloud_metadata.py
```diff
@@ -226,3 +229,34 @@ def get_instance_id(agentConfig):
...
+            return CloudFoundry.host_aliases
+        if os.environ.get("DD_BOSH_ID"):
+            CloudFoundry.host_aliases.append(os.environ.get("DD_BOSH_ID"))
+        if agentConfig.get("bosh_id"):
+            CloudFoundry.host_aliases.append(agentConfig.get("bosh_id"))
+        if len(CloudFoundry.host_aliases) == 0:
+            # Only use this if the prior one fails
+            # The reliability of the socket hostname is not assured
+            CloudFoundry.host_aliases.append(socket.gethostname())
+        return []
```
olivielpeau: this should return `CloudFoundry.host_aliases` no?

utils/cloud_metadata.py
```diff
@@ -226,3 +229,34 @@ def get_instance_id(agentConfig):
...
+    host_aliases = []
+
+    @staticmethod
+    def get_host_aliases(agentConfig):
+        if not CloudFoundry.is_cloud_foundry(agentConfig):
+            return CloudFoundry.host_aliases
+        if os.environ.get("DD_BOSH_ID"):
+            CloudFoundry.host_aliases.append(os.environ.get("DD_BOSH_ID"))
+        if agentConfig.get("bosh_id"):
+            CloudFoundry.host_aliases.append(agentConfig.get("bosh_id"))
```
olivielpeau: should we append this one even if `DD_BOSH_ID` has already been appended?

utils/cloud_metadata.py
```diff
@@ -226,3 +229,34 @@ def get_instance_id(agentConfig):
...
+        if len(CloudFoundry.host_aliases) == 0:
+            # Only use this if the prior one fails
+            # The reliability of the socket hostname is not assured
+            CloudFoundry.host_aliases.append(socket.gethostname())
+        return []
+
+    @staticmethod
+    def is_cloud_foundry(agentConfig):
+        if agentConfig.get("cloud_foundry"):
+            return True
```
olivielpeau: quick additional question: why is this test not enough? are there cases where the `cloud_foundry` flag would not be present on an agent running on cloudfoundry and installed with the `boshrelease`?

checks/collector.py
```diff
@@ -750,6 +750,15 @@ def _get_hostname_metadata(self):
                 metadata["socket-hostname"] = socket.gethostname()
             except Exception:
                 pass
+
+        if not metadata.get("host_aliases"):
+            metadata["host_aliases"] = []
+
+        try:
+            metadata["host_aliases"] += CloudFoundry.get_host_aliases
```
gmmeyer: boop yes

checks/collector.py
```diff
@@ -750,6 +750,15 @@ def _get_hostname_metadata(self):
                 metadata["socket-hostname"] = socket.gethostname()
             except Exception:
                 pass
+
+        if not metadata.get("host_aliases"):
```
gmmeyer: I can do this!

utils/cloud_metadata.py
gmmeyer: What we should do is check for inclusion!

utils/cloud_metadata.py
gmmeyer: Yup you're right!

utils/cloud_metadata.py
gmmeyer: There are ways to use Cloud Foundry without Bosh, so I wanted to make it a bit more robust of a check. However, since the only thing we add is a Bosh ID it won't necessarily do anything now.

gmmeyer: For example, SUSE has a Cloud Foundry offering where the whole of Cloud Foundry runs in a Kubernetes cluster. (I'm not sure how it then does the additional container orchestration, I guess Docker in Docker but still...)

olivielpeau: ok, thanks for the explanation.

So the one thing I want to make sure of here is that we're not too "optimistic" in our detection of a cloud foundry environment, i.e. we don't detect that we're running on cloud foundry when we're not. Even though for now this is only used to add a host alias (including the socket hostname if no bosh-related alias is defined), given the name of the function I could see this being used in the future for other things :)

Or, we can keep the detection "optimistic" and add a comment that makes that explicit

PR #3407, opened 2017-06-30
[dogstatsd6][beta] enable on agent5

truthbk:
### What does this PR do?

Will enable the `beta` feature of running dogstatsd6 on agent5.

### Motivation

Requested.

### Test

Manual testing on ubuntu 16.04, working as expected. Supervisor behaving correctly, init scripts too.


truthbk:
Some minor work might be necessary on the init scripts to enhance the `dogstatsd6` support.

datadog.conf.example
```diff
@@ -158,6 +158,13 @@ gce_updated_hostname: yes
 
 # If you don't want to enable the DogStatsd server, set this option to no
 # use_dogstatsd: yes
+#
+# Note: this is a beta feature
+# Enables dogstatsd6 on agent5. Will disable the default dogstatsd daemon.
+# dogstatsd6_enable: false
+#
+# Port from which to collect expvar statistics
+# dogstatsd6_stat_port: 5000
```
olivielpeau: I'd prefer `stats` to `stat`, not sure though

packaging/osx/supervisor.conf
```diff
@@ -44,6 +44,14 @@ startsecs=5
 startretries=3
 priority=998
 
+[program:dogstatsd6]
+command=/opt/datadog-agent/bin/dogstatsd6 start -l -f /etc/dd-agent/datadog.conf
```
olivielpeau: nit: on macOS, the config file path is `/opt/datadog-agent/etc/datadog.conf`

packaging/supervisor.conf
```diff
@@ -48,6 +48,15 @@ startretries=3
...
 user=dd-agent
 
+[program:dogstatsd6]
+command=/opt/datadog-agent/bin/dogstatsd6 start -l -f /etc/dd-agent/datadog.conf
+stdout_logfile=NONE
+stderr_logfile=NONE
+startsecs=2
+startretries=3
+priority=998
+user=dd-agent
```
olivielpeau: I _think_ (haven't actually checked what the binary does) that we also need to set `exitcodes=0` like the trace-agent does. By default supervisor also regards `2` as a successful exit code, and I think the go runtime can exit with `2` when an actual error occurs.

dogstatsd.py
```diff
@@ -474,10 +479,47 @@ def run(self):
...
-    def info(self):
+    def info(self, cfg=None):
         logging.getLogger().setLevel(logging.ERROR)
+        if cfg and cfg.get('dogstatsd6_enable', False):
+            dsd6_status = Dogstatsd._get_dsd6_stats(cfg)
+            if dsd6_status:
+                dsd6_status.render()
+                return 0
+            else:
+                return -1
```
olivielpeau: can we also print something here when dogstatsd6 is enabled but can't be reached/its response is invalid? No need to format it very nicely

Same thing for the flare

dogstatsd.py
```diff
@@ -493,6 +535,11 @@ def init(config_path=None, use_watchdog=False, use_forwarder=False, args=None):
         sleep(4)
         sys.exit(0)
 
+    if _is_affirmative(c.get('dogstatsd6_enabled')):
```
olivielpeau: typo: should be `dogstatsd6_enable`

utils/http.py
```diff
@@ -12,3 +12,15 @@ def retrieve_json(url, timeout=DEFAULT_TIMEOUT, verify=True):
     r = requests.get(url, timeout=timeout, verify=verify)
     r.raise_for_status()
     return r.json()
+
+# Get expvar stats
+def get_expvar_stats(key, port):
+    try:
+        json = retrieve_json("http://127.0.0.1:{port}/debug/vars".format(port=port))
```
olivielpeau: I think we should use `config['bind_host']` instead of `127.0.0.1` here. Could be passed as a parameter of this function

truthbk: We're not binding to `bind_host` on the dogstatsd6 side, but to localhost or `0.0.0.0`, so I think we should maybe keep it as `localhost`. But, I don't have a strong opinion either.

olivielpeau: ok, so since dogstatsd6 uses `localhost`, can we use `localhost` instead of `127.0.0.1`?

There's some logic in the agent5 (https://github.com/DataDog/dd-agent/blob/5.14.1/config.py#L269-L275) that falls back to `127.0.0.1` only if localhost can't be resolved, so if agent6 always uses `localhost` should we just stick to `localhost` here?

That said I don't think this is that important, and at some point we'll probably want to improve the logic on dogstatsd6's side

tests/core/test_dogstatsd.py
```diff
@@ -32,18 +37,29 @@ def test_get_socket_address(self):
...
+    def test_init6(self, gc):
+        cfg = defaultdict(str)
+        cfg['api_key'] = "deadbeeffeebdaed"
+        cfg['use_dogstatsd'] = True
+        cfg['dogstatsd6_enable'] = True
+        gc.return_value = cfg
+
+        path, _, env = init6(gc)
+        self.assertNotEqual(env, os.environ)
+        self.assertEqual(env.get('DD_API_KEY'), cfg['api_key'])
```
olivielpeau: let's make sure we also pass the value of `dogstatsd6_stats_port` in the env to dogstatsd6

jmxfetch.py
```diff
@@ -330,7 +279,7 @@ def _start(self, path_to_java, java_run_opts, jmx_checks, command, reporter, too
             JMXFiles.write_status_file(invalid_checks)
             raise
         except Exception:
-            log.exception("Couldn't launch JMXFetch")
+            log.info("unable to launch JMXFetch")
```
olivielpeau: should remain an `exception` I think

daemon.py
```diff
@@ -86,6 +90,80 @@ def _handle_sigterm(cls, signum, frame):
...
+            signal.signal(signal.SIGTERM, self._handle_sigterm)
+
+            # Handle Keyboard Interrupt
+            signal.signal(signal.SIGINT, self._handle_sigterm)
+
+        except ValueError:
+            log.exception("Unable to register signal handlers.")
+
+    def execute(self, process_args, redirect_std_streams=None, env=None):
+        # Launch JMXfetch subprocess manually, w/o get_subprocess_output(), since it's a special case
```
olivielpeau: nit: let's remove all mentions of jmxfetch in this file :)

dogstatsd.py
```diff
@@ -474,12 +484,73 @@ def run(self):
...
+            else:
+                message = DogstatsdStatus._dogstatsd6_unavailable_message()
+                exit_code = -1
+
+        sys.stdout.write(message)
+        return exit_code
+
+    @classmethod
+    def _get_dsd6_stats(self, cfg={}):
+        port = cfg.get('dogstatsd6_stat_port', 5000)
```
olivielpeau: typo: should be `dogstatsd6_stats_port`

dogstatsd.py
```diff
@@ -578,15 +694,34 @@ def main(config_path=None):
...
+            if c.get('dogstatsd6_enable'):
+                message = 'Status unavailable for dogstatsd6'
+                log.warning(message)
+                sys.stderr.write(message)
+            else:
+                daemon.status()
         elif command == 'info':
-            return Dogstatsd.info()
+            if c.get('dogstatsd6_enable'):
+                Dogstatsd6.info(c)
```
olivielpeau: let's `return` this

dogstatsd.py
```diff
@@ -578,15 +694,34 @@ def main(config_path=None):
             return 1
 
         if command == 'start':
-            daemon.start()
+            if dsd6:
+                dsd6.execute([dsd6_path, command], env)
```
olivielpeau: I don't think we should support these commands with dogstatsd6, except for the `info` one.

Dogstatsd with dogstatsd6 enabled should either be run in the foreground (by supervisor generally), i.e. with no `command`, or with the `info` command.

tests/core/test_dogstatsd.py
olivielpeau: also, I think we should configure `dogstatsd6` so that it logs to a `dogstatsd6.log` file in the same directory as the one that's configured for `dogstatsd` (if it's the same file I could see some weird stuff happening with the log rotation of `dogstatsd.py` vs `dogstatsd6`... May be completely fine to do that though)

jmxfetch.py
truthbk: We log the exception in the underlying class I believe.... I'll double-check

tests/core/test_dogstatsd.py
truthbk: it's currently logging to `agent.log` - but that's just confusing.

jmxfetch.py
olivielpeau: you're right 👍

dogstatsd.py
```diff
@@ -556,20 +671,31 @@ def main(config_path=None):
...
+                            cnf.get('autorestart', False))
+            in_developer_mode = cnf.get('developer_mode')
 
     # If no args were passed in, run the server in the foreground.
     if not args:
-        daemon.start(foreground=True)
-        return 0
+        if dsd6:
+            logging.info("Launching Dogstatsd6 - logging to dogstatsd6.log")
+            logging.shutdown()
```
truthbk: True, this was something I had been experimenting with...

utils/logger.py
```diff
@@ -27,6 +29,19 @@ def wrapper(*args, **kwargs):
         return wrapper
     return decorator
 
+class DisableLoggerInit():
```
truthbk: Worked fine, yup. I'll keep it as-is for now.

dogstatsd.py
```diff
@@ -542,6 +620,45 @@ def init(config_path=None, use_watchdog=False, use_forwarder=False, args=None):
...
+        # and thus can exit cleanly.
+        sleep(4)
+        sys.exit(0)
+
+    if not _is_affirmative(c.get('dogstatsd6_enable', False)):
+        log.debug("Dogstatsd v6 is disabled")
+        return None, None, None
+
+
+    env = copy.deepcopy(os.environ)
```
xvello: If running alongside agent5, shouldn't we specify `DD_ENABLE_METADATA_COLLECTION = false` to tell dsd6 not to send host metadata?

olivielpeau: yes we should, actually I have a branch (`dogstatsd6-beta`) based on `5.16.0` + the changes in this PR + a few fixes (including disabling metadata collection), and I'm running an Agent built from that branch on staging. Will work on pushing these fixes to this branch.

olivielpeau:
@truthbk I've pushed a few additional commits (including `5c29159` that requires https://github.com/DataDog/datadog-agent/pull/475), could you have a quick look through them?

I think we're ready for a merge.

PR #3409, opened 2017-06-30
pkg: Add process-agent to supervisor and flares

conorbranagan:
### What does this PR do?

Adds the [datadog-process-agent](https://github.com/DataDog/datadog-process-agent) to the supervisor file so it will run as part of the Agent. As with the trace-agent, you must enable a flag in the config for the Agent to actually run without immediately stoppign.

The omnibus change is https://github.com/DataDog/dd-agent-omnibus/pull/181

### Testing Guidelines

Once both this and omnibus build are in place we will be able to validate it's running. If you run the omnibus build on this and the omnibus branch you will get a pkg that installs things correctly.

cc @shang-wang


conorbranagan:
This is a good idea. I'll open a ticket on our end to look at doing this and probably follow the trace style. Thanks for reviewing!

PR #3411, opened 2017-06-30
[core] Collect OpenStack metadata and tags

mtougeron:
### What does this PR do?

- Collects the OpenStack instance's `uuid` and `name` for `host_aliases` from `meta_data.json`
- Collects the OpenStack instance's `meta` key/value pairs as agent `tags` from `meta_data.json`
- Allows you to use the OpenStack instance's `uuid` instead of its `instance-id`

### Motivation

- We need the OpenStack instance's metadata tagging to work similar to the AWS EC2 tag collection.
- We want the `uuid` and `name` to appear in the "also known as" part of the UI.
- OpenStack nova recycles the `instance-id` so it isn't guaranteed to be unique where the `uuid` is. The `uuid` value is also what you use when calling the nova api.

### Additional Notes

Created a config option called `openstack_use_metadata_tags` (defaults to true) in case someone doesn't want those tags to be agent tags.

### DIFF
--- a/datadog.conf.example
+++ b/datadog.conf.example
@@ -67,6 +67,13 @@ api_key:
 # when not specified, default: no
 gce_updated_hostname: yes
 
+# when using OpenStack, use the instance's "uuid"
+# the OpenStack instance-id can be recycled but the uuid is not
+# openstack_use_uuid: no
+
+# when using OpenStack, use the instance's metadata as tags
+# openstack_use_metadata_tags: yes
+
 # Set the threshold for accepting points to allow anything
 # within recent_point_threshold seconds (default: 30)
 # recent_point_threshold: 30
--- a/utils/cloud_metadata.py
+++ b/utils/cloud_metadata.py
@@ -142,6 +142,7 @@ class EC2(object):
     TIMEOUT = 0.1  # second
     DEFAULT_PREFIXES = [u'ip-', u'domu']
     metadata = {}
+    is_openstack = None
 
     class NoIAMRole(Exception):
         """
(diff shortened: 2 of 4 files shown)


mtougeron:
FYI I had to change the `TestEC2` to allow up to `1.25` seconds because there was one more possible `0.1` second timeout from the OpenStack metadata check.

utils/cloud_metadata.py
```diff
@@ -218,6 +238,31 @@ def get_metadata(agentConfig):
...
+                # Set the "also known as" metadata similar to AWS EC2
+                EC2.metadata['host_aliases'] = [
+                    openstack_metadata['uuid'],
+                    openstack_metadata['name'],
+                ]
+
+                # the OpenStack instance-id can be recycled but the uuid will always be unique
+                if agentConfig['openstack_use_uuid']:
+                    log.info(u"Using the instance's uuid for instance-id")
+                    EC2.metadata["instance-id"] = openstack_metadata['uuid']
```
gmmeyer: It might make sense to send this up as another host alias, instead of overwriting the ec2 instance id. Instance id's can be important to people for various reasons, as it's a stable (external) identifier, Amazon knows what to do with it in the case of a support ticket, and so on. I don't think it's a good idea to overwrite it.

Instead of you could just make a new metadata piece that sends it up as a separate host alias, setting it as the main hostname if this is enabled.

datadog.conf.example
```diff
@@ -47,7 +47,8 @@ api_key:
 # and dice per monitored app (= running Agent Check) on Datadog's backend.
 # create_dd_check_tags: no
 
-# Collect AWS EC2 custom tags as agent tags (requires an IAM role associated with the instance)
+# Collect AWS or OpenStack EC2 custom tags as agent tags
```
gmmeyer: Does OpenStack need to be mentioned here or can you move it all down to the openstack config options? I think it would make more sense down there.

gmmeyer:
I added two small comments to the PR. Overall I think it's pretty good, this PR makes sense and it's well done. Thank you! Let me know what you think of the comments. :)

utils/cloud_metadata.py
mtougeron: I am setting it as a host alias in the lines above.

The reason I'm overriding it is because the default `i-####` value in openstack is *not* stable and is recycled by Nova. I did this as a flag so that it is backwards compatible though I'd hope that Nova would stop recycling the values. :(

datadog.conf.example
mtougeron: No, it doesn't need to mention openstack here. I just thought it should be updated because now it impacts both cloud providers instead of just AWS. I'll update when I get into the office today.

mtougeron:
@gmmeyer Pushed up the revert for the `datadog.conf.example` message.

lmk if you are okay with the `openstack_use_uuid` flag.

gmmeyer:
Hey @mtougeron sorry this fell through the cracks! Can you resolve the conflicts? I think after that it should be good to merge

mtougeron:
@gmmeyer Updated, thanks!

gmmeyer:
great! I'll merge it now! Thank you for your contribution!

mtougeron:
Thanks!

PR #3413, opened 2017-07-04
[kube] disable kube_service tagging on 4 403 errors from apiserver

xvello:
### What does this PR do?

- Move from log.exception to log.warning (less scary)
- On 8th 403 error, print log.error on configuration change then disable class to avoid spamming apiserver
- Add unit test for backoff mechanism

### DIFF
--- a/utils/kubernetes/pod_service_mapper.py
+++ b/utils/kubernetes/pod_service_mapper.py
@@ -26,14 +26,14 @@ class PodServiceMapper:
         self._pod_services_mapping = defaultdict(list)      # {pod_uid:[service_uid]}
 
         self._403_errors = 0             # Count how many 403 errors we got from apiserver
-        self._403_backoff = False        # Disable the service mapper because of 403 errors
+        self._403_disable = False        # Disable the service mapper because of 403 errors
 
     def _fill_services_cache(self):
         """
         Get the list of services from the kubelet API and store the label selector dicts.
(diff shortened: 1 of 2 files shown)


utils/kubernetes/pod_service_mapper.py
```diff
@@ -41,7 +50,16 @@ def _fill_services_cache(self):
                 self._service_cache_selectors[uid] = selector
             self._service_cache_invalidated = False
         except Exception as e:
-            log.exception('Unable to read service list from apiserver: %s', e)
+            if isinstance(e, requests.exceptions.HTTPError) and str(e).startswith("403"):
```
hkaj: `str(e).startswith("403")` --> `e.response.status_code == 403` (it's more robust)

utils/kubernetes/pod_service_mapper.py
```diff
@@ -22,11 +25,17 @@ def __init__(self, kubeutil_object):
         self._pod_labels_cache = defaultdict(dict)          # {pod_uid:{label}}
         self._pod_services_mapping = defaultdict(list)      # {pod_uid:[service_uid]}
 
+        self._403_errors = 0             # Count how many 403 errors we got from apiserver
+        self._403_backoff = False        # Disable the service mapper because of 403 errors
```
hkaj: backoff makes it sound like there's a retry mechanism, but once _403_backoff is True we actually stop retrying. Let's change the name.

PR #3415, opened 2017-07-05
[auto-discovery] re-pipe configurations after JMXFetch restart

truthbk:
### What does this PR do?

If JMXFetch happens to go down, use the `atime` stat date on the JMX launch file to decide if the auto-discovery configurations should be re-piped to JMX. JMXFetch will touch that file when coming up so we should have a relatively good idea of when JMXFetch started - alternatively, we could look at the process table to figure this stuff out (but that would be computationally more expensive).

### Motivation

Customer bug.

### Testing Guidelines

### Additional Notes

Sister PR: https://github.com/DataDog/jmxfetch/pull/143

### DIFF
--- a/agent.py
+++ b/agent.py
@@ -345,9 +345,12 @@ class Agent(Daemon):
                         self._agentConfig).crawl_config_template()
 
                     # JMXFetch restarts should prompt reload
-                    jmx_launch = JMXFetch._get_jmx_launchtime()
-                    if self.last_jmx_piped and self.last_jmx_piped < jmx_launch:
-                        self.sd_backend.reload_check_configs = True
+                    try:
+                        jmx_launch = JMXFetch._get_jmx_launchtime()
+                        if self.last_jmx_piped and self.last_jmx_piped < jmx_launch:
+                            self.sd_backend.reload_check_configs = True
+                    except OSError as e:
+                        log.debug("could not stat JMX lunch file: %s", e)
 
                 except Exception as e:
                     log.warn('Something went wrong while looking for config template changes: %s' % str(e))
--- a/agent.py
+++ b/agent.py
@@ -219,6 +219,11 @@ class Agent(Daemon):
         logging.getLogger().setLevel(logging.ERROR)
         return CollectorStatus.print_latest_status(verbose=verbose)
 
+    def sd_pipe_jmx_configs(self, hostname):
+        jmx_sd_configs = generate_jmx_configs(self._agentConfig, hostname)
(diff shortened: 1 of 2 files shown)


bai:
It seems that SD-based jmx metric collection is borked at the moment on Kubernetes. Friendly question: Is there any ETA on merging this?

jmxfetch.py
```diff
@@ -69,6 +69,8 @@
     'list_limited_attributes': "List attributes that do match one of your instances configuration but that are not being collected because it would exceed the number of metrics that can be collected",
     JMX_COLLECT_COMMAND: "Start the collection of metrics based on your current configuration and display them in the console"}
 
+JMX_LAUNCH_FILE = 'jmx.launch'
```
truthbk: maybe it should be a hidden file.

truthbk:
@bai this just barely missed the window for the upcoming `5.15.0` release, but will hopefully make it into the next bugfix version.

jmxfetch.py
olivielpeau: since it's in jmxfetch's temp directory I'd say a regular file is okay

agent.py
```diff
@@ -342,6 +343,15 @@ def run(self, config=None):
                 try:
                     self.sd_backend.reload_check_configs = get_config_store(
                         self._agentConfig).crawl_config_template()
+
+                    # JMXFetch restarts should prompt reload
+                    try:
+                        jmx_launch = JMXFetch._get_jmx_launchtime()
+                        if self.last_jmx_piped and self.last_jmx_piped < jmx_launch:
+                            self.sd_backend.reload_check_configs = True
```
olivielpeau: I'm not sure this jmxfetch-specific code block should be here exactly: `JMXFetch` restarts aren't really related to the updates in the config template store (which is what the parent block is about).

On top of that, when JMXFetch restarts, it'd be nice to reload only the jmx-related check configs instead of all the check configs.

So:
1. Could we move this logic to outside this `if` block (i.e. right after `self.reload_configs_flag = False`)?
2. To avoid reloading all the check configs, we could do what's done here: https://github.com/DataDog/dd-agent/blob/5.15.0/agent.py#L281-L285 (which could be out in a separate function) instead of setting `self.sd_backend.reload_check_configs = True`

Let me know how that sounds to you, I might have missed something.

truthbk:
Thanks for the review @olivielpeau, insightful comments. I scrambled to put this together ASAP and wasn't as careful as I should have. Let me know if you prefer the current logical path.

PR #3419, opened 2017-07-06
allows integrations core to download jmx fetch from rakefile

gmmeyer:
### What does this PR do?

Integrations core does not use a proper venv when testing, meaning that when you call the rake task, it doesn't download jmx fetch. This will allow it to download JMX fetch

### Motivation

I was working on a JMX Check

### Testing Guidelines

### Additional Notes

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -48,7 +48,9 @@ task 'setup_libs' do
   in_venv = system "python -c \"import sys ; exit(not hasattr(sys, 'real_prefix'))\""
   raise 'Not in dev venv/CI environment/Integrations - bailing out.' if !in_venv && !ENV['CI'] && !ENV['SDK_HOME']
 
-  jmx_version = `python -c "import config ; print config.JMX_VERSION"`
+  Rake::Task["setup_env"].invoke
+
+  jmx_version = `venv/bin/python -c "import config ; print config.JMX_VERSION"`
   jmx_version = jmx_version.delete("\n")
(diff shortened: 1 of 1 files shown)


Rakefile
```diff
@@ -46,9 +46,11 @@ end
 desc 'Grab libs'
 task 'setup_libs' do
   in_venv = system "python -c \"import sys ; exit(not hasattr(sys, 'real_prefix'))\""
-  raise 'Not in dev venv/CI environment - bailing out.' if !in_venv && !ENV['CI']
+  raise 'Not in dev venv/CI environment/Integrations - bailing out.' if !in_venv && !ENV['CI'] && !ENV['SDK_HOME']
```
truthbk: I don't think we set `SDK_HOME` in `dd-agent`, do we? We still need to be able to setup the libs here, so I'm not sure if we can rely on `SDK_HOME` like this. Can you double-check?

gmmeyer: We don't, but we do in the rake tasks for the sdk

truthbk: So this would break `setup_libs` for the dd-agent. I think we should change this - correct me if I'm wrong.

gmmeyer: I don't think it would break it? Unless I misunderstand the logic there, it's saying, raise this if none of these things are present. But if it is present, then it wouldn't raise. So, if it's in the venv, then it's fine.

truthbk: You're right, my bad.

PR #3421, opened 2017-07-10
Add container name tag in auto discovery

hkaj:
if you have not yet done so.*

### What does this PR do?

Rework #3365 to move back to a different template variable name.

### Motivation

This is something we actually don't want included by default in `%%tags%%`. It should be used with care as it could flood the tag auto complete menu with a useless tag if used carelessly (typically: in a highly volatile environment where containers churn often and have different names each time).

### Testing Guidelines

### Additional Notes

Sorry for the back and forth @subuk, your first version was actually the good one!

### DIFF
--- a/utils/service_discovery/sd_docker_backend.py
+++ b/utils/service_discovery/sd_docker_backend.py
@@ -104,6 +104,7 @@ class SDDockerBackend(AbstractSDBackend):
             'host': self._get_host_address,
             'pid': self._get_container_pid,
             'port': self._get_port,
+            'container-name': self._get_container_name,
             'tags': self._get_additional_tags,
         }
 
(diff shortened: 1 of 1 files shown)


utils/service_discovery/sd_docker_backend.py
```diff
@@ -337,8 +338,13 @@ def get_tags(self, state, c_id):
 
         return tags
 
+    def _get_container_name(self, state, c_id, tpl_var):
```
hush-hush: `tpl_var` isn't used at all, is it normal ?

hkaj: yeah it's passed to all tag extractors by https://github.com/DataDog/dd-agent/blob/5.14.1/utils/service_discovery/sd_docker_backend.py#L489 but only used by some.

PR #3426, opened 2017-07-11
[auto-discovery][jmx] send terminator after configs + fix execption handling

truthbk:
### What does this PR do?

We need to submit the config terminator to acknowledge we've piped everything out. Done in it's own `write()` to ensure atomicity of the write. The jmxfetch will continue to greedily `read()` until the configuration terminator is received.

### Motivation
Feedback in the sister PR: https://github.com/DataDog/jmxfetch/pull/145

### Additional Notes

The local FIFO nature of the named pipes should make this process safe, a `write()` should not be lost and the configuration on the other end should be received in full. Additionally, the delivery of the config terminator - which marks the end of the "piping" procedure, should be guaranteed so there should be no real possibility of having the jmxfetch side fall in a busy-loop waiting for a terminator that will never be delivered.

### DIFF
--- a/agent.py
+++ b/agent.py
@@ -74,6 +74,7 @@ JMX_SUPERVISOR_ENTRY = 'datadog-agent:jmxfetch'
 JMX_GRACE_SECS = 2
 SERVICE_DISCOVERY_PREFIX = 'SD-'
 SD_CONFIG_SEP = "#### SERVICE-DISCOVERY ####\n"
+SD_CONFIG_TERM = "#### SERVICE-DISCOVERY TERM ####\n"
 
 DEFAULT_SUPERVISOR_SOCKET = '/opt/datadog-agent/run/datadog-supervisor.sock'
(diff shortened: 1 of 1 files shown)


agent.py
```diff
@@ -440,18 +441,21 @@ def _submit_jmx_service_discovery(self, jmx_sd_configs):
...
-            os.write(self.sd_pipe, buffer)
+            if buffer:
+                # will block if len(buffer) > OS pipe buffer.
+                # JMX will unblock when it reads on the other end.
+                os.write(self.sd_pipe, buffer)
+                os.write(self.sd_pipe, SD_CONFIG_TERM)
+        except Exception as e:
+            log.exception("unable to submit YAML via pipe: %s", e)
+        else:
+            log.info("JMX SD Config via named pipe %s successfully.", name)
```
olivielpeau: logging `name` here doesn't really make sense :)
Maybe you could log all the names in this entry?

PR #3438, opened 2017-07-13
Collect Kubernetes daemonset readiness metric

mnussbaum:
### What does this PR do?

This PR collects the `kube_daemonset_status_number_ready` metric as part of the Kubernetes State Metrics integration.

### Motivation

We need the `kube_daemonset_status_number_ready` metric to alert when an instance of one of our Kubernetes Daemonsets goes unhealthy.

@maggie-lou and I worked on this together.

### DIFF
--- a/utils/kubernetes/kube_state_processor.py
+++ b/utils/kubernetes/kube_state_processor.py
@@ -47,6 +47,7 @@ class KubeStateProcessor:
             'kube_daemonset_status_current_number_scheduled': NAMESPACE + '.daemonset.scheduled',
             'kube_daemonset_status_number_misscheduled': NAMESPACE + '.daemonset.misscheduled',
             'kube_daemonset_status_desired_number_scheduled': NAMESPACE + '.daemonset.desired',
+            'kube_daemonset_status_number_ready': NAMESPACE + '.daemonset.ready',
             # pods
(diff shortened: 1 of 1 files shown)


hkaj:
Hi @mnussbaum 
Thanks for the PR. This file is deprecated starting from 5.15 (which will go out tomorrow), we're going to remove it in a future version. It is replaced by the more generic [prometheus check](https://github.com/DataDog/dd-agent/blob/5.14.1/checks/prometheus_check.py) that can parse any prometheus metrics.

Could you port this PR to integrations-core instead? [Here](https://github.com/DataDog/integrations-core/blob/5.15.0/kubernetes_state/check.py#L40) would work. If you could also update [metrics metadata](https://github.com/DataDog/integrations-core/blob/5.15.0/kubernetes_state/metadata.csv) that would be awesome. You can find documentation about it [here](https://docs.datadoghq.com/guides/integration_sdk/#metadata-csv).

Thanks again.

mnussbaum:
Ah, it looks like the `kube_daemonset_status_number_ready` metric is [already present](https://github.com/DataDog/integrations-core/blob/5.15.0/kubernetes_state/check.py#L44) in the new repo. I guess we'll be good to go with the next release, thanks!

PR #3440, opened 2017-07-17
Remove deprecated kube_state_processor

hkaj:
*Note: Please remember to review the Datadog [Contribution Guidelines](https://github.com/DataDog/dd-agent/blob/master/CONTRIBUTING.md)
if you have not yet done so.*

### What does this PR do?

🔥  `kube_state_processor.py`
### Motivation

We switched to the more generic PrometheusCheck.

### Testing Guidelines

An overview on [testing](https://github.com/DataDog/dd-agent/blob/master/tests/README.md)
is available in our contribution guidelines.

### Additional Notes

Let's wait for 5.16. I'll add a deprecation warning to 5.15.x

PR #3442, opened 2017-07-17
[auto-discovery] tags could be a dict of KV pairs

truthbk:
### What does this PR do?

Tags could be dicts.

Also, I believe `kube_state_processor` has been deprecated after the migration to prometheus checks.

### Motivation

Encountered this issue testing.


utils/kubernetes/__init__.py
```diff
@@ -4,6 +4,5 @@
 
 from .kube_event_retriever import KubeEventRetriever  # noqa: F401
 from .pod_service_mapper import PodServiceMapper   # noqa: F401
-from .kube_state_processor import NAMESPACE  # noqa: F401
```
xvello: already addressed in https://github.com/DataDog/dd-agent/pull/3440

PR #3446, opened 2017-07-20
[emitter] Split check runs from main agent payload

remh:
### What does this PR do?

This PR extracts the check runs from the legacy payload and posts them to their API endpoint.

### Motivation

This would help in the backend.

### Testing Guidelines

Added a test on the split function but this will need to be well tested on staging


remh:
@Viq111 @dd-caleb

remh:
Let's switch the milestone to 5.16. We actually need to do some changes in the backend to be ready for this. 
Thanks!

olivielpeau:
@remh We're planning to release 5.17.0 in about a week (feature freeze on Monday), are the backend changes related to this ready?

PR #3448, opened 2017-07-21
[win32] Remove `run_check`

olivielpeau:
### What does this PR do?

Removes the `run_check` method.

### Motivation

`run_check` dates back from pre-5.12 windows agents, when we weren't
shipping a full python interpreter and needed this specific method
to allow users to run custom checks from `shell.exe`.

This doesn't make sense anymore now that we ship a full python
interpreter that the agent processes use. That interpreter has the
necessary modules to run checks. The correct way to run checks
(shipped by default or custom) is now similar to Linux/macOS, i.e.:

```
# From the `<install_dir>\agent` directory
> ..\embedded\python.exe agent.py check <check_name>
```

`shell.exe` should only be used to troubleshoot the python environment
that the windows service and the windows GUI use.

### DIFF
--- a/checks/__init__.py
+++ b/checks/__init__.py
@@ -33,7 +33,6 @@ from config import AGENT_VERSION, _is_affirmative
 from util import get_next_id, yLoader
 from utils.hostname import get_hostname
 from utils.proxy import get_proxy
-from utils.platform import Platform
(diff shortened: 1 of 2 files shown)


nmuesch:
👍 LGTM

PR #3467, opened 2017-08-04
[windows wmi]  When the WMI type returned is a UINT32, python struggles

derekwbrown:
because python doesn't have the concept of unsigned integers. So, it
just converts to a negative number.

If the type returned is supposed to be UINT32, convert it accordingly.

### Motivation

Customer report.

### DIFF
--- a/checks/libs/wmi/sampler.py
+++ b/checks/libs/wmi/sampler.py
@@ -560,7 +560,7 @@ class WMISampler(object):
                 try:
                     fval = float(wmi_property.Value)
                     if self._property_data_types is not None:
-                        if self._property_data_types[wmi_property.Name]:
+                        if wmi_property.Name in self._property_data_types:
                             if self._property_data_types[wmi_property.Name] == "uint32":
                                 val = wmi_property.Value
                                 if val < 0:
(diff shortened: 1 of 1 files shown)


checks/libs/wmi/sampler.py
```diff
@@ -509,12 +509,11 @@ def _parse_results(self, raw_results, includes_qualifiers):
...
                 )
 
-                if should_get_qualifier_type:
-
-                    # Can't index into "Qualifiers_" for keys that don't exist
-                    # without getting an exception.
-                    qualifiers = dict((q.Name, q.Value) for q in wmi_property.Qualifiers_)
+                # Can't index into "Qualifiers_" for keys that don't exist
+                # without getting an exception.
+                qualifiers = dict((q.Name, q.Value) for q in wmi_property.Qualifiers_)
```
olivielpeau: If I understand correctly this change makes the logic access the Qualifiers at each iteration, which is costly according to the comment 15 lines above:

> IMPORTANT: To improve performance, only access the Qualifiers if the "CounterType" hasn't already been cached.

If that concern is valid, what do you think about caching the results of the test `CIMTYPE in qualifiers` in a dict on the `sampler` object and access that dict after the first iteration? (having a similar logic as what's done for `CounterType` basically)

PR #3470, opened 2017-08-08
Db/proxy cfg cmdline

derekwbrown:
### What does this PR do?

Allows configuration of proxy parameters to be configured on the windows installer command line

### Motivation

Customer request

### Additional Notes

Needs companion PR https://github.com/DataDog/dd-agent-omnibus/pull/192

### DIFF
--- a/utils/windows_configuration.py
+++ b/utils/windows_configuration.py
@@ -17,13 +17,28 @@ SDK_REG_PATH = WINDOWS_REG_PATH + '\\Integrations\\'
 
 log = logging.getLogger(__name__)
 
+config_attributes = ['api_key', 'tags', 'hostname', 'proxy_host', 'proxy_port', 'proxy_user', 'proxy_password']
+
+def remove_registry_conf():
+    try:
+        with _winreg.OpenKey(_winreg.HKEY_LOCAL_MACHINE,
+                             WINDOWS_REG_PATH) as reg_key:
+            for attribute in config_attributes:
+                try:
+                    _winreg.DeleteValue(reg_key, attribute)
+                except Exception:
+                    # it's ok if it's not there.
+                    pass
+    except Exception:
+        # also OK if the whole tree isn't there
+        pass
 
 def get_registry_conf(agentConfig):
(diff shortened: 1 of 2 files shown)


utils/windows_configuration.py
```diff
@@ -17,13 +17,29 @@
 
 log = logging.getLogger(__name__)
 
+config_attributes = ['api_key', 'tags', 'hostname', 'proxy_host', 'proxy_port', 'proxy_user', 'proxy_password']
```
gmmeyer: 🍰

win32/service.py
```diff
@@ -134,7 +134,12 @@ def _update_config_file(self, config):
         config.update(registry_conf)
         if registry_conf:
             log.info('Updating conf file options: %s', registry_conf.keys())
-            update_conf_file(registry_conf, get_config_path())
+            try:
+                update_conf_file(registry_conf, get_config_path())
+                log.info('update succeeded, deleting old values')
+                remove_registry_conf()
```
gmmeyer: Just curious, why do we want to remove it from the registry?

utils/windows_configuration.py
```diff
@@ -17,13 +17,29 @@
...
+
+def remove_registry_conf():
+    try:
+        with _winreg.OpenKey(_winreg.HKEY_LOCAL_MACHINE,
+                             WINDOWS_REG_PATH, 0, _winreg.KEY_WRITE) as reg_key:
+            for attribute in config_attributes:
+                try:
+                    _winreg.DeleteValue(reg_key, attribute)
+                except Exception as e:
```
gmmeyer: Will it error if the key doesn't exists? In that case, it seems like we would want to ignore that error instead of logging it.

win32/service.py
derekwbrown: I wanted to remove the values from the registry so we don't end up with "two masters"... you put something on the install config.  Then the install works.  Then you change the config, but this is still lingering in the registry, waiting to be written back out to the config file (and killing your changes)

PR #3473, opened 2017-08-12
Db/5.10.x install proxy params

derekwbrown:
### What does this PR do?

Allows proxy configuration to be set on the installer command line for the 5.11 line of the  agent.

Back-ports the configuration code which checks the registry and applies the config on first run of the agent, rather than using "FindReplace" at install time to write the entries.  

Fixes old problem of not including correct license in the license dialog.

### Motivation

Customer request.

### Additional Notes


packaging/datadog-agent/win32/build.ps1
```diff
@@ -39,7 +39,7 @@ cp -R ..\..\..\dist\gohai install_files\files\gohai
 cp -R install_files\guidata install_files\files
 
 # Copy the license file
-cp ..\..\..\LICENSE install_files\license.rtf
+cp ..\..\..\LICENSE.rtf install_files\license.rtf
```
masci: Is this path correct? Not sure but could it be `..\license.rtf`?

derekwbrown: It's correct; because earlier in the script we changed directories several levels down (line 22)
		 cd packaging\datadog-agent\win32

PR #3478, opened 2017-08-18
[autodiscovery] annotations with multi instances bug

CharlyF:
### What does this PR do?
Bug fix of https://github.com/DataDog/dd-agent/pull/3341

Annotations with multi instances configuration leads to check level tags be added to the container level tags array.

While adding container tags with this [method](https://github.com/DataDog/dd-agent/blob/master/utils/service_discovery/sd_docker_backend.py#L490), performs a shallow copy of the container tag array to avoid adding tags from the previous instance ( which would come from [this for loop](https://github.com/DataDog/dd-agent/blob/master/utils/service_discovery/sd_docker_backend.py#L438)), contains its check tags.

### Motivation

Issue encountered by a customer.

### Testing Guidelines

Instantiate a pod with annotations containing several instances (along the lines of [this example](https://gist.github.com/CharlyF/c1220097e45ce9caf77bde567a066723))
The agent will detect the check but it will add the config level tags to the container level ones.
e.g.:

```
 (u'num.empl',
  1503048087,
  300024.0,
  {'hostname': 'i-0xxx',
   'tags': [u'kube_deployment:repl',
            u'kube_namespace:default',
            u'kube_replica_set:repl',
            u'env:prd',
            u'dbinstanceidentifier:dbi2',
            u'name:mysql',
            u'account:analytics',
            u'dbinstanceidentifier:dbi',
            u'team:data',
            u'image_tag:3.0',
            u'docker_image:gcr.io/google_containers/pause-amd64:3.0',
            u'image_name:gcr.io/google_containers/pause-amd64',
            u'table:playcount',
            u'dbinstanceidentifier:dbi2',
            u'env:prd',
            u'account:analytics',
            u'team:data'],
   'type': 'gauge'}),
```

As per the above pod yaml, this metric should only be tagged with `dbinstanceidentifier:dbi2`. 
With the change, we have:

```
 (u'num.empl',
  1503050652,
  300024.0,
  {'hostname': 'i-0xxx',
   'tags': [u'kube_deployment:repl',
            u'kube_namespace:default',
            u'kube_replica_set:repl',
            u'env:prd',
            u'dbinstanceidentifier:dbi2',
            u'name:mysql',
            u'account:analytics',
            u'team:data',
            u'image_tag:3.0',
            u'docker_image:gcr.io/google_containers/pause-amd64:3.0',
            u'image_name:gcr.io/google_containers/pause-amd64',
            u'table:playcount',
            u'dbinstanceidentifier:dbi2',
            u'env:prd',
            u'account:analytics',
            u'team:data'],
   'type': 'gauge'}),
```

### DIFF
--- a/utils/service_discovery/sd_docker_backend.py
+++ b/utils/service_discovery/sd_docker_backend.py
@@ -487,13 +487,17 @@ class SDDockerBackend(AbstractSDBackend):
 
         return templates
 
-    def _fill_tpl(self, state, c_id, instance_tpl, variables, tags=None):
+    def _fill_tpl(self, state, c_id, instance_tpl, variables, c_tags=None):
         """Add container tags to instance templates and build a
            dict from template variable names and their values."""
         var_values = {}
         c_image = state.inspect_container(c_id).get('Config', {}).get('Image', '')
 
-        # add default tags to the instance
+        # add only default c_tags to the instance to avoid duplicate tags from conf
+        if c_tags:
+            tags = c_tags[:] # shallow copy of the c_tags array
+        else:
+            tags = []
         if tags:
             tpl_tags = instance_tpl.get('tags', [])
             if isinstance(tpl_tags, dict):


utils/service_discovery/sd_docker_backend.py
```diff
@@ -487,13 +488,14 @@ def _get_config_templates(self, identifier, **platform_kwargs):
...
-    def _fill_tpl(self, state, c_id, instance_tpl, variables, tags=None):
+    def _fill_tpl(self, state, c_id, instance_tpl, variables, c_tags=None):
         """Add container tags to instance templates and build a
            dict from template variable names and their values."""
         var_values = {}
         c_image = state.inspect_container(c_id).get('Config', {}).get('Image', '')
 
-        # add default tags to the instance
+        # add only default c_tags to the instance to avoid duplicate tags from conf
+        tags = copy.copy(c_tags)
```
hkaj: Good catch! Thanks for digging in this issue.

If that's the only place we use it I'd rather not import copy and do `tags = c_tags[:]` instead. It will create a new list with all the elements of c_tags.

aerostitch: Tested with @hkaj's suggestion and it fixed the issue I had with way too much tags on my "query" metrics for the mysql integration configured in multi-instances mode via k8s annotations.

utils/service_discovery/sd_docker_backend.py
```diff
@@ -487,13 +487,17 @@ def _get_config_templates(self, identifier, **platform_kwargs):
...
            dict from template variable names and their values."""
         var_values = {}
         c_image = state.inspect_container(c_id).get('Config', {}).get('Image', '')
 
-        # add default tags to the instance
+        # add only default c_tags to the instance to avoid duplicate tags from conf
+        if c_tags:
+            tags = c_tags[:] # shallow copy of the c_tags array
+        else
+            tags = []
```
hkaj: what about this?
```python
def _fill_tpl(self, state, c_id, instance_tpl, variables, c_tags=[]):
    ...
    tags = c_tags[:]
```

PR #3480, opened 2017-08-20
[WIP] Dogstreams check for log file on each iteration

nmuesch:
Re-open file on each check of dogstreams

if you have not yet done so.*

### What does this PR do?

Removes the check to see if the tail object is None. Ensures that we build the tailFile object each time the check is run. That way the file can be created after the Agent starts. Before this, the file dogstreams it tailing needed to be created prior to the Agent starting up, or else it needed to be restarted after file creation. 

### Motivation

Customer Requested

### Testing Guidelines

### Additional Notes

It looks like the file is closed if already open in the TailFile class, and I've looked on a test machine and there doesn't seem to be any fd leak for the ddagent process. But perhaps we should explicitly close this elsewhere? Before we only were calling that tail function (which opens the specified log file) once, now we call it at each check iteration.


checks/datadog.py
```diff
@@ -175,8 +175,7 @@ def check(self, agentConfig, move_end=True):
             self._events = []
 
             # Build our tail -f
-            if self._gen is None:
-                self._gen = TailFile(self.logger, self.log_path, self._line_parser).tail(line_by_line=False, move_end=move_end)
+            self._gen = TailFile(self.logger, self.log_path, self._line_parser).tail(line_by_line=False, move_end=move_end)
```
truthbk: I don't think we need to create a TailFile object every iteration, probably just make sure we call `tail()`.... Can we look into doing that properly? The way I see it maybe that should just be in the constructor.... Let me know if you see problems with that approach.

jalaziz:
How is this better than #2815? This PR re-opens the file every time `check` is called. This would result in missing lines between calls to `check()`.

nmuesch:
Going to close this in favor of @jalaziz s PR. Thanks for making that! This one was still a WIP 🙂

PR #3481, opened 2017-08-21
[Core] Utilize sys.exit in configcheck

nmuesch:
if you have not yet done so.*

### What does this PR do?

Utilizes the sys.exit() function in the `configcheck` command utility. This allows the exit code to be determined based on whether or not the yaml configs are good/bad. 

### Motivation

Customer Request. Now when running the one off config check script/command, you can utilize `$?` to determine if there are any bad yaml files. 

Previously, if the configcheck finished and printed that there were errors in the yaml files, the following command:

```
echo $?
```

would output `0. Now this outputs 1, indicating there was a failure. 

### Testing Guidelines

### Additional Notes

### DIFF
--- a/agent.py
+++ b/agent.py
@@ -602,8 +602,8 @@ def main():
                     check.stop()
 
     elif 'configcheck' == command or 'configtest' == command:
-        if configcheck() or sd_configcheck(agentConfig):
-            sys.exit(1)
+        sd_configcheck(agentConfig)
+        return configcheck()
 
     elif 'jmx' == command:
         jmx_command(args[1:], agentConfig)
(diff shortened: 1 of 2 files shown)


agent.py
```diff
@@ -602,8 +602,8 @@ def parent_func():
                     check.stop()
 
     elif 'configcheck' == command or 'configtest' == command:
-        configcheck()
-        sd_configcheck(agentConfig)
+        if configcheck() or sd_configcheck(agentConfig):
```
olivielpeau: `sd_configcheck` doesn't return anything so I don't think it makes sense to have it in the condition here. I'm not sure we could make that function return a sensible value that reflects the status of the service discovery confs...

So instead of this, I think we should simply make this code return the value that `configcheck` returns.  The part of the code under `__main__` will then exit with the correct status code.

olivielpeau:
Thanks @nmuesch! Merged, this should be included in 5.17.0

PR #3483, opened 2017-08-21
[packaging] Fix bashism in init script

olivielpeau:
### What does this PR do?

Only has a real impact on the debian init script, but let's fix that
everywhere.

Introduced in #3301.

### Additional Notes

The bug would only affect Debian by:
- making the `DATADOG_ENABLED` var not work as expected
- always printing a pretty ugly error message


truthbk:
We have to add the `bashism` check to the CI

PR #3488, opened 2017-08-23
[info] `unknown` ver if failed module load, `custom` if custom check.

truthbk:
### What does this PR do?

Fixes a couple of issues with the info page:
- if a module fails to load - `unknown` version + show exception in info page.
- if a custom check is used - report `custom` as the version.

### Motivation
https://github.com/DataDog/integrations-core/issues/669

### DIFF
--- a/checks/__init__.py
+++ b/checks/__init__.py
@@ -379,14 +379,17 @@ class AgentCheck(object):
     def set_manifest_path(self, manifest_path):
         self.manifest_path = manifest_path
 
-    def set_check_version(self, manifest=None):
-        version = AGENT_VERSION
+    def set_check_version(self, version=None, manifest=None):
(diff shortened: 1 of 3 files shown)


checks/__init__.py
```diff
@@ -379,14 +379,17 @@ def __init__(self, name, init_config, agentConfig, instances=None):
     def set_manifest_path(self, manifest_path):
         self.manifest_path = manifest_path
 
-    def set_check_version(self, manifest=None):
-        version = AGENT_VERSION
+    def set_check_version(self, version=None, manifest=None):
+        _version = AGENT_VERSION
+
+        if version:
```
masci: this could be `_version = version or AGENT_VERSION`

PR #3489, opened 2017-08-23
[systemd] apply sysconfig settings on launch too.

truthbk:
### What does this PR do?

Honor sysconfig settings when launching with systemd service.

### Motivation

Not working in the mentioned scenario: https://github.com/DataDog/dd-agent/pull/3301

### DIFF
--- a/packaging/start_agent.sh
+++ b/packaging/start_agent.sh
@@ -3,15 +3,11 @@
 # All rights reserved
 # Licensed under Simplified BSD License (see LICENSE)
 
-AGENTSYSCONFIG="/etc/sysconfig/datadog-agent"
 PATH=/opt/datadog-agent/embedded/bin:/opt/datadog-agent/bin:$PATH
 
-# pull in sysconfig settings
-[ -f $AGENTSYSCONFIG ] && . $AGENTSYSCONFIG
-
 if [ "$DATADOG_ENABLED" = "no" ]; then
-        echo "Disabled via $AGENTSYSCONFIG. Exiting."
-        exit 0
+    echo "Disabled via $AGENTSYSCONFIG. Exiting."
+    exit 0
 fi
 
 exec /opt/datadog-agent/bin/supervisord -c /etc/dd-agent/supervisor.conf
--- a/packaging/start_agent.sh
+++ b/packaging/start_agent.sh
@@ -3,6 +3,15 @@
 # All rights reserved
 # Licensed under Simplified BSD License (see LICENSE)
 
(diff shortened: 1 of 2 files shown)


packaging/start_agent.sh
```diff
@@ -5,4 +5,9 @@
 
 PATH=/opt/datadog-agent/embedded/bin:/opt/datadog-agent/bin:$PATH
 
+if [ "$DATADOG_ENABLED" = "no" ]; then
+    echo "Disabled via $AGENTSYSCONFIG. Exiting."
```
olivielpeau: `$AGENTSYSCONFIG` doesn't exist here, I'll update the PR

olivielpeau:
@truthbk This seems to work well but the way systemd logs what our script echoes is a bit weird when `DATADOG_ENABLED` is set to `no`: 

This is what systemd logs once `service datadog-agent start` is called on an ubuntu 16.04 VM. The agent does not actually start (as expected) but the logs might be misleading:
```
$ journalctl -u datadog-agent
Aug 23 18:46:29 ubuntu-xenial systemd[1]: Starting "Datadog Agent"...
Aug 23 18:46:29 ubuntu-xenial supervisorctl[28008]: ERROR: unix:///opt/datadog-agent/run/datadog-supervisor.sock no such file (already shut down?)
Aug 23 18:46:29 ubuntu-xenial systemd[1]: Started "Datadog Agent".
```
whereas `/var/log/syslog` does contain what we echo from `start_agent.sh`:
```
Aug 23 18:46:29 ubuntu-xenial systemd[1]: Starting "Datadog Agent"...
Aug 23 18:46:29 ubuntu-xenial start_agent.sh[28007]: Disabled via DATADOG_ENABLED env var. Exiting.
Aug 23 18:46:29 ubuntu-xenial supervisorctl[28008]: ERROR: unix:///opt/datadog-agent/run/datadog-supervisor.sock no such file (already shut down?)
Aug 23 18:46:29 ubuntu-xenial systemd[1]: Started "Datadog Agent".
```

I _think_ this is related to the `forking` type of our service, which might make systemd only log what the _forked_ processes echo to stdout.

Not sure this is a blocker, but let me know what you think.

truthbk:
The error: `ERROR: unix:///opt/datadog-agent/run/datadog-supervisor.sock no such file (already shut down?)` we see is due to `ExecStop` getting called. I think that since the script is exiting cleanly, the service is considered to have been up. Then systemd detects the process is gone, and calls `ExecStop`.

olivielpeau:
Going to go ahead and merge this, we can fix the slightly misleading log entry later.

PR #3500, opened 2017-08-29
Charly/adding docker inspect to flare

CharlyF:
### What does this PR do?

This PR adds the `docker inspect` of the Datadog containerized agent and redacts the API_KEY(s).
Placed as a separate file in the tar.bz2 called docker_inspect.log

### Motivation

When investigating, it's important to have a good grasp of the environment on top of the logs captured by the agent. With the inspect of the agent, it'll be easier to identify if the core components are correctly configured (socket, volumes, cgroups...)

### Testing Guidelines

Reproduction in my local environment.
Launching a flare in local creates the log correctly and it contains all the expected information.

### DIFF
--- a/utils/configcheck.py
+++ b/utils/configcheck.py
@@ -6,6 +6,10 @@
 import glob
 import os
 
+#3p
+import re
+import simplejson as json
+
 # project
 from config import (
     check_yaml,
@@ -51,6 +55,34 @@ def sd_configcheck(agentConfig):
         print("\nLoading check configurations...\n\n")
         configs = load_check_directory(agentConfig, get_hostname(agentConfig))
         get_sd_configcheck(agentConfig, configs)
+
+def agent_container_inspect():
+    # Self inspection based on cgroups
+    # On all platforms, the container ID is the last part of the path.
+    REGEX_PATTERN = '(.*/)+([a-z0-9]{64})$'
+
+    dockerutil = DockerUtil()
+    cgroup_path = '/proc/self/cgroup'
+    container_id = None
+
+    with open(cgroup_path, 'r') as f:
+        for ind in f:
+            id_match = re.search(REGEX_PATTERN, ind)
+            if id_match:
+                container_id = id_match.group(2)
+                break
+    if container_id is None:
+        print("The container_id could not be found. Refer to the docker log of the container running the agent")
+        return 1
+    try:
+        inspect = dockerutil.inspect_container(container_id)
+        key_indices = [i for i, k in enumerate(inspect['Config']['Env']) if 'API_KEY' in k]
+        for ind in key_indices:
+            inspect['Config']['Env'][ind] = '%s=%s' % (inspect['Config']['Env'][ind].split('=', 1)[0], 'redacted')
+        print json.dumps(inspect, indent=4)
+        return 0
+    except Exception as e:
+        print "Could not inspect container: %s" % e
 
 
 def get_sd_configcheck(agentConfig, configs):
--- a/utils/flare.py
+++ b/utils/flare.py
@@ -50,7 +50,7 @@ from utils.hostname import get_hostname
 from utils.jmx import jmx_command, JMXFiles
 from utils.platform import Platform
 from utils.sdk import load_manifest
-from utils.configcheck import configcheck, sd_configcheck
+from utils.configcheck import configcheck, sd_configcheck, agent_container_inspect
 from utils.windows_configuration import get_sdk_integration_paths
 
 from utils.logger import DisableLoggerInit
@@ -177,6 +177,10 @@ class Flare(object):
         self._permissions_file.close()
         self._add_file_tar(self._permissions_file.name, 'permissions.log',
                            log_permissions=False)
+
+        # Only add docker inspect if dockerized agent
+        if Platform.is_containerized():
+            self._add_command_output_tar('docker_inspect.log', agent_container_inspect)
 
     # Set the proxy settings, if they exist
     def set_proxy(self, options):


utils/configcheck.py
```diff
@@ -52,6 +52,31 @@ def sd_configcheck(agentConfig):
         configs = load_check_directory(agentConfig, get_hostname(agentConfig))
         get_sd_configcheck(agentConfig, configs)
 
+def docker_inspect():
+    print("-- docker_inspect -- \n")
+    dockerutil = DockerUtil()
+    containers = dockerutil.client.containers()
+    for cont in containers:
+        c_image = dockerutil.image_name_extractor(cont)
+        if 'datadog/docker-dd-agent' in c_image:
```
hkaj: That only works if the image is named like that.
- Some users replicate images they depend on in their self-hosted repository where docker-dd-agent may be named differently. We should check all RepoTags in this case.
- Trickier: if we can make this work even when users run a custom version of our image it would be best.

CharlyF: @hkaj indeed, I figured I would do 2 loops, one for the image where we only inspect if we find it (as inspect is expensive) and then, if indeed the user is running a custom image then I [inspect all containers searching for the dd-agent env variable](https://github.com/DataDog/dd-agent/blob/5c6469d3a3128dfb6763a8c8702b19b0ffc2f5f6/utils/configcheck.py#L72-L79)
And I stop as we find one.

I can look into the RepoTags too.

remh: The flare command is run within the container. You can get the container id from within a container by parsing
`/proc/self/cgroup`

e.g. of output of `/proc/self/cgroup`

```
11:devices:/docker/928221b873eb753b220c6011c4cabb041771dd3508080c3cffe67989b1115b58
10:freezer:/docker/928221b873eb753b220c6011c4cabb041771dd3508080c3cffe67989b1115b58
9:perf_event:/docker/928221b873eb753b220c6011c4cabb041771dd3508080c3cffe67989b1115b58
8:hugetlb:/docker/928221b873eb753b220c6011c4cabb041771dd3508080c3cffe67989b1115b58
7:cpu,cpuacct:/docker/928221b873eb753b220c6011c4cabb041771dd3508080c3cffe67989b1115b58
6:net_cls,net_prio:/docker/928221b873eb753b220c6011c4cabb041771dd3508080c3cffe67989b1115b58
5:blkio:/docker/928221b873eb753b220c6011c4cabb041771dd3508080c3cffe67989b1115b58
4:memory:/docker/928221b873eb753b220c6011c4cabb041771dd3508080c3cffe67989b1115b58
3:cpuset:/docker/928221b873eb753b220c6011c4cabb041771dd3508080c3cffe67989b1115b58
2:pids:/docker/928221b873eb753b220c6011c4cabb041771dd3508080c3cffe67989b1115b58
1:name=systemd:/docker/928221b873eb753b220c6011c4cabb041771dd3508080c3cffe67989b1115b58
```

utils/configcheck.py
```diff
@@ -52,6 +56,29 @@ def sd_configcheck(agentConfig):
...
 
+def agent_container_inspect():
+    print("-- docker_inspect -- \n")
+    dockerutil = DockerUtil()
+    # Self inspection based on cgroups
+
+    proc_path = dockerutil._docker_root + '/proc/self/cgroup'
+    with open(proc_path, 'r') as fp:
+        lines = map(lambda x: x.split(':'), fp.read().splitlines())
+        subsystems = dict(zip(map(lambda x: x[1], lines), map(dockerutil._parse_subsystem, lines)))
```
hkaj: you don't need these 2 lines I think, you can just work on the full line. Just do
```python
for line in fp.read().splitlines():
    ...
```

utils/configcheck.py
```diff
@@ -52,6 +56,29 @@ def sd_configcheck(agentConfig):
...
+        for k in subsystems.keys():
+            cgroup_regex_list = re.search('(.*/)+(.+?)$', subsystems[k])
```
hkaj: the regex should be a module constant. Also you can precise the group for the container id with `[a-z0-9]{64}`. Length is constant as you can see [here](https://github.com/moby/moby/blob/e23965d620e1655f44d47edd8b5e08c2c384fd03/pkg/stringid/stringid.go#L44-L59)

utils/configcheck.py
```diff
@@ -52,6 +56,29 @@ def sd_configcheck(agentConfig):
...
+            num_gr = len(cgroup_regex_list.groups())
+            containerID = cgroup_regex_list.group(num_gr) # On all platforms, the container ID is the last part of the path.
+    try:
+        inspect = dockerutil.client.inspect_container(containerID)
```
hkaj: please use `dockerutil.inspect_container` instead. It's just an alias right now but if we add a cache we want to only update this method to use it.

utils/configcheck.py
```diff
@@ -52,6 +56,29 @@ def sd_configcheck(agentConfig):
...
+        key = [i for i, k in enumerate(inspect['Config']['Env']) if 'API_KEY' in k]
```
hkaj: for improved clarity: key --> api_key_indices or similar

utils/configcheck.py
```diff
@@ -52,6 +56,29 @@ def sd_configcheck(agentConfig):
...
+        for ind in key:
+            inspect['Config']['Env'][ind] = "API_KEY=redacted"
```
hkaj: the key might not be API_KEY (it can be DD_API_KEY, there can be api keys for other tools when the agent is shipped in a pod, etc.)
```python
inspect['Config']['Env'][ind] = '%s=%s' % (inspect['Config']['Env'][ind].split('=', 1)[0], 'redacted')

File: utils/configcheck.py [hunk 7]
```diff
@@ -52,6 +56,29 @@ def sd_configcheck(agentConfig):
...
+    # Self inspection based on cgroups
+
+    proc_path = dockerutil._docker_root + '/proc/self/cgroup'
+    with open(proc_path, 'r') as fp:
+        lines = map(lambda x: x.split(':'), fp.read().splitlines())
+        subsystems = dict(zip(map(lambda x: x[1], lines), map(dockerutil._parse_subsystem, lines)))
+        for k in subsystems.keys():
+            cgroup_regex_list = re.search('(.*/)+(.+?)$', subsystems[k])
+            num_gr = len(cgroup_regex_list.groups())
+            containerID = cgroup_regex_list.group(num_gr) # On all platforms, the container ID is the last part of the path.
```
hkaj: as you mentioned, we can break early here

utils/configcheck.py
```diff
@@ -52,6 +56,29 @@ def sd_configcheck(agentConfig):
...
+    try:
+        inspect = dockerutil.client.inspect_container(containerID)
+        key = [i for i, k in enumerate(inspect['Config']['Env']) if 'API_KEY' in k]
+        for ind in key:
+            inspect['Config']['Env'][ind] = "API_KEY=redacted"
+        print json.dumps(inspect, indent=4)
+        return 1
```
hkaj: we generally return 0 on successful execution, != 0 means an error occurred (you can use it in case of exception for example)

masci:
Can we port this feature to v6?

CharlyF:
@hkaj Tested with --pid=host, the flare contains the docker_inspect.log and shares the host's process tree as expected. Also tested with the agent on the host (ran a redis container for the sake of testing) and the flare does not contain the docker_inspect.log, as expected.
I will then proceed to rebase and merge.

utils/configcheck.py
```diff
@@ -52,6 +56,32 @@ def sd_configcheck(agentConfig):
...
+
+    with open(cgroup_path, 'r') as fp:
+        cgroup_list = fp.readlines()
+        for ind in cgroup_list:
+            id_match = re.search(REGEX_PATTERN, ind)
+            if id_match:
+                num_gr = len(id_match.groups())
+                container_id = id_match.group(num_gr)
+                break
+    try:
```
xvello: We should have a logline and graceful return if `container_id is None`

PR #3507, opened 2017-09-06
[docker] catch docker.errors.NotFound for nonexisting images

hkaj:
if you have not yet done so.*

### What does this PR do?

Inspecting non existant images raise a NotFound error that spams the collector logs. This PR catches it and return None. If the image doesn't exist the process stops here.

### Motivation

Sometimes a container is stopped, and its image is forcibly removed. In this case when we try and extract the image info from the container, we hit a NotFound error that was not expected. Customers using Nomad seem to hit this issue often. Not sure if it's Nomad that force removes images, or users' scripts, but we can and should handle it.

### Testing Guidelines

### Additional Notes

### DIFF
--- a/utils/dockerutil.py
+++ b/utils/dockerutil.py
@@ -598,8 +598,8 @@ class DockerUtil:
                         return name
                     except (LookupError, TypeError) as e:
                         log.debug("Failed finding image name in RepoTag and RepoDigests: %s", e)
-            except Exception:
-                log.error("Exception getting docker image name")
+            except Exception as ex:
+                log.error("Exception getting docker image name: %s" % str(ex))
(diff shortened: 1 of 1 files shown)


utils/dockerutil.py
```diff
@@ -595,7 +599,7 @@ def image_name_resolver(self, image):
                     except (LookupError, TypeError) as e:
                         log.debug("Failed finding image name in RepoTag and RepoDigests: %s", e)
             except Exception:
-                log.exception("Exception getting docker image name")
+                log.error("Exception getting docker image name")
```
xvello: we should print `% str(e)` though, to get the error

PR #3509, opened 2017-09-07
[Core] Update the Main Credential regex to support

nmuesch:
multiple types of API input in datadog.conf

if you have not yet done so.*

### What does this PR do?

Modified the regex used when checking for the API Key in datadog.conf or other locations. Currently we only support the syntax `api_key: <KEY>` However, this doesn't allow for a space before the `:` or using `api_key = <KEY>`

### Motivation

Currently the API key isn't being fully redacted in the flare if it doesn't use the aforementioned format. It seems like K8s Agents use the `=` syntax by default.

### Testing Guidelines

Performed a series of manual tests locally based on the above syntaxes.

### Additional Notes

With this PR, the flare will be unsuccessful if the key is <YOUR_API_KEY_HERE> with a seemingly helpful message saying the API key is invalid. So the flare will not be sent at all, but the user sees this message. 

Initially attempted to remove lambdas to make it a bit cleaner, however this doesn't work for the case where there are multiple API keys, comma delimited. We would need another function to be called in the replacement piece that split the matched group into segments.

### DIFF
--- a/utils/flare.py
+++ b/utils/flare.py
@@ -94,13 +94,8 @@ class Flare(object):
     ]
     MAIN_CREDENTIALS = [
         CredentialPattern(
-            re.compile('^\s*api_key\s*[=, :]( *\w+(\w{5}) ?,?)+$'),
-            lambda matchobj:  'api_key: ' + ', '.join(map(
-                lambda key: '*' * 26 + key[-5:],
-                map(lambda x: x.strip(),
-                    matchobj.string.split("api_key")[1].split(',')
-                    )
-            )),
+            re.compile(r'^\s*api_key\s*[=, :]\s*\w{27}(\w{5})$'),
+            r'api_key: ***************************\1',
             'api_key'
         ),
         CredentialPattern(
--- a/utils/flare.py
+++ b/utils/flare.py
@@ -93,7 +93,7 @@ class Flare(object):
         ),
     ]
     MAIN_CREDENTIALS = [
-         CredentialPattern(
+        CredentialPattern(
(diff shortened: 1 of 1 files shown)


utils/flare.py
```diff
@@ -94,11 +94,11 @@ class Flare(object):
     ]
     MAIN_CREDENTIALS = [
         CredentialPattern(
-            re.compile('^\s*api_key:( *\w+(\w{5}) ?,?)+$'),
+            re.compile('^\s*api_key\s*[=, :]( *\w+(\w{5}) ?,?)+$'),
```
masci: I think this piece of code is overly complex, your patch is good but I'd make it a little bit simpler.

1) we can change the regex to only capture in a group the last 5 digits of the api key 
2) instead of nesting maps and lambdas, proceed with a string replace using a backreference to the group from above

the following piece of code should do the same thing (untested):
```
CredentialPattern(
  re.compile(r'^\s*api_key\s*[=, :]\s*\w{27}(\w{5})$'),
  r'api_key: ***************************\1',
  'api_key'
)
```

xvello:
Thanks @nmuesch, that new format is introduced in `docker-dd-agent` 12.0.5160+ 's new entrypoint. It would be great to add to 5.17.1.

utils/flare.py
nmuesch: I definitely agree here. Removing the lambdas and just replacing everything but the last 5 chars does indeed do the trick. Thanks!

nmuesch: Looks like it doesn't work for comma separated API keys, left more details in the notes based on our discussions

PR #3521, opened 2017-09-14
Rename library from wmi_check.py -> winwmi_check.py as the class that it

derekwbrown:
provides is WinWMICheck, and having the name wmi_check.py was
confusing the python check loader in agent6 between this file and wmi_check.py (the
actual check).  However, since the checks are shared (integrations-core)
the corresponding change has to be made here

Requires 3 PRs to go in together
https://github.com/DataDog/integrations-core/pull/752
https://github.com/DataDog/datadog-agent/pull/568
https://github.com/DataDog/dd-agent/pull/3521


checks/wmi_check.py
```diff
@@ -1,344 +1,13 @@
...
-        timezone = None
-
-    return year, month, day, hours, minutes, seconds, microseconds, timezone
+# provided for backward compatibility.  wmi_check.py renamed to winwmi_check.py
+# to prevent collision with the actual wmi check; provide the redirect for
+# any agent check that uses the base library
+#
+# this file will be deprecated in Agent6
+from checks.winwmi_check import (
+    InvalidWMIQuery,
```
olivielpeau: you're missing one import for full BC: `WMIMetric`

PR #3528, opened 2017-09-20
do not crash if ENV in CONF is None

CharlyF:
### What does this PR do?

If envvar is NoneType, we want to return the tags directly instead of crashing.

### Motivation

Support case brought up by the community.

### Testing Guidelines

Set up a DCOS cluster v 1.9.2, deployed the agent running 5.17.2 and witnessed the successful auto discovery of a redis container along with the collection of Mesos/Marathon specific tags.


alexcristi:
👍

PR #3529, opened 2017-09-21
[Core] Update the normalize to also convert @ symbol to _

nmuesch:
if you have not yet done so.*

### What does this PR do?

Some device names can have the `@` symbol in them. This PR converts that symbol to a `_` using the standard normalize method.

### Motivation

Since some deice names can have this symbol, and we aren't currently sanitizing, it can cause undesired behaviour in the UI, since its not one of our supported tag chars. 

### Testing Guidelines

### Additional Notes

This is to go along with an integrations core PR for the disk check so that can use this updated normalize method.

### DIFF
--- a/checks/__init__.py
+++ b/checks/__init__.py
@@ -93,7 +93,7 @@ class Check(object):
         """Turn a metric into a well-formed metric name
         prefix.b.c
         """
-        name = re.sub(r"[,\+\*\-/()\[\]{}\s]", "_", metric)
+        name = re.sub(r"[,\@\+\*\-/()\[\]{}\s]", "_", metric)
         # Eliminate multiple _
         name = re.sub(r"__+", "_", name)
         # Don't start/end with _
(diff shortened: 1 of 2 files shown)


checks/__init__.py
```diff
@@ -889,7 +889,7 @@ def normalize(self, metric, prefix=None, fix_case=False):
             if prefix is not None:
                 prefix = self.convert_to_underscore_separated(prefix)
         else:
-            name = re.sub(r"[,\+\*\-/()\[\]{}\s]", "_", metric_name)
+            name = re.sub(r"[,\@\+\*\-/()\[\]{}\s]", "_", metric_name)
```
truthbk: Could you also check if we should fix, `normalize` for the main Check class (https://github.com/DataDog/dd-agent/blob/ea4b687c87719e7bb5d8992df059d414ad6d2a22/checks/__init__.py#L92) here as well? In reality most checks should inherit from `AgentCheck` but it might make sense for legacy custom checks.

nmuesch: I don't think it needs to be updated there since its a deprecated class, but it doesn't hurt to add it :)

truthbk: I was asking in case customers had implemented custom checks based on the deprecated class :)

nmuesch: Ahh, that makes sense. Good call, thanks!

PR #3530, opened 2017-09-22
Docker_daemon puts tags ECS's task-id (task:arn)

h-imaoka:
### What does this PR do?

Via docker_daemon with ecs orchestration, put ecs's task-id (=task-arn) as a tag.

### Motivation

We want to draining ecs tasks that contains NG container.
Now, I found NG container's name & instance-id, but no task-id(arn). 

### Testing Guidelines

Unit test, `tests/core/test_ecsutil.py` passed `nosetests tests/core/test_ecsutil.py` 
No config-file change.

### DIFF
--- a/tests/core/test_ecsutil.py
+++ b/tests/core/test_ecsutil.py
@@ -23,11 +23,16 @@ class TestECSUtil(unittest.TestCase):
         util.agent_url = 'http://dummy'
 
         mock_get.reset_mock()
-        mock_get.return_value = MockResponse({"Tasks": [{"Family": "dd-agent-latest", "Version": "12",
-                                                         "Containers": [{"DockerId": CO_ID}]}]}, 200)
+        mock_get.return_value = MockResponse(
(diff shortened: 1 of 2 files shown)


h-imaoka:
I found other way to get task-arn via `docker inspect`.
dd-agent has `collect_labels_as_tags` for it.
```
# docker_daemon.yml
collect_labels_as_tags: ["com.amazonaws.ecs.task-arn"]
``` 

When ecs runs tasks (= task similar k8s's pod, task contains any containers), publish task-id (task-arn). ECS operation (e.g. kill task) with that `task-id`.Even if we know docker's container-id or  container-name, can not ECS's operation (without  ssh login to container instances and  docker kill container-id `rough` commands. )

e.g. 
task-id x  has  container A and B
task-id y has  container M and N

Datadog detect container A's issue, but if we know ecs's task-id, we can not kill task-id x via ecs operation.

xvello:
Hi @h-imaoka that looks like a safe workaround for now. We'll add it to high-cardinality collection for agent6.

PR #3531, opened 2017-09-22
Added: env var for config path override

agamdua:
### What does this PR do?

Adds a way to specify a separate path for a config file.

### Motivation

Would like to manage the config file at a separate location from the default.

Specifically, I would like to mount the config file to a container separate from the whole directory. I cannot create a separate docker image with a config file baked in since it needs to be templated at runtime dependent on what the host is. This is managed by the CM tool at the moment.

### Testing Guidelines

I am happy to write tests if there's some sort of acceptance on the approach going forward.

### Additional Notes

I am not sure if
* this is the best place to make this change (it seemed like it)
* there will be any interference with a variable of the same name in `dogstatsd`. I see that we leverage `os.environ` to fetch the variable there and thought injecting it here may help in defining that. Happy to work in a different place if my assumptions are incorrect.

### DIFF
--- a/config.py
+++ b/config.py
@@ -238,6 +238,11 @@ def get_config_path(cfg_path=None, os_name=None):
     if cfg_path is not None and os.path.exists(cfg_path):
         return cfg_path
 
+    env_path_override = os.environ.get('DD_CONF_PATH')
+
+    if env_path_override is not None and os.path.exists(env_path_override):
+        return env_path_override
+
     # Check if there's a config stored in the current agent directory
     try:
         path = os.path.realpath(__file__)


gmmeyer:
I think this is a change that makes sense! DD_CONF_PATH is a way of passing the variable to the upcoming Dogstatsd version 6 (part of the larger Agent v6). I think it makes sense to set it in the same way as the other one!

If you'd like to write a test, please do! That would be great! But I think this pull request makes sense. I also think it makes sense for the cli setting to take precedence over the environment variable setting.

gmmeyer:
Also, I forgot, I just want to thank you very much for doing this @agamdua! This is a great contribution, and all of our community contributions are super valuable. 😄

agamdua:
Great! Let me try to find some time to work on it this week.

Do you know if there's anything I need to do to handle Windows or will this location hit all three? Its been a while since I looked at the codebase and I see thats a question I had even at the time of opening.

gmmeyer:
it should also work on windows, though on windows config paths are a bit different than they are on Linux.

PR #3535, opened 2017-10-02
[windows] Fix problem with non-English Windows.

derekwbrown:
Many PDH counter names are localized.  However, there's a translation
table in the registry.  Implements the lookup table so agent runs on
non-English windows

### DIFF
--- a/checks/libs/win/winpdh.py
+++ b/checks/libs/win/winpdh.py
@@ -1,17 +1,43 @@
 import time
 import win32pdh
+import _winreg
 
 DATA_POINT_INTERVAL = 0.01
 
 class WinPDHCounter(object):
+    # store the dictionary of pdh counter names
+    pdh_counter_dict = {}
+
     def __init__(self, class_name, instance_name):
-        self._class_name = class_name
-        self._instance_name = instance_name
+        self._get_counter_dictionary()
+        self._class_name = win32pdh.LookupPerfNameByIndex(None, int(WinPDHCounter.pdh_counter_dict[class_name]))
+        self._instance_name = win32pdh.LookupPerfNameByIndex(None, int(WinPDHCounter.pdh_counter_dict[instance_name]))
         self.hq = win32pdh.OpenQuery()
 
     def __del__(self):
         if(self.hq):
             win32pdh.CloseQuery(self.hq)
+
+    def _get_counter_dictionary(self):
+        if WinPDHCounter.pdh_counter_dict:
+            # already populated
+            return
+
+        try:
+            val, t = _winreg.QueryValueEx(_winreg.HKEY_PERFORMANCE_DATA, "Counter 009")
+        except Exception as e:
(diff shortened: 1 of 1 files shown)


checks/libs/win/winpdh.py
```diff
@@ -1,18 +1,44 @@
...
 
+    def _get_counter_dictionary(self):
+        if WinPDHCounter.pdh_counter_dict:
+            # already populated
+            return
+
+        try:
+            val, t = _winreg.QueryValueEx(_winreg.HKEY_PERFORMANCE_DATA, "Counter 009")
+        except Exception as e:
+            print "Exception %s" % str(e)
```
hush-hush: Should we print the exception if we raise again ? If yes, maybe should we print some context ?

checks/libs/win/winpdh.py
```diff
@@ -1,18 +1,44 @@
...
+            raise
+
+        # create a table of the keys to the counter index, because we want to look up
+        # by counter index.
+        idx = 0
+        idx_max = len(val)
+        while idx < idx_max:
+            WinPDHCounter.pdh_counter_dict[val[idx+1]] = val[idx]
```
hush-hush: Could you add a comment explaining why `val[idx+1]` is the key and `val[idx]` the value ?

checks/libs/win/winpdh.py
```diff
@@ -1,18 +1,44 @@
...
         if(self.hq):
             win32pdh.CloseQuery(self.hq)
 
+    def _get_counter_dictionary(self):
+        if WinPDHCounter.pdh_counter_dict:
+            # already populated
+            return
+
+        try:
+            val, t = _winreg.QueryValueEx(_winreg.HKEY_PERFORMANCE_DATA, "Counter 009")
```
truthbk: Can we document what this _"Counter 009"_ thing is - maybe a url (is it this https://msdn.microsoft.com/en-us/library/windows/desktop/aa373181(v=vs.85).aspx ?)

checks/libs/win/winpdh.py
```diff
@@ -1,18 +1,44 @@
 import time
 import win32pdh
+import _winreg
 
 DATA_POINT_INTERVAL = 0.01
 
 class WinPDHCounter(object):
+    # store the dictionary of pdh counter names
```
truthbk: Could we add some more info into the comment about how the dictionary translates from "US-English" name to the index identifier for the counter?

checks/libs/win/winpdh.py
derekwbrown: Yeah... that's windows for you.  009 is the locale code, but there's no definition for it.  The docs (I added to the comment) actually say to add "Counter 009" as the parameter :(

truthbk:
Changes superseded by #3536, closing this one. Thanks @derekwbrown

PR #3546, opened 2017-10-13
Avoid starting the agent if the api key is too short

masci:
if you have not yet done so.*

### What does this PR do?

Performs a sanity check before starting the agent to avoid collecting anything if the api key is obviously wrong.

Enhance the message on the status page when an api key is invalid

### Motivation

The agent might start collecting metrics but never flush if users paste a wrong api key

### DIFF
--- a/checks/check_status.py
+++ b/checks/check_status.py
@@ -131,15 +131,15 @@ def validate_api_key(config):
             timeout=3, verify=(not config.get('skip_ssl_validation', False)))
 
         if r.status_code == 403:
-            return "API Key is invalid"
+            return "[ERROR] API Key is invalid"
 
         r.raise_for_status()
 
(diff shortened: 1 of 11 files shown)


xvello:
Fixes https://github.com/DataDog/docker-dd-agent/issues/250

ISauve:
Note: the `process-agent` and `trace-agent` treat a missing/empty/shortened API key differently than the `dd-agent` does, and since supervisor does not have the concept of “dependencies” each process is on its own. We observed that an empty API key lead the `process-agent` to start as usual, an API key that was too short made both processes start, and a missing API key (line completely commented on in conf.d) meant neither started. This PR doesn't affect those agents at all, so the observed behaviour is not a bug that introduced here - it's just how the other agents work.

PR #3553, opened 2017-10-25
[AD] Fix port parsing

hkaj:
if you have not yet done so.*

### What does this PR do?

We shouldn't rely on `ExposedPorts`. Let's just work with `Ports`. 

### Motivation

This is confusing and ExposedPorts is only indicative anyway. Plus if we find something there we skip the Kubernetes pod lookup which we should do by all means.

### Testing Guidelines

### Additional Notes

### DIFF
--- a/utils/service_discovery/sd_docker_backend.py
+++ b/utils/service_discovery/sd_docker_backend.py
@@ -95,7 +95,7 @@ class SDDockerBackend(AbstractSDBackend):
                 self.kubeutil = KubeUtil()
             except Exception as ex:
                 log.error("Couldn't instantiate the kubernetes client, "
-                    "subsequent kubernetes calls will fail as well. Error: %s" % str(ex))
+                          "subsequent kubernetes calls will fail as well. Error: %s" % str(ex))
 
         self.metadata_collector = MetadataCollector()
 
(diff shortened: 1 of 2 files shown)


utils/service_discovery/sd_docker_backend.py
```diff
@@ -225,43 +225,37 @@ def _extract_ip_from_networks(self, ip_dict, tpl_var):
...
+        ports = []
         try:
-            ports = map(lambda x: x.split('/')[0], container_inspect['NetworkSettings']['Ports'].keys())
-            if len(ports) == 0: # There might be a key Port in NetworkSettings but no ports so we raise IndexError to check in ExposedPorts
+            ports = [x.split('/')[0] for x in container_inspect['NetworkSettings']['Ports'].keys()]
+            if len(ports) == 0:
                 raise IndexError
         except (IndexError, KeyError, AttributeError):
-            # try to get ports from the docker API. Works if the image has an EXPOSE instruction
-            ports = map(lambda x: x.split('/')[0], container_inspect['Config'].get('ExposedPorts', {}).keys())
```
xvello: That line might get AD working for one customer with an exotic version. Instead of removing it, I'd go with
```
if Platform.is_k8s():
    kube_way
else:
   exposedports


Just as a precaution

hkaj: that's a good idea

hkaj: @xvello fixed

PR #3554, opened 2017-10-26
Enable the apm service based on the configuration file

derekwbrown:
### What does this PR do?

Checks datadog.conf. If `apm_enabled` is set to true, then it starts the trace agent service.  Trace agent service will be installed by parallel checkin


win32/service.py
```diff
@@ -176,6 +177,13 @@ def SvcDoRun(self):
         for proc in self.procs.values():
             proc.start()
 
+        # check to see if apm is enabled.
+        if self.config["apm_enabled"]:
```
masci: Nit: it'd be safer to set `self.config = {}` in the `__init__` and change this to `if self.config.get('apm_enabled'):`

PR #3555, opened 2017-10-30
[wheels] allow running checks packaged as wheels

truthbk:
if you have not yet done so.*

### What does this PR do?

The goal of this PR is to allow running checks packages as pip wheels just like any other check. These checks would typically packaged as follows such that they will be installed to `./datadog/<integration>/` site in `site-packages`. 

It should remain completely backward compatible as far as previous functionality goes.

### Motivation

Improved packaging of checks, improved developer experience.

### Testing Guidelines

### DIFF
--- a/config.py
+++ b/config.py
@@ -21,6 +21,7 @@ import string
 import sys
 import traceback
 from urlparse import urlparse
+from importlib import import_module
 
 # 3p
 import simplejson as json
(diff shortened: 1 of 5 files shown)


config.py
```diff
@@ -1069,16 +1095,28 @@ def load_check_from_places(check_config, check_name, checks_places, agentConfig)
...
+        prev_failures = bool(load_failure)
+        check_is_valid, check_class, load_failure = get_valid_check_class(check_name, check_path, from_site=is_wheel)
         if not check_is_valid:
+            load_error = load_failure.get(check_name, {}).get('error')
+            if is_wheel and not prev_failures and isinstance(load_error, ImportError):
+                load_failure = {}
             continue
 
-        if manifest_path:
+        if manifest_path or is_wheel:
```
masci: if `manifest_path` is false, we still call `validate_sdk_check(manifest_path)` and produce a warning, is this expected?

truthbk: the idea was to ship the manifest with the wheels, but it's not that straight-forward...

masci: If we delegate dependency management to pip (and assume an incompatible check can't just be installed) there's no need to ship/use the manifest at runtime

PR #3557, opened 2017-10-31
Switch most of container host tags to host metadata

hkaj:
if you have not yet done so.*

### What does this PR do?

Move all host tags coming from docker and orchestrators to host metadata with the exception of node labels in k8s because they're legit.

### Motivation

This metadata is not really useful as tags and it's mostly for analysis purpose. Therefore metadata is a better fit that tags - especially because host metadata doesn't churn contexts when updated.

### Testing Guidelines

### Additional Notes

This will require some work in the app to display it. Still figuring that out.

### DIFF
--- a/utils/kubernetes/kubeutil.py
+++ b/utils/kubernetes/kubeutil.py
@@ -480,7 +480,7 @@ class KubeUtil:
             meta['kubelet_version'] = version[1:]
         except Exception as ex:
             log.debug("Error getting Kubelet version: %s" % str(ex))
-            
+
         return meta
 
 
@@ -496,7 +496,7 @@ class KubeUtil:
             _, node_name = self.get_node_info()
             if not node_name:
                 raise ValueError("node name missing or empty")
-            
+
             request_url = "%s/nodes/%s" % (self.kubernetes_api_url, node_name)
             node_info = self.retrieve_json_auth(request_url).json()
             node_labels = node_info.get('metadata', {}).get('labels', {})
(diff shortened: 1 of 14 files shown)


utils/orchestrator/baseutil.py
```diff
@@ -22,6 +22,8 @@ class BaseUtil:
       - _get_transient_tags: tags can change and won't be cached (TODO)
       - invalidate_cache: custom cache invalidation logic
       - is_detected (staticmethod)
+      - get_host_tags: list of tags that are applied to the host in Datadog
+      - get_host_metadata: dict of container-related host metadata (typically node labels)
```
xvello: Bad example, aren't we exporting node labels as tags and not metadata?

utils/kubernetes/kubeutil.py
```diff
@@ -526,10 +548,6 @@ def extract_event_tags(self, event):
             tags.append('node_name:%s' % event['source']['host'])
         if 'kind' in event.get('involvedObject', {}):
             tags.append('object_type:%s' % event['involvedObject'].get('kind', '').lower())
-        if 'name' in event.get('involvedObject', {}):
```
xvello: Probably a bad rebase (these are quite new), please keep them

utils/orchestrator/baseutil.py
hkaj: oops yeah wrong line

hkaj:
@xvello can you have a second look 🙏

PR #3563, opened 2017-11-02
[kube] restore kube_deployment tag on k8s 1.8+ clusters

xvello:
### What does this PR do?

Kubernetes 1.8 changed the generation of ReplicaSet names, they’re no longer digits only, and this broke the agent’s metric collection code, see https://github.com/kubernetes/kubernetes/pull/51538/files and https://github.com/DataDog/integrations-core/issues/837

### Testing Guidelines

Updated unit test cases to cover valid and invalid cases for this new format.

### DIFF
--- a/utils/kubernetes/kubeutil.py
+++ b/utils/kubernetes/kubeutil.py
@@ -40,6 +40,9 @@ CREATOR_KIND_TO_TAG = {
 DEFAULT_INIT_RETRIES = 0
 DEFAULT_RETRY_INTERVAL = 20  # seconds
 
+# Taken from https://github.com/kow3ns/kubernetes/blob/96067e6d7b24a05a6a68a0d94db622957448b5ab/staging/src/k8s.io/apimachinery/pkg/util/rand/rand.go#L76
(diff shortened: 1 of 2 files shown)


utils/kubernetes/kubeutil.py
```diff
@@ -374,9 +377,17 @@ def get_deployment_for_replicaset(self, rs_name):
         https://github.com/kubernetes/kubernetes/blob/release-1.6/pkg/controller/deployment/sync.go#L299
         But it might change in a future k8s version. The other way to match RS and deployments is
         to parse and cache /apis/extensions/v1beta1/replicasets, mirroring PodServiceMapper
+        In 1.8, the hash generation logic changed: https://github.com/kubernetes/kubernetes/pull/51538/files
```
hkaj: i'd like to see a note here that we may have a few false positive. It's okay that we do until we have the cluster agent, but we should document it somewhere and what better place than where it happens?

xvello: Added comment

xvello:
Successfully tested on a 1.8.1 cluster

PR #3565, opened 2017-11-02
change restart to stop/start with deps

derekwbrown:
Changes gui implementation to deal with the fact that the datadog-agent service now may have dependent services.  Stop them all, and then restart (or just stop them all when stopping).

### DIFF
--- a/win32/gui.py
+++ b/win32/gui.py
@@ -735,7 +735,7 @@ def restartServiceWithDeps(serviceName):
                     for restart_dep in stopped_deps:
                         try:
                             win32serviceutil.StartService(restart_dep)
-                        except:
+                        except Exception:
                             # this is best effort, just continue restarting
                             pass
                     raise
@@ -746,16 +746,16 @@ def restartServiceWithDeps(serviceName):
         hs = win32service.OpenService(hscm, serviceName, win32service.SERVICE_ALL_ACCESS)
         try:
             win32serviceutil.__StopServiceWithTimeout(hs)
-        except:
+        except Exception:
             # if this fails to stop, then restart the dependent services we
             # already stopped.
             for restart_dep in stopped_deps:
                 try:
                     win32serviceutil.StartService(restart_dep)
-                except:
+                except Exception:
                     # this is best effort, just continue restarting
                     pass
-                raise
+            raise
         finally:
             win32service.CloseServiceHandle(hs)
     finally:
--- a/win32/gui.py
+++ b/win32/gui.py
@@ -717,11 +717,12 @@ def check_yaml_syntax(content):
 def service_manager(action):
     try:
         if action == 'stop':
-            win32serviceutil.StopService(DATADOG_SERVICE)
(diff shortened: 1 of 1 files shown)


olivielpeau:
We discussed this offline: `win32serviceutil.RestartService` doesn't only do a `Stop` and a `Start` but also catches some expected errors (see https://github.com/SublimeText/Pywin32/blob/0d2174cb24e6eac613a62b932c7562d10cc49bc4/lib/x64/win32/lib/win32serviceutil.py#L423-L441), we should have the same kind of logic in our restart

win32/gui.py
```diff
@@ -713,15 +715,67 @@ def check_yaml_syntax(content):
...
+            try:
+                win32serviceutil.__StopServiceWithTimeout(hs)
+                stopped_deps.insert(0, dep)
+            except pywintypes.error as exc:
+                if exc.winerror != winerror.ERROR_SERVICE_NOT_ACTIVE:
+                    # walk through and start all the services that have been stopped
+                    for restart_dep in stopped_deps:
+                        try:
+                            win32serviceutil.StartService(restart_dep)
+                        except:
```
olivielpeau: nit: can you change all the `except:` to `except Exception:`? Always better to avoid the catch-all 👍

win32/gui.py
```diff
@@ -713,15 +715,67 @@ def check_yaml_syntax(content):
...
+        except:
+            # if this fails to stop, then restart the dependent services we
+            # already stopped.
+            for restart_dep in stopped_deps:
+                try:
+                    win32serviceutil.StartService(restart_dep)
+                except:
+                    # this is best effort, just continue restarting
+                    pass
+                raise
```
olivielpeau: the `raise` here should be one indent lower (i.e. outside the `for` loop)

derekwbrown: Good catch.

derekwbrown: I've been able to test the mainline code, starting and stopping and restarting dd-agent with and without trace agent installed.  Haven't been able to test the edge cases, as the services all start and stop the way they're supposed to

win32/gui.py
```diff
@@ -713,15 +715,67 @@ def check_yaml_syntax(content):
...
+                            pass
+                    raise
+            finally:
+                win32service.CloseServiceHandle(hs)
+
+        # all the deps are stopped.  Stop the one we really wanted to stop
+        hs = win32service.OpenService(hscm, serviceName, win32service.SERVICE_ALL_ACCESS)
+        try:
+            win32serviceutil.__StopServiceWithTimeout(hs)
+        except:
```
olivielpeau: `except:` -> `except Exception:` here too

win32/gui.py
```diff
@@ -713,15 +715,67 @@ def check_yaml_syntax(content):
...
+            # if this fails to stop, then restart the dependent services we
+            # already stopped.
+            for restart_dep in stopped_deps:
+                try:
+                    win32serviceutil.StartService(restart_dep)
+                except:
```
olivielpeau: and here :)

PR #3571, opened 2017-11-04
[kubernetes] Allow overriding the pod name

darnaut:
### What does this PR do?

Allows the default pod name to be set by an environment variable named `KUBERNETES_POD_NAME`.

### Motivation

Currently in Kubernetes, when a pod is created, its hostname is the Pod's `metadata.name` value. As a result, the Kubernetes check assumes that the pod's hostname is the same as the pod's name when locating pod metadata information.

But that's not necessarily always the case as the hostname can be set to any other value using a Pod annotation. If that's the case, the Kubernetes check is not able to locate pod _and_ node information as it expects the hostname to be the same as the pod name.

This is especially problematic when a pod uses host networking, where the pod's hostname is the same as the host's hostname.

### DIFF
--- a/tests/core/test_kubeutil.py
+++ b/tests/core/test_kubeutil.py
@@ -34,6 +34,14 @@ class KubeTestCase(unittest.TestCase):
         json_array = cls._load_json_array(names)
         return map(lambda x: MockResponse(x, 200), json_array)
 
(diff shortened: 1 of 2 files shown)


hkaj:
CI failure is unrelated, merging.

PR #3575, opened 2017-11-06
Nick/normalize device name

nmuesch:
if you have not yet done so.*

### What does this PR do?

Converts the device name tag to confirm to tag standards. The existing normalize function will strip out the device name's `\`, so I've created its own function `normalize_device_name` in the Agent Check class. 

### Motivation

Having the @ symbol in the tag seems to cause issues in app.

### Testing Guidelines

### Additional Notes

### DIFF
--- a/tests/core/test_common.py
+++ b/tests/core/test_common.py
@@ -119,7 +119,9 @@ class TestCore(unittest.TestCase):
 
         self.assertEqual(self.ac.normalize("PauseTotalNs", "prefix", fix_case = True), "prefix.pause_total_ns")
         self.assertEqual(self.ac.normalize("Metric.wordThatShouldBeSeparated", "prefix", fix_case = True), "prefix.metric.word_that_should_be_separated")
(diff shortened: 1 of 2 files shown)


tests/core/test_common.py
```diff
@@ -119,6 +119,7 @@ def test_name(self):
 
         self.assertEqual(self.ac.normalize("PauseTotalNs", "prefix", fix_case = True), "prefix.pause_total_ns")
         self.assertEqual(self.ac.normalize("Metric.wordThatShouldBeSeparated", "prefix", fix_case = True), "prefix.metric.word_that_should_be_separated")
+        self.assertEqual(self.ac.normalize_device_name("//device@name"), "//device_name")
```
hush-hush: Since `normalize_device_name`  replace a lot more than just `@` we should test them also.

something like:
```python
self.assertEqual(self.ac.normalize_device_name(",@+*-()[]{}//device@name"), "___________//device_name")
```

PR #3581, opened 2017-11-10
[Windows] Update pdh library to support generic pdh based checks

derekwbrown:
Updates the PDH library.  Provide a generic PDHBaseCheck to build PDH based system checks.

### DIFF
--- a/checks/libs/win/pdhbasecheck.py
+++ b/checks/libs/win/pdhbasecheck.py
@@ -1,0 +1,64 @@
+# (C) Datadog, Inc. 2013-2016
+# All rights reserved
+# Licensed under Simplified BSD License (see LICENSE)
+
+# project
+from checks import AgentCheck
+from utils.containers import hash_mutable
+# datadog
+try:
+    from checks.libs.win.winpdh import WinPDHCounter
+except ImportError:
+    class WinPDHCounter(*args, **kwargs):
+        return
+
+class PDHBaseCheck(AgentCheck):
+    """
+    PDH based check.  check.
+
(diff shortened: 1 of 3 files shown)


checks/libs/win/pdhbasecheck.py
```diff
@@ -0,0 +1,64 @@
...
+# Licensed under Simplified BSD License (see LICENSE)
+
+# project
+from checks import AgentCheck
+from utils.containers import hash_mutable
+# datadog
+try:
+    from checks.libs.win.winpdh import WinPDHCounter
+except ImportError:
+    def WinPDHCounter(*args, **kwargs):
```
gmmeyer: this should be a class instead of a function? You'd probably want to have the check error out if it can't import it, but not on import. That will be ugly

checks/libs/win/winpdh.py
```diff
@@ -8,24 +8,24 @@ class WinPDHCounter(object):
     # store the dictionary of pdh counter names
     pdh_counter_dict = {}
 
-    def __init__(self, class_name, instance_name, log):
+    def __init__(self, class_name, counter_name, log, instance_name = None):
```
gmmeyer: Will making these changes affect the pdh check in integrations core?

derekwbrown: No; the pdh check always will supply no instance name. The problem is confusion between what PDH calls countersets, counters, and instances, and what we generally refer to as instances.  The generic PDH counter does all instances of the counters, so it's `instance_name` param is always none.

PR #3586, opened 2017-11-16
[kube] fix deployment tagging for 1.8+

xvello:
my bad, RS new names' lenght is not constant. Let's try to validate every suffix more than 2 runes long

See https://github.com/DataDog/integrations-core/issues/837#issuecomment-344911890

### DIFF
--- a/utils/kubernetes/kubeutil.py
+++ b/utils/kubernetes/kubeutil.py
@@ -388,7 +388,7 @@ class KubeUtil:
         if end > 0 and rs_name[end + 1:].isdigit():
             # k8s before 1.8
             return rs_name[0:end]
-        if end > 0 and len(rs_name[end + 1:]) == 10:
+        if end > 0 and len(rs_name[end + 1:]) > 2:
             # k8s 1.8+ maybe? Check contents
             for char in rs_name[end + 1:]:
                 if char not in ALLOWED_ENCODESTRING_ALPHANUMS:
--- a/utils/kubernetes/kubeutil.py
+++ b/utils/kubernetes/kubeutil.py
@@ -380,9 +380,13 @@ class KubeUtil:
         to parse and cache /apis/extensions/v1beta1/replicasets, mirroring PodServiceMapper
         In 1.8, the hash generation logic changed: https://github.com/kubernetes/kubernetes/pull/51538/files
 
-        As we are matching both patterns without checking the apiserver version, we might have
-        some false positives. For agent6, we plan on doing this pod->replicaset->deployment matching
-        in the cluster agent, with replicaset data from the apiserver. This will address that risk.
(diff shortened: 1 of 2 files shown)


utils/kubernetes/kubeutil.py
```diff
@@ -388,7 +388,7 @@ def get_deployment_for_replicaset(self, rs_name):
         if end > 0 and rs_name[end + 1:].isdigit():
             # k8s before 1.8
             return rs_name[0:end]
-        if end > 0 and len(rs_name[end + 1:]) == 10:
+        if end > 0 and len(rs_name[end + 1:]) > 2:
```
hkaj: If there's a dash in the replica set and at least 2 runes after it we'll assume it's a deployment. That's not great, but I don't see any other short-term fix either :/
Maybe we should document that (as a comment)? I don't expect this to be an issue, but we need to be able to point users at some doc if it is.

xvello: At least three indeed, but that's a risk. I added a comment in the test, but I'll copy it in kubeutil too

xvello: BTW for now if there's a dash and two numbers the old matching pattern will trigger too

andor44:
Based on the comment I can see you don't want to change this too much in the python agent, but the ownership can be easily resolved through an object's `ownerReferences`, eg:

```sh
:: ~
» kubectl get pod kube-dns-new-2650501873-f745g -o json | jq ".metadata.ownerReferences"
[
  {
    "apiVersion": "extensions/v1beta1",
    "blockOwnerDeletion": true,
    "controller": true,
    "kind": "ReplicaSet",
    "name": "kube-dns-new-2650501873",
    "uid": "ec9f6b8e-992d-11e7-abf1-b8ca3a685d0c"
  }
]
:: ~
» kubectl get pod kube-dns-new-2650501873-f745g -o json | jq ".metadata.ownerReferences[0].name" | xargs kubectl get rs -o json | jq ".metadata.ownerReferences[0].name"
"kube-dns-new"
:: ~
» kubectl get pod dd-agent-7ftnv -o json | jq ".metadata.ownerReferences"
[
  {
    "apiVersion": "extensions/v1beta1",
    "blockOwnerDeletion": true,
    "controller": true,
    "kind": "DaemonSet",
    "name": "dd-agent",
    "uid": "04eff791-9865-11e7-abf1-b8ca3a685d0c"
  }
]
```

xvello:
@hkaj updated tests pass, the failing one is the dsd one.
```
test_deployment_illegal_name (tests.core.test_kubeutil.TestKubeUtilDeploymentTag) ... ok
test_deployment_name_k8s_1_8 (tests.core.test_kubeutil.TestKubeUtilDeploymentTag) ... ok
test_deployment_name_nominal (tests.core.test_kubeutil.TestKubeUtilDeploymentTag) ... ok
```

PR #3609, opened 2017-12-08
Add kube container name in service discovery

JulienBalestra:
### What does this PR do?

Extract the tag `kube_container_name` in `sd_docker_backend` and `kubeutil`.

Also improves the coverage of these two packages.

### DIFF
--- a/utils/service_discovery/sd_docker_backend.py
+++ b/utils/service_discovery/sd_docker_backend.py
@@ -314,6 +314,11 @@ class SDDockerBackend(AbstractSDBackend):
             namespace = pod_metadata.get('namespace')
             tags.append('kube_namespace:%s' % namespace)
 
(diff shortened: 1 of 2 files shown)


utils/service_discovery/sd_docker_backend.py
```diff
@@ -314,6 +314,10 @@ def get_tags(self, state, c_id):
             namespace = pod_metadata.get('namespace')
             tags.append('kube_namespace:%s' % namespace)
 
+            # get kubernetes container name
+            kube_container_name = state.get_kube_container_name(c_id)
```
xvello: Although probably not plausible, `get_kube_container_name` could return `None` as we get() in a dict. You should add a `if kube_container_name` before appening the tag

PR #3611, opened 2017-12-11
[aggregator] Ignore all dist metrics (metric type starting with 'd')

olivielpeau:
### What does this PR do?

Ignore all dist metrics (i.e. metric types starting with 'd')

### Motivation

Avoid dropping packets, just ignore these metrics.

### Testing

unit test + no noticeable impact on performance (used benchmark in `benchmark_aggregator`)


olivielpeau:
cc @jeerim

PR #3614, opened 2017-12-18
kubeutil: get nodeName and nodeIP even if the agent isn't inside a Pod

JulienBalestra:
### What does this PR do?

Improve the way we are setting the nodeIP / nodeName in kubeutil.

If the agent doesn't run inside a kubernetes Pod, we cannot [get](https://github.com/DataDog/integrations-core/blob/bf7c8aa7dea0af631572f31e84c4fdabd80ccf15/kubernetes/check.py#L517) the hostname for the [events](https://github.com/DataDog/integrations-core/blob/bf7c8aa7dea0af631572f31e84c4fdabd80ccf15/kubernetes/check.py#L554).

All events will have **None** as `node_name` value.

All Pods inside the PodList are running on the same node/kubelet so in theory we could pick any of them and get the `spec.nodeName` for the `node_name` / hostname.

Sadly the static Pods don't have the associated `status.hostIP` so we need to manage that.

### DIFF
--- a/utils/kubernetes/kubeutil.py
+++ b/utils/kubernetes/kubeutil.py
@@ -543,18 +543,6 @@ class KubeUtil:
             log.warning("Unable to retrieve pod list %s. Not fetching host data", str(e))
             return
 
-        for pod in pod_items:
-            metadata = pod.get("metadata", {})
-            name = metadata.get("name")
-            if name == self.pod_name:
-                status = pod.get('status', {})
-                spec = pod.get('spec', {})
-                # if not found, use an empty string - we use None as "not initialized"
-                self._node_ip = status.get('hostIP', '')
-                self._node_name = spec.get('nodeName', '')
-                return
-
-        # At this point the agent isn't inside the podList.
         # Take the first Pod with a status:
         # all running pods have the adapted '.spec.nodeName'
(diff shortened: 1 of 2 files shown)


utils/kubernetes/kubeutil.py
```diff
@@ -552,7 +552,24 @@ def _fetch_host_data(self):
                 # if not found, use an empty string - we use None as "not initialized"
                 self._node_ip = status.get('hostIP', '')
                 self._node_name = spec.get('nodeName', '')
-                break
+                return
+
+        # At this point the agent isn't inside the podList.
+        # Take the first Pod with a status:
+        # all running pods have the adapted '.spec.nodeName'
+        # static pods doesn't have the '.status.hostIP'
```
hkaj: Do we need to keep the previous logic then? It looks like your block could be a replacement instead of an addition.

JulienBalestra: @hkaj I removed the old block and keep the associated test to confirm this change.

JulienBalestra:
Beware if there is no running Pod on the agent's host, the agent will not be able to resolve the `nodeName` because of an empty PodList.
This will be taken care of in the [agent6](https://github.com/DataDog/datadog-agent).

PR #3617, opened 2017-12-22
Fix prometheus check summary and histogram, replace the text parsing

JulienBalestra:
### What does this PR do?

This PR fix the way we are extracting the count, sum on histogram and summary on `text/plain` payload.
They were overrides by the last metric in the case of the histogram / summary have multiples metrics (labels).

It also replace the text parsing by an official prometheus library.

### Additional Notes

I set inline comment in the PR to ease the review.

We need to wait [dd-agent-omnibus](https://github.com/DataDog/dd-agent-omnibus/pull/209) and [omnibus-software](https://github.com/DataDog/omnibus-software/pull/172) are merged before merge this PR.


checks/prometheus_check.py
```diff
@@ -284,7 +281,7 @@ def process_metric(self, message, send_histograms_buckets=True, custom_tags=None
         except AttributeError as err:
             self.log.debug("Unable to handle metric: {} - error: {}".format(message.name, err))
 
-    def poll(self, endpoint, pFormat=PrometheusFormat.PROTOBUF, headers={}):
+    def poll(self, endpoint, pFormat=PrometheusFormat.PROTOBUF, headers=None):
```
JulienBalestra: We shouldn't pass a mutable object in a default parameter.

tests/core/test_prometheus.py
```diff
@@ -129,17 +131,17 @@ def test_parse_metric_family_text(self):
...
-            {'ct':1359194,'sum':199427281.0, 'lbl': {'system':'recursive'},
-                'buckets':{0.0: 0, 512.0:520924, 1024.0:520924, 1500.0:520924,
-                    2048.0:520924, float('+Inf'):520924}},
-            {'ct':1359194,'sum':199427281.0, 'lbl': {'system':'reverse'},
-                'buckets':{0.0: 0, 512.0:67648, 1024.0:67648, 1500.0:67648,
-                    2048.0:67648, float('+Inf'):67648}},
+            {'ct': 1359194, 'sum': 199427281.0, 'lbl': {'system': 'auth'},
+             'buckets': {0.0: 0, 512.0: 1359194, 1024.0: 1359194,
+                         1500.0: 1359194, 2048.0: 1359194, float('+Inf'): 1359194}},
+            {'ct': 520924, 'sum': 41527128.0, 'lbl': {'system': 'recursive'},
```
JulienBalestra: the `ct` and `sum` were incorrect regarding the fixtures.

tests/core/test_prometheus.py
```diff
@@ -129,17 +131,17 @@ def test_parse_metric_family_text(self):
...
+             'buckets': {0.0: 0, 512.0: 520924, 1024.0: 520924, 1500.0: 520924,
+                         2048.0: 520924, float('+Inf'): 520924}},
+            {'ct': 67648, 'sum': 6075182.0, 'lbl': {'system': 'reverse'},
```
JulienBalestra: ditto

checks/prometheus_check.py
```diff
@@ -120,25 +124,54 @@ def parse_metric_family(self, buf, content_type):
...
+    def get_metric_value_by_labels(messages, _metric, _m, metric_suffix):
+        """
+        :param messages: dictionary as metric_name: {labels: {}, value: 10}
+        :param _metric: dictionary as {labels: {le: '0.001', 'custom': 'value'}}
+        :param _m: str as metric name
+        :param metric_suffix: str must be in (count or sum)
+        :return: value of the metric_name matched by the labels
+        """
+        metric_name = '{}_{}'.format(_m, metric_suffix)
+        unwanted_labels = ["le", "quantile"]  # are specifics keys for prometheus itself
```
JulienBalestra: I can move `unwanted_labels` outside this scope, it's here only for clarity

JulienBalestra:
Rebased because of #3619

checks/prometheus_check.py
```diff
@@ -120,25 +124,54 @@ def parse_metric_family(self, buf, content_type):
...
-            for line in buf.splitlines():
-                self._extract_metrics_from_string(line, messages, obj_map, obj_help)
-
-            # Add type overrides:
-            for m_name, m_type in self.type_overrides.iteritems():
-                if m_type in self.METRIC_TYPES:
-                    obj_map[m_name] = m_type
-                else:
-                    self.log.debug("type override %s for %s is not a valid type name" % (m_type,m_name))
+            for metric in text_string_to_metric_families(buf):
```
xvello: Instead of loading the whole response body in memory, we should stream the body, use [iter_lines](http://docs.python-requests.org/en/master/api/#requests.Response.iter_lines) as a generator to be fed into [text_fd_to_metric_families](https://github.com/prometheus/client_python/blob/master/prometheus_client/parser.py#L146)

checks/prometheus_check.py
xvello: indeed, that's an unwanted allocation in a loop, let's move it as package constant

JulienBalestra:
I run some quick benchmarks against each parsing (protobuf and text).
```python
for _ in xrange(100000000):
    self.check.parse_metric_family(mock)
```
The results are quite the same (around 29s each):
**proto**: 29.7692170143s 
**text**: 29.1101601124s

checks/prometheus_check.py
```diff
@@ -83,11 +94,7 @@ def prometheus_metric_name(self, message, **kwargs):
...
         pass
 
-    class UnknownFormatError(Exception):
-        def __init__(self, arg):
-            self.args = arg
-
-    def parse_metric_family(self, buf, content_type):
+    def parse_metric_family(self, response):
         """
         Gets the output data from a prometheus endpoint response along with its
```
xvello: input parameters changed

checks/prometheus_check.py
```diff
@@ -118,26 +126,56 @@ def parse_metric_family(self, buf, content_type):
...
-                    obj_map[m_name] = m_type
-                else:
-                    self.log.debug("type override %s for %s is not a valid type name" % (m_type,m_name))
+        elif 'text/plain' in response.headers['Content-Type']:
+            messages = defaultdict(list)  # map with the name of the element (before the labels)
+            # and the list of occurrences with labels and values
+
+            obj_map = {}  # map of the types of each metrics
+            obj_help = {}  # help for the metrics
+            for metric in text_fd_to_metric_families(response.iter_lines(chunk_size=1024*10)):
```
xvello: 10kB seems OK for the chuck size, could you move it to a module constant + add comment so we don't have to do to the `requests` apidoc to know the unit and compute the value?

checks/prometheus_check.py
```diff
@@ -29,7 +33,14 @@ class PrometheusFormat:
     PROTOBUF = "PROTOBUF"
     TEXT = "TEXT"
 
+
+class UnknownFormatError(TypeError):
+    pass
+
+
 class PrometheusCheck(AgentCheck):
+    unwanted_labels = ["le", "quantile"]  # are specifics keys for prometheus itself
```
xvello: most of agent5 set constants as module constants, see https://github.com/DataDog/dd-agent/blob/master/utils/dockerutil.py#L44
Could you please adjust?

PR #3623, opened 2018-01-02
Generic label joins + node to hostname for the Prometheus Check

mfpierre:
### What does this PR do?

This PR adds 2 generic mechanisms to the prometheus check:
- A `label_to_hostname` setting that allows a specified label to override hostname if present
- A `label_joins` setting that allows to join 1:1 labels 🏷 from one metric to another

### Motivation

Sometimes some metrics come from a different node and the node is usually specified in an associated label, hence the utility of the `label_to_hostname` setting.

For the `label_joins`: in prometheus it's common to have some placeholder metric holding all the labels because it's possible to do some label joins in the prometheus query language, that's why we need a mechanism to do joins some labels during processing, so the metric can be sent with all labels/tags wanted.

### DIFF
--- a/checks/prometheus_check.py
+++ b/checks/prometheus_check.py
@@ -357,7 +357,6 @@ class PrometheusCheck(AgentCheck):
                         # Set this label value as active
                         if label.name not in self._active_label_mapping:
                             self._active_label_mapping[label.name] = {}
-
                         self._active_label_mapping[label.name][label.value] = True
                         # If mapping found add corresponding labels
                         try:
@@ -425,15 +424,7 @@ class PrometheusCheck(AgentCheck):
         """
         if message.type < len(self.METRIC_TYPES):
             for metric in message.metric:
-                # if hostname is not specified, look at label_to_hostname setting
-                if hostname is None and self.label_to_hostname is not None:
-                    custom_hostname = hostname
-                    for label in metric.label:
-                        if label.name == self.label_to_hostname:
-                            custom_hostname = label.value
-                            break
-                else:
-                    custom_hostname = hostname
+                custom_hostname = self._get_hostname(hostname, metric)
                 if message.type == 4:
                     self._submit_gauges_from_histogram(metric_name, metric, send_histograms_buckets, custom_tags, custom_hostname)
                 elif message.type == 2:
@@ -444,6 +435,17 @@ class PrometheusCheck(AgentCheck):
 
         else:
             self.log.error("Metric type {} unsupported for metric {}.".format(message.type, message.name))
+
+    def _get_hostname(self, hostname, metric):
+        """
+        If hostname is None, look at label_to_hostname setting
+        """
+        if hostname is None and self.label_to_hostname is not None:
+            for label in metric.label:
+                if label.name == self.label_to_hostname:
+                    return label.value
+
+        return hostname
 
     def _submit_gauge(self, metric_name, val, metric, custom_tags=None, hostname=None):
         """
--- a/checks/prometheus_check.py
+++ b/checks/prometheus_check.py
@@ -328,45 +328,46 @@ class PrometheusCheck(AgentCheck):
 
         `send_histograms_buckets` is used to specify if yes or no you want to send the buckets as tagged values when dealing with histograms.
         """
+
+        # If targeted metric, store labels
+        if message.name in self.label_joins:
+            matching_label = self.label_joins[message.name]['label_to_match']
+            for metric in message.metric:
+                labels_list = []
+                matching_value = None
+                for label in metric.label:
+                    if label.name == matching_label:
+                        matching_value = label.value
+                    elif label.name in self.label_joins[message.name]['labels_to_get']:
+                        labels_list.append((label.name, label.value))
+                try:
+                    self._label_mapping[matching_label][matching_value] = labels_list
+                except KeyError:
+                    if matching_value is not None:
+                        self._label_mapping[matching_label] = {matching_value: labels_list}
+
+        if message.name in self.ignore_metrics:
+            return  # Ignore the metric
+
+        # Filter metric to see if we can enrich with joined labels
+        if self.label_joins:
+            for metric in message.metric:
+                for label in metric.label:
+                    if label.name in self._watched_labels:
+                        # Set this label value as active
+                        if label.name not in self._active_label_mapping:
+                            self._active_label_mapping[label.name] = {}
+
+                        self._active_label_mapping[label.name][label.value] = True
+                        # If mapping found add corresponding labels
+                        try:
+                            for label_tuple in self._label_mapping[label.name][label.value]:
+                                extra_label = metric.label.add()
+                                extra_label.name, extra_label.value = label_tuple
+                        except KeyError:
+                            pass
+
         try:
-            # If targeted metric, store labels
-            if message.name in self.label_joins:
-                matching_label = self.label_joins[message.name]['label_to_match']
-                for metric in message.metric:
-                    labels_list = []
-                    matching_value = None
-                    for label in metric.label:
-                        if label.name == matching_label:
-                            matching_value = label.value
-                        elif label.name in self.label_joins[message.name]['labels_to_get']:
-                            labels_list.append((label.name, label.value))
-                    try:
-                        self._label_mapping[matching_label][matching_value] = labels_list
-                    except KeyError:
-                        if matching_value is not None:
-                            self._label_mapping[matching_label] = {matching_value: labels_list}
-
-            if message.name in self.ignore_metrics:
-                return  # Ignore the metric
-
-            # Filter metric to see if we can enrich with joined labels
-            if self.label_joins:
-                for metric in message.metric:
-                    for label in metric.label:
-                        if label.name in self._watched_labels:
-                            # Set this label value as active
-                            if label.name not in self._active_label_mapping:
-                                self._active_label_mapping[label.name] = {}
-
-                            self._active_label_mapping[label.name][label.value] = True
-                            # If mapping found add corresponding labels
-                            try:
-                                for label_tuple in self._label_mapping[label.name][label.value]:
-                                    extra_label = metric.label.add()
-                                    extra_label.name, extra_label.value = label_tuple
-                            except KeyError:
-                                pass
-
             if not self._dry_run:
                 try:
                     self._submit(self.metrics_mapper[message.name], message, send_histograms_buckets, custom_tags)
@@ -426,6 +427,7 @@ class PrometheusCheck(AgentCheck):
             for metric in message.metric:
                 # if hostname is not specified, look at label_to_hostname setting
                 if hostname is None and self.label_to_hostname is not None:
+                    custom_hostname = hostname
                     for label in metric.label:
                         if label.name == self.label_to_hostname:
                             custom_hostname = label.value
--- a/checks/prometheus_check.py
+++ b/checks/prometheus_check.py
@@ -77,7 +77,7 @@ class PrometheusCheck(AgentCheck):
         # }
         self._label_mapping = {}
 
-        # `_active_label_mapping` holds a dictionnary of label values found during the run
+        # `_active_label_mapping` holds a dictionary of label values found during the run
         # to cleanup the label_mapping of unused values, example:
         # self._active_label_mapping = {
         #     'pod': {
@@ -96,7 +96,7 @@ class PrometheusCheck(AgentCheck):
         # skipped without a 'Unable to handle metric' debug line in the logs
         self.ignore_metrics = []
 
-        # If the `labels_mapper` dictionnary is provided, the metrics labels names
+        # If the `labels_mapper` dictionary is provided, the metrics labels names
         # in the `labels_mapper` will use the corresponding value as tag name
         # when sending the gauges.
         self.labels_mapper = {}
@@ -105,7 +105,7 @@ class PrometheusCheck(AgentCheck):
         # will just not be added as tags when submitting the metric.
         self.exclude_labels = []
 
-        # `type_overrides` is a dictionnary where the keys are prometheus metric names
+        # `type_overrides` is a dictionary where the keys are prometheus metric names
         # and the values are a metric type (name as string) to use instead of the one
         # listed in the payload. It can be used to force a type on untyped metrics.
         # Note: it is empty in the mother class but will need to be
@@ -295,11 +295,11 @@ class PrometheusCheck(AgentCheck):
         response = self.poll(endpoint)
         try:
             # no dry run if no label joins
-            if len(self.label_joins) == 0:
+            if not self.label_joins:
                 self._dry_run = False
-            elif len(self._watched_labels) == 0:
+            elif not self._watched_labels:
                 # build the _watched_labels set
-                for metric, val in self.label_joins.items():
+                for metric, val in self.label_joins.iteritems():
                     self._watched_labels.add(val['label_to_match'])
 
             tags = []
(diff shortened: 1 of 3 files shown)


checks/prometheus_check.py
```diff
@@ -63,6 +63,39 @@ def __init__(self, name, init_config, agentConfig, instances=None):
...
+        # `_label_mapping` holds the additionals label info to add for a specific
+        # label value, example:
+        # self._label_mapping = {
+        #     'pod': {
+        #         'dd-agent-9s1l1': [("node","yolo"),("host_ip","yey")]
+        #     }
+        # }
+        self._label_mapping = {}
+
+        # `_active_label_mapping` holds a dictionnary of label values found during the run
```
JulienBalestra: s/dictionnary/dictionary

checks/prometheus_check.py
```diff
@@ -266,11 +303,28 @@ def process(self, endpoint, send_histograms_buckets=True, instance=None):
         """
         response = self.poll(endpoint)
         try:
+            # no dry run if no label joins
+            if len(self.label_joins) == 0:
+                self._dry_run = False
+            elif len(self._watched_labels) == 0:
+                # build the _watched_labels set
+                for metric, val in self.label_joins.items():
```
JulienBalestra: s/items/iteritems

checks/prometheus_check.py
```diff
@@ -284,12 +338,45 @@ def process_metric(self, message, send_histograms_buckets=True, custom_tags=None
...
+                    matching_value = None
+                    for label in metric.label:
+                        if label.name == matching_label:
+                            matching_value = label.value
+                        elif label.name in self.label_joins[message.name]['labels_to_get']:
+                            labels_list.append((label.name, label.value))
+                    if matching_label in self._label_mapping:
+                        self._label_mapping[matching_label][matching_value] = labels_list
+                    elif matching_value is not None:
+                        self._label_mapping[matching_label] = {matching_value : labels_list}
```
JulienBalestra: PEP8: "matching_value :" -> "matching_value:"

checks/prometheus_check.py
```diff
@@ -266,11 +303,28 @@ def process(self, endpoint, send_histograms_buckets=True, instance=None):
         """
         response = self.poll(endpoint)
         try:
+            # no dry run if no label joins
+            if len(self.label_joins) == 0:
```
JulienBalestra: could be just:
```python
if not self.label_joins:
```

checks/prometheus_check.py
```diff
@@ -266,11 +303,28 @@ def process(self, endpoint, send_histograms_buckets=True, instance=None):
...
+                self._dry_run = False
+            elif len(self._watched_labels) == 0:
```
JulienBalestra: ditto:
```python
if not self._watched_labels
```

checks/prometheus_check.py
```diff
@@ -284,12 +338,45 @@ def process_metric(self, message, send_histograms_buckets=True, custom_tags=None
...
+
             if message.name in self.ignore_metrics:
                 return  # Ignore the metric
-            if message.name in self.metrics_mapper:
-                self._submit(self.metrics_mapper[message.name], message, send_histograms_buckets, custom_tags)
-            else:
-                getattr(self, message.name)(message, **kwargs)
+
+            # Filter metric to see if we can enrich with joined labels
+            if len(self.label_joins) > 0:
```
JulienBalestra: can be replaced by:

```python
if self.label_join:
```

checks/prometheus_check.py
```diff
@@ -284,12 +338,45 @@ def process_metric(self, message, send_histograms_buckets=True, custom_tags=None
...
+                        if label.name in self._watched_labels:
+                            # Set this label value as active
+                            if label.name not in self._active_label_mapping:
+                                self._active_label_mapping[label.name] = {}
+                            self._active_label_mapping[label.name][label.value] = True
+                            # If mapping found add corresponding labels
+                            if label.name in self._label_mapping and label.value in self._label_mapping[label.name]:
+                                for label_tuple in self._label_mapping[label.name][label.value]:
+                                    extra_label = metric.label.add()
+                                    extra_label.name, extra_label.value = label_tuple[0], label_tuple[1]
```
JulienBalestra: This block could be simplified by dropping the condition:

```python
                            try:
                                for label_tuple in self._label_mapping[label.name][label.value]:
                                    extra_label = metric.label.add()
                                    extra_label.name, extra_label.value = label_tuple[0], label_tuple[1]
                            except KeyError:
                                pass
```
What do you think ? 

> it's easier to ask for forgiveness than permission

checks/prometheus_check.py
```diff
@@ -284,12 +338,45 @@ def process_metric(self, message, send_histograms_buckets=True, custom_tags=None
...
+
+            # Filter metric to see if we can enrich with joined labels
+            if len(self.label_joins) > 0:
+                for metric in message.metric:
+                    for label in metric.label:
+                        if label.name in self._watched_labels:
+                            # Set this label value as active
+                            if label.name not in self._active_label_mapping:
+                                self._active_label_mapping[label.name] = {}
+                            self._active_label_mapping[label.name][label.value] = True
```
JulienBalestra: By using a collection defaultdict(dict) you can skip the if block before setting the bool.

checks/prometheus_check.py
JulienBalestra: Can also be [EAFP](https://docs.python.org/3.6/glossary.html#term-eafp)
```python
                    try:
                        self._label_mapping[matching_label][matching_value] = labels_list
                    except KeyError:
                        if matching_value is not None:
                            self._label_mapping[matching_label] = {matching_value: labels_list
```

checks/prometheus_check.py
```diff
@@ -284,12 +329,45 @@ def process_metric(self, message, send_histograms_buckets=True, custom_tags=None
...
+                            if label.name in self._label_mapping and label.value in self._label_mapping[label.name]:
+                                for label_tuple in self._label_mapping[label.name][label.value]:
+                                    extra_label = metric.label.add()
+                                    extra_label.name, extra_label.value = label_tuple[0], label_tuple[1]
+
+            if not self._dry_run:
+                if message.name in self.metrics_mapper:
+                    self._submit(self.metrics_mapper[message.name], message, send_histograms_buckets, custom_tags)
+                else:
+                    getattr(self, message.name)(message, **kwargs)
```
JulienBalestra: EAFP 

```python
                try:
                    self._submit(self.metrics_mapper[message.name], message, send_histograms_buckets, custom_tags)
                except KeyError:
                    getattr(self, message.name)(message, **kwargs)
```

checks/prometheus_check.py
```diff
@@ -342,18 +420,25 @@ def _submit(self, metric_name, message, send_histograms_buckets=True, custom_tag
         """
         if message.type < len(self.METRIC_TYPES):
             for metric in message.metric:
+                # if hostname is not specified, look at label_to_hostname setting
+                if hostname == None and self.label_to_hostname != None:
```
JulienBalestra: can be changed by
```python
if hostname is None and self.label_to_hostname is not None:
```

tests/core/test_prometheus.py
```diff
@@ -268,7 +270,37 @@ def test_submit_gauge_with_labels(self):
         _l2.value = 'my_2nd_label_value'
         self.check._submit(self.check.metrics_mapper[self.ref_gauge.name], self.ref_gauge)
         self.check.gauge.assert_called_with('prometheus.process.vm.bytes', 39211008.0,
-                                            ['my_1st_label:my_1st_label_value', 'my_2nd_label:my_2nd_label_value'])
+                                            ['my_1st_label:my_1st_label_value', 'my_2nd_label:my_2nd_label_value'],
+                                            hostname=None)
+
+    def test_submit_gauge_with_labels_and_hostname_override(self):
+        ''' submitting metrics that contain labels should result in tags on the gauge call '''
```
JulienBalestra: We should use triple double quotes for docstrings [PEP 257](https://www.python.org/dev/peps/pep-0257/)

> For consistency, always use """triple double quotes""" around docstrings.

checks/prometheus_check.py
```diff
@@ -284,12 +329,49 @@ def process_metric(self, message, send_histograms_buckets=True, custom_tags=None
...
+                            # Set this label value as active
+                            if label.name not in self._active_label_mapping:
+                                self._active_label_mapping[label.name] = {}
+
+                            self._active_label_mapping[label.name][label.value] = True
+                            # If mapping found add corresponding labels
+                            try:
+                                for label_tuple in self._label_mapping[label.name][label.value]:
+                                    extra_label = metric.label.add()
+                                    extra_label.name, extra_label.value = label_tuple[0], label_tuple[1]
```
JulienBalestra: Can be unpacked like this:
```python
extra_label.name, extra_label.value = label_tuple
```

checks/prometheus_check.py
```diff
@@ -284,12 +329,49 @@ def process_metric(self, message, send_histograms_buckets=True, custom_tags=None
         `send_histograms_buckets` is used to specify if yes or no you want to send the buckets as tagged values when dealing with histograms.
         """
         try:
+            # If targeted metric, store labels
+            if message.name in self.label_joins:
```
JulienBalestra: We can also here drop the if statement with a KeyError:
```python
            try:
                matching_label = self.label_joins[message.name]['label_to_match']
                for metric in message.metric:
                    labels_list = []
                    matching_value = None
                    for label in metric.label:
                        if label.name == matching_label:
                            matching_value = label.value
                        elif label.name in self.label_joins[message.name]['labels_to_get']:
                            labels_list.append((label.name, label.value))
                    try:
                        self._label_mapping[matching_label][matching_value] = labels_list
                    except KeyError:
                        if matching_value is not None:
                            self._label_mapping[matching_label] = {matching_value: labels_list}
            except KeyError:
                pass
```

checks/prometheus_check.py
```diff
@@ -284,12 +329,49 @@ def process_metric(self, message, send_histograms_buckets=True, custom_tags=None
...
+                            matching_value = label.value
+                        elif label.name in self.label_joins[message.name]['labels_to_get']:
+                            labels_list.append((label.name, label.value))
+                    try:
+                        self._label_mapping[matching_label][matching_value] = labels_list
+                    except KeyError:
+                        if matching_value is not None:
+                            self._label_mapping[matching_label] = {matching_value: labels_list}
+
             if message.name in self.ignore_metrics:
```
JulienBalestra: Do you think it's useful to move this condition at the top ?

checks/prometheus_check.py
```diff
@@ -342,18 +424,26 @@ def _submit(self, metric_name, message, send_histograms_buckets=True, custom_tag
...
         if message.type < len(self.METRIC_TYPES):
             for metric in message.metric:
+                # if hostname is not specified, look at label_to_hostname setting
+                if hostname is None and self.label_to_hostname is not None:
+                    for label in metric.label:
+                        if label.name == self.label_to_hostname:
+                            custom_hostname = label.value
+                            break
+                else:
+                    custom_hostname = hostname
```
JulienBalestra: Are you not afraid of an `UnboundLocalError` in case of the label.name is never matched ?

checks/prometheus_check.py
```diff
@@ -342,18 +424,26 @@ def _submit(self, metric_name, message, send_histograms_buckets=True, custom_tag
         """
         if message.type < len(self.METRIC_TYPES):
             for metric in message.metric:
+                # if hostname is not specified, look at label_to_hostname setting
+                if hostname is None and self.label_to_hostname is not None:
+                    for label in metric.label:
+                        if label.name == self.label_to_hostname:
+                            custom_hostname = label.value
```
JulienBalestra: Are you not afraid of `UnboundLocalError` for **custom_hostname** if label.name == self.label_to_hostname is never True ?

checks/prometheus_check.py
```diff
@@ -284,12 +329,49 @@ def process_metric(self, message, send_histograms_buckets=True, custom_tags=None
...
+                                    extra_label.name, extra_label.value = label_tuple
+                            except KeyError:
+                                pass
+
+            if not self._dry_run:
+                try:
+                    self._submit(self.metrics_mapper[message.name], message, send_histograms_buckets, custom_tags)
+                except KeyError:
+                    getattr(self, message.name)(message, **kwargs)
         except AttributeError as err:
```
JulienBalestra: I think we can reduce the scope of the catch `AttributeError` block to avoid any side effects.

checks/prometheus_check.py
```diff
@@ -2,7 +2,6 @@
 # All rights reserved
 # Licensed under Simplified BSD License (see LICENSE)
 
-import re
```
JulienBalestra: thanks for that

checks/prometheus_check.py
mfpierre: I'll probably leave the `if` on this on I found the `try` block too big

checks/prometheus_check.py
mfpierre: I placed it intentionally bellow because we might want to target a metric for joining label on it but ignoring it for sending (it's the case of the `kube_pod_info` for example)

checks/prometheus_check.py
mfpierre: There's an `else` bellow setting the variable

mfpierre: modified the `else` bellow to `if not custom_hostname`

mfpierre: Added a default value above loop

checks/prometheus_check.py
mfpierre: done

checks/prometheus_check.py
```diff
@@ -342,18 +425,27 @@ def _submit(self, metric_name, message, send_histograms_buckets=True, custom_tag
         """
         if message.type < len(self.METRIC_TYPES):
             for metric in message.metric:
+                # if hostname is not specified, look at label_to_hostname setting
+                if hostname is None and self.label_to_hostname is not None:
```
JulienBalestra: To improve readability we could factorize this outside:

```python
diff --git a/checks/prometheus_check.py b/checks/prometheus_check.py
index 920731fd..47d9d425 100644
--- a/checks/prometheus_check.py
+++ b/checks/prometheus_check.py
@@ -425,15 +425,7 @@ class PrometheusCheck(AgentCheck):
         """
         if message.type < len(self.METRIC_TYPES):
             for metric in message.metric:
-                # if hostname is not specified, look at label_to_hostname setting
-                if hostname is None and self.label_to_hostname is not None:
-                    custom_hostname = hostname
-                    for label in metric.label:
-                        if label.name == self.label_to_hostname:
-                            custom_hostname = label.value
-                            break
-                else:
-                    custom_hostname = hostname
+                custom_hostname = self._get_hostname(hostname, metric)
                 if message.type == 4:
                     self._submit_gauges_from_histogram(metric_name, metric, send_histograms_buckets, custom_tags, custom_hostname)
                 elif message.type == 2:
@@ -445,6 +437,20 @@ class PrometheusCheck(AgentCheck):
         else:
             self.log.error("Metric type {} unsupported for metric {}.".format(message.type, message.name))
 
+    def _get_hostname(self, hostname, metric):
+        """
+        If hostname is None, look at label_to_hostname setting
+        :param hostname:
+        :param metric:
+        :return:
+        """
+        if hostname is None and self.label_to_hostname is not None:
+            for label in metric.label:
+                if label.name == self.label_to_hostname:
+                    return label.value
+
+        return hostname
+
     def _submit_gauge(self, metric_name, val, metric, custom_tags=None, hostname=None):
         """
         Submit a metric as a gauge, additional tags provided will be added to```

PR #3636, opened 2018-01-16
[iostat] show device mapper names on linux

ofek:
### What does this PR do?

This displays the `/dev/mapper` names of LVM devices on Linux e.g. `dm-0` might become `vg-backup`.

ref. https://linux.die.net/man/1/iostat

### Motivation

Customer request

### DIFF
--- a/checks/system/unix.py
+++ b/checks/system/unix.py
@@ -119,7 +119,7 @@ class IO(Check):
         io = {}
         try:
             if Platform.is_linux():
-                stdout, _, _ = get_subprocess_output(['iostat', '-d', '1', '2', '-x', '-k'], self.logger)
+                stdout, _, _ = get_subprocess_output(['iostat', '-d', '1', '2', '-x', '-k', '-N'], self.logger)
 
                 #                 Linux 2.6.32-343-ec2 (ip-10-35-95-10)   12/11/2012      _x86_64_        (2 CPU)
                 #


checks/system/unix.py
```diff
@@ -119,7 +119,7 @@ def check(self, agentConfig):
         io = {}
         try:
             if Platform.is_linux():
-                stdout, _, _ = get_subprocess_output(['iostat', '-d', '1', '2', '-x', '-k'], self.logger)
+                stdout, _, _ = get_subprocess_output(['iostat', '-d', '1', '2', '-x', '-k', '-N'], self.logger)
```
truthbk: Although this would clearly work as advertised according to the manpage, my concern is that applying this by default in the general case may have unwanted side-effects (like context churn in the backend or broken dashboards/monitors due to the change in name to the tags). I am wondering if it would perhaps be more prudent to add this option behind a feature flag such that one could enable this behavior only if desired.

ofek: We could indeed make this optional. Though, to play the devil's advocate for a moment, if a user has registered device mapper names, they likely desire this behavior as that would require manual symlinking.

How would you like to expose this option?

ParthKolekar:
This is a great feature. I guess this should be put behind a configurable option so that it does not break existing monitoring.

checks/system/unix.py
ParthKolekar: I am not sure how `agentConfig` works, but that seems like the right way of enabling this to be "more optional".

ParthKolekar: Bump @truthbk

olivielpeau:
Version `6.10` of the Agent will add a `device_label` tag on these metrics. This feature will not be backported to Agent v5. Closing this.

PR #3648, opened 2018-01-29
add jmx_custom_jars option to datadog.yaml

xvello:
This enables users to load a custom jar at jmxfetch start,
to allow using them reliably in AD templates

Only supporting a single string in agent5 to enable the use of
the DD_CONF_JMX_CUSTOM_JARS envvar binding, agent6 will support
a string array

`datadog/dev-dd-agent:backport_3648` allows to test it

### DIFF
--- a/datadog.conf.example
+++ b/datadog.conf.example
@@ -157,8 +157,9 @@ gce_updated_hostname: yes
 # Enable JMX checks for service discovery
 # sd_jmx_enable: no
 #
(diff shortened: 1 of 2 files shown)


xvello:
The DD_CONF_* envvars are automatically added to `datadog.conf`, so we don't need to  create a dedicated envvar binding in the entrypoint. The agent6 one will be called `DD_JMX_CUSTOM_JAR` though, as the convention changed.

PR #3649, opened 2018-01-30
jmxfetcher reporter host change to separete 

cw-sakamoto:
### What does this PR do?
- In datadog.conf.example there is an option to specify the forward destination of statsd, but in jmxfetch.py the reporter host is set to 'bind_host'

- To separate the host from which metrics are to be acquired and the report host, I would like to change the report host(statsd host) to 'statsd_forward_host' 

### Motivation
- As daemonset of kubernetes, running dd-agent, statsd server is one for each worker and it is necessary to be able to specify statsd server host


cw-sakamoto:
I'd like you to see this PR,  do I have something to do to next steps?

PR #3650, opened 2018-01-30
update for wheels

ofek:
### What does this PR do?

This updates the setup script to use wheels.

### Motivation

Outdated

### DIFF
--- a/packaging/datadog-agent/source/setup_agent.sh
+++ b/packaging/datadog-agent/source/setup_agent.sh
@@ -447,11 +447,16 @@ then
   INTEGRATIONS=$(ls $DD_HOME/integrations/)
   for INT in $INTEGRATIONS; do
     INT_DIR="$DD_HOME/integrations/$INT"
+    pushd $INT_DIR
+
     if [ -f "$INT_DIR/requirements.txt" ]; then
       "$DD_HOME/agent/utils/pip-allow-failures.sh" "$INT_DIR/requirements.txt"
     fi
-    if [ -f "$INT_DIR/check.py" ]; then
-      cp "$INT_DIR/check.py" "$DD_HOME/agent/checks.d/$INT.py"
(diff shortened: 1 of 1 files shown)


packaging/datadog-agent/source/setup_agent.sh
```diff
@@ -447,11 +447,19 @@ then
   INTEGRATIONS=$(ls $DD_HOME/integrations/)
   for INT in $INTEGRATIONS; do
     INT_DIR="$DD_HOME/integrations/$INT"
+    pushd $INT_DIR
```
olivielpeau: since you're using `pushd` here you can remove all the `$INT_DIR` prefixes to the paths you use below (unless I'm missing something...)

packaging/datadog-agent/source/setup_agent.sh
```diff
@@ -447,11 +447,19 @@ then
...
+
     if [ -f "$INT_DIR/requirements.txt" ]; then
       "$DD_HOME/agent/utils/pip-allow-failures.sh" "$INT_DIR/requirements.txt"
     fi
-    if [ -f "$INT_DIR/check.py" ]; then
-      cp "$INT_DIR/check.py" "$DD_HOME/agent/checks.d/$INT.py"
+    if [ -f "$INT_DIR/setup.py" ]; then
+      $PYTHON_CMD -m pip install .
```
olivielpeau: you should use `VENV_PIP_CMD` directly

ofek: Thanks!

packaging/datadog-agent/source/setup_agent.sh
ofek: It looks much cleaner now!

ofek:
Thanks! I think I'll merge the fix now and we'll improve it later.

PR #3651, opened 2018-01-31
Send region tag when running on GCE

arbll:
### What does this PR do?

Send region tag when the agent runs on Google compute engine.

### Motivation

Customer request.

### Additional Notes

- https://cloud.google.com/compute/docs/regions-zones/ to understand `zone` and `region`.

- `host_metadata['instance']` does not contain any region key.


arbll:
This will need a feature flag. Closing to avoid getting it merged.

Issue #3642, opened 2018-01-24
Kubernetes 1.9 breaks pod owner lookup


andor44:
The code [here](https://github.com/DataDog/dd-agent/blob/774a58204238b8029b075fa937eb43ed3ab77897/utils/kubernetes/kubeutil.py#L649-L663) to find the pod's creator no longer works because in https://github.com/kubernetes/kubernetes/pull/54445 they removed the pod annotation that the agent code is looking for. They advise on using the stabilized `metadata.ownerReferences` field instead. The above patch is in 1.9, therefore any pods created by k8s 1.9 and later will be missing owner tags on metrics (e.g. `kube_deployment`, `kube_daemonset`, etc.)

CharlyF:
Hey @andor44, it does seem like it indeed - Thank you very much for bringing this up to our attention!

I have added a task in our backlog so we can further investigate.
While we'll do our best to solve this as soon as possible, should you want to be updated on the progress made, feel free to reach out to our support team (support@datadoghq.com) with a reference to this issue so we can best communicate on our efforts.

Cheers,
.C

PR #3667, opened 2018-02-09
[kubernetes] fix logic to collect ownerreference for k8 1.9

antoinepouille:
### What does this PR do?

The previous code was looking the pod's creator in the annotations, while [kubernetes/kubernetes#54445](https://github.com/kubernetes/kubernetes/pull/54445) removed the pod annotation that the agent code is looking for, from version 1.9.
It now uses the stabilized metadata.ownerReferences field instead, and falls back on the annotations if they are missing.

### Motivation

This issue came up in https://github.com/DataDog/dd-agent/issues/3642 

### Testing Guidelines

Changed tests relatively to the new way to get ownerreferences

- Tested on GKE cluster with k8 1.9.2
- Tested on minikube with k8 1.7.3

### DIFF
--- a/tests/core/test_kubeutil.py
+++ b/tests/core/test_kubeutil.py
@@ -98,7 +98,8 @@ class TestKubeUtilCreatorTags(KubeTestCase):
 class TestKubeUtilCreatorTagsNoAnnotation(KubeTestCase):
     @classmethod
     def _fake_pod(cls, creator_kind, creator_name):
-        return {'ownerReferences': {'kind': creator_kind, 'name': creator_name}}
+        ownerReferences_entry = [{'kind': creator_kind, 'name': creator_name}]
+        return {'ownerReferences': ownerReferences_entry}
 
     def test_with_replicaset(self):
         self.assertEqual(['kube_replica_set:test-5432', 'kube_deployment:test'],
--- a/tests/core/test_kubeutil.py
+++ b/tests/core/test_kubeutil.py
@@ -93,6 +93,31 @@ class TestKubeUtilCreatorTags(KubeTestCase):
         self.assertEqual([], self.kube.get_pod_creator_tags({}))
 
 
+class TestKubeUtilCreatorTagsNoAnnotation(KubeTestCase):
+    @classmethod
+    def _fake_pod(cls, creator_kind, creator_name):
+        payload = '{"reference": {"kind":"%s", "name":"%s"}}' % (creator_kind, creator_name)
+        return {'ownerReferences': {'kind': creator_kind, 'name': creator_name}}
+
+    def test_with_replicaset(self):
+        self.assertEqual(['kube_replica_set:test-5432', 'kube_deployment:test'],
+                         self.kube.get_pod_creator_tags(self._fake_pod("ReplicaSet", "test-5432")))
+
+    def test_with_statefulset(self):
+        self.assertEqual(['kube_stateful_set:test-5432'],
+                         self.kube.get_pod_creator_tags(self._fake_pod("StatefulSet", "test-5432")))
+
+    def test_with_legacy_repcontroller(self):
+        self.assertEqual(['kube_daemon_set:test', 'kube_replication_controller:test'],
+                         self.kube.get_pod_creator_tags(self._fake_pod("DaemonSet", "test"), True))
+
+    def test_with_unknown(self):
+        self.assertEqual([], self.kube.get_pod_creator_tags(self._fake_pod("Unknown", "test")))
+
+    def test_invalid_input(self):
+        self.assertEqual([], self.kube.get_pod_creator_tags({}))
+
+
 class TestKubeGetNodeInfo(KubeTestCase):
     @staticmethod
     def mocked_requests_get_without_agent_but_ok(*args, **kwargs):
--- a/tests/core/test_kubeutil.py
+++ b/tests/core/test_kubeutil.py
@@ -68,6 +68,7 @@ class TestKubeUtilDeploymentTag(KubeTestCase):
         self.assertEqual('front-end', self.kube.get_deployment_for_replicaset('front-end-768dd754b7'))
 
 
+# Creator Tags tests for k8 version < 1.9: getting them from the annotation
 class TestKubeUtilCreatorTags(KubeTestCase):
     @classmethod
     def _fake_pod(cls, creator_kind, creator_name):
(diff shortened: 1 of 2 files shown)


utils/kubernetes/kubeutil.py
```diff
@@ -654,13 +654,21 @@ def get_pod_creator(self, pod_metadata):
...
-            creator_name = created_by.get('reference', {}).get('name')
+            ownerReferences_entry = pod_metadata['ownerReferences'][0]
+            creator_kind = ownerReferences_entry['kind']
+            creator_name = ownerReferences_entry['name']
             return (creator_kind, creator_name)
         except Exception:
-            log.debug('Could not parse creator for pod ' + pod_metadata.get('name', ''))
-            return (None, None)
+            try:
+                log.debug('Could not parse creator for pod ' + pod_metadata.get('name', '') +
```
JulienBalestra: We should use the common logger interpolation here like:
```python

                log.debug('Could not parse creator for pod %s through `OwnerReferences`, '
                          'falling back to annotation: %s', pod_metadata.get('name', ''), type(e))
```
This is because the logger do some [lazy evaluation](https://en.wikipedia.org/wiki/Lazy_evaluation), see stack overflow [thread](https://stackoverflow.com/questions/4148790/lazy-logger-message-string-evaluation).

I also added the exception type in the message `<type 'exceptions.KeyError'>`

utils/kubernetes/kubeutil.py
```diff
@@ -654,13 +654,21 @@ def get_pod_creator(self, pod_metadata):
...
+                          ' through `OwnerReferences`, falling back to annotation')
+                created_by = json.loads(pod_metadata['annotations']['kubernetes.io/created-by'])
+                creator_kind = created_by.get('reference', {}).get('kind')
+                creator_name = created_by.get('reference', {}).get('name')
+                return (creator_kind, creator_name)
```
JulienBalestra: Ditto, you can remove the parenthesis here.

utils/kubernetes/kubeutil.py
```diff
@@ -654,13 +654,21 @@ def get_pod_creator(self, pod_metadata):
...
+            except Exception:
```
JulienBalestra: I suggest to alias the exception to display its type in the log message:
```python
expect Exception as e:
```

utils/kubernetes/kubeutil.py
```diff
@@ -654,13 +654,21 @@ def get_pod_creator(self, pod_metadata):
...
+                log.debug('Could not parse creator for pod ' + pod_metadata.get('name', ''))
```
JulienBalestra: Ditto, the logger should be like:
```python
log.debug('Could not parse creator for pod "%s": %s', pod_metadata.get('name', ''), type(e))
```

utils/kubernetes/kubeutil.py
```diff
@@ -654,13 +654,21 @@ def get_pod_creator(self, pod_metadata):
...
+                return (None, None)
```
JulienBalestra: ditto, it's a tuple here: 
```python
return None, None
```

tests/core/test_kubeutil.py
```diff
@@ -68,6 +68,7 @@ def test_deployment_name_k8s_1_8(self):
         self.assertEqual('front-end', self.kube.get_deployment_for_replicaset('front-end-768dd754b7'))
 
 
+# Creator Tags tests for k8 version < 1.9: getting them from the annotation
```
JulienBalestra: The documentation should be conform to the [PEP257](https://www.python.org/dev/peps/pep-0257/)
Like:
```python
class MyClass(object):
    """
    My cool comment
    """
```

Then, I could use some tools to directly build the documentation, they'll call `MyClass.__doc__` to get `My cool comment`

tests/core/test_kubeutil.py
```diff
@@ -93,6 +94,32 @@ def test_invalid_input(self):
         self.assertEqual([], self.kube.get_pod_creator_tags({}))
 
 
+# Creator Tags tests for k8 version >= 1.9: getting them from the metadata 'ownerReferences'
```
JulienBalestra: Ditto for the comment.

tests/core/test_kubeutil.py
```diff
@@ -93,6 +94,32 @@ def test_invalid_input(self):
...
+class TestKubeUtilCreatorTagsNoAnnotation(KubeTestCase):
+    @classmethod
+    def _fake_pod(cls, creator_kind, creator_name):
+        ownerReferences_entry = [{'kind': creator_kind, 'name': creator_name}]
```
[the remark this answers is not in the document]
antoinepouille: 👍

tests/core/test_kubeutil.py
```diff
@@ -93,6 +94,32 @@ def test_invalid_input(self):
         self.assertEqual([], self.kube.get_pod_creator_tags({}))
 
 
+# Creator Tags tests for k8 version >= 1.9: getting them from the metadata 'ownerReferences'
+class TestKubeUtilCreatorTagsNoAnnotation(KubeTestCase):
+    @classmethod
```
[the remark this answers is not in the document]
antoinepouille: I copied it from the test case from just over it. I'm not sure it's absolutely necessary, but it doesn't seem to harm

utils/kubernetes/kubeutil.py
```diff
@@ -654,13 +654,21 @@ def get_pod_creator(self, pod_metadata):
         This allows for consitency across code path
         """
         try:
-            created_by = json.loads(pod_metadata['annotations']['kubernetes.io/created-by'])
-            creator_kind = created_by.get('reference', {}).get('kind')
-            creator_name = created_by.get('reference', {}).get('name')
+            ownerReferences_entry = pod_metadata['ownerReferences'][0]
+            creator_kind = ownerReferences_entry['kind']
+            creator_name = ownerReferences_entry['name']
             return (creator_kind, creator_name)
```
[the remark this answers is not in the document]
antoinepouille: 👍

utils/kubernetes/kubeutil.py
```diff
@@ -654,13 +654,21 @@ def get_pod_creator(self, pod_metadata):
...
         except Exception:
```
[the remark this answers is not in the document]
antoinepouille: Indeed, sounds better. `LookupError` then maybe?

utils/kubernetes/kubeutil.py
antoinepouille: 👍

tests/core/test_kubeutil.py
antoinepouille: 👍

PR #3676, opened 2018-02-14
[setup_agent.sh] install integrations as wheels

ofek:
### What does this PR do?

This better supports checks with multiple files.


packaging/datadog-agent/source/setup_agent.sh
```diff
@@ -455,7 +455,8 @@ then
       "$DD_HOME/agent/utils/pip-allow-failures.sh" "requirements.txt"
     fi
     if [ -f "setup.py" ]; then
-      $VENV_PIP_CMD install .
+      $PYTHON_CMD "setup.py" bdist_wheel
```
truthbk: we need to make sure `datadog_checks_base` is installed before any other wheel. It would meet this criteria but integrations in the glob that were to appear before it would fail to install.

PR #3685, opened 2018-02-22
Add case for swarm image name with @sha256

antoinepouille:
### What does this PR do?
Fixes the parsing of swarm images such as `datadog/docker-dd-agent:latest@sha256:769418c18c3e9e0b6ab2c18147c3599d6e27f40fb3dee56418bf897147ff84d0` when gathering tags in dockerutil. The issue here was with the `@sha256`

### Motivation

A user was missing his tags on his hosts.

### Testing Guidelines
Added a test case

### Additional Notes

Similar to the [maching agent 6 code](https://github.com/DataDog/datadog-agent/blob/c0d3dc9bdd3ce0dbc379c01b010487e927464095/pkg/util/docker/util_common.go#L90-L93)

### DIFF
--- a/utils/dockerutil.py
+++ b/utils/dockerutil.py
@@ -588,7 +588,7 @@ class DockerUtil:
         return image
 
     def image_name_resolver(self, image):
-        if image.startswith('sha256:') or '@sha256:' in image:
+        if image.startswith('sha256:'):
             # Some orchestrators setup containers with image checksum instead of image name
             try:
                 if image in self._image_sha_to_name_mapping:
@@ -613,6 +613,10 @@ class DockerUtil:
                         log.debug("Failed finding image name in RepoTag and RepoDigests: %s", e)
             except Exception as ex:
                 log.error("Exception getting docker image name: %s" % str(ex))
+        elif '@sha256:' in image: # Swarm case where image = name:tag@sha256SHA, keep only name:tag
+            pos = image.rfind('@sha256')
+            if pos > 0:
+                return image[0:pos]
         else:
             return image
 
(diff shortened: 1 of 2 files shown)


utils/dockerutil.py
```diff
@@ -588,7 +588,7 @@ def image_name_extractor(self, co, short=False):
         return image
 
     def image_name_resolver(self, image):
-        if image.startswith('sha256:') or '@sha256:' in image:
```
xvello: I'm afraid we're breaking the kubernetes `redis@sha256:474fb41b08bcebc933c6337a7db1dc7131380ee29b7a1b64a7ab71dad03ad718` case here, WDTY?

antoinepouille: Indeed, I wasn't aware of that particular case. I updated the code appropriately to fetch the tag in the _sha256_ when absent before the `@sha256`+

utils/dockerutil.py
```diff
@@ -588,8 +588,17 @@ def image_name_extractor(self, co, short=False):
...
+        pos_sha = image.rfind('@sha256')
+        if pos_sha > 0:
+            stripped_image = image[:pos_sha]
+            # a column indicates a tag only after the last slash (for cases as 'reg.local:5000/name:tag@sha256SHA')
+            pos_colon = stripped_image.rfind(':')
+            pos_slash = stripped_image.rfind('/')
+            if pos_colon > pos_slash and pos_colon > 0:
+                return stripped_image # keep only name:tag
+        # Handle the case in some orchestrators which setup containers with image checksum instead of image name
         if image.startswith('sha256:') or '@sha256:' in image:
```
xvello: Can this test be merged with  `if pos_sha > 0:`, to fail faster towards `else: return image`?
in that case, you should test `if pos_sha > -1:` to handle `startswith` (which is equivalent to pos==0)

Issue #3700, opened 2018-03-06
dogstreams does not pick up newly created log files when using wildcard


jonauman:
**Additional environment details (Operating System, Cloud provider, etc):**
OS: Linaro (Ubuntu)
Arch: armhf

**Steps to reproduce the issue:**
1. Add this to datadog.conf:
`dogstreams: /home/cbuser/.cortex/log/player.json.*:/opt/datadog-agent/agent/cortex_log_plays.py:cortex_parser`
2. restart the agent

**Describe the results you received:**
As new logs appear (logrotate) over time, they do not get picked up by dogstreams.

**Describe the results you expected:**
I would expect that all logs matching the wildcard would get processed, not only the logs matching at time of agent startup.

**Additional information you deem important (e.g. issue happens only occasionally):**
Datadog support has provided a fix for collector.py (358-259):

`
357        if self._dogstream is not None:
358            if (self.run_count % 10) == 0:
359                self._dogstream = Dogstreams.init(log, self.agentConfig)
`

I have tested this patch and it works. Can we get this added into the codebase?

Since we are on armhf architecture, we must install the agent from source and would like to avoid having to patch the file every time we upgrade the agent. 

Zendesk ticket 131296 was opened (and resolved) in reference to this issue.

nsushain:
Can we get his merged? facing the same issue

eddjlam:
is there any update on this? currently facing this issue as well.

PR #3701, opened 2018-03-07
Fixes issue #3700

jonauman:
### What does this PR do?

re-instantiates the dogstream every 10 runs (approx. every 2 minutes)

### Motivation

Fixes issue #3700

### DIFF
--- a/checks/collector.py
+++ b/checks/collector.py
@@ -355,6 +355,11 @@ class Collector(object):
         if self._ganglia is not None:
             payload['ganglia'] = self._ganglia.check(self.agentConfig)
         if self._dogstream is not None:
+            # every 10 run (~2min) we reload the list of files watched by
+            # dogstream
+            if (self.run_count % 10) == 0:
+                log.info("reloading list of files watched by Dogstreams")
+                self._dogstream = Dogstreams.init(log, self.agentConfig)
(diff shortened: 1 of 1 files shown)


checks/collector.py
```diff
@@ -355,6 +355,8 @@ def run(self, checksd=None, start_event=True, configs_reloaded=False):
         if self._ganglia is not None:
             payload['ganglia'] = self._ganglia.check(self.agentConfig)
         if self._dogstream is not None:
+            if (self.run_count % 10) == 0:
```
hush-hush: Could we add a comment explaining why we reload dogstream every 10 run ?

checks/collector.py
```diff
@@ -355,6 +355,8 @@ def run(self, checksd=None, start_event=True, configs_reloaded=False):
...
+                self._dogstream = Dogstreams.init(log, self.agentConfig)
```
hush-hush: Also could we add a debug log line saying we're reloading dogstream files ?

olivielpeau:
Unfortunately, this change causes a file descriptor leak because new `init`s re-open all the tailed files without closing the previous file descriptors. Revert: #3777, planned for `5.27.2`.

Happy to review a new PR that'd implement this while avoiding the leak!

PR #3703, opened 2018-03-12
[windows] change default datatype for pdh queries to double. Integer

derekwbrown:
loses fidelity, especially with values between 0 and 1.
### What does this PR do?
change default pdh counter precision to float from int.
Allows configuration to set specific counter types

### Motivation
customer request

### DIFF
--- a/checks/libs/win/pdhbasecheck.py
+++ b/checks/libs/win/pdhbasecheck.py
@@ -67,6 +67,15 @@ class PDHBaseCheck(AgentCheck):
                         self.log.error("Failed to make remote connection %s" % str(e))
                         return
 
+                ## counter_data_types allows the precision with which counters are queried
+                ## to be configured on a per-metric basis. In the metric instance, precision
+                ## should be specified as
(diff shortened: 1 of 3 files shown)


checks/libs/win/pdhbasecheck.py
```diff
@@ -8,12 +8,22 @@
 from utils.containers import hash_mutable
 # datadog
 try:
-    from checks.libs.win.winpdh import WinPDHCounter
+    from checks.libs.win.winpdh import WinPDHCounter, DATA_TYPE_INT, DATA_TYPE_DOUBLE
 except ImportError:
-    from checks.libs.win.winpdh_stub import WinPDHCounter
+    from checks.libs.win.winpdh_stub import WinPDHCounter,DATA_TYPE_INT, DATA_TYPE_DOUBLE
```
truthbk: I don't think `DATA_TYPE_INT`, `DATA_TYPE_DOUBLE` has been defined in the stub.

PR #3704, opened 2018-03-14
Avoid failing the configcheck command if the instances is empty

hush-hush:


utils/configcheck.py
```diff
@@ -11,8 +11,8 @@
 import simplejson as json
 
 # project
+from util import NoInstancesFound, check_yaml
 from config import (
-    check_yaml,
```
ofek: Nice catch!

PR #3710, opened 2018-03-17
Slim down the README and redirect to datadog-agent

masci:
### What does this PR do?

Makes more clear that the reference repo for the Agent development should be https://github.com/DataDog/datadog-agent

### Motivation

Avoid confusion, see https://github.com/DataDog/datadog-agent/issues/1405

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,8 @@
 
 # Important note
 
-This repository contains the source code for the Datadog Agent until major version 5.
+This repository contains the source code for the Datadog Agent up to and including
+major version 5.
 Although still supported, no major feature is planned for this release line and we
 encourage users and contributors to refer to the new Agent codebase, introduced
 with the release of version 6.0.0 and tracked in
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,83 +1,40 @@
 [![Build Status](https://travis-ci.org/DataDog/dd-agent.svg?branch=master)](https://travis-ci.org/DataDog/dd-agent)
 
-The Datadog Agent faithfully collects events and metrics and brings
-them to [Datadog](https://app.datadoghq.com) on your behalf so that
-you can do something useful with your monitoring and performance data.
+# Important note
 
-You're looking at the source code right now. We provide a number of
-pre-packaged binaries for your convenience for both [Agent 5](https://app.datadoghq.com/account/settings?agent_version=5#agent) and [Agent 6]((https://app.datadoghq.com/account/settings#agent)). 
+This repository contains the source code for the Datadog Agent until major version 5.
```
olivielpeau: `up to and including` sounds better to me than `until`, but don't know why exactly 🤔

README.md
```diff
@@ -1,83 +1,40 @@
...
-Exceptions are marked with LICENSE-xxx where xxx is the component name.
-If you do **not** agree with the licensing terms and wish to contribute code nonetheless,
-please email us at <info@datadoghq.com> before submitting your
-pull request.
-
-# [Integration SDK](https://github.com/DataDog/integrations-core)
-
-All checks have been moved to the [Integration SDK](https://github.com/DataDog/integrations-core). Please look there to submit related issues, PRs, or review the latest changes.
-
-## Setup your environment
```
olivielpeau: not sure this section is documented anywhere else. Should we keep it in the README here? (we could repeat that development/contributions are encouraged on the Agent 6 repo instead of here, but it's good to keep the A5 base instructions for reference somewhere IMO)

README.md
```diff
@@ -1,83 +1,40 @@
...
-separately. See
-[dd-trace-agent](https://github.com/DataDog/datadog-trace-agent) for
-instructions.
+Please refer to the [Change log](https://github.com/DataDog/dd-agent/blob/master/CHANGELOG.md)
+for more details about the changes introduced at each release.
 
-# [Change log](https://github.com/DataDog/dd-agent/blob/master/CHANGELOG.md)
+## Integrations
 
-# How to contribute code
```
olivielpeau: you've removed this section entirely, is that to indicate that we don't accept contributions anymore? (we should make that clear in that case)

README.md
masci: If this the only place we have these instructions they should be moved elsewhere, maybe the official docs.

README.md
masci: Yep that was the goal. I'll rephrase keeping the basic info without the emphasis

README.md
masci: because it's less ambiguous? Indeed it's not clear if `until` has to be considered `<=` or `<`

olivielpeau: yup exactly

README.md
jeremy-lq: Is there a reason we wouldn't accept a bug or security fix? May be worth emphasizing that all feature work should happen in Agent 6, and only PRs for large impact bug fixes and security issues should be submitted for Agent 5.

PR #3712, opened 2018-03-20
Remove API key if included in processes metadata payload

dixonscottr:
if you have not yet done so.*

### What does this PR do?

Removes API key if included as part of the `processes` field in the metadata payload when log level is set to `DEBUG`

### Motivation

Noticed an API key, as part of the processes field, could be sent if `DEBUG` mode is enabled

### Testing Guidelines

### Additional Notes

### DIFF
--- a/checks/collector.py
+++ b/checks/collector.py
@@ -529,6 +529,8 @@ class Collector(object):
             del data['metrics']
             del data['events']
             del data['service_checks']
+            if data.get('processes'):
+                data['processes']['apiKey'] = '*************************' + data['processes'].get('apiKey', '')[-5:]
             log.debug("Metadata payload: %s", json.dumps(data))
 
         # Persist the status of the collection run.


checks/collector.py
```diff
@@ -529,6 +529,8 @@ def run(self, checksd=None, start_event=True, configs_reloaded=False):
             del data['metrics']
             del data['events']
             del data['service_checks']
+            if data.get('processes'):
+                data['processes']['apiKey'] = '*************************' + data['processes'].get('apiKey', '')[-5:]
```
ofek: Minor nit: maybe do `'*' * 27` to make the scrubbed key the proper length (32)

ofek: Don't worry since the flare also doesn't output the proper length https://github.com/DataDog/dd-agent/blob/1d0dc34ce7dd3f114ddeed99540b655d97ee879b/utils/flare.py#L105

PR #3718, opened 2018-04-05
Set default value for AGENT_VERSION

hush-hush:
### What does this PR do?

Set a default version in the dev env for the agent. The real version will be updated by the package builder.


hush-hush:
We're going another way

PR #3720, opened 2018-04-10
Fix source install

arbll:
### What does this PR do?

The migration to wheels added hashes in the integrations `requirements.txt`s. The source install was not made to handle theses hashes. As a quick fix, this PR uses `requirements.in` instead.

This also allows wheel installations to fail.

### Motivation

Install from source was broken.


olivielpeau:
discussed IRL: we need to handle older versions of `integrations-core` too, so the script should use `requirements.in` only if it exists

PR #3730, opened 2018-05-05
Add message about check_rate

nmuesch:
if you have not yet done so.*

### What does this PR do?

If the check command is run without `check_rate` then print a message saying that is an available option.

### Motivation

If an integration only submits rates, it can be confusing to see 0 metrics when just running the `check` command.

### Testing Guidelines

### Additional Notes

### DIFF
--- a/agent.py
+++ b/agent.py
@@ -603,7 +603,7 @@ def main():
                         cs = Collector.run_single_check(check, verbose=True)
                         print CollectorStatus.render_check_status(cs)
                     else:
-                        print "Check has run only once, if some metrics are missing you can try again with check_rate appended at the end to see any other metrics if available."
+                        print "Check has run only once, if some metrics are missing you can run the command again with the 'check_rate' argument appended at the end to see any other metrics if available."
                     check.stop()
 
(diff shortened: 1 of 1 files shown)


agent.py
```diff
@@ -602,7 +602,8 @@ def parent_func():
                         time.sleep(1)
                         cs = Collector.run_single_check(check, verbose=True)
                         print CollectorStatus.render_check_status(cs)
-
+                    else:
+                        print "Check has run only once, if some metrics are missing you can try again with check_rate appended at the end to see any other metrics if available."
```
olivielpeau: nit: maybe change to `if some metrics are missing you can run the command again with the 'check_rate' argument appended at the end to see any other metrics if available.`

up to you though 👍

nmuesch: I really just copied and pasted from the line we used for Agent 6 to keep things in sync :)

olivielpeau: mostly I'd add that `check_rate` is an argument, so something like `try again with the 'check_rate' argument appended at the end ` would work for me :)

nmuesch: Ahh good point. I changed the wording a bit. Let me know what you think

PR #3735, opened 2018-05-10
[docker] Add host tag for docker swarm node role.

devonboyer:
if you have not yet done so.*

### What does this PR do?

This adds a host tag to docker swarm nodes containing the node role (manager or worker). 

### Testing Guidelines

screen shot 2018-05-10 at 10 14 54 am

screen shot 2018-05-10 at 10 15 43 am

### Additional Notes

To determine if the node is a manager/worker this checks the `ControlAvailable` flag in the response from `docker info` the same it is done by the [docker cli](https://github.com/docker/cli/blob/c889ea1bca95f78d10e9e8f7869d5cfd755850eb/cli/command/system/info.go#L259).

### DIFF
--- a/tests/core/test_dockerutil.py
+++ b/tests/core/test_dockerutil.py
@@ -185,6 +185,25 @@ class TestDockerUtil(unittest.TestCase):
         self.assertEqual({'docker_version': '1.13.1', 'docker_swarm': 'active'}, DockerUtil().get_host_metadata())
         mock_version.assert_called_once()
 
+    def test_docker_host_tags_ok(self):
+        du = DockerUtil()
+        mock_isswarm = mock.MagicMock(name='is_swarm', return_value=False)
+        du._client = mock.MagicMock()
+        du.is_swarm = mock_isswarm
+
+        self.assertEqual([], DockerUtil().get_host_tags())
+
+    def test_docker_host_tags_swarm_ok(self):
+        du = DockerUtil()
+        mock_info = mock.MagicMock(name='info', return_value={'Swarm': {'ControlAvailable' : True}})
+        mock_isswarm = mock.MagicMock(name='is_swarm', return_value=True)
+        du._client = mock.MagicMock()
+        du._client.info = mock_info
+        du.is_swarm = mock_isswarm
+
+        self.assertEqual(['docker_swarm_node_role:manager'], DockerUtil().get_host_tags())
+        mock_info.assert_called_once()
+
     def test_docker_are_tags_filtered(self):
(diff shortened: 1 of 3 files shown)


hkaj:
The core difference between host tags and host metadata is that host tags are part of the context of metrics coming from said host, whereas host metadata is not. So a change in host tags invalidates all metrics coming from the host and will require creating new time series for these metrics. This is why we avoid using host tags for things that can change frequently and/or all at once (or in a short time frame) across many hosts (like the docker version for example).

We also allow filtering by host tags in queries, but not by host metadata.

In this case it's fine to use tags as hosts don't flap between worker/manager.

devonboyer:
@hkaj 

I did some testing to determine what would happen with the tags if a worker node was promoted to a manager and/or manager demoted to a worker.

Since the host metadata (and host tags) is collected [every 4 hours by default](https://github.com/DataDog/dd-agent/blob/master/checks/collector.py#L176), it can take a long time for the tags to get updated for metrics (unless the agent is restarted).

If it makes sense to have the `docker_swarm_node_role` tag to be updated right away for metrics I think we would need to use the [tag extractor](https://github.com/DataDog/integrations-core/blob/master/docker_daemon/datadog_checks/docker_daemon/docker_daemon.py#L141) @CharlyF mentioned and also maybe include this information in host metadata so it's easy for people to check in the infrastructure list.

xvello:
Please note that tags from the Agent5 orchestrator package / Agent6 Tagger are only applied to docker / kubernetes metrics and AutoDiscovery checks (but are constant once the check template is resolved).
They would not be added to `system.cpu.usage` for example.

hkaj:
This is a good point.

However host tags allow us to factor this info in a single payload instead of attaching it to each metric (it's common for hosts to have 10's of tags x thousands of metrics).

And promoting a worker to manager shouldn't happen often in prod. Immutability dictates that one should rather create a new node built and configured for that purpose. I don't think it's too bad to impose the 4h delay in this context.

devonboyer:
@xvello Could you elaborate on what you mean by 

> but are constant once the check template is resolved

I am seeing metrics for `system.cpu.usage` in Datadog tagged with `docker_swarm_node_role` which is not a Docker/Kubernetes metric.

devonboyer:
@hkaj I'm not sure where we left off in the discussion for this PR. I am not sure I understand @xvello's last comment. 

It sounds like we agree that keeping this as a `host_tag` is fine.

xvello:
@devonboyer indeed, host tag will be better suited than container tag for this use case, let's go with that code

PR #3749, opened 2018-06-06
Sets apm enabled by default (if there's no indication in the config

derekwbrown:
file)

### What does this PR do?

Sets the default state for APM to be enabled if not set in the config file

### Motivation

Make behavior consistent between agent5 and agent6

Windows agent uses this value to determine whether to start the apm service on main service startup (the service will not be started automatically by the service control manager).


olivielpeau:
Great, thanks for the confirmation @palazzem! Let's merge this for 5.25.0 then

PR #3750, opened 2018-06-06
Remove utils that are available in datadog_checks

nmuesch:
if you have not yet done so.*

### What does this PR do?

Remove a lot of code that is already available in datadog_checks_base. 

### Motivation

Keep only one set of code to make things easier to update. 

### Testing Guidelines

### Additional Notes

The way the imports work on the datadog_checks lib means that in order to test the wmi checks, it requires the win32 extension to be available. So we now conditionally import the wmi_calculator stuff so that it will pass if it doesn't exist. This should cause the tests to fail if we are on a windows machine, but be properly skipped if we aren't (had to do this to get Travis to be 🆗 )

This PR assumes this is in datadog-checks-base https://github.com/DataDog/integrations-core/pull/1897


nmuesch:
@olivielpeau I was able to build the Agent and tested the imports continue to work on Windows and Linux specifically. Could someone from croissant take another pass? Would love to get this merged for the next minor release.

PR #3767, opened 2018-07-26
Fix copying integrations' data files for source install

rishabh:
### What does this PR do?

Fixes the paths of each integrations' data files when installing from source.

### Motivation

To make integrations work.

### Testing Guidelines

An overview on [testing](https://github.com/DataDog/dd-agent/blob/master/tests/README.md) is available in our contribution guidelines.


nmuesch:
This looks good to me, thanks for the fix! We should get a review from @DataDog/agent-core

PR #3787, opened 2018-10-30
[windows] change calculation of % interrupt time from NT kernel function

derekwbrown:
to pdh based counter

### Motivation

Get data from windows performance counters; begin moving all of the system checks to windows performance counters; match agent6 implementation


remicalixte:
Better than to rely on an undocumented part of the windows API

PR #3794, opened 2018-11-27
Report amazon linux in metadata using the distro module

remicalixte:
if you have not yet done so.*

### Motivation

amazon linux is not reported corrrectly with the builtin `platform` module. We use the `distro` module which is the recommended replacement since `platform.dist` and `platform.linux_distribution` are deprecated and will be removed in python 3.8: https://docs.python.org/3.7/library/platform.html#platform.linux_distribution.

We change `amzn` to `amazon` to match what `gopsutil` reports in agent 6: https://github.com/shirou/gopsutil/blob/0f70a4a06f7a1039305ec9b3ba9c2800f8929fba/host/host_linux.go#L403.

### DIFF
--- a/config.py
+++ b/config.py
@@ -25,6 +25,7 @@ from importlib import import_module
 
 # 3p
 import simplejson as json
+import distro
 
 # project
 from util import check_yaml, config_to_yaml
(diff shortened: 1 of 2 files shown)


olivielpeau:
(➕ on Jaime's comment, with just a tiny correction: `requirements.txt` is actually used for one Agent 5 install method: the install from source 😉)

PR #3805, opened 2019-01-07
Fix status page cleaning up warning message after the first run

hush-hush:
### What does this PR do?

Fix status page cleaning up warning message after the first run

This also add a feature flag to disable the linter feature if needed.

New status page is now much less verbose when checks have multiple instances. We also only log Python3 error once for all instance.
```
  Checks
  ======
  
    test (custom)
    -------------
      - instance #0 [WARNING]
          Warning: check is not compatible with Python3 (see log for more information)
      - instance #1 [WARNING]
          Warning: check is not compatible with Python3 (see log for more information)
      - instance #2 [WARNING]
          Warning: check is not compatible with Python3 (see log for more information)
      - instance #3 [WARNING]
          Warning: check is not compatible with Python3 (see log for more information)
      - instance #4 [WARNING]
          Warning: check is not compatible with Python3 (see log for more information)
      - Collected 0 metrics, 0 events & 0 service checks
```

### DIFF
--- a/checks/__init__.py
+++ b/checks/__init__.py
@@ -350,6 +350,7 @@ class AgentCheck(object):
         self.service_checks = []
         self.instances = instances or []
         self.warnings = []
+        self.persistent_warnings = []
         self.check_version = None
         self.library_versions = None
         self.last_collection_time = defaultdict(int)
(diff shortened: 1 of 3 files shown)


checks/__init__.py
```diff
@@ -350,6 +350,7 @@ def __init__(self, name, init_config, agentConfig, instances=None):
         self.service_checks = []
         self.instances = instances or []
         self.warnings = []
+        self.persistant_warnings = []
```
truthbk: nit: `persistent_warnings`

PR #3809, opened 2019-01-09
Add telemetry to the python3 linter and but display warnings in the status page for now

hush-hush:
### What does this PR do?

Add telemetry to the linter but don't display warnings in the status page for now

### DIFF
--- a/config.py
+++ b/config.py
@@ -57,6 +57,9 @@ SD_PIPE_WIN_PATH = "\\\\.\\pipe\\{pipename}"
 UNKNOWN_WHEEL_VERSION_MSG = 'Unknown Wheel'
 CUSTOM_CHECK_VERSION_MSG = 'custom'
 A7_COMPATIBILITY_ATTR = 'a7_compatible'
+A7_COMPATIBILITY_READY = 'ready'
+A7_COMPATIBILITY_NOT_READY = 'not_ready'
+A7_COMPATIBILITY_UNKNOWN = 'unknown'
 
 log = logging.getLogger(__name__)
 
@@ -130,7 +133,6 @@ def get_parsed_args():
                                 'verbose': False,
                                 'profile': False}), []
     return options, args
-
 
 def get_version():
     return AGENT_VERSION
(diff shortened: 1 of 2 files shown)


config.py
```diff
@@ -131,6 +134,10 @@ def get_parsed_args():
                                 'profile': False}), []
     return options, args
 
+def a7_compatible_to_int(status):
+    if status == A7_COMPATIBILITY_READY:
+        return 1
+    return 0
```
masci: nit: since this is only used in `collector.py`, can we move it there and spare the import?

config.py
```diff
@@ -1189,23 +1191,27 @@ def load_check_from_places(check_config, check_name, checks_places, agentConfig)
         if not agentConfig.get("disable_py3_validation", False):
             if version_override in (UNKNOWN_WHEEL_VERSION_MSG, CUSTOM_CHECK_VERSION_MSG):
                 log.info('Validating {} for Python 3 compatibility'.format(check_path))
+                a7_compatible = A7_COMPATIBILITY_READY
                 try:
                     output, _, _ = get_subprocess_output(['a7_validate', check_path], log)
```
truthbk: We made the same mistake we almost made with A6. This is a blocking call, and a bit dangerous to have in such a sensitive area of the code. The likelihood of this being a problem is small, but maybe we should do something like this to `get_subprocess_output`:

```python
# other imports
from threading import Timer

def subprocess_output(command, raise_on_empty_output, timeout=None):
    """
    Run the given subprocess command and return its output. This is a private method
    and should not be called directly, use `get_subprocess_output` instead.
    """

    # Use tempfile, allowing a larger amount of memory. The subprocess.Popen
    # docs warn that the data read is buffered in memory. They suggest not to
    # use subprocess.PIPE if the data size is large or unlimited.
    with tempfile.TemporaryFile() as stdout_f, tempfile.TemporaryFile() as stderr_f:
        try:
            proc = subprocess.Popen(command, stdout=stdout_f, stderr=stderr_f)
            if timeout:
                timer = Timer(timeout, proc.kill)
        finally:
            if timeout:
                timer.cancel()

        proc.wait()
        stderr_f.seek(0)
        err = stderr_f.read()
        stdout_f.seek(0)
        output = stdout_f.read()

    if not output and raise_on_empty_output:
        raise SubprocessOutputEmptyError("get_subprocess_output expected output but had none.")

    return (output, err, proc.returncode)
```

config.py
```diff
@@ -1189,23 +1191,27 @@ def load_check_from_places(check_config, check_name, checks_places, agentConfig)
         if not agentConfig.get("disable_py3_validation", False):
             if version_override in (UNKNOWN_WHEEL_VERSION_MSG, CUSTOM_CHECK_VERSION_MSG):
                 log.info('Validating {} for Python 3 compatibility'.format(check_path))
```
olivielpeau: this should be a debug log

config.py
```diff
@@ -1189,23 +1191,27 @@ def load_check_from_places(check_config, check_name, checks_places, agentConfig)
...
+                a7_compatible = A7_COMPATIBILITY_READY
                 try:
                     output, _, _ = get_subprocess_output(['a7_validate', check_path], log)
                 except Exception as e:
                     log.error("error executing a7_validate on custom check: %s", e)
```
olivielpeau: this should be a warning instead of an error

PR #3838, opened 2019-04-25
[py2exe] exclude system DLLs we were accidentally shipping

derekwbrown:
### What does this PR do?

Exclude system DLLs from shipping in the agent 5 package

### Motivation

Agent wasn't starting on win2k8

### Testing Guidelines

Verify installation and startup on win2k8, 2012, 16


truthbk:
Ran a few tests on 2008. DLLs were not present and psutil appeared to working fine, thus I'm assuming the system libs are getting picked up properly (dependencywalker was successfully able to resolve and find the DLLs at the expected `c:\windows\system32` path). 

Merging!

PR #3839, opened 2019-04-30
[requirements] bumping pyyaml to 5.1

truthbk:
### What does this PR do?

Bumps PyYaml to the latest 5.1.

### Motivation

Enable new `full_load()` option for customers perhaps wanting to use that after `load_all()` was patched for security reasons.

### Testing Guidelines

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,5 +7,5 @@ gem 'rake'
 gem 'rubocop', '~>0.50.0'
 
 # Upper versions are incompatible with ruby version on Travis (1.9.3)
-gem 'public_suffix', '~>1.4.0'
 gem 'parallel', '~>1.13.0'
+gem 'public_suffix', '~>1.4.0'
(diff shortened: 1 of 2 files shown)


hydrosquall:
Chiming in b/c I've been waiting on updating pyyaml in my own python package for a little while and have been following this issue closely since the 4.x release last summer, - since DD enforces an API boundary in `ddyaml.py` which monkey-patches yaml.load to be safe, it looks like the upgrade doesn't require any internal API rewrite. 

However, there are some possible internal small internal breakages that will be patched in 5.2, as indicated in this release note: https://github.com/yaml/pyyaml/issues/265

truthbk:
Note: Further context on why the monkey patch has been kept in place follows. The pyyaml authors go on to say the following about the 5.1 release https://github.com/yaml/pyyaml/pull/257):

```
We still recommend that people choose SafeLoader for untrusted data, but
aribitrary code execution will no longer be possible using yaml.load() with
the default loader (FullLoader). FullLoader will instantiate objects of classes
that you have imported. Since object instantiation runs the class's constructor code, that may be exploitable.
```

Because the `FullLoader` still presents some exploitable vector, we will wait for 6.0, when the `yaml.load()` will raise an exception if no loader is specified in the calling code, to remove the monkey patch.

PR #3846, opened 2019-05-27
[CI] Use agent-v5 branch of integrations-core

olivielpeau:
### What does this PR do?

Switch to use the `agent-v5` branch of `integrations-core`

### Motivation

The code of Agent v5 is meant to work with the `agent-v5` branch of integrations-core now.

### Testing

As long as the CI passes on the PR we should be good

### Additional Notes

Should've done this when we created the `agent-v5` branches on the other repos, but I missed it at the time.


olivielpeau:
(CI is still failing, I'll have to continue looking for the actual root cause)

PR #3852, opened 2019-09-23
Return total count value as a metric

dpricha89:
# What does this PR do?

Sends the total metric count value of histogram metrics.

### Motivation

When migrating from regular statsd to dogstatsd we noticed that the count was aggregated differently.

### Testing Guidelines

Hoping the PR creation will run the tests.

### DIFF
--- a/aggregator.py
+++ b/aggregator.py
@@ -268,6 +268,7 @@ class Histogram(Metric):
             ('avg', avg, MetricTypes.GAUGE),
             ('sum', sum_, MetricTypes.GAUGE),
             ('count', self.count/interval, MetricTypes.RATE),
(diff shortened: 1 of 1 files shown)


olivielpeau:
Hi @dpricha89, I believe querying the `.count`-prefixed metric with the `as_count()` modifier will give you the same result as the `.total`-prefixed metric that you want to add in this PR.

If not, do you have more details on what you're seeing in-app vs. what you expect to see? Feel free to open a support case at support@datadoghq.com if you wish to share non-public information.

olivielpeau:
Closing this PR as I believe the functionality it implements is already available in Datadog. Feel free to comment here if you have additional thoughts on this.

PR #3854, opened 2019-10-14
Release 5.32.5

albertvaka:
Bump AGENT_VERSION and update changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,14 @@ Datadog Agent 6 has been officially released and the release notes can be found 
 Changes
 =======
 
+# 5.32.5 / 2019-10-15
+
+**Windows only**
+
+### Changes
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,14 @@ Datadog Agent 6 has been officially released and the release notes can be found
 Changes
 =======
 
+# 5.32.5 / 2019-10-14
+
+**Windows and Source Install**
```
olivielpeau: I don't think it makes sense to release for `Source Install` since the change only affects Windows, and the `Source Install` is only available on Unix platforms on Agent v5.

Let's mention Windows only instead?

```suggestion
**Windows only**
```

CHANGELOG.md
```diff
@@ -3,6 +3,14 @@ Datadog Agent 6 has been officially released and the release notes can be found
...
+
+### Changes
+
+* [BUGFIX] Fix trace agent failing to start on Windows.
```
olivielpeau: could you link the PR? :)

PR #3856, opened 2019-10-16
Call pylint directly instead of using a7

olivielpeau:
### What does this PR do?

Makes the Agent call pylint directly in a subprocess instead of using a7.

### Testing Guidelines

Checked this worked on local dev env and custom check.

This will have to be QA'ed on Windows as well.

### Additional Notes

Also renamed the constants since they're actually related to python 3.

Requires https://github.com/DataDog/dd-agent-omnibus/pull/383

### DIFF
--- a/config.py
+++ b/config.py
@@ -56,10 +56,10 @@ SD_PIPE_UNIX_PATH = '/opt/datadog-agent/run'
 SD_PIPE_WIN_PATH = "\\\\.\\pipe\\{pipename}"
 UNKNOWN_WHEEL_VERSION_MSG = 'Unknown Wheel'
 CUSTOM_CHECK_VERSION_MSG = 'custom'
-A7_COMPATIBILITY_ATTR = 'a7_compatible'
-A7_COMPATIBILITY_READY = 'ready'
-A7_COMPATIBILITY_NOT_READY = 'not_ready'
-A7_COMPATIBILITY_UNKNOWN = 'unknown'
+PY3_COMPATIBILITY_ATTR = 'py3_compatible'
+PY3_COMPATIBILITY_READY = 'ready'
+PY3_COMPATIBILITY_NOT_READY = 'not_ready'
+PY3_COMPATIBILITY_UNKNOWN = 'unknown'
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


config.py
```diff
@@ -1198,26 +1194,29 @@ def load_check_from_places(check_config, check_name, checks_places, agentConfig)
...
             if version_override in (UNKNOWN_WHEEL_VERSION_MSG, CUSTOM_CHECK_VERSION_MSG):
-                a7_compatible = A7_COMPATIBILITY_READY
+                py3_compatible = PY3_COMPATIBILITY_UNKNOWN
+                warnings = []
                 try:
                     file_path = os.path.realpath(check_path.decode(sys.getfilesystemencoding()))
-                    warnings = a7.validate.validate_py3(file_path)
+                    output, _, _ = get_subprocess_output(
+                        [sys.executable, "-m", "pylint", "-f", "json", "--py3k", "-d", "W1618", "--persistent", "no", "--exit-zero", file_path], log)
+                    warnings = json.loads(output)
```
therve: I think you need to add a condition here, because on old versions of pylint, the output can be '' instead of '[]' when there is no errors.

config.py
```diff
@@ -1198,26 +1194,29 @@ def load_check_from_places(check_config, check_name, checks_places, agentConfig)
...
-                    warnings = []
-                    a7_compatible = A7_COMPATIBILITY_UNKNOWN
 
                 for w in warnings:
                     message = w.get('message')
                     if message:
                         # for now we don't display anything in the status page
                         # log.warn("check '{}' is not Python3 compatible: {}".format(check_name, message))
-                        a7_compatible = A7_COMPATIBILITY_NOT_READY
+                        py3_compatible = PY3_COMPATIBILITY_NOT_READY
```
therve: If think you can simplify by saying `if warnings`, no need to iterate through all.

config.py
olivielpeau: thanks, addressed this. Note that `get_subprocess_output` raises `SubprocessOutputEmptyError` if the output is empty, and the logic handles this exception

config.py
olivielpeau: done

PR #3859, opened 2019-10-22
Disable use of PEP517 in setup_agent

belidzs:
### What does this PR do?

It will disable the use of PEP517 during installing python packages with pip.

### Motivation

Execution failed on Raspberry Pi 4B with BackendUnavailable during installation of the dependency `lazy-object-proxy`

### Testing Guidelines

None

### DIFF
--- a/packaging/datadog-agent/source/setup_agent.sh
+++ b/packaging/datadog-agent/source/setup_agent.sh
@@ -412,7 +412,7 @@ print_done
 
 print_console "* Installing requirements"
 $DOWNLOADER "$DD_HOME/requirements.txt" "$BASE_GITHUB_URL/requirements.txt"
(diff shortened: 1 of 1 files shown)


darthnater007:
Hi Team, I've discovered the root issue of this pep517 Backend Unavailable stacktrace is a bug within pip 19.x

https://github.com/pypa/pip/issues/6164

Source install is broken by default on some machines

The above issue indicated that the pip v19.x is the issue, and version 18.1 of pip will resolve this:

Steps
I pulled down a local repo of the agent, and set $PIP_VERSION=“18.1” in dd-agent/packaging/datadog-agent/source/setup_agent.sh
(https://github.com/DataDog/dd-agent/blob/master/packaging/datadog-agent/source/setup_agent.sh#L17)

This has allowed the requirements portion fo the install to complete without issue.

PR #3898, opened 2020-10-27
Log a warning that file-configs override service-discovery configs

sodabrew:
### What does this PR do?

Logs a warning if a service-discovery configured check will be ignored if the same-named check was already configured from a file.

### Motivation

I have an haproxy.yaml file pointing at one instance of haproxy that I want to monitor on all servers. I am using Docker service discovery to configure other containers for service checks. This week I wanted to deploy a new use case for an haproxy container, and so I diligently added the Docker labels for an haproxy check. Nothing happened. Radio silence in the logs files.

### Testing Guidelines

### Additional Notes

A better fix would be to merge the `instances` list from the file with the service discovered-instances list. Ironically, there is code in the service discovery config module to do exactly this, and then it is dutifully discarded if a file config exists. But first I want to solve the head-banging-on-wall problem of failing to log the discarded config.

### DIFF
--- a/config.py
+++ b/config.py
@@ -1268,6 +1268,7 @@ def load_check_directory(agentConfig, hostname):
         if check_name in initialized_checks or \
                 check_name in init_failed_checks or \
                 check_name in JMX_CHECKS:
+            log.warning('Ignoring config for service discovery check %s because it was already configured by file.' % check_name)
             continue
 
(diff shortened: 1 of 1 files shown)


sodabrew:
Do Agent 6 or 7 allow checks defined in multiple sources per my use case?

PR #3909, opened 2021-03-18
[build] Ensure pip version is Python2 compatible

sgnn7:
### What does this PR do?

Old code was fetching and using the latest pip which no longer supports
Python2. This change ensures that we only use Python v2-compatible
version for our CI/CD. while this doesn't fix all the issues with the pipeline,
it fixes the problems with pip and gets most of the CircleCI builds passing

### Motivation

CI/CD failures

### Testing Guidelines

### Additional Notes


olivielpeau (approved):
Thanks, can you create a follow up card for the remaining failures?

sgnn7:
@olivielpeau Followup cards created

PR #3916, opened 2021-07-13
[install_agent] Fix APT keyring permissions

bkabrda:
### What does this PR do?

On Ubuntu/Debian systems, there is an `_apt` user which reads the APT keyrings to verify signatures. Our APT keyrings are owned by root and so if they're not world-readable, the `_apt` user fails to read them and thus can't verify the signatures.

### Motivation

### Testing Guidelines

### Additional Notes


olivielpeau (approved):
LGTM if you've been able to confirm that the install script works well with these changes.

It's OK to make this keyring world-readable since it will only ever contain public keys. FTR, related to https://github.com/DataDog/datadog-agent/pull/8609.

bkabrda:
Yeah, I've tested in on Ubuntu 14.04 and 18.04 and everything works even with `umask 007`, which was the case that broke this. I'm merging.

PR #3922, opened 2021-12-28
Setting up codescanning for dd-agent

ganeshkumarsv:
### What does this PR do?

Static code analysis on dd-agent using Github's Code Scanning.

This PR has Github workflow configs to trigger the Code Scanning on every PR and on push to master branch. As of now only default security based CodeQL queries are configured. We can gradually add queries for Code Quality later. I have added the `tests` directories to ignore while code scanning. paths-ignore works only for interpreted languages like Python/Javascript. Feel free to keep / add / remove the entries from codeql_config.yml.
### Motivation

### Testing Guidelines

### Additional Notes

### DIFF
--- a/.github/codeql_config.yml
+++ b/.github/codeql_config.yml
@@ -1,0 +1,4 @@
+name: "CodeQL config"
+
+paths-ignore: 
+  - tests
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,68 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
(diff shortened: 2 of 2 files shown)


djmitche (commented):
I don't have a problem with merging this, but I'm not sure it's useful.  This repo is for an ancient version of the agent that is more-or-less in an "extended maintenance" phase.  Looking back over the last year, the only changes have been to packaging and build mechanics.  So, if we were to find a code quality issue here, I'm not sure it should be fixed.  We might just be making extra work for ourselves here.  @truthbk what do you think?

truthbk:
Indeed, the truth is this is pretty much unnecessary. This repository is plagued by numerous issues we will not likely be addressing given the current support policy for Agent 5. Users should be encouraged to migrate to Agent 6/7 to address many of the issues and shortcomings here, in Agent 5.

ganeshkumarsv:
@djmitche @truthbk I agree with your point. But as long as customers can install this and as long as we are supporting it, compliance wants us to regularly perform security code scanning on the repositories.

cc: @stephengroat-dd

PR #3943, opened 2025-06-04
High-level diagram of dd-agent's code base

ivanmilevtues:
### What does this PR do?

Generated documentation to help new developers to get to know the python-slack-sdk (from the stand point of contributors).

Let me know what is your opinion on the generated diagrams, do they fit in your existing process of on-boarding people to new codebases?

Any feedback is more than welcome!

Full disclosure: we're trying to turn this into a startup, but we're still in a very early stage and figuring out what will actually be useful for people.

### Motivation

As an open-source project, we think that there are a lot of people out there that want to get to know it better and possibly contribute. We would love to help you and other OSS to improve on the process and are trying to collect as much feedback as possible to improve what we are doing and make something that people use and like!

### Testing Guidelines

No tests are needed as no additional logic is added.

### Additional Notes

Again any sort of feedback is more than welcome, if you want to get in touch with me please do so.
Looking forward to see if this sparks any interest!

### DIFF
--- a/.codeboarding/Agent Self-Monitoring.md
+++ b/.codeboarding/Agent Self-Monitoring.md
@@ -1,0 +1,129 @@
+```mermaid
+graph LR
+    AgentStatus["AgentStatus"]
+    CollectorStatus["CollectorStatus"]
+    DogstatsdStatus["DogstatsdStatus"]
+    ForwarderStatus["ForwarderStatus"]
+    get_ntp_info["get_ntp_info"]
+    validate_api_key["validate_api_key"]
+    get_jmx_status["get_jmx_status"]
+    get_jmx_instance_status["get_jmx_instance_status"]
+    style["style"]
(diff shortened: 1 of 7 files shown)


ivanmilevtues:
Thank you for the response and especially for the feedback @sgnn7!

The idea for the document to be up to date with the project's content is to be regenerated often. This documentation is generated by leveraging Static Analysis and LLMs. We want it to be solely based on the code as this is the actual ground truth. Some people suggested us that we should create a Github workflow, which you can add to your repository, this could address the outdated docs issue.

This generation is with a semi-weak LLM, but from your feedback it seems like it is not enough. Thank you for the .codeboarding directory comment, usually I try to open discussions before PR's as it is less invasive.

All the other comments (linking, not useful comments, UI refinement) you made make sense and we will try to improve on those and when we have better version of the product will probably open a discussion in your new repository!

Again thanks for the response, it really helps us!

Best, Ivan

PR #3963, opened 2026-03-10
feat(runbooks): unified scripts for sectigo CA rotation (all OS versions)

louis-cqrl:
## Summary

Following review feedback, the separate legacy-OS scripts have been merged into the main scripts. There are now two scripts covering all supported platforms:

- **`linux.sh`** — all Linux distributions (EL5/6, EL7+, Ubuntu, Debian, Fedora, …)
- **`windows.ps1`** — all supported Windows versions (2008 R2, 2012 R2, 2016+)

Each script auto-detects the environment at runtime and applies the appropriate compatibility adjustments, so operators no longer need to choose between multiple scripts.

---

### `linux.sh` — unified Linux script

Merges `el5-6/linux.sh` into the standard `linux.sh`. On RHEL/CentOS/Oracle Linux 5 and 6, the script automatically enables **EL5/6 compatibility mode**:

- **bash 3.1 compatible** — no bash 4+ features; `pipefail` is supported from bash 3.0+
- **Portable `truncate`** — coreutils 5.97 (EL5) does not ship `truncate`; falls back to `sudo sh -c "> \"$f\""`
- **Two-stage download** — tries standard TLS first; falls back to `--insecure` / `--no-check-certificate` if the system CA bundle is too old to verify `raw.githubusercontent.com` (common on EL5), with a clear operator warning
- **No `journalctl`** — EL5/6 use SysV init; the check is skipped on legacy EL
- **OS detection** — reads `/etc/redhat-release` to identify RHEL vs CentOS vs Oracle Linux (including early OL5 `"Enterprise Linux"` strings) and set compatibility mode accordingly

Options available on all distros:
- `-c <cert_file>` — local copy of `datadog-cert.pem` for air-gapped hosts
- `-p <agent_directory>` — custom Agent installation path

Script-level hardening: `set -eo pipefail`.

**Reliability fixes:**
- Connectivity checks are now **non-fatal**: if the pre-restart cert verification or the post-restart log scan cannot confirm connectivity (network restriction, firewall, no curl/wget), the script prints a warning and continues to completion rather than aborting — the certificate is still installed
- `grep | head -10 || true` — prevents `set -o pipefail` from aborting the script when `head` closes the pipe early (SIGPIPE → exit 141) if there are more than 10 matching SSL error lines
- `journalctl | grep | head -1 || true` — same pipefail guard for the journald scan in `test_connectivity_since_restart`; result is captured into a variable and matched with `-n`
- `update_datadog_config` now writes a timestamped backup (`${CONF_FILE}.pre-cert-update-<ts>`) before the `sed -i` mutation, mirroring the Windows behavior and the existing log-rotation backup pattern
- Final **summary block** clearly tells the operator what was done and what (if anything) to verify manually

### `windows.ps1` — unified Windows script

Merges `windows-2008-2012/windows.ps1` into the standard `windows.ps1`. Compatible with PowerShell 2.0+ and .NET 3.5+:

- **Primary downloader is `System.Net.WebClient`** (.NET 2.0) — works on PS 2.0 (default on 2008 R2); `Invoke-WebRequest` and BITS are kept as fallbacks
- **`Get-Content -Raw` replaced** with `[System.IO.File]::ReadAllText()` (.NET 2.0)
- **Manual epoch calculation** — avoids `[DateTimeOffset]::ToUnixTimeSeconds()` (.NET 4.6+)
- **TLS 1.2 via integer `3072`** — avoids `[Net.SecurityProtocolType]::Tls12` named enum (.NET 4.5+)
- **OS architecture detection** fixed — uses `PROCESSOR_ARCHITECTURE` / `PROCESSOR_ARCHITEW6432` env vars instead of `[IntPtr]::Size` (which reports PowerShell process bitness, not OS bitness, in 32-bit WOW64 sessions) or `[Environment]::Is64BitOperatingSystem` (.NET 4.0+)
- **`Test-Certificate`** validates PEM format before the connectivity check; the connectivity check uses the machine trust store (the definitive check is the post-restart log scan)
- `-CertFile` / `-c` option for air-gapped hosts
- `-AgentDirectory` / `-p` option for custom installation paths

**Reliability fixes:**
- Connectivity checks are now **non-fatal**: `Test-Certificate` connectivity failure and SSL errors detected in `Test-ConnectivitySinceRestart` set a warning flag instead of calling `Error-Exit` — the script completes and the operator is told what to verify manually
- All `Test-Path` calls now use `-LiteralPath` so paths containing `[` or `]` are not misinterpreted as glob patterns
- Final **summary block** (same as Linux) with manual verification steps when needed

### `README.md`

- One script per OS (`linux.sh` / `windows.ps1`) — the legacy-OS subfolders are gone
- Per-OS download instructions updated; the 2008 R2 `System.Net.WebClient` download snippet is retained
- **Expected Output** section now shows both outcomes: "verified successfully" and "cert replaced, verify manually", with representative output for each
- **Configuration Changes** section notes that both Linux and Windows now create a timestamped backup before modifying `datadog.conf`
- **Troubleshooting** section updated to reflect that connectivity verification failures are warnings, not script failures


pgimalac:
@codex review

runbooks/sectigo-root-ca-rotation-2025/README.md
```diff
@@ -22,20 +22,56 @@ This runbook provides automated scripts for both Linux and Windows that will:
...
 ## Available Scripts
 
-### Linux: `linux.sh`
+### Which script should I use?
 
-- **Supported distributions**: Ubuntu/Debian, RHEL/CentOS, Fedora, and similar
+| Your OS | Script to use |
+|---|---|
+| RHEL / CentOS / Oracle Linux **5 or 6** | [`el5-6/linux.sh`](el5-6/linux.sh) |
+| RHEL / CentOS / Oracle Linux **7+** | [`linux.sh`](linux.sh) |
```
pgimalac: As discussed IRL probably best if we can have a single script and determine the version within the script to decide what to do

runbooks/sectigo-root-ca-rotation-2025/README.md
```diff
@@ -22,20 +22,56 @@ This runbook provides automated scripts for both Linux and Windows that will:
...
+### `linux.sh` (EL 7+, Ubuntu, Debian, Fedora, and similar)
+
+- **Supported distributions**: Ubuntu/Debian, RHEL/CentOS/Oracle Linux 7+, Fedora, and similar
+- **Requirements**: Root or sudo access, `curl` or `wget`
+
+### `el5-6/linux.sh` (RHEL / CentOS / Oracle Linux 5 and 6)
+
+- **Supported distributions**: RHEL 5/6, CentOS 5/6, Oracle Linux 5/6
 - **Requirements**: Root or sudo access
-- **Automatic features**: Will attempt to install `curl` temporarily if neither `curl` nor `wget` is available
```
pgimalac, line 29: The script doesn't install curl automatically anymore if it's missing ?

fabbing (approved):
No additional comment beside what Codex said, and I'm not familiar with Powershell.

runbooks/sectigo-root-ca-rotation-2025/README.md
```diff
@@ -22,20 +22,56 @@ This runbook provides automated scripts for both Linux and Windows that will:
...
+### Which script should I use?
 
-- **Supported distributions**: Ubuntu/Debian, RHEL/CentOS, Fedora, and similar
+| Your OS | Script to use |
+|---|---|
+| RHEL / CentOS / Oracle Linux **5 or 6** | [`el5-6/linux.sh`](el5-6/linux.sh) |
+| RHEL / CentOS / Oracle Linux **7+** | [`linux.sh`](linux.sh) |
+| Ubuntu / Debian / Fedora / any modern distro | [`linux.sh`](linux.sh) |
+| Windows Server **2008 R2 or 2012 R2** | [`windows-2008-2012/windows.ps1`](windows-2008-2012/windows.ps1) |
+| Windows Server **2016+** | [`windows.ps1`](windows.ps1) |
```
pgimalac: Same for windows if possible, determining the version in the script avoids user mistakes and be more convenient

runbooks/sectigo-root-ca-rotation-2025/el5-6/linux.sh
```diff
@@ -0,0 +1,344 @@
...
+# Covers: RHEL 5/6, CentOS 5/6, Oracle Linux 5/6.
+#
+# Differences from ../linux.sh (EL 7+ script):
+#   1. Uses >/dev/null 2>&1 instead of &> (safer on bash 3.1 shipped with RHEL 5)
+#   2. Replaces `truncate` with a portable sudo sh -c redirect (truncate is absent
+#      from coreutils 5.97 shipped with RHEL 5 / CentOS 5 / OL 5)
+#   3. Falls back to --insecure / --no-check-certificate if the system CA bundle
+#      is too old to verify raw.githubusercontent.com (common on RHEL/CentOS 5)
+#   4. Removes the journalctl check entirely — RHEL/CentOS/OL 5 and 6 use SysV
+#      init and do not ship systemd/journald
```
pgimalac: Do we need a separate script for that ? Why not just update the original script to use the alternatives directly ? (I guess the alternatives would work on more recent versions as well)

pgimalac: In particular if it's a copy paste of the other script except for those changes, directly updating the other script would avoid having to maintain two versions and ease reviews

runbooks/sectigo-root-ca-rotation-2025/el5-6/linux.sh
```diff
@@ -0,0 +1,344 @@
+#!/bin/bash
```
pgimalac: Should the script set some options, eg. to fail if any command fails or if pipe fails ?

runbooks/sectigo-root-ca-rotation-2025/README.md
louis-cqrl: Not anymore, IIRC Jeremy asked it to be removed as it should be on customer side to do it

runbooks/sectigo-root-ca-rotation-2025/README.md
louis-cqrl: Done — both scripts now auto-detect the OS/environment at runtime. `linux.sh` reads `/etc/redhat-release` to identify RHEL/CentOS/Oracle Linux and activates EL5/6 compatibility mode automatically on versions 5 and 6. `windows.ps1` detects OS bitness via `PROCESSOR_ARCHITECTURE`/`PROCESSOR_ARCHITEW6432` env vars and the right cert path based on Agent version.

runbooks/sectigo-root-ca-rotation-2025/README.md
louis-cqrl: Done — `windows.ps1` now auto-detects OS bitness (via `PROCESSOR_ARCHITECTURE`/`PROCESSOR_ARCHITEW6432`) and Agent version (by checking whether the `agent\` subdirectory exists under the install path). A single script handles all supported Windows versions without any choice required from the operator.

runbooks/sectigo-root-ca-rotation-2025/el5-6/linux.sh
louis-cqrl: Done — `el5-6/linux.sh` has been removed. The main `linux.sh` now reads `/etc/redhat-release` at startup and automatically applies EL5/6 compatibility mode (portable log truncation, insecure download fallback, no journalctl) when running on RHEL/CentOS/Oracle Linux 5 or 6.

louis-cqrl: Agreed — no duplication anymore. Single unified script with the OS-specific paths conditional on the detected version.

runbooks/sectigo-root-ca-rotation-2025/el5-6/linux.sh
louis-cqrl: Done — the unified `linux.sh` has `set -eo pipefail` on line 13. We also added `|| true` to the `grep | head -10` pipeline to prevent `pipefail` from aborting on SIGPIPE when `head` closes the pipe early after 10 lines.

runbooks/sectigo-root-ca-rotation-2025/windows-2008-2012/windows.ps1
```diff
@@ -0,0 +1,406 @@
...
+        else {
+            return "$CUSTOM_DD_AGENT_DIR\files\datadog-cert.pem"
+        }
+    }
+
+    # [IntPtr]::Size -eq 8 works on PS 2.0 + .NET 2.0+.
+    # [Environment]::Is64BitOperatingSystem requires .NET 4.0 and is not available
+    # on Windows 2008 R2 with the default PS 2.0 / .NET 3.5 stack.
+    $is64 = [IntPtr]::Size -eq 8
+    if ($is64) {
```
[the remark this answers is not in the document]
louis-cqrl: Fixed in the unified `windows.ps1` — OS bitness is now detected via `PROCESSOR_ARCHITECTURE`/`PROCESSOR_ARCHITEW6432` environment variables instead of `[IntPtr]::Size`, which reports the PowerShell process bitness rather than the OS bitness (a 32-bit PS host on a 64-bit OS would have returned the wrong path). The old `windows-2008-2012/windows.ps1` file has been removed.

runbooks/sectigo-root-ca-rotation-2025/windows-2008-2012/windows.ps1
```diff
@@ -0,0 +1,406 @@
...
+
+function Test-Certificate {
+    param([string]$CertFile)
+    Write-Host "Verifying the downloaded certificate..."
+    $testUrl = "https://app.datadoghq.com"
+    Enable-Tls12
+    try {
+        $request = [System.Net.WebRequest]::Create($testUrl)
+        $request.Timeout = 10000
+        $response = $request.GetResponse()
```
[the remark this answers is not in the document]
louis-cqrl: Addressed in the unified `windows.ps1`. `Test-Certificate` now validates PEM format first (file exists and contains `BEGIN CERTIFICATE`), then does a reachability check using the machine trust store. The code comment explicitly documents this design choice: the connectivity check confirms network access, not cert correctness. The definitive correctness check is the Agent log scan after restart — any wrong or corrupt cert will surface as SSL errors there. Using the cert file as a custom trust anchor in .NET on PS 2.0 would require `ServicePointManager.ServerCertificateValidationCallback` wiring, which is overly complex for a runbook script.

runbooks/sectigo-root-ca-rotation-2025/linux.sh
```diff
@@ -225,14 +350,45 @@ test_connectivity_since_restart() {
...
   check_downloader
   ensure_target_directory
-  download_certificate
+  if [ -n "$LOCAL_CERT_FILE" ]; then
+    install_local_certificate
+  else
+    download_certificate
+  fi
   verify_certificate
   update_datadog_config
```
pgimalac, line 362: Is there any backup for the datadog config ?
This is not introduced in this PR but still 😄

runbooks/sectigo-root-ca-rotation-2025/linux.sh
```diff
@@ -179,42 +296,50 @@ test_connectivity_since_restart() {
...
     echo "  Checking journald logs..."
-    # Prefer epoch form; fall back to a readable UTC timestamp if needed
+    # Prefer epoch form; fall back to a readable UTC timestamp if needed.
     local SINCE_ARG
     SINCE_ARG="@${PRE_TS_UNIX}"
-    if ! sudo journalctl --since "$SINCE_ARG" -n 0 &>/dev/null; then
+    if ! sudo journalctl --since "$SINCE_ARG" -n 0 >/dev/null 2>&1; then
       SINCE_ARG="$PRE_TS_READABLE"
     fi
     if sudo journalctl -u datadog-agent --since "$SINCE_ARG" --no-pager 2>/dev/null | grep -qiE "$pattern"; then
```
pgimalac, line 326: I think this will need an `|| true` just like line 308 above, to avoid pipefail issue

pgimalac (approved):

jeremy-hanna (approved):

